PR #80, opened 2015-01-02
Add states to zoom level 4

rmarianski:


rmarianski:
We wanted to start showing state boundaries at zoom level 4. We already have an appropriate index for this.

@bcamper, @hkrishna: any objections?

bcamper:
works for me

PR #82, opened 2015-01-18
First attempt to add waterway lines (fix #81)

### LINKED ISSUE #81: Waterway lines?
Hi :)

I don't see the river lines in the vector tiles (i.e. `waterway=river` or `waterway=canal` for example).

Am I missing it, or is it on purpose, or is it a "to be done" issue? :)

This layer would be needed even in a generic style to display the river labels along the river line (now, having only polygons, we can't).

If it's to be done, I may work on a PR.

Thanks!

Yohan

yohanboniface:
So here is a shot :)

`river` and `canal` are made visible starting at zoom 9, `dam` and `stream` starting at zoom 13 and `ditch` and `drain` at zoom 16.

I'm not sure if we need a dedicated index, maybe a btree on waterway plus way. I'm happy to have your feedback on this :)

Thanks for reviewing!

Yohan

Ref: #81

### DIFF
--- a/queries/water-z11.pgsql
+++ b/queries/water-z11.pgsql
@@ -35,6 +35,24 @@ FROM
         AND way_area::bigint > 25600 -- 4px
         AND mz_way12 && !bbox!
 
+    --
+    -- Water line geometries
+    --
+    UNION
+
+    SELECT name,
+           NULL AS area,
+           waterway AS kind,
+           'openstreetmap.org' AS source,
+           way AS __geometry__,
+           mz_id AS __id__
+
+    FROM planet_osm_line
+
+    WHERE
+        waterway IN ('canal', 'river')
+        AND way && !bbox!
+
 ) AS water_areas
 
 ORDER BY
--- a/queries/water-z16.pgsql
+++ b/queries/water-z16.pgsql
@@ -41,7 +41,7 @@ FROM
 
     SELECT name,
            NULL AS area,
-           COALESCE("waterway", "natural", "landuse") AS kind,
+           waterway AS kind,
            'openstreetmap.org' AS source,
            way AS __geometry__,
            mz_id AS __id__
--- a/queries/water-z16.pgsql
+++ b/queries/water-z16.pgsql
@@ -34,6 +34,24 @@ FROM
         mz_is_water = TRUE
         AND way && !bbox!
 
+    --
+    -- Water line geometries
+    --
+    UNION
+
+    SELECT name,
+           NULL AS area,
+           COALESCE("waterway", "natural", "landuse") AS kind,
+           'openstreetmap.org' AS source,
+           way AS __geometry__,
+           mz_id AS __id__
+
+    FROM planet_osm_line
+
+    WHERE
+        waterway IN ('canal', 'dam', 'ditch', 'drain', 'river', 'stream')
+        AND way && !bbox!
+
 ) AS water_areas
 
 ORDER BY
--- a/queries/water-z9.pgsql
+++ b/queries/water-z9.pgsql
@@ -35,6 +35,24 @@ FROM
         AND way_area::bigint > 102400 -- 4px
         AND mz_way12 && !bbox!
 
+    --
+    -- Water line geometries
+    --
+    UNION
+
+    SELECT name,
+           NULL AS area,
+           waterway AS kind,
+           'openstreetmap.org' AS source,
+           way AS __geometry__,
+           mz_id AS __id__
+
+    FROM planet_osm_line
+
+    WHERE
+        waterway IN ('canal', 'river')
+        AND way && !bbox!
+
 ) AS water_areas
 
 ORDER BY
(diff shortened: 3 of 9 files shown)


queries/water-z11.pgsql
```diff
@@ -35,6 +35,24 @@ FROM
         AND way_area::bigint > 25600 -- 4px
         AND mz_way12 && !bbox!
 
+    --
+    -- Water line geometries
+    --
+    UNION
+
+    SELECT name,
+           NULL AS area,
```
rmarianski: We have access to the area here. Might as well use it for all the queries.

    way_area::bigint AS area

queries/water-z9.pgsql
```diff
@@ -35,6 +35,24 @@ FROM
...
+           NULL AS area,
+           waterway AS kind,
+           'openstreetmap.org' AS source,
+           way AS __geometry__,
+           mz_id AS __id__
+
+    FROM planet_osm_line
+
+    WHERE
+        waterway IN ('canal', 'river')
```
rmarianski: For the other queries, we typically have a filter on the area. The idea is that at the lower zoom levels, like perhaps this one, we wouldn't show waterways unless they were of a certain size. I'm not sure if that makes sense to do here, as it could be that the tag value itself would already provide the sufficient filtering, just pointing it out for consideration.

rmarianski:
Thanks for the pull request!

Like you mentioned, we'll want an index on waterway if it doesn't already exist. I'm thinking a partial index where the waterway is not null, or perhaps where the waterway needs to be one of the values we care about across all queries, the latter being more helpful if there are other values of waterways that we aren't interested in showing. That would go around here:

https://github.com/mapzen/vector-datasource/blob/master/data/apply-planet_osm_line.sql#L28

The source for that function is in functions.sql. Like @bcamper mentioned in the issue though, I'm happy to add that in separately.

queries/water-z11.pgsql
yohanboniface: Humm, but what would be the area of a line? The envelope?

queries/water-z9.pgsql
yohanboniface: Good point, that can filter a bit. The only thing is that, at the opposite of the areas, those lines are very often many pieces of a same river or canal, so we just need to check that we don't create hole in a river stream.
I'll have a look.

rmarianski: Also, are there ever any waterways that we'd want to show that don't have a name set?

If we only want to show those that have names, we can add a filter where `name IS NOT NULL` and include that in the partial index. This would most likely help with index size and performance.

yohanboniface: > Also, are there ever any waterways that we'd want to show that don't have a name set?

I'm not sure about this. I think that generally speaking that would certainly be smart way of reducing data size to only what matters, but I'm afraid about edge cases, specifically when a piece of a river for example would be missing the name, and so we would also miss the line (in the case we render the line geometries). So I'm balanced about this, and I think it's more of an editorial choice here, so I'll let you choose :)

queries/water-z11.pgsql
rmarianski: Good question. We have a column for the area in our tables, but I'm not sure how it's calculated. I was assuming that it would have been the line length, but let me run a query to see if there's actually any helpful data in the database.

If there isn't, then a condition in the where clause won't be helpful either.

queries/water-z9.pgsql
rmarianski: Ah, that's a great point about the holes. I think that might be reason enough not to filter on area, even if we did have useful data in there.

yohanboniface:
I've added a first version of the index for waterway, only filtering on `waterway NOT NULL` for now, I'll update it if we decide to add name condition in the WHERE clause.

queries/water-z11.pgsql
bcamper: Probably least confusing to leave it null rather than mixing areas and lengths, IMO.

queries/water-z16.pgsql
```diff
@@ -34,6 +34,24 @@ FROM
...
         AND way && !bbox!
 
+    --
+    -- Water line geometries
+    --
+    UNION
+
+    SELECT name,
+           NULL AS area,
+           COALESCE("waterway", "natural", "landuse") AS kind,
```
bcamper: If `waterway` is guaranteed to not be null based on the `WHERE` clause, then there's no need for this `COALESCE`, since it will never get past the first column, right?

yohanboniface: Good catch, I've removed the COALESCE from the other zooms. Will fix :)

queries/water-z9.pgsql
bcamper: I don't think we want to filter geometries that don't have a name, as those are still valid to render.

The idea of filtering out very short lines is interesting. For area filters, we are often only filtering very small geometries, so small that they are smaller than 1-2 pixels (an exception is for buildings, we filter a bit more aggressively at zoom 13 and lower since it can pull in millions of geometries in places like NYC). So theoretically if we filtered on lines smaller than 1-pixel at the current zoom, there would not be a hole or other visible effect. I am unsure how common this situation is though, ideally we'd have some data. For now we could leave them in and see if we get cases of tiles with many many small line segments?

A better, more involved approach might be pre-join/simplify the line segments into larger ones at the lower zooms. I believe @kevinkreiser has some experience with this (and may have some code for that handy). Likely outside the scope of this PR, but something to think about for the future.

bcamper:
@yohanboniface thanks for the PR, looks good to me! @rmarianski feel free to merge whenever.

PR #83, opened 2015-01-18
Support more places

yohanboniface:
Adds `island` starting at zoom 9, and `village`, `suburb`, `hamlet` and `isolated_dwelling` starting at 12.

(Not sure about the space/tab usage, I see both in the query files.)

Thanks!

Yohan

### DIFF
--- a/queries/places-z12.pgsql
+++ b/queries/places-z12.pgsql
@@ -11,11 +11,16 @@ WHERE name IS NOT NULL
 AND place IN (
 	'city',
 	'county',
-	'province',
+    'province',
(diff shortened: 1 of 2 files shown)


rmarianski:
Regarding the whitespace, I've been trying to modify queries as I've been been making changes to use spaces. I'll probably convert them all to just use spaces once these latest pull requests are resolved.

This looks good to me, and we already have an index in place that would support this. What do you think @bcamper?

bcamper:
Looks good to me, feel free to merge. Thanks!

Issue #94, opened 2015-04-22
railway=preserved is not included in the vector tiles


mlundblad:
Ways with the tag railway=preserved are not included in the vector tiles.
If I understand how things work, including this tag in the mz_calculate_road_level function to set a minimum zoom level should solve it.

bcamper:
While I know there is an avid historical mapping community in OSM, the aim of these tiles is to provide a simplified, curated view that includes the most common features for everyday mapping. By that criteria, I am not sure that preserved railways qualify. I think those would be more appropriate for a different tile set, possibly focused on historical data. We might be interested in exploring that in the future.

PR #95, opened 2015-04-27
Update road level tags

rmarianski:


bcamper:
I like including `highway=living_street` since it reflects active everyday use, but please see my comment on `railway=preserved` in #94 -- I would prefer to hold off on those for now until we have a better sense of how we would handle historical data in general.

rmarianski:
Ok, will only include `highway=living_street`

PR #101, opened 2015-05-20
query NE 10m datasets for earth/water starting at zoom level 7

sevko:
* add `water-z7` and `earth-z7` queries
  * use them for zoom levels 7 and 8 in `tilestache.cfg`
  * add the `ne_10m_land` dataset (converted to mercator and split into a grid of tiles, since it's initially just a single massive polygon and thus kills intersection query speed)
  * add `data/split_to_tiles.sql`, the SQL script used to partition the `ne_10m_land` dataset into tiles


rmarianski:
Looks good to me, @baldur: any thoughts?

baldur:
:ship:

sevko:
Let's hold off on merging until tomorrow morning; I'd like to spot check the tile size some more.

sevko:
Alright, I've looked the sizes over once more and everything's in order. I'll merge and let @rmarianski take it from here.

PR #102, opened 2015-05-21
fix tiles' internal seams by unifying geometries

sevko:
`st_union()` the Natural Earth geometries in all `earth` queries, and `water` queries only for zoom levels 0, 4, and 7. This should resolve the issue of tiles' "internal" seams, which are caused when a continuous earth/water feature is composed of multiple different polygons at the data level, which, when simplified, slightly diverge away from one another and create a gap. By creating a union of the polygons, we can ensure that simplification will only alter the exterior of that feature.

### DIFF
--- a/queries/earth-z0.pgsql
+++ b/queries/earth-z0.pgsql
@@ -1,6 +1,8 @@
 SELECT
     'base' AS land,
-    the_geom AS __geometry__,
-    gid AS __id__
+    st_union(the_geom) AS __geometry__,
+    max(gid) AS __id__
 FROM
     ne_110m_land
+WHERE
+    the_geom && !bbox!
--- a/queries/earth-z4.pgsql
+++ b/queries/earth-z4.pgsql
@@ -1,6 +1,8 @@
 SELECT
     'base' AS land,
-    the_geom AS __geometry__,
-    gid AS __id__
+    st_union(the_geom) AS __geometry__,
+    max(gid) AS __id__
 FROM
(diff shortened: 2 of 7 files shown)


sevko:
@baldur , @rmarianski , check this out when you get a chance.

rmarianski:
:+1:

bcamper:
Let's give it a whirl in dev.

rmarianski:
I updated the dev environment to contain this change. Next week, we should be able to see the results. If we're satisfied, let's merge this branch into master then.

rmarianski:
This is running on dev at the moment, and low zoom levels have been regenerated.

sevko:
The seams that I spotted appear to be fixed. @bcamper ?

bcamper:
So this sounds expected, but I'm seeing some seams, this one looks like it may be on a tile border. Do we have a plan to address those?

http://tangrams.github.io/tangram/#mapzen-dev,40.63375667842965,-74.17556762695314,12

screen shot 2015-05-28 at 5 19 12 pm

sevko:
Right, seams on tile borders are still an issue. The two images below illustrate the problem pretty well: some tiles have very fine detail near their borders, and when it gets simplified away, the geometry may no longer line up against the edge:

non_simplified_no_seam
simplified_seam

I haven't done any work towards fixing those, yet.

bcamper:
Good to go, will tackle tile seams as a separate issue. Thanks!

PR #104, opened 2015-06-03
Set simplify_before_intersect=true for earth/water.

sevko:
This is necessary for the code-changes that fix tile seams in mapzen/TileStache#34 and mapzen/tilequeue#16 to kick in.


rmarianski:
:+1:

PR #111, opened 2015-07-14
add add-external-data script to data folder to pull in polygons

leblowl:
Internal data script to help prepare the database instead of having to download a gist as it currently describes in the wiki. mapzen/vector-datasource#108

### DIFF
--- a/data/add-external-data.sh
+++ b/data/add-external-data.sh
@@ -1,0 +1,21 @@
+# water polygons
+#
+wget http://data.openstreetmapdata.com/water-polygons-split-3857.zip
+unzip water-polygons-split-3857.zip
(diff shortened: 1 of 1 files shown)


rmarianski:
The other commands in that gist also load the shape files that come from the repo, and make other updates to the database. Should these just be mentioned in the wiki as separate steps? Or should there be a wrapper script that adds the external data and runs these other steps too?

leblowl:
The other commands were already listed in the wiki as separate steps, so the original gist was repeating steps. All commands could be combined into one script - I like the separateness for clarity tho.

rmarianski:
I hadn't realized that you updated the wiki to add those steps explicitly already. Thanks for doing that!

Issue #163, opened 2015-09-03
add missing POIs for office and craft


nvkelso:
Mapzen's SF office doesn't show up because we don't include that feature in tiles – we should.

Same for home service type businesses that are hiding in `craft`.

Issue #164, opened 2015-09-04
Add POI tags for cuisine, sport, and religion


nvkelso:
Religion is needed to draw the correct church icon.

Most restaurants include a cuisine, helpful for refining the restaurant. 

Sport is required to show the appropriate sportful icon.

PR #167, opened 2015-09-04
Add POIs and landuse AOIs

nvkelso:
We were missing craft and office POIs and landuse=military AOIs. We were also missing some special tags on these.


nvkelso:
Fixes #163, #164, #165, and #165

PR #169, opened 2015-09-04
Use adm0cap to provide capital values.

zerebubuth:


queries/places.jinja2
```diff
@@ -9,7 +9,7 @@ SELECT
     scalerank,
     labelrank,
     pop_max::text AS population,
-    NULL AS capital,
+    CASE WHEN adm0cap = 1 THEN 'yes' ELSE 'no' END AS capital,
```
nvkelso: CASE WHEN featurecla = 'Admin-0 capital' THEN 'yes' ELSE 'no' END AS capital,

queries/places.jinja2
```diff
@@ -9,7 +9,7 @@ SELECT
...
     NULL as state_capital,
```
nvkelso: CASE WHEN featurecla = 'Admin-1 capital' THEN 'yes' ELSE 'no' END AS state_capital,

zerebubuth:
Removing in favour of https://github.com/mapzen/TileStache/pull/38

Issue #153, opened 2015-09-02
US:I is tagged as `kind: minor_road`


meetar:
Everything tagged as "network: US:I" also seems to be tagged as "kind: minor_road" – Should probably be `highway`, no?

nvkelso:
Let's remove the duplicate Interstate features and instead apply the network and ref from the relation directly onto the original way feature.

(Only the duplicate features show this bug.)

nvkelso:
Related to #121.

nvkelso:
Example (search for 80, as in I-80): 
http://vector.mapzen.com/osm/all/15/5251/12659.topojson?api_key=vector-tiles-HqUVidw

You'll see two segments, twice. Once as highway, once as minor_road.

rmarianski:
Yea, the relation doesn't have the appropriate `kind` set. We could update the road kind logic, but this should implicitly get fixed when #121 does.

nvkelso:
Fixed in #170.

Issue #121, opened 2015-08-04
Add network attributes to road, transit lines


nvkelso:
We kind of have this in the ref attribute on roads now, but it's not normalized (and the network attribute on the road's route relation is, more often than not).

We need this for:
1. Road shields
2. Improve road connectivity based on highway designation, not just road surface / width classification.

For instance: US 101 goes thru San Francisco, but for 1/2 it's length thru the northern part of the city it's shown only as highway:primary and kind:major_road. The southern 1/2 is marked highway:motorway and kind:highway.

For the road shields it seems like the network attribute is very useful.

To improve road connectivity we might be able to override the kind to show as highway?

nvkelso:
For **road shields**, do we start exporting: tags->'network' as network?

Looks like we do a little **connectivity** testing like this for US Interstates already:
https://github.com/mapzen/vector-datasource/blob/master/data/functions.sql#L96

Could add network_val='US:US' and see where that gets us?

rmarianski:
We already do include the `network` attribute on road features, like you just mentioned. How about changing the query to instead of querying on the `network` tag, instead include `type='route' and route='road'` features starting at z9. And when determining the `kind`, we can identify these relations as `highway`. We'll probably end up returning back duplicate geometries in these cases though, one for the underlying way and another for the relation, but maybe that's what we want to do?

This looks to be the relevant openstreetmap wiki page:

http://wiki.openstreetmap.org/wiki/Interstate_Highway_relations

nvkelso:
Let's try to avoid duplicating line work (lifework, thanks autocorrect). Should we limit this to adding the network_val='US:US' test (like we do now for US Interstates)?

rmarianski:
The duplicate geometries come from including both the ways and the relations. It already started happening when we included `network=US:I` features.

I would expect that querying on type/route would yield more results because according to the wiki page there are more network values besides `US:I` and `US:US`. I'm assuming we would want to include those as well.

rmarianski:
Should we add in the additional relation features for now, and consider how to remove any duplication in the future?

meetar:
Another test case of relation/feature property conflict: the Brooklyn-Battery connection is a relation (https://www.openstreetmap.org/relation/1860086), much of which is a tunnel (https://www.openstreetmap.org/way/5681925) – the roads layer currently includes both features, with the tunnel section properly marked as `is_tunnel : yes` but the whole relation marked as `is_tunnel: no`.

[...]

PR #170, opened 2015-09-04
Project the network tag from the relation.

zerebubuth:
Project the network tag, for some classes of roads, from the relations which contain them. Also, don't return relation geometry. This should mean that the ways get all the relevant network information, and also keep the per-way information like brunnel status.


zerebubuth:
Refs #121.

nvkelso:
Rob, please have a look at this Tuesday.

rmarianski:
:+1: 

Because we're not going to be returning back any relations, seems like we should also get rid of this `network` check in the index, and no longer have the function accept the `network` value. I think that part was only returning relations, and is superseded by this pull request.

https://github.com/mapzen/vector-datasource/blob/project-network-tag-from-relations/data/functions.sql#L115

zerebubuth:
@rmarianski thanks for spotting that! I've added dcd44cb & 666726c which hopefully do what you were suggesting.

rmarianski:
:zap: :+1:

nvkelso:
Closes #153 and #121

Issue #145, opened 2015-08-26
Don't use 110m Natural Earth data (prefer 50m), adjust NE to OSM changeover


nvkelso:
The 50m could be used just as well from [0,5] and 10m from [5,8] and OSM [8,*] for almost all themes.

Once simplification is turned off, we should verify this hypothesis. VERIFIED.

nvkelso:
Turning off Natural Earth simplification is covered in #152. That's done (but needs verification).

nvkelso:
Now that #152 is done, Natural Earth looks better. But it's still too generalized at some zooms so we **do** want to proceed with fixing this issue by updating the changeover for various input datasets.

zerebubuth:
I've made the changes described above; 50m from [0,5), 10m from [5, 9), then OSM. This results in the very low zooms (0,1,2) increasing in size by a factor of between 4 and 8. I'd suggest we keep the 110m for zoom <= 2. How does that sound?

nvkelso:
Were they super small before? How large are they (before, after) compared to a zoom 15 tile in London?

nvkelso:
Another option is to simplify the 50m at zooms(0,1,2) instead of using 110m data (which is just too generalized for this purpose).

zerebubuth:
A zoom 15 tile over Manhattan ([15/9647/12321](http://a.tile.openstreetmap.org/15/9647/12321.png) to be precise) is 73kB for the topojson and 24kB for the MVT. For comparison, that PNG tile is 18kB. Looking at the size of the zoom (0, 1, 2) tiles, showing the size of the file using 110m and 50m NE data in terms of the size of the zoom 15 tile, and the ratio of 50m size to 110m size, for all layers at that zoom:

```
tile   110m   50m     ratio
0/0/0  83.3%  356.7%  4.3x
1/0/0  35.9%  158.0%  4.4x
1/0/1  10.6%   39.2%  3.7x
1/1/0  54.6%  171.9%  3.2x
1/1/1  16.9%   54.2%  3.2x
2/0/0   8.8%   19.7%  2.2x
2/0/1  14.2%   67.8%  4.8x
2/0/2   1.3%    7.4%  5.7x
2/0/3   3.8%    8.3%  2.2x
2/1/0   7.0%   20.4%  2.9x
2/1/1  28.3%   94.2%  3.3x
2/1/2  10.7%   27.0%  2.5x
2/1/3   4.2%   11.2%  2.7x
[...]

PR #172, opened 2015-09-04
Tweak NE transition zooms

zerebubuth:
For layers containing 110m NE data, use 110m only at zoom 2 and below. Transition from 50m to 10m at soom 5.

Refs #145.


nvkelso:
LGTM

Issue #154, opened 2015-09-02
Explicit landuse sort order


zerebubuth:
Add a server-generated parameter for sort order based on the existing sort keys in the stylesheet for landuse and water. This will help simplify the stylesheet and can be used to decide which landuse polygon "wins" for the purposes of projecting its attributes onto intercut features (see #136).

rmarianski:
:+1: Feel free to merge in.

I put in a minor comment https://github.com/mapzen/TileStache/pull/39#discussion-diff-38975673

zerebubuth:
Thanks! Merged.

PR #173, opened 2015-09-07
Use landuse sort key filter for landuse & landuse label layers.

zerebubuth:
Refs #154. Depends on mapzen/TileStache#39.


rmarianski:
:+1:

Issue #136, opened 2015-08-17
Add landuse kind to roads


nvkelso:
When a road intersects a landuse polygon, split the road up and tag the segment inside the landuse polygon with the polygon's landuse kind property. 

Light grey roads are hard to see crossing a park, and white casings that look good over generic earth background pop too much over green parks.

bcamper:
We'd do this server-side by splitting the lines? Sounds cool.

nvkelso:
Yes, there'd need to be an extra (PostGIS) processing step.

zerebubuth:
When splitting a road up into parts which intersect and parts which don't, should the feature ID of the two parts be the same, or different? If different, how should we allocate new IDs -- do they have to be unique?

rmarianski:
I'll defer to @nvkelso, but my understanding is that we should use the same feature id.

nvkelso:
Let's keep the same OSM feature ID for the new feature parts as the original feature. Because of tile boundaries, people already have to deal with multiple features in view with the same ID, and if someone needs to isolate / optimize for the bits they can create a effective ID by indexing on ID and kind?

rmarianski:
@zerebubuth All looks good to me :zap:. I put in a couple of comments/questions on this pr:  mapzen/TileStache#37. Circle complained about a couple of pep8 issues on mapzen/tilequeue#22, but that should be easy to fix.

We might also want to apply different "cutting" strategies too. In particular, the intercutting with the buildings layer might have some other logic. But maybe we should just address that when working on that issue (#142) and not worry about it now.

Do you think that the n^2 intersections approach might be "fast enough"? I'm wondering if we can get away with this approach for the time being, or if we'll need to worry about that sooner rather than later. Might be worth measuring the impact, ideally on a tile that has lots of landuse and road features.

Btw, do you have a pr for the config updates to queries.yaml in vector-datasource?

zerebubuth:
@rmarianski Thanks for the review! I'll fix/answer those things now.

I'm not sure I quite follow what we'd want to do differently with the buildings layer. Do you mean we might want to "cut" against only one background polygon, so as not to actually split the feature?

I really don't know whether the O(N^2) will be fast enough - my gut feeling is that it'll be okay on most tiles, but there's always bound to be a few nasty corner cases. The question is whether it's worth mitigating the corner cases by introducing an indexing step which might slow down the average case.
I'll take some measurements and see if I can quantify what the overhead is, so that we can make an informed decision.

[...]

PR #183, opened 2015-09-09
Add config to enable intercut of landuse and roads.

zerebubuth:
This will enable the code, already merged, for intercut and apply the landuse polygon's kind tag as "landuse_kind" on the roads. The roads will be split at landuse polygon boundaries.

Refs #136.

I almost forgot about this in the confusion. Just the config update to enable intercutting.


zerebubuth:
@rmarianski, @nvkelso: are we good to merge this and get it running in dev?

rmarianski:
:+1:

PR #184, opened 2015-09-10
Handle the case where tags is NULL.

zerebubuth:
Ooops. Should have handled this the first time around, but didn't check whether `tags` was NULLable. Turns out it is, at least on my osm2pgsql database, so `mz_rel_get_tag()` needs to not explode when it encounters one.


rmarianski:
Nice catch! :+1:

Issue #140, opened 2015-08-19
Add explicate coastline (ocean, lake, * water) to tiles


nvkelso:
We need to stroke the water coastlines directly.

Right now stroking the existing water polygons doesn't work because that adds blue lines at tile boundaries / feature chunks.

bcamper:
To clarify, in Tangram, that should not add lines at tile boundaries (unless explicitly turned on in the stylesheet), however, it **does** add lines where different water polygons are joined. So yeah, probably necessary for that case.

rmarianski:
Natural earth coastline data ok to use here?

* [0, 4) -> 110m
* [4, 7) -> 50m
* [7, *) -> 10m

And should simplification just be turned off completely here for all zoom levels?

nvkelso:
Yes, we shouldn't simply Natural Earth at all (it's pre simplified). 

I'm wondering if we should be using the 110m data at all – I think the 50m could be used just as well from [0,5] and 10m from [5,7] and OSM [8,*] for almost all themes (including coastline). Once simplification is turned off, we should verify this hypothesis. 

- For Natural Earth there's an explicate Coastline geometry to source from.
- For OSM looks like the equivalent is: natural:coastline.

But... this doesn't necessarily solve the problem because waterway:riverbank and natural:coastline still meet at overlaps like: https://www.openstreetmap.org/edit?#map=17/40.74168/-74.12203 in New Jersey. 

I wonder if we can take the linestrings from each of OSM sources (coastline, lakeshore & etc), and then take the Symmetric Difference to remove the overlapping parts?

- http://toblerity.org/shapely/manual.html#object.symmetric_difference
- http://postgis.net/docs/ST_SymDifference.html

nvkelso:
Other areas like this one in California show that water polygons will overlap sometimes, so we'd need to erase the overlapping bits of the polygons, then do the symmetric difference of their boundary linestring.

https://www.openstreetmap.org/edit#map=17/39.46597/-121.59752

rmarianski:
> I'm wondering if we should be using the 110m data at all – I think the 50m could be used just as well from [0,5] and 10m from [5,7] and OSM [8,*] for almost all themes (including coastline). Once simplification is turned off, we should verify this hypothesis. 

Generally speaking, it's faster to query against the 110m datasets. We can run a test to verify. I think we would also want to align with the intervals that we use in the water/earth queries, which are at the steps that I mentioned above.

[...]

PR #185, opened 2015-09-10
Add explicit water boundaries

zerebubuth:
This is the config part, enabling the changes in mapzen/TileStache#43.

Refs #140.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -9,6 +9,7 @@ all:
   - pois
   - boundaries
   - transit
+  - water_boundaries
 layers:
   water:
     template: water.jinja2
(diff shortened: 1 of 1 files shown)


data/functions.sql
```diff
@@ -174,6 +174,11 @@ DECLARE
   lo CONSTANT integer := array_lower(tags, 1);
   hi CONSTANT integer := array_upper(tags, 1);
 BEGIN
+  -- if tags is null, then we're not going to find any values
+  -- in it, so just short-circuit and return NULL.
+  IF tags IS NULL THEN
+    RETURN NULL;
+  END IF;
```
rmarianski: Guessing we don't need this if we rebase onto master, yes?

rmarianski:
:+1:

data/functions.sql
zerebubuth: Yup, looks like I'm getting confused with my working copy again... Too many branches! :confused: Sorry!

Issue #174, opened 2015-09-08
San Francisco place exists twice


gronke:
San Francisco exists twice in Zoom-Level 14 with slightly different coordinates.

http://vector.mapzen.com/osm/places/14/2620/6333.json
```json
{
    "features": [
        {
            "geometry": {
                "coordinates": [
                    -122.417169,
                    37.769196
                ],
                "type": "Point"
            },
            "properties": {
                "id": 7255,
                "kind": "Populated place",
                "labelrank": 1,
                "name": "San Francisco",
                "population": "3450000",
                "scalerank": 1,
                "source": "naturalearthdata.com"
            },
            "type": "Feature"
        },
        ...
    ],
    "type": "FeatureCollection"
}
```

http://vector.mapzen.com/osm/places/14/2620/6332.json
```json
{
    "features": [
        {
            "geometry": {
                "coordinates": [
                    -122.41927,
                    37.779277
                ],
                "type": "Point"
            },
            "properties": {
                "id": 26819236,
                "kind": "city",
                "name": "San Francisco",
                "name:hi": "\u0938\u0948\u0928 \u092b\u094d\u0930\u093e\u0902\u0938\u093f\u0938\u094d\u0915\u094b",
                "name:ja": "\u30b5\u30f3\u30d5\u30e9\u30f3\u30b7\u30b9\u30b3",
                "name:kn": "\u0cb8\u0cbe\u0ca8\u0ccd \u0cab\u0ccd\u0cb0\u0cbe\u0ca8\u0ccd\u0cb8\u0cbf\u0cb8\u0ccd\u0c95\u0cca",
                "name:pt": "S\u00e3o Francisco",
                "name:ru": "\u0421\u0430\u043d-\u0424\u0440\u0430\u043d\u0446\u0438\u0441\u043a\u043e",
                "name:ta": "\u0bb8\u0bbe\u0ba9\u0bcd \u0baa\u0bcd\u00b2\u0bb0\u0bbe\u0ba9\u0bcd\u0bb8\u0bbf\u0bb8\u0bcd\u0b95\u0bca",
                "name:zh": "\u820a\u91d1\u5c71\u5e02\u8207\u7e23",
                "population": "815358",
                "short_name": "SF",
                "source": "openstreetmap"
            },
            "type": "Feature"
        },
      ...
    ],
    "type": "FeatureCollection"
}
```

zerebubuth:
[...]

PR #190, opened 2015-09-10
Switch from NE to OSM between zooms 8 and 9.

zerebubuth:
Refs #174. Change from `UNION` of the two datasets to choosing one or the other based on zoom.


nvkelso:
LGTM

The same as you did with the water edges issue, can you start including before / after screenshots in the PR?

Issue #181, opened 2015-09-08
Add addr_housenumber points


nvkelso:
We're only capturing a minority of addresses from OSM when they are attached to a building shape. Most OSM addresses are separate points with the addr_housenumber tag. These should probably go into the buildings layer.

nvkelso:
@bcamper will including addr_housenumber points in the building layer cause any downstream problems in Tangram?

bcamper:
Nope! Just makes the tiles a bit heavier, but better to have the data IMO.

zerebubuth:
Do we want to try to interpolate house number lines as well?

nvkelso:
I tend to think no, but tell me more about what the result would look like. Are the features generated guaranteed to match ground truth in terms of position & count (would there be more of them generated than there are in fact)?

zerebubuth:
I don't think the individual addr:housenumber points are guaranteed to match ground truth in terms of position and count - there will be plenty which are out of position or missing, especially if they were imported.

As for interpolation lines; it depends where they came from. The ones I've entered have always been "full" and probably accurate enough (assuming buildings in a terrace are equal width, which they approximately are). Ones which have come in as block interpolations (e.g: from TIGER) will probably run the full X00-X99 regardless of how many houses are there.

We have a few options:

1. Don't include any interpolated stuff.
2. Include the interpolation lines.
3. Break the interpolation lines into points.
4. Break the interpolation lines into points, but mark them as having been interpolated.

I agree that the TIGER-style block interpolations are pretty useless from a display standpoint, but would think that they might come in handy for other users of the data, particularly delivery-oriented apps.

nvkelso:
Yeah, let's skip the line interpolation for now. Side note: Pelias doesn't do any line interpolation for addresses now, so we're not the odd kid out here. I fee like routing to an interpolated address is fine, but I'm nervous about showing them on a map if they are likely to not be in the right spot.

The PR was merged. Closing this one out.

PR #193, opened 2015-09-10
Add address points to buildings layer.

zerebubuth:
Refs #181.

Adds address point features to buildings layer. At present, doesn't do address line interpolation, nor expose the address interpolation lines.

Before:

address_points_before

After:

address_points_after


nvkelso:
Cool! 

/cc @rmarianski 
Do we need to be selecting all the extra attributes in `queries/buildings.jinja2`? I'm not opposed to some/all of them, but should we collapse some of them into `landuse_kind` like we did for the roads intercut?

rmarianski:
That's a good point. Are you thinking of coalescing the `amenity`, `shop`, and `tourism` properties? The others look significant on their own right to me.

https://github.com/mapzen/vector-datasource/blob/5fb743a6f4e957e1c128bbca0e529026198b45e3/queries/buildings.jinja2#L2-L26

nvkelso:
Are the total list of exported attributes on address and building features comparable? Should they be? 

I would expect there to be *more* attributes on buildings than on addresses. For instance, it doesn't seem worthwhile to export building shape attributes on the address points at all. And we should coalesce the typical list of keys into the generic kinds namespace to be consistent with our other feature classes.

zerebubuth:
As I understand it, each jinja2 template is a single query, and the selected attributes have to be the same across all `union`ed sub-queries. Have I misunderstood? Are you suggesting dropping some attributes in the processing stages?

I would generally expect to put address points in a separate layer from buildings. The geometry (point / polygon) is completely different, the attributes are pretty different, and the style applied to them is going to be pretty different.

rmarianski:
> As I understand it, each jinja2 template is a single query, and the selected attributes have to be the same across all unioned sub-queries. Have I misunderstood? Are you suggesting dropping some attributes in the processing stages?

We usually handle that by unioning the list of properties, and just setting them to null in the query itself for the query where it's irrelevant. We drop all properties where the value is null pretty early on in the pipeline.

nvkelso:
Here's an example of the nulling Rob mentioned from the places layer:

https://github.com/mapzen/vector-datasource/blob/master/queries/places.jinja2#L9-L15

zerebubuth:
@nvkelso I've dropped it down to only selecting `addr_housenumber`, `addr_street`, `name`, `id` and `amenity`, `shop` and `tourism` -- these last 3 are almost always NULL anyway, and it might make it easier to show a proper icon for these if they're on the address. Review, please?

nvkelso:
LGTM. 

We're already not collapsing the `amenity`, `shop`, `tourism` keys into a single `kind`, but that doesn't need to be fixed in this issue.

nvkelso:
See also: 

* https://github.com/mapzen/vector-datasource/issues/200

PR #196, opened 2015-09-11
Config update to take advantage of new boundaries code.

zerebubuth:
Depends on mapzen/TileStache#46.

Added example screenshot in parent #140.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -159,4 +159,14 @@ post_process:
     target_attribute: landuse_kind
   TileStache.Goodies.VecTiles.transform.exterior_boundaries:
     base_layer: water
-    new_layer_name: water_boundaries
+    prop_transform:
+      kind:
+        ocean: shoreline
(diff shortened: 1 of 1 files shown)


nvkelso:
In keeping with the manta `less layers, more stuff in the layers`, I think these should end up in the same `water` layer they source from. 

I'll have a separate comment about the `kind` transformations.

zerebubuth:
The new features stay in the `water` layer, since I removed the `new_layer_name` parameter. See mapzen/TileStache#46.

nvkelso:
- [x] List of features matches those we test for in [mz_calculate_is_water](https://github.com/mapzen/vector-datasource/blob/e29bf15a68370ac6da093449e99e157fa474c042/data/functions.sql#L130-L140). 

- [x] Result goes back into the water layer.

- [x] We add a new `boundary=yes` property.

Issue #198, opened 2015-09-11
Export landuse sort order in tiles


nvkelso:
We're calculating a landuse sort order on the server for the intercut features. Let's expose that downstream so the stylesheet can make use of it, too. 

See also:
* https://github.com/mapzen/vector-datasource/issues/154
* https://github.com/mapzen/vector-datasource/pull/173

nvkelso:
See related discussion in: https://github.com/mapzen/vector-datasource/issues/186

nvkelso:
This needs to match the `sort_key` property used on roads.

Issue #142, opened 2015-08-20
Add landuse kind to buildings


nvkelso:
Like we are planning for roads: https://github.com/mapzen/vector-datasource/issues/136

rmarianski:
This one might be a little bit trickier, because currently we don't clip buildings. If a building is completely contained within a landuse polygon it's fine, but if it's just a portion of it, what should we do?

nvkelso:
Thinking out loud:

- If 100% of a building is contained within a landuse polygon, **do get** the landuse attribute.
- If 80% of a building is contained within a landuse polygon, **do get** the landuse attribute.
- If 60% of a building is contained within a landuse polygon, **don't get** the landuse attribute.
- If 10% of a building is contained within a landuse polygon, **don't get** the landuse attribute.

80% seems like a good fuzzy containment. The linework can be a little squishy (not line up perfect) but it'll still match. What do you think?

How does this work for buildings that extend past tile edges? We only know post processing for the part in the tile that has overlap with landuse? Do we attempt to resolve the whole shape?

rmarianski:
> 80% seems like a good fuzzy containment. The linework can be a little squishy (not line up perfect) but it'll still match. What do you think?

Sounds reasonable to me.

> How does this work for buildings that extend past tile edges? We only know post processing for the part in the tile that has overlap with landuse? Do we attempt to resolve the whole shape?

Yes, I was assuming that we would use the whole shape.

zerebubuth:
I couldn't seem to get this to work / display properly in "skin-and-bones", but here's an example showing it in "simple-demo". Before:

![building_landuse_kind_before](https://cloud.githubusercontent.com/assets/271360/9814806/cbd937d8-5888-11e5-9bcb-e58444e31776.png)

After mapzen/TileStache#45:

![building_landuse_kind_after](https://cloud.githubusercontent.com/assets/271360/9814826/fe4d1414-5888-11e5-9771-ef9b58eeefbf.png)

Note that the purple buildings are theatres, even when they're in the park, because we're pulling in the building polygon as an "amenity" landuse. To me, this seems a bit weird, and we might want to see if we can be a bit more precise about what "landuse" actually means.

The scene file fragment for the above is:

```yaml
    buildings:
        data: { source: osm }
        draw:
            polygons:
                order: 6
                color: '#404040'
        in_park:
          filter: { landuse_kind: park }
          draw:
            polygons:
              color: '#608040'
        in_garden:
[...]

PR #207, opened 2015-09-14
Add sorting to landuse intercut.

zerebubuth:
Refs #198.

Also adds missing config from #142, also updated to do the sorting based on the existing `order` parameter.


rmarianski:
:+1:

Issue #203, opened 2015-09-11
Source water boundaries from Natural Earth for low zooms


nvkelso:
Natural Earth includes prebuilt ocean coastlines, let's use those and reuse the lakes as lake boundaries instead of spending time calculating them.

nvkelso:
See also:

* https://github.com/mapzen/vector-datasource/issues/208

nvkelso:
Fixed in: https://github.com/mapzen/vector-datasource/pull/209

PR #209, opened 2015-09-14
Add natural earth coastlines

zerebubuth:
Refs #203.

Add Natural Earth coastline shapefiles to the repository, update scripts to import them and the queries to fetch them. This requires some code changes, see mapzen/tilequeue#28 and mapzen/TileStache#49.

### DIFF
--- a/queries/water.jinja2
+++ b/queries/water.jinja2
@@ -3,7 +3,8 @@
     '{{ kind }}' AS kind,
     'naturalearthdata.com' AS source,
     {% filter geometry %}st_union(the_geom){% endfilter %} AS __geometry__,
-    max(gid) AS __id__
+    max(gid) AS __id__,
+    NULL AS boundary
 {% endmacro %}
 
 {% if zoom < 9 %}
@@ -37,6 +38,28 @@ FROM ne_10m_lakes
 
 WHERE {{ bounds|bbox_filter('the_geom') }}
 GROUP BY name
+
+UNION
+
+SELECT
+    NULL AS name,
+    NULL AS area,
+    'shoreline' AS kind,
+    'naturalearthdata.com' AS source,
+    the_geom AS __geometry__,
+    gid AS __id__,
+    'yes' AS boundary
+
+FROM
+{% if zoom < 2 %}
+    ne_110m_coastline
+{% elif 2 <= zoom < 5 %}
+    ne_50m_coastline
+{% elif 5 <= zoom < 9 %}
+    ne_10m_coastline
+{% endif %}
+
+WHERE {{ bounds|bbox_filter('the_geom') }}
 
 {% if 4 <= zoom < 9 %}
 UNION
--- a/queries/water.jinja2
+++ b/queries/water.jinja2
@@ -90,9 +90,9 @@ SELECT
     name,
     {{ ne_water_cols('playa') }}
 
-{% if 4 <= zoom < 7 %}
+{% if 4 <= zoom < 5 %}
 FROM ne_50m_playas
-{% elif 7 <= zoom < 9 %}
+{% elif 5 <= zoom < 9 %}
 FROM ne_10m_playas
 {% endif %}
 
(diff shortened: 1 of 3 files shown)


nvkelso:
We should also reference the Natural Earth lake polygons already loaded and convert them to boundaries in the query, too?

zerebubuth:
Thanks, @nvkelo, I completely forgot about lakes and playas. Added those in 06bcfe8. See the screenshot below for how this looks with a very simple style at zoom 8 (all NE).

water_boundaries_z8

In the upper left you can see the shores of lake Erie and some of the minor lakes which surround it. Here's how NYC looks zoomed in a bit more to 10 (mix of NE & OSM data):

water_boundaries_z10

I'm not sure whether the detail visible in the landuse areas and all the small lakes means we should look at reducing detail at this zoom, but that's a matter for another issue.

zerebubuth:
Note: the cut across the Hudson to the north of that last screenshot doesn't seem to be the same issue as the previous one. I'm investigating further.

zerebubuth:
Okay, the bug above is orthogonal to this PR, and should be fixed in mapzen/TileStache#50.

@nvkelso how's it looking?

nvkelso:
The new 1:10,000,000 coastline file (merc) has an error linking the Siberian coastline from the far right side of the map to the far left side. The offending vertex on the far left side should be deleted.

screen shot 2015-09-15 at 07 55 08

queries/water.jinja2
```diff
@@ -54,6 +99,26 @@ FROM ne_10m_playas
...
+    NULL AS name,
+    NULL AS area,
+    'playa' AS kind,
+    'naturalearthdata.com' AS source,
+    st_boundary(the_geom) AS __geometry__,
+    gid AS __id__,
+    'yes' AS boundary
+
+FROM
+{% if 4 <= zoom < 7 %}
```
nvkelso: We should be consistent with the other water levels of detail: 
{% if 2 <= zoom < 5 %}

queries/water.jinja2
```diff
@@ -54,6 +99,26 @@ FROM ne_10m_playas
...
+    ne_50m_playas
+{% elif 7 <= zoom < 9 %}
```
nvkelso: We should be consistent with the other water levels of detail: 
{% elif 5 <= zoom < 9 %}

queries/water.jinja2
zerebubuth: Yup, well spotted, thanks! Looks like the _playas_ only kick in at zoom 4, though. Otherwise, fixed in 65f727d.

nvkelso:
> Here's how NYC looks zoomed in a bit more to 10 (mix of NE & OSM data):

@zerebubuth: Can you explain this more, why are we seeing a mix of Natural Earth and OSM data at zoom 10 – shouldn't that be OSM only?

nvkelso:
Once the 1:10m coastline file has been fixed, I think this is ready to go.

zerebubuth:
> Can you explain this more, why are we seeing a mix of Natural Earth and OSM data at zoom 10 – shouldn't that be OSM only?

@nvkelso, you're right - I just saw `source != "openstreetmap.org"` and thought it was NE, but it's actually OSM we're using for the ocean tiles. Forget I said that about the mix of NE & OSM.

rmarianski:
This looks good to me too :zap:

zerebubuth:
> The new 1:10,000,000 coastline file (merc) has an error linking the Siberian coastline from the far right side of the map to the far left side. The offending vertex on the far left side should be deleted.

Hmm... that came straight from `ogr2ogr`. Must have been a precision / wraparound error. Fixed in 00429901 by removing all the new segments it had created by intersecting with the existing coastline.

nvkelso:
@zerebubuth I'm still seeing the rogue coastline in the SHP in the new ZIP. Can you include a screenshot here showing your fix?

nvkelso:
Adding issue to reducing LOD at some zooms (less water, less landuse):

* https://github.com/mapzen/vector-datasource/issues/213

zerebubuth:
@nvkelso Here's the screenshot. Not sure why that's useful, though - if you're seeing the issue still, with the latest version of the file, then we need to figure out why you're seeing it and I'm not.

fixed_dateline_line

nvkelso:
Weird. Everything checks out today, nothing more to see here :)

Issue #180, opened 2015-09-08
OSM boundary funk (ways, relations)


nvkelso:
Reviewing this tile along the California and Oregon boundary it looks like we're returning relations (note the negative OSM ids) and ways.

http://vector.mapzen.com/osm/all/10/159/380.topojson?api_key=vector-tiles-HqUVidw

But the ways (mostly) don't have names! 

We should generate a "Left - Right" synthetic name for basic labeling, and an explicate name:left and name:right pair for advanced labeling on the ways, and stop exporting the relations. Bonus points for preserving the localized names.

nvkelso:
Related: https://github.com/tangrams/tangram/pull/177

zerebubuth:
After mapzen/TileStache#51 and #216, we have data like this. For example, just looking at the name attribute:

![boundaries_names_inline](https://cloud.githubusercontent.com/assets/271360/9909130/c48e8d3e-5c8e-11e5-9415-36f900798ad8.png)

Or, using the translations in the `left:name:*` and `right:name:*`:

![boundaries_left_right_translation](https://cloud.githubusercontent.com/assets/271360/9909152/d58b211a-5c8e-11e5-93fa-68cc2cab7cf0.png)

Using the following scene file fragment:

```yaml
    boundaries:
      data: { source: osm }
      draw:
        lines:
          order: 4
          color: '#808080'
          width: 2px
      admin_4:
	filter: { admin_level: '4' }
        draw:
          lines:
            width: 6px
          text:
            text_source: |
              function() {                                                                  
                return feature['left:name:el'] + "/" + feature['right:name:el']             
              }                                                                             
            font:
              fill: '#404040'
              typeface: 500 18px Helvetica
              stroke: { color: '#88bbee', width: 4 }
```

zerebubuth:
@nvkelso could you review, please?

nvkelso:
@zerebubuth In #216 it looks like you added the Natural Earth selection for the left and right handed names to the **50m** selection – it should be on the **10m** select instead.

The Natural Earth 1:50m data should have this data, but doesn't. I've filed a bug on that over there: 

* https://github.com/nvkelso/natural-earth-vector/issues/146

nvkelso:
Should we store the left and right sided names as `name:left` and `name:right` and their star variants? That keeps the tags closer together in alpha sorted name space. Is there a downside to this?

zerebubuth:
We've got the names as infix now - see #218 for a JSON sample.

[...]

PR #216, opened 2015-09-16
Boundary left / right names

zerebubuth:
Query for geometry and attributes both sides of a boundary.

WARNING: This might have a profound change on what data is available in the tiles.

Previously, the boundaries came from the lines table, where they were hard to link back to the relation via the "rels" table, and didn't provide easy left/right distinction without intensive efforts to reconstruct the whole boundary topology.

Instead, this works around the issue by extracting the admin area polygons at matching `admin_level`s, ensures that the boundary of one is correctly oriented and uses the shared portions of boundary to reconstruct the original. This _should_ work, as long as it is done before any simplification steps.

For example, just looking at the name attribute:

boundaries_names_inline

Or, using the translations in the `left:name:*` and `right:name:*`:

boundaries_left_right_translation

Using the following scene file fragment:

```yaml
    boundaries:
      data: { source: osm }
      draw:
        lines:
          order: 4
          color: '#808080'
          width: 2px
      admin_4:
	filter: { admin_level: '4' }
        draw:
          lines:
            width: 6px
          text:
            text_source: |
              function() {                                                                  
                return feature['left:name:el'] + "/" + feature['right:name:el']             
              }                                                                             
            font:
              fill: '#404040'
              typeface: 500 18px Helvetica
              stroke: { color: '#88bbee', width: 4 }
```

Refs #180 & requires mapzen/TileStache#51.

### DIFF
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -57,16 +57,37 @@ WHERE
 
 {% else %}
 
+WITH admin_polygons AS (
+  SELECT *
+  FROM planet_osm_polygon
+  WHERE
+    {{ bounds|bbox_filter('way') }}
+  AND boundary='administrative'
+)
+
 SELECT
-    osm_id AS __id__,
-    {% filter geometry %}way{% endfilter %} AS __geometry__,
-    tags->'border_type' AS kind,
-    name,
-    %#tags AS tags
+    rhs.osm_id AS __id__,
+-- this is taking the right hand polygon, forcing the direction of its
+-- boundary to be clockwise (i.e: polygon on the right). this makes
+-- sure we have a consistent idea of what is to the right. it then
+-- tries to intersect that with each of the left-hand polygons (now
+-- direction doesn't matter, as it should keep the first operand's
+-- direction). it then merges the result to try and keep a smaller
+-- number of line segments.
+    {% filter geometry %}st_linemerge(st_intersection(st_boundary(st_forcerhr(rhs.way)), st_boundary(lhs.way))){% endfilter %} AS __geometry__,
+    coalesce(rhs.tags->'border_type', lhs.tags->'border_type') AS kind,
+    rhs.admin_level as admin_level,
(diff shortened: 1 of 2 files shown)


nvkelso:
Let's also include the name_l and name_r data from the 10m Natural Earth data (using the OSM name:left, name:right property names already in this PR).

zerebubuth:
> Let's also include the name_l and name_r data from the 10m Natural Earth data (using the OSM name:left, name:right property names already in this PR).

There's no `name:left` or `name:right` on these from OSM. The query synthesizes a `lhs_name` and `rhs_name` attribute from the left and right polygon `name` attributes, and adds the translations in `left:name:*` and `right:name:*`.

I've added the NE 10m admin-1 equivalents in 0e6a4cd. I didn't see `name_l` or `name_r` attributes in the other boundaries shapefiles, though.

queries/boundaries.jinja2
```diff
@@ -57,16 +59,37 @@ WHERE
...
-    %#tags AS tags
+    rhs.osm_id AS __id__,
+-- this is taking the right hand polygon, forcing the direction of its
+-- boundary to be clockwise (i.e: polygon on the right). this makes
+-- sure we have a consistent idea of what is to the right. it then
+-- tries to intersect that with each of the left-hand polygons (now
+-- direction doesn't matter, as it should keep the first operand's
+-- direction). it then merges the result to try and keep a smaller
+-- number of line segments.
+    {% filter geometry %}st_linemerge(st_intersection(st_boundary(st_forcerhr(rhs.way)), st_boundary(lhs.way))){% endfilter %} AS __geometry__,
```
rmarianski: :zap:

rmarianski:
:+1:

nvkelso:
@zerebubuth It looks like you added the Natural Earth selection for the left and right handed names to the **50m** selection – it should be on the **10m** select instead.

The Natural Earth 1:50m data should have this data, but doesn't. I've filed a bug on that over there: 

* https://github.com/nvkelso/natural-earth-vector/issues/146

Issue #195, opened 2015-09-11
Include feature area in POI zoom calculation


nvkelso:
Today we show *all* features from a kind class at the same initial zoom, but this results in a boring and busy, poorly balanced map. 

For instance, we start showing hospitals first, but we end up showing a mix a large & important hospitals and small and less important clinics at the same initial zoom. Instead only the large & important hospitals should be shown at first, and then the smaller, less important clinics should be added progressively in later zooms.

Many POIs are technically labeling landuse AOIs and building featuers and we can leverage the parent's area to throttle the POIs zoom (first adopting the same landuse / building area thresholds and then refining them to be better for POIs – for instance, we see some delta between landuse area, landuse-labels area, and building area for calculating label visibility today.

nvkelso:
I'm seeing some more POIs, earlier (good), but we need more and earlier – at least in San Francisco. Let's try adjusting the area thresholds &/or the zoom offsets in the earlier #222 PR.

zerebubuth:
Here's some stuff I've pulled out. Note that the text is all busy because I've enabled a text label for _every single_ point, so you can see everything that's in the data. I've also disabled all the landuse areas, because I haven't merged the logic for those yet.

![195-sf-z12](https://cloud.githubusercontent.com/assets/271360/10080159/e60b55b6-62e5-11e5-90f2-16a72b164db4.png)

![195-sf-z13](https://cloud.githubusercontent.com/assets/271360/10080162/ea03c14e-62e5-11e5-9da5-0a690b2fc71b.png)

![195-sf-z14](https://cloud.githubusercontent.com/assets/271360/10080165/ec5c046a-62e5-11e5-9abf-1bf8ec1a81fc.png)

zerebubuth:
@nvkelso, it would be really helpful if you could give me some feedback on the images above and whether stuff is coming in too early, too late, or the wrong stuff is visible, etc... Just bear in mind that it doesn't include any of the landuse area centroids - I'm working on that and hopefully will have some screenshots of that soon.

nvkelso:
Generally this is a huge improvement, thank you! 

Should we roll this out incrementally? (What you have already seems good enough to move to prod if you think the code is good enough.)

I've marked up the images. Here's the summary, images down farther.

**Please make these types of features visible earlier**, as annotated on the images:

1. Big churches
2. Museums
3. Convention centers
4. Rail stations (earlier than subway stations)
5. Subway stations (they are often underground)
6. Large shopping centers
[...]

PR #222, opened 2015-09-18
Extract POIs from polygons and filter zoom by area.

zerebubuth:
Previously, the POIs had just been coming from the `planet_osm_point` table, but some POIs are represented as polygons. These are included via their centroid, and the same test function `mz_calculate_poi_level`.

That function has been modified to adjust the zoom based on the feature's area. Note that this will probably mean we want to take another look at the zoom levels that some of these get assigned in the function and adjust based on the distribution of areas for those feature types.

Finally, the `mz_calculate_poi_level` function gets called almost a million times for a mid-zoom tile and, even though it's pretty quick, that's just too slow. The POI `min_zoom` is reified as a column in both the points and polygons tables and kept up to date with a trigger (yuck...). That brings the query time for a mid-zoom tile down from 4s to about 5ms.

Refs #195.

### DIFF
--- a/data/apply-planet_osm_point.sql
+++ b/data/apply-planet_osm_point.sql
@@ -6,7 +6,9 @@ BEGIN
 --------------------------------------------------------------------------------
 
 CREATE INDEX planet_osm_point_place_index ON planet_osm_point(place) WHERE name IS NOT NULL AND place IN ('borough', 'city', 'continent', 'country', 'county', 'district', 'farm', 'hamlet', 'island', 'isolated_dwelling', 'lake', 'locality', 'neighbourhood', 'ocean', 'province', 'quarter', 'sea', 'state', 'suburb', 'town', 'village');
-CREATE INDEX planet_osm_point_level_index ON planet_osm_point(mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real)) WHERE mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real) IS NOT NULL;
+-- NOTE: removed index on mz_poi_min_zoom, possibly will replace with
+-- a partial index at some point in the future, depends on how the
+-- query performs.
 
 ALTER TABLE planet_osm_point ADD COLUMN mz_poi_min_zoom SMALLINT;
 -- the coalesce here is just an optimisation, as the poi level
--- a/data/apply-planet_osm_point.sql
+++ b/data/apply-planet_osm_point.sql
@@ -6,6 +6,14 @@ BEGIN
 --------------------------------------------------------------------------------
 
 CREATE INDEX planet_osm_point_place_index ON planet_osm_point(place) WHERE name IS NOT NULL AND place IN ('borough', 'city', 'continent', 'country', 'county', 'district', 'farm', 'hamlet', 'island', 'isolated_dwelling', 'lake', 'locality', 'neighbourhood', 'ocean', 'province', 'quarter', 'sea', 'state', 'suburb', 'town', 'village');
-CREATE INDEX planet_osm_point_level_index ON planet_osm_point(mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway")) WHERE mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway") IS NOT NULL;
+CREATE INDEX planet_osm_point_level_index ON planet_osm_point(mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real)) WHERE mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real) IS NOT NULL;
+
+ALTER TABLE planet_osm_point ADD COLUMN mz_poi_min_zoom SMALLINT;
+-- the coalesce here is just an optimisation, as the poi level
+-- will always be NULL if all of the arguments are NULL.
+UPDATE planet_osm_point SET
+    mz_poi_min_zoom = mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real)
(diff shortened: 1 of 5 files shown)


data/functions.sql
```diff
@@ -41,12 +41,15 @@ CREATE OR REPLACE FUNCTION mz_calculate_poi_level(
...
+    way_area real
 )
 RETURNS SMALLINT AS $$
+DECLARE
+  zoom smallint;
 BEGIN
-    RETURN (
-        CASE WHEN aeroway_val IN ('aerodrome', 'airport') THEN 9
+  zoom =
+        CASE WHEN aeroway_val IN ('aerodrome', 'airport') THEN 13
```
rmarianski: Just checking, did you mean to update this from 9 -> 13? If so, should we just move this into the level 13 "section"?

data/functions.sql
```diff
@@ -105,8 +108,13 @@ BEGIN
...
              WHEN (amenity_val IN ('bench', 'waste_basket')) THEN 18
-             ELSE NULL END
-    );
+             ELSE NULL END;
+  RETURN (CASE
+    WHEN way_area > 1.0e7 THEN zoom - 4
+    WHEN way_area > 1.0e6 THEN zoom - 3
+    WHEN way_area > 1.0e5 THEN zoom - 2
+    WHEN way_area > 1.0e4 THEN zoom - 1
+    ELSE zoom END);
```
rmarianski: We'll probably want to update the `start_zoom` in queries.yaml to reflect the lowest possible zoom level that we can return here now.

queries/pois.jinja2
```diff
@@ -6,48 +6,40 @@ SELECT
...
+    planet_osm_point
+
+  WHERE
+    {{ bounds|bbox_filter('way') }}
+    AND mz_poi_min_zoom <= {{ zoom }}
+
+UNION ALL
+
+  SELECT
+    name, st_centroid(way) AS way, way_area, mz_poi_min_zoom AS min_zoom,
```
rmarianski: Can we use the pre-calculated `mz_centroid` column here instead of calling `st_centroid`?

data/apply-planet_osm_point.sql
```diff
@@ -6,6 +6,14 @@ BEGIN
...
-CREATE INDEX planet_osm_point_level_index ON planet_osm_point(mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway")) WHERE mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway") IS NOT NULL;
+CREATE INDEX planet_osm_point_level_index ON planet_osm_point(mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real)) WHERE mz_calculate_poi_level("aerialway", "aeroway", "amenity", "barrier", "craft", "highway", "historic", "leisure", "lock", "man_made", "natural", "office", "power", "railway", "shop", "tourism", "waterway", 0::real) IS NOT NULL;
```
rmarianski: Should we drop this index, and instead create one on `mz_poi_min_zoom`? I think this index only existed to support the `pois` queries, and we're filtering on this new column now.

data/functions.sql
zerebubuth: Yup. Airports can be quite large features and having them pop up at zoom < 9 didn't seem right. I suspect we'll have to revisit these levels anyway in a future task.

data/functions.sql
zerebubuth: I don't know - given we switch from NE to OSM at 9 (soon to be 8) do we want to show some of this stuff at pre-OSM zooms? I reckon that might be a decision to punt on until we get some feedback from @nvkelso on the new way POIs are coming in as zoom changes.

queries/pois.jinja2
zerebubuth: Well spotted! I forgot this, even after changing the trigger logic to ensure it's there... Fixed in  2c4e47505.

data/apply-planet_osm_point.sql
zerebubuth: This index wasn't been used on the POIs queries anyway, since Postgres was choosing to use the geo index instead. Should we just drop it, or create a series of partial indexes on successive zooms?

data/functions.sql
rmarianski: That's a good point, especially if we're going to be looking at tweaking these values anyway.

data/apply-planet_osm_point.sql
rmarianski: Is that always be true? Or could there be certain cases where the query planner would elect to use both indexes? I've generally taken the approach of adding indexes on any filter. If that doesn't make sense, we might want to revisit all the current indexes we have to see if any can be dropped.

What would the partial indexes on successive zooms look like? Would that have a performance improvement over just a partial index on the column?

zerebubuth: As I understand it, Postgres can only [combine index scans](http://www.postgresql.org/docs/9.3/static/indexes-bitmap-scans.html) under some pretty specific circumstances and I haven't seen it do so on my database. I think the generally-accepted method would be to create something like the following:

```SQL
CREATE INDEX planet_osm_point_poi_zoom_9 ON planet_osm_point USING gist(way) WHERE mz_poi_min_zoom <= 9;
CREATE INDEX planet_osm_point_poi_zoom_10 ON planet_osm_point USING gist(way) WHERE mz_poi_min_zoom <= 10;
CREATE INDEX planet_osm_point_poi_zoom_11 ON planet_osm_point USING gist(way) WHERE mz_poi_min_zoom <= 11;
CREATE INDEX planet_osm_point_poi_zoom_12 ON planet_osm_point USING gist(way) WHERE mz_poi_min_zoom <= 12;
CREATE INDEX planet_osm_point_poi_zoom_13 ON planet_osm_point USING gist(way) WHERE mz_poi_min_zoom <= 13;
```

Each index is trading off some improved runtime performance for slower updates, so it's probably not worth doing every single zoom, perhaps only 9, 11 & 13.

rmarianski: It tends not to do it on my database either, but I think I've seen it combine indexes on a database with the full planet. We could run some tests here to see.

About the multiple partial indexes, would that really help in this case? I'd expect that they wouldn't really do very much over just a single index on the column. Wouldn't postgresql internally organize the index in a similar way to what the partial indexes are doing?

rmarianski:
:+1:

data/apply-planet_osm_point.sql
zerebubuth: Running tests would be good to get some real info.

My gut feeling is that because we've got two effectively unrelated items; the row's geographic position and its `mz_poi_min_zoom`. An index on one has no correlation to the other, and vice-versa. What we want is a combined index where lower values of `mz_poi_min_zoom` will be "higher up" in the tree structure, but (AFAIK) no such index exists.

As I understand it, to do a bitmap index scan would be conceptually similar to:

```SQL
create temporary table geo_rows as
select __row_id__ from table
where way && geo_condition;

create temporary table poi_zoom_rows as
select __row_id__ from table
where poi_min_zoom <= zoom_condition;

select * from table
where __row_id__ in (
  select * from geo_rows union
  select * from poi_zoom_rows);
```

Which means that all the index pages for everything matching `way && geo_condition` *or* `poi_min_zoom <= zoom_condition` need to be loaded. Whereas for one index + filter, it only needs to load one set of index pages, although it'll end up loading more data pages from disk. The optimal is to have a partial index on both conditions... but then we'd be paying the cost on update rather than query.

Hard data for what the actual relative speeds of each are would be very useful! :smile:

PR #224, opened 2015-09-18
Use UNION ALL instead of UNION.

zerebubuth:
Turns out, it was always appropriate to use `UNION ALL` in our queries, as we were either selecting from different tables with definitely different data, or selecting different attributes anyway.


rmarianski:
Great catch! :zap: :+1:

Issue #205, opened 2015-09-11
Tweak NE to OSM transition to zoom 8 (all but places)


nvkelso:
We adjusted the internal NE data zoom transitions with https://github.com/mapzen/vector-datasource/pull/172 and https://github.com/mapzen/vector-datasource/issues/145 but we still need to move the overall Natural Earth to OSM zoom transition from zoom 9 to zoom 8.

This needs to be for all features in all layers (there are other tickets to adjust a couple individual features earlier).

zerebubuth:
All features in all layers, or all features in all layers but places? (Is the text body or title correct?)

nvkelso:
Needs to start showing OSM at zoom 8 (the last NE test needs to be <8 not <9):

- [ ] boundaries
- [ ] landuse - generally we need to move this to 8, but some other issues reference a subset of feature that should show earlier.
- [ ] landuse_labels – same deal was landuse changes above
- [ ] roads

These are optional. Visually it's better to leave as-is (transition at 9 not 8), but may be required to match up with changes above.

- [ ] earth
- [ ] water

OSM should remain at zoom 9 (the last NE test should remain <9):
- [x] places - This one needs more and different work, no change now
- [x] pois - these only come in at zoom 9 now for airports only, no change needed

Not relevant/ignore:
- [x] buildings - these only come in at zoom 13 now
- [x] transit - some things come in at zoom 6 now, no change

Additionally, we're should be calculating `landuse-labels` for `parks` in earlier zooms from Natural Earth, that's a separate issue: 

- [ ] landuse_labels

nvkelso:
Back to @zerebubuth to coordinate changes with https://github.com/mapzen/vector-datasource/issues/149.

zerebubuth:
See screenshots in #229 for cross-fading of urban areas. Unfortunately, you can only see the limited subset of stuff loaded on my box - dev will show this better.

nvkelso:
Primary roads need to be visible at z8, too. Else gaps are visible. Otherwise looks good.

PR #225, opened 2015-09-21
Tweak NE to OSM transition

zerebubuth:
Refs #205.

This changes the zoom test for NE from `<9` to `<8` for landuse, landuse labels and roads. Boundaries were already `<8`.

Additionally, made a couple of fixes to existing layers:

* Minor speed-up to boundaries by doing including a bbox `&&` test as part of the join condition. This helps at lower zooms where there are many county boundaries. Better would be to cut them out at, say, zoom 10 and lower if we don't need to render them.
* Several queries in the water layer weren't returning their responses `ST_AsBinary()`, but this didn't seem to cause any problems. In any case, for consistency, they all now have that.

@rmarianski, could you review, please?

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -122,7 +122,7 @@ CREATE OR REPLACE FUNCTION mz_calculate_road_level(highway_val text, railway_val
 RETURNS SMALLINT AS $$
 BEGIN
     RETURN (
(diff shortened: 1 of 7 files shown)


rmarianski:
:+1: 

Good idea with the bbox filter on the join, and great catch on the `ST_AsBinary`. My guess is that postgresql inferred to make it binary if any of the queries in the union had that set, but like you say, it's better to be consistent.

Issue #149, opened 2015-08-26
Include national parks and other significant OSM landuse earlier


nvkelso:
- [x] National Parks at zoom 4
- [x] National Forest like things at zoom 5+
- [x] Keep Natural Earth urban areas as zoom 8

Punting this one:
- [x] ~~Urban / Rural at zoom 6+~~

rmarianski:
> National Parks at zoom 4

We include the 50m urban areas and zoom 4. Is it ok to mix the 10m parks dataset with this one?

> National Forest like things at zoom 5+

Is there a dataset that you had in mind for this one? I browsed through ne but didn't find any that seemed to match.

> Urban / Rural at zoom 6+

We're including urban areas starting at zoom 6. Is there an equivalent dataset for rural areas? I couldn't find one in ne.

nvkelso:
> We include the 50m urban areas and zoom 4. Is it ok to mix the 10m parks dataset with this one?

Yes. Let's preview this on dev and see how it looks. This will only add National Parks for the USA at first, but that may be enough for first cut. We'd need to do a non-trivial amount of work to pull out just National Parks from OSM for the same purpose world wide, which is beyond the scope of this issue.

> Is there a dataset that you had in mind for this one? I browsed through ne but didn't find any that seemed to match.

This should come from OSM, but make sure it's generalized.

> We're including urban areas starting at zoom 6. Is there an equivalent dataset for rural areas? I couldn't find one in ne.

For OSM, let's investigate only including Urban polygons earlier and make sure they are generalized. (Urban and rural are all tied up in Urban in Natural Earth so the distinction isn't important.)

zerebubuth:
> For OSM, let's investigate only including Urban polygons earlier and make sure they are generalized. 

OSM doesn't really have the concept of "urban polygons". We could try using various values of landuse polygons, but I'm not sure the quality will be any good. Most landuse polygons in OSM are pretty small and their coverage is patchy. See below the area around NYC:

![landuse_nyc](https://cloud.githubusercontent.com/assets/271360/9998046/5a967016-6087-11e5-9a80-7781b8364325.png)

The red/purple/blue are landuse values associated with urban areas (e.g: residential, commercial, industrial) and the green-ish ones are associated with rural areas (e.g: forest, conservation, farm). The brown are neither. As you can see, the coverage is patchy, and many landuse polygons have voids between them that we would need to fill in, which will make for expensive queries or complex preprocessing.

[...]

PR #227, opened 2015-09-21
Added OSM-based national park polygons at zooms 4-8.

zerebubuth:
This adds those polygons, removing the NE ones which were in some of the queries. The new polygons merge in the `boundary` attribute and update the `mz_calculate_is_landuse` function to treat national park polygons as a type of landuse.

It also exposes an attribute for the OSM `operator` tag, which can be used to distinguish US National Parks Service areas from those run by the US Forestry Service.

Refs #149.

Before, zooms 4-6:

149-before-z4

149-before-z5

149-before-z6

After:

149-after-z4

149-after-z5

149-after-z6

With the following snippet of `scene.yaml`:

```yaml
    landuse:
        data: { source: osm }
        draw:
            polygons:
                order: 1
                color: '#aaaaa0'
        urban:
          filter: { kind: urban area }
          draw:
            polygons:
              order: 2
              color: '#808080'

        park or protected land:
          filter: { kind: national_park }
          draw:
            polygons:
              order: 2
              color: '#80aa80'

          national_forest:
            filter: { operator: [United States Forest Service, US Forest Service, National \
Forest Service] }
            draw: { polygons: { color: '#609060' } }

          national_park:
            filter: { operator: [United States National Park Service, United States Park Se\
rvice, US National Park Service, National Park Service] }
            draw: { polygons: { color: '#50a060' } }
```

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -21,7 +21,7 @@ BEGIN
      OR man_made_val IN ('pier', 'wastewater_plant', 'works', 'bridge', 'tower',
                          'breakwater', 'water_works', 'groyne', 'dike', 'cutline')
      OR power_val IN   ('plant', 'generator', 'substation', 'station', 'sub_station')
-     OR boundary_val IN ('national_park');
+     OR boundary_val IN ('national_park', 'protected_area');
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
 
@@ -304,5 +304,65 @@ CREATE OR REPLACE FUNCTION mz_one_pixel_zoom(
 RETURNS real AS $$
 BEGIN
   RETURN (17.256-ln(way_area)/ln(4));
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+-- calculate the collapse of several properties onto one string
+-- 'kind'. this involves a series of precedence choices, as
+-- it's possible for a feature to have values in several of
+-- these categories. in general, the "most important" should go
+-- first, or its presence should modify the later.
+CREATE OR REPLACE FUNCTION mz_calculate_landuse_kind(
+  landuse_val text,
+  leisure_val text,
+  natural_val text,
+  highway_val text,
+  aeroway_val text,
+  amenity_val text,
+  tourism_val text,
+  man_made_val text,
+  power_val text,
+  boundary_val text)
+RETURNS text AS $$
+BEGIN
+  RETURN
+    CASE
+      WHEN boundary_val IN (
+        'national_park', 'protected_area')
+	THEN boundary_val
+      WHEN landuse_val IN (
+        'park', 'forest', 'residential', 'retail', 'commercial', 'industrial',
+	'railway', 'cemetery', 'grass', 'farmyard', 'farm', 'farmland', 'wood',
+	'meadow', 'village_green', 'recreation_ground', 'allotments', 'quarry',
+	'urban', 'rural', 'military')
+        THEN landuse_val
+      WHEN leisure_val IN (
+        'park', 'garden', 'playground', 'golf_course', 'sports_centre', 'pitch',
+	'stadium', 'common', 'nature_reserve')
+	THEN leisure_val
+      WHEN natural_val IN (
+        'wood', 'land', 'scrub', 'wetland', 'glacier')
+	THEN natural_val
+      WHEN highway_val IN (
+        'pedestrian', 'footway')
+	THEN highway_val
+      WHEN amenity_val IN (
+        'university', 'school', 'college', 'library', 'fuel', 'parking',
+	'cinema', 'theatre', 'place_of_worship', 'hospital')
+	THEN amenity_val
+      WHEN aeroway_val IN (
+        'runway', 'taxiway', 'apron', 'aerodrome')
+	THEN aeroway_val
+      WHEN tourism_val IN (
+        'zoo')
+	THEN tourism_val
+      WHEN man_made_val IN (
+        'pier', 'wastewater_plant', 'works', 'bridge', 'tower', 'breakwater',
+	'water_works', 'groyne', 'dike', 'cutline')
+	THEN man_made_val
+      WHEN power_val IN (
+        'plant', 'generator', 'substation', 'station', 'sub_station')
+	THEN power_val
+      ELSE NULL END;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -53,6 +53,11 @@ BEGIN
       WHEN boundary_val IN (
         'national_park', 'protected_area')
 	THEN boundary_val
+      -- promote this above landuse as it's more specific, and we
+      -- don't want to lump nature reserves in with all of the
+      -- generic forests.
+      WHEN leisure_val = 'nature_reserve'
+        THEN leisure_val
       WHEN landuse_val IN (
         'park', 'forest', 'residential', 'retail', 'commercial', 'industrial',
 	'railway', 'cemetery', 'grass', 'farmyard', 'farm', 'farmland', 'wood',
@@ -61,7 +66,7 @@ BEGIN
         THEN landuse_val
       WHEN leisure_val IN (
         'park', 'garden', 'playground', 'golf_course', 'sports_centre', 'pitch',
-	'stadium', 'common', 'nature_reserve')
+	'stadium', 'common')
 	THEN leisure_val
       WHEN natural_val IN (
         'wood', 'land', 'scrub', 'wetland', 'glacier')
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1,3 +1,5 @@
+-- IF YOU UPDATE THIS, PLEASE UPDATE mz_calculate_landuse_kind
+-- BELOW!
 CREATE OR REPLACE FUNCTION mz_calculate_is_landuse(
     landuse_val text, leisure_val text, natural_val text, highway_val text,
     amenity_val text, aeroway_val text, tourism_val text, man_made_val text,
@@ -22,6 +24,69 @@ BEGIN
                          'breakwater', 'water_works', 'groyne', 'dike', 'cutline')
      OR power_val IN   ('plant', 'generator', 'substation', 'station', 'sub_station')
      OR boundary_val IN ('national_park', 'protected_area');
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+-- calculate the collapse of several properties onto one string
+-- 'kind'. this involves a series of precedence choices, as
+-- it's possible for a feature to have values in several of
+-- these categories. in general, the "most important" should go
+-- first, or its presence should modify the later.
+--
+-- IF YOU UPDATE THIS, PLEASE UPDATE mz_calculate_is_landuse
+-- ABOVE!
+CREATE OR REPLACE FUNCTION mz_calculate_landuse_kind(
+  landuse_val text,
+  leisure_val text,
+  natural_val text,
+  highway_val text,
+  aeroway_val text,
+  amenity_val text,
+  tourism_val text,
+  man_made_val text,
+  power_val text,
+  boundary_val text)
+RETURNS text AS $$
+BEGIN
+  RETURN
+    CASE
+      WHEN boundary_val IN (
+        'national_park', 'protected_area')
+	THEN boundary_val
+      WHEN landuse_val IN (
+        'park', 'forest', 'residential', 'retail', 'commercial', 'industrial',
+	'railway', 'cemetery', 'grass', 'farmyard', 'farm', 'farmland', 'wood',
+	'meadow', 'village_green', 'recreation_ground', 'allotments', 'quarry',
+	'urban', 'rural', 'military')
+        THEN landuse_val
+      WHEN leisure_val IN (
+        'park', 'garden', 'playground', 'golf_course', 'sports_centre', 'pitch',
+	'stadium', 'common', 'nature_reserve')
+	THEN leisure_val
+      WHEN natural_val IN (
+        'wood', 'land', 'scrub', 'wetland', 'glacier')
+	THEN natural_val
+      WHEN highway_val IN (
+        'pedestrian', 'footway')
+	THEN highway_val
+      WHEN amenity_val IN (
+        'university', 'school', 'college', 'library', 'fuel', 'parking',
+	'cinema', 'theatre', 'place_of_worship', 'hospital')
+	THEN amenity_val
+      WHEN aeroway_val IN (
+        'runway', 'taxiway', 'apron', 'aerodrome')
+	THEN aeroway_val
+      WHEN tourism_val IN (
+        'zoo')
+	THEN tourism_val
+      WHEN man_made_val IN (
+        'pier', 'wastewater_plant', 'works', 'bridge', 'tower', 'breakwater',
+	'water_works', 'groyne', 'dike', 'cutline')
+	THEN man_made_val
+      WHEN power_val IN (
+        'plant', 'generator', 'substation', 'station', 'sub_station')
+	THEN power_val
+      ELSE NULL END;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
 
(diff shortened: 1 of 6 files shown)


zerebubuth:
Note - waiting on whether we're rolling back bits of #225 (as discussed in parent #149) before actually trying to review this.

queries/landuse.jinja2
```diff
@@ -29,19 +48,27 @@ WHERE
 UNION ALL
 
 SELECT
-    name,
-    {{ ne_landuse_cols('park or protected land') }}
+  name,
+  operator,
+  way_area::bigint AS area,
+  COALESCE("boundary", "landuse", "leisure", "natural", "highway", "aeroway", "amenity", "tourism", "man_made", "power") AS kind,
```
nvkelso: We've been adding the new coalesce values at the end of the list. 

Should we:

* Keep that pattern, adding "boundary" after "power
* Alpha sort them all (new and old)
* Add at the beginning of the list (as you propose)

queries/landuse.jinja2
```diff
@@ -53,21 +80,8 @@ FROM
...
-    {% elif zoom == 12 %}
-    AND way_area::bigint > 6400
-    {% elif zoom == 13 %}
-    AND way_area::bigint > 1600
-    {% elif zoom == 14 %}
-    AND way_area::bigint > 400
-    {% elif zoom == 15 %}
-    AND way_area::bigint > 100
+    {% if zoom <= 15 %}
+    AND (17.256-ln(way_area)/ln(4)) <= {{ zoom }}
```
nvkelso: I like the terseness, but it's also not as legible as what we had before.

queries/landuse.jinja2
nvkelso: Seems like we should follow `#1` as it has the least impact on other code.

nvkelso:
Why are `national forests` and `national parks` getting the same [kind=national_park](http://wiki.openstreetmap.org/wiki/Tag:boundary=national_park) value? Maybe it's my US bias, but that seems wrong – is this just an OSM-ism? 

As you suggest, we can use operator to suss out the difference, but relying on the operator does feel country specific – is there another attribute we could use like `admin_level` that's more generic between countries?

Should we also include [boundary=protected_area](http://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area) as well? The wiki page is more complete than for the `boundary=national_park`, and looks like data adoption / coverage is global.

Here's [TagInfo's national_park](https://taginfo.openstreetmap.org/tags/boundary=national_park#map) map:

taginfo_national_park

And [TagInfo's protected_area](https://taginfo.openstreetmap.org/tags/boundary=protected_area#map) map:

taginfo_protected_area

nvkelso:
Looks like some `boundary=national_park` have a [protect_class](https://taginfo.openstreetmap.org/keys/protect_class) tag.

queries/landuse.jinja2
zerebubuth: Adding it at the end might not have the effect that you want; many of these `boundary=national_park` areas also have `landuse`, `leisure` or `natural` tags, and so the `boundary` value would be shadowed by these other values. This would mean that most `national_park` areas would be indistinguishable from other areas. That's why I added it at the beginning.

Using `COALESCE` like this is basically dropping a whole bunch of useful information, and the side-effect is that the values in the `COALESCE` need to be sorted by "importance". Any key that appears before another must be more important than that other, and if the first has a value then the second's value (or lack of it) will be ignored.

If we hide the `national_park` value at the back of the list, it's likely that very few polygons will get this tag. Let me know if that's still the preferred option.

queries/landuse.jinja2
zerebubuth: Yup, good point! I've added a function for that in b8e748d which hopefully makes it more readable while still being succinct.

zerebubuth:
> Why are national forests and national parks getting the same kind=national_park value? Maybe it's my US bias, but that seems wrong – is this just an OSM-ism?

We can use a different kind for those, if you like? The data in OSM, upon inspection, uses the `boundary=national_park` tag for all of the national-level conservation areas that, in the USA, are called National Parks and National Forests. But let's not get confused about terminology; what OSM calls `boundary=national_park` appears to include both US National Parks and US National Forests. The difference between the two is highly country- and cultural-specific, and doesn't exist in most other parts of the world.

> As you suggest, we can use operator to suss out the difference, but relying on the operator does feel country specific – is there another attribute we could use like admin_level that's more generic between countries?
> Looks like some boundary=national_park have a protect_class tag.

[The `protect_class` values](http://wiki.openstreetmap.org/wiki/Tag:boundary%3Dprotected_area#Protect_classes_for_various_countries) also seem pretty country-specific, are available on only 18% of `boundary=national_park` polygons and seem to refer to some arbitrary non-physical designations elsewhere. But seems like a better way to go than looking at `operator`.

> Should we also include boundary=protected_area as well? The wiki page is more complete than for the boundary=national_park, and looks like data adoption / coverage is global.

I excluded these on purpose because all of the wilderness areas that we discussed yesterday, and you didn't want to appear, turned out to be `boundary=protected_area`. I'll add them back in now.

zerebubuth:
Now it looks like this:

149-better-parks-forests

With only a minimum level of fudging, which hopefully we can get rid of as soon as the `protect_class` tag has wider accepance:

```yaml
        national park:
          filter: { kind: national_park }
          draw:
            polygons:
              order: 2
              color: '#80aa80'

          national forest:
            filter: |
              function() {                                                                  
                return feature["name"] && (feature["name"].indexOf("Forest") >= 0);         
              }                                                                             
            draw:
              polygons:
                order: 2
                color: '#889988'

        protected land:
          filter: { kind: protected_area }
          draw:
            polygons:
              order: 2
              color: '#aa9980'
```

@nvkelso how's it looking?

queries.yaml
```diff
@@ -53,7 +53,11 @@ layers:
     template: landuse.jinja2
     start_zoom: 4
     geometry_types: [Polygon, MultiPolygon]
+<<<<<<< HEAD
```
nvkelso: Rebase funk hanging out?

queries/landuse.jinja2
```diff
@@ -10,16 +10,39 @@
...
 SELECT
     '' AS name,
+    NULL AS protect_class,
     {{ ne_landuse_cols('urban area') }}
 FROM
     ne_50m_urban_areas
 WHERE
     {{ bounds|bbox_filter('the_geom') }}
 
+<<<<<<< HEAD
```
nvkelso: More rebase funk?

queries/landuse.jinja2
```diff
@@ -29,19 +52,27 @@ WHERE
...
+  {% filter geometry %}way{% endfilter %} AS __geometry__,
+  osm_id AS __id__
 FROM
-    ne_10m_parks_and_protected_lands
+  planet_osm_polygon
 WHERE
-    {{ bounds|bbox_filter('the_geom') }}
+  {{ bounds|bbox_filter('way') }}
+  AND boundary IN ('national_park', 'protected_area')
+  AND mz_one_pixel_zoom(way_area) <= {{ zoom }}
```
nvkelso: We've discussed changing 1 pixel to something more configurable. The name should reflect that, maybe `mz_minimum_pixel_size_zoom(way_area)`? That's doesn't roll of the tongue, proposals?

data/functions.sql
```diff
@@ -292,3 +293,77 @@ BEGIN
...
+	'urban', 'rural', 'military')
+        THEN landuse_val
+      WHEN leisure_val IN (
+        'park', 'garden', 'playground', 'golf_course', 'sports_centre', 'pitch',
+	'stadium', 'common', 'nature_reserve')
+	THEN leisure_val
+      WHEN natural_val IN (
+        'wood', 'land', 'scrub', 'wetland', 'glacier')
+	THEN natural_val
+      WHEN highway_val IN (
```
nvkelso: Why are highway tags being included here?

data/functions.sql
```diff
@@ -292,3 +293,77 @@ BEGIN
...
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+-- returns the (possibly fractional) zoom at which the given way
+-- area will be one square pixel nominally on screen (assuming
+-- that tiles are 256x256px at integer zooms). sadly, features
+-- aren't always rectangular and axis-aligned, but this should
+-- still give a reasonable approximation to the zoom that it
+-- would be appropriate to show them.
+CREATE OR REPLACE FUNCTION mz_one_pixel_zoom(
```
nvkelso: See comment below about function name.

nvkelso:
```
national forest:
            filter: |
              function() {                                                                  
                return feature["name"] && (feature["name"].indexOf("Forest") >= 0);         
              }            
```

@bcamper and @zerebubuth Brett, is this the performant thing to do, or is an exact filter match better in Tangram? For instance, if we export the features `operator` tag in the vector tile, an exact match could be performed.

I also wonder about including the `operator` tag we had earlier in this thread for renderers that can't evaluate JS inside their filters.

nvkelso:
@zerebubuth The results are looking great!

Can you include screenshots of Europe, Africa, South America, Asia, and Australia at the same zooms to confirm the changes are working outside of USA?

bcamper:
@nvkelso @zerebubuth yes an exact match will be faster.

For Tangram JS, *all* filters (even ones with YAML syntax) get compiled into JS functions internally anyway. But an equality check (`===`) is still going to be faster than an `indexOf` string match.

For Tangram ES, exact matches will definitely be better, since there's a higher cost to calling a JS function (eventually we want better benchmarks on that to get a sense of the cost, @blair1618 @hjanetzek).

queries/landuse.jinja2
zerebubuth: Ah, this is something slightly different; rather than the number of pixels it'll be at any zoom, this function returns the (fractional) zoom at which the feature will be one pixel in size. This allows us to adjust it by changing the comparison (e.g: `<= {{ zoom }} - 1` if we want it to be at least 4 square pixels). I would call it `mz_zoom_at_which_this_feature_is_one_square_pixel`, but that doesn't exactly roll off the tongue. :wink:

data/functions.sql
zerebubuth: Because `highway` was previously included in the `COALESCE` (e.g: [here](https://github.com/mapzen/vector-datasource/pull/227/files#diff-21e9ac7608f051b94d3add540d651c99L44)), and is selected as part of [mz_calculate_is_landuse](https://github.com/mapzen/vector-datasource/pull/227/files#diff-85235d701f80586d926a4211261dbbc1L16).

queries.yaml
zerebubuth: D'oh! Thanks for catching that. I thought `git` had a lint for that... obviously not. Sorry. Fixed in 162f1bd.

queries/landuse.jinja2
nvkelso: Cool! Let's keep as-is.

data/functions.sql
nvkelso: So we were already limiting the highway selection to things that are more "area" in nature, so good as-is. Thanks for checking.

zerebubuth: Yup, and all this is coming out of `planet_osm_polygon` anyway. So any highway features ending up in there will be polygonal.

rmarianski: This sort of thing can also be done at the transform layer in python. The query would just need to be modified to pass through all the necessary tags.

zerebubuth: Right, but because this is so closely tied to `mz_calculate_is_landuse` I thought it made sense to keep them together. What do you think?

zerebubuth:
Here's some screenshots showing global coverage. Some areas are better than others, but it's not clear whether that's because some are better mapped, or the tagging is different, or they have a completely different system. A quick look at the national parks in Africa suggests it might be that they're not mapped (yet).

149-east-usa

149-central-america

149-south-america

149-peru

149-europe

149-uk

14-africa

149-south-africe

149-asia

149-korea-japan

149-oceania

149-south-australia

data/functions.sql
rmarianski: Yup, makes sense to me. Maybe we should move them together to make it more likely to update them in tandem?

zerebubuth: Great idea, thanks! Fixed in 0cd2099.

nvkelso:
Looks like a lot of details is hiding in **France**, **Span**, and most of continental Europe. Assuming some are `National Park` like in importance, can some of them be promoted a zoom (so a subset is visible earlier)?

Surprised that nothing is showing up in Japan, what's up there?

Also surprised there's nothing in **Brazil**.

nvkelso:
**Brazil** looks to be like France and Spain where it's all hiding in the brown feature class.

zerebubuth:
Japan seems to not have national parks mapped except as `landuse=forest`. I searched for a few which are mentioned [here](https://www.env.go.jp/en/nature/nps/park/index.html), but I couldn't find any matching features in OSM tagged with anything that meant more than just "here are some trees".

nvkelso:
It looks like OSM.org shows more National Park like features at zoom 7 than we are.

Is this the [relevant bit](https://github.com/gravitystorm/openstreetmap-carto/blob/master/project.mml#L1183-L1206) in the default style?

```
      "name": "nature-reserve-boundaries",
      "srs-name": "900913",
      "geometry": "polygon",
      "class": "",
      "id": "nature-reserve-boundaries",
      "srs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0.0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs +over",
      "Datasource": {
        "extent": "-20037508,-20037508,20037508,20037508",
        "table": "(SELECT\n    way,\n    name,\n    boundary,\n    way_area/NULLIF(!pixel_width!::real*!pixel_height!::real,0) AS way_pixels\n  FROM planet_osm_polygon\n  WHERE (boundary = 'national_park' OR leisure = 'nature_reserve')\n    AND building IS NULL\n    AND way_area > 0.01*!pixel_width!::real*!pixel_height!::real\n) AS national_park_boundaries",
        "geometry_field": "way",
        "type": "postgis",
        "key_field": "",
        "dbname": "gis"
      },
      "extent": [
        -180,
        -85.05112877980659,
        180,
        85.05112877980659
      ],
      "properties": {
        "minzoom": 7
      },
      "advanced": {}
    },
```

Looks like they're also pulling out `leisure=nature_reserve`.

nvkelso:
@zerebubuth Japan National Parks are showing up in OSM.org at [zoom 9](https://www.openstreetmap.org/#map=9/36.0968/138.6378):

osm org-z9-japan

But it's pretty hard to select the things in iD to query their attributes :(

zerebubuth:
@nvkelso try emptying your cache:

japan-parks

Perhaps they were deleted recently, but they're not showing up for me in the current data, nor in the data from a couple of weeks ago that's loaded in my database.

zerebubuth:
[This one](http://www.openstreetmap.org/relation/5145361) seems to have survived, and is just about visible on the tiles at zoom 6:

149-japan-national-park

nvkelso:
Go figure, I'm seeing what you see now.

nvkelso:
@zerebubuth Let's add back the `operator` tag and call this done enough. 

Let's live with the change (+ `operator`) and any other mods can be done via a new issue.

zerebubuth:
Here's what it looks like now, with nature reserves and protected areas available at lower zooms. The protection class and operator attributes are available for discriminating various types and importances of area. Note that there are still some (generally smaller) areas missing because my local database only has these features for _relations_ loaded, and actually some are _ways_ - we'll need to wait until this hits dev to get a better idea what the actual coverage is, because my desktop can't handle loading much more...

149-spain

149-europe-2

149-africa2

149-south-america-2

Example `scene.yaml` fragment:

```yaml
    landuse:
        data: { source: osm }
        draw:
            polygons:
                order: 1
                color: '#aaaaa0'
        urban:
          filter: { kind: urban area }
          draw:
            polygons:
              order: 2
              color: '#999990'

        national park:
          filter: { kind: national_park }
          draw:
            polygons:
              order: 2
              color: '#889988'

          national parks service:
            filter: { operator: [ "United States National Park Service", "US National Park Service", "National Park Service", "United States Park Service", "US Park Service" ] }
            draw:
              polygons:
                order: 3
                color: '#60b060'

        protected land:
          filter: { kind: protected_area }
          draw:
            polygons:
              order: 2
              color: '#a89860'

        nature reserve:
          filter: { kind: nature_reserve }
          draw:
            polygons:
              order: 2
              color: '#408030'
```

zerebubuth:
@nvkelso looks good?

nvkelso:
LGTM, let's get this in dev.

PR #229, opened 2015-09-22
More tweaks to NE/OSM transition

zerebubuth:
Primary roads clutter the map at zoom 8, push them back to zoom 9. Also the `motorway`/`trunk` roads seem to be closer to the road selection on NE.

Keep NE urban areas a couple more zooms [as discussed](https://github.com/mapzen/vector-datasource/issues/149#issuecomment-142132701).

**NOTE**: changes to road level function require dropping & recreating that index.

205_after_z7

205_after_z8

205_after_z9

Note that the red bits of road in those images are not just the bits which are outside urban "landuse", but also those which clip a different kind of landuse which I didn't put a rendering rule in for, so they're not intersection errors.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -122,7 +122,8 @@ CREATE OR REPLACE FUNCTION mz_calculate_road_level(highway_val text, railway_val
 RETURNS SMALLINT AS $$
 BEGIN
     RETURN (
-        CASE WHEN highway_val IN ('motorway', 'trunk', 'primary', 'motorway_link') THEN 8
+        CASE WHEN highway_val IN ('motorway', 'trunk', 'motorway_link') THEN 8
+             WHEN highway_val IN ('primary') THEN 9
              WHEN highway_val IN ('secondary') THEN 10
              WHEN (highway_val IN ('tertiary')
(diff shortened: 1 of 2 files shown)


rmarianski:
:+1:

zerebubuth:
Ooops, should have said: Refs #205.

data/functions.sql
```diff
@@ -122,7 +122,8 @@ CREATE OR REPLACE FUNCTION mz_calculate_road_level(highway_val text, railway_val
 RETURNS SMALLINT AS $$
 BEGIN
     RETURN (
-        CASE WHEN highway_val IN ('motorway', 'trunk', 'primary', 'motorway_link') THEN 8
+        CASE WHEN highway_val IN ('motorway', 'trunk', 'motorway_link') THEN 8
+             WHEN highway_val IN ('primary') THEN 9
```
nvkelso: Primary roads need to be visible at z8, too. Else gaps are visible.

zerebubuth: All roads need to be visible, else gaps are visible. There are places where a primary road down-grades to a secondary which will show as gaps. I looked at the road network at z8 with primary included, and it was far too busy. If you'd like this changed, please open a new issue.

Issue #210, opened 2015-09-14
Improve data for admin boundary in the water


nvkelso:
Right now vector tiles don't indicate if a boundary (country, state, region, county, etc) is over the water or not. We should indicate this case to improve styling (color them blue, hide them, etc).

nvkelso:
See also: 

* https://github.com/mapzen/vector-datasource/issues/180

nvkelso:
Looks like the existing [boundary_type](http://wiki.openstreetmap.org/wiki/Key:border_type) property in the boundary query doesn't give us enough information for all maritime boundaries.

We need to export tags for [boundary%3Dmaritime](http://wiki.openstreetmap.org/wiki/Tag:boundary%3Dmaritime)

zerebubuth:
The work for #180 makes this way more difficult than just looking at a tag. Also, a maritime boundary isn't the same thing as a boundary over water. For example, [this border](http://www.openstreetmap.org/way/61602969) is entirely in the Hudson River, but isn't a maritime boundary.

If we want boundaries over water distinguished from boundaries over land for rendering purposes, then perhaps we need to intercut the water layer and boundaries layer. If we want maritime boundaries, then possibly that data can be recovered by intersecting the LineStrings, but I'm not 100% confident that it'll work.

@nvkelso, is it the former or latter that we're interested in?

nvkelso:
We do need maritime to deal with that case, thanks for bringing this up!

zerebubuth:
Hmm... looks like the boundary up the Hudson River is marked `maritime=yes` (e.g: ways [61602961](http://www.openstreetmap.org/way/61602961) and [61602969](http://www.openstreetmap.org/way/61602969)). In previous discussions, we'd talked about _not_ wanting to lose that boundary when intersecting with any water, so I think we'll need to think a bit more on what exactly it is that we want shown (or not shown) in the tiles.

nvkelso:
Let's bootstrap this using the maritime=yes tag alone and see how that goes.

zerebubuth:
After #236 and mapzen/TileStache#54, this is what it looks like. Maritime boundaries (with `maritime=yes`) are in blue, other boundaries in red:

![210-nyc-2](https://cloud.githubusercontent.com/assets/271360/10052239/cac79ca6-621d-11e5-9e45-a81814a11cd3.png)

[...]

PR #236, opened 2015-09-23
Add a `maritime` tag onto boundaries

zerebubuth:
Add a `maritime` tag onto boundaries, which is `yes` when the original boundary way from OSM had a `maritime=yes` tag.

Note that this doesn't correspond to only the "3 nautical mile" boundaries to international waters, it looks like a very large number (but not all) of the boundaries which occur in water or rivers are marked this way too.

Note also that this also adds some run-time overhead. Because we're using geometry from the admin polygon boundaries to supply boundaries with correct left/right orientation and name properties, we need to operate on the original, unsimplified geometry in the post-processor and requires mapzen/TileStache#54.

Refs #210. For screenshots, see the parent issue.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -174,3 +174,6 @@ post_process:
     attribute: kind
     target_attribute: landuse_kind
     cutting_attrs: { sort_key: 'order', reverse: True }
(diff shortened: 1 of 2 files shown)


rmarianski:
:+1:

PR #240, opened 2015-09-24
Fix logarithm of zero bug.

zerebubuth:
We can't take the logarithm of zero, and some features do have incredibly tiny areas. Instead, cut off so that 16 is the highest zoom we'll return.


rmarianski:
:+1:

PR #241, opened 2015-09-24
Re-order fields in UNION so that they are consistent.

zerebubuth:
This was causing an issue on tiles:

```
ERROR:  UNION types text and bigint cannot be matched at character 828
```

Or similar.


rmarianski:
:+1:

PR #245, opened 2015-09-25
fixes #194 add underground tag to buildings (and addr points)

### LINKED ISSUE #194: add underground tag to buildings
There are a couple of BART stations which are underground: https://www.openstreetmap.org/way/155378773
https://www.openstreetmap.org/way/160331904

But the tiles have them as just vanilla buildings at street level: 
http://tangrams.github.io/explorer/#17.5/37.7896/-122.4013/name/Montgomery%20Stat

https://trello-attachments.s3.amazonaws.com/55f1d2933cfa613ec481fe6c/978x816/e9e8d46c82c10fdd49e3d4af117272b5/Screen_Shot_2015-09-10_at_2.55.37_PM.png

nvkelso:
Adds `layer` and `location` to building features (from tags as those aren't included in the default import style file), and since the buildings layer includes address points also adds `addr_floor` and `level` (from tags).

### DIFF
--- a/queries/buildings.jinja2
+++ b/queries/buildings.jinja2
@@ -11,6 +11,8 @@ SELECT
     "building:min_levels",
     height,
     min_height,
+    tags->'layer' AS layer,
+    tags->'location' AS location,
     {% filter geometry %}way{% endfilter %} AS __geometry__,
 {% if zoom >= 15 %}
     "roof:color" AS roof_color,
@@ -22,6 +24,8 @@ SELECT
 {% if zoom >= 17 %}
     "addr:housenumber" AS addr_housenumber,
     "addr:street" AS addr_street,
+    NULL AS addr_floor,
+    NULL AS level,
 {% endif %}
     %#tags AS tags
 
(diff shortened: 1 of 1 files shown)


queries/buildings.jinja2
```diff
@@ -73,6 +79,8 @@ SELECT
     NULL AS roof_orientation,
     "addr:housenumber" AS addr_housenumber,
     "addr:street" AS addr_street,
+    tags->'addr:floor' AS addr_floor,
+    tags->'level' as level,
```
zerebubuth: Taginfo suggests that only [0.02% of nodes](http://taginfo.openstreetmap.org/keys/addr%3Ahousenumber?filter=nodes#combinations) tagged with `addr:housenumber` also have `level` set. Were there some specific nodes you were looking at where `level` was important?

nvkelso: Usually its not a problem until it is-large apartment towers were suddenly there are too many address points to display.

If this introduces too much cost to the query we should drop it for now, was trying to edge case / future proof.

_n

zerebubuth: Sorry, I didn't make my point very well. What I meant to say was: this seems like a very rarely-used tag, that doesn't provide a lot of extra information in the vast majority of cases. If you've got some examples of large apartment blocks with too many address points then we should check to see if `level` has been defined on any of those, so we can check that it'll provide the information for the edge-cases that you want it for.

nvkelso: Let's drop the add_floor and level bits from this until there's more of a use case.

zerebubuth:
:+1: looks good to me.

PR #246, opened 2015-09-25
add kind property to building layer features

### LINKED ISSUE #200: Building features should collapse amenity, shop, tourism into kind property
To be consistent with the rest of the tile queries (landuse, pois, etc), we should not export raw OSM attributes for `amenity`, `shop`, and `tourism`. Instead those should be coalesced into a single `kind` property that references the expanded list of keys from the landuse and poi queries.

### LINKED ISSUE #242: Add kind property to addr points
We recently added address points (addr:*) to the buildings layer, but those features are missing a `kind` attribute. Let's set that to `addr`.

nvkelso:
1. fixes #200 (coalesce kinds on buildings) 
2. fixes #242 (set kind to addr)


queries/buildings.jinja2
```diff
@@ -55,6 +52,7 @@ UNION ALL
 SELECT
     name,
     NULL AS area,
+    'addr' AS kind,
     osm_id AS __id__,
     NULL AS "building:part",
     NULL AS building,
```
zerebubuth: Both sides of a `UNION` must have the same columns, so you'll need to remove "building", "amenity", "shop" and "tourism" from the address points as well.

queries/buildings.jinja2
```diff
@@ -1,12 +1,9 @@
 SELECT
     name,
     way_area::bigint AS area,
+    COALESCE("building", "amenity", "shop", "tourism") AS kind,
```
zerebubuth: [Over 80% of buildings](http://taginfo.openstreetmap.org/keys/building#values) are tagged `building=yes`, which will mean that over 80% of building polygons in the tiles will have a kind of "yes", which is probably okay because they're in a separate layer, but doesn't seem to me like it's a useful value. Would it be better to keep the "building", "amenity", "shop" and "tourism" columns in the query and normalise them in a Python filter? That way we could express slightly more complex and useful logic about collapsing these values.

bcamper: I believe in some cases (I thought including buildings, could also be a
regression?), we were only including non-"yes" values. Because it's not
useful as Matt said, it takes up space (a lot on low zooms!), and it's
ambiguous when coalescing with other tags with possible "yes" values.

rmarianski: Yea, this is happening in a transform at the moment to account for that.

https://github.com/mapzen/TileStache/blob/0e6acb222c32653b2622e732d2745652ee208f52/TileStache/Goodies/VecTiles/transform.py#L156-L164

Maybe what was going on was for these features the values for `amenity`, `shop`, and `tourism` were all null so no kind value was being set.

nvkelso:
Depends on: https://github.com/mapzen/TileStache/pull/59

queries/buildings.jinja2
```diff
@@ -57,10 +58,11 @@ SELECT
     NULL AS area,
     osm_id AS __id__,
     NULL AS "building:part",
+    "address" AS kind,
     NULL AS building,
-    amenity,
-    shop,
-    tourism,
+    NULL AS amenity,
```
nvkelso: The `address` nature is important for addr points, the amenity, shop, tourism is not.

queries/buildings.jinja2
```diff
@@ -57,10 +58,11 @@ SELECT
     NULL AS area,
     osm_id AS __id__,
     NULL AS "building:part",
+    "address" AS kind,
```
nvkelso: Using `address` instead of `addr` because the extra characters aren't that precious.

nvkelso:
@rmarianski Can you have a look?

queries/buildings.jinja2
bcamper: More readable... could actually add up to a decent amount in a dense
building tile... but will also compress well.

rmarianski: @nvkelso do we want to update the [transform](https://github.com/mapzen/TileStache/blob/7ca4b5e2f87e6218522a981315e00c379d80bdc8/TileStache/Goodies/VecTiles/transform.py#L156-L164) to leave the `kind` alone if it's already set?

Reading through it, it should work as is because all the other fields would be null anyway, but we might want to be more explicit in the transform just for clarity.

rmarianski: :facepalm: Nm ... I just saw your comment below about updating the transform.

rmarianski:
This looks good to me. :+1:

queries/buildings.jinja2
zerebubuth: You'll want `'address'` instead of `"address"` if it's supposed to be a string. PostgreSQL uses double quotes for column and table names, single quotes for strings.

nvkelso:
Per Matt's comment, fixed the `'address'` bit.

PR #256, opened 2015-09-25
Temporarily roll back boundaries changes

rmarianski:


nvkelso:
We should roll back to what's in prod instead: 4c59ac235e7fa7059f3c1db3d5b46f4bb0466bb2. That's working for existing map demos and is less disruptive to the code base.

PR #258, opened 2015-09-25
Property rename from order -> sort_key

rmarianski:
Refs #198

Depends on https://github.com/mapzen/TileStache/pull/57


nvkelso:
LGTM

PR #261, opened 2015-09-28
Configure transform to create landuse_labels

rmarianski:
Refs https://github.com/mapzen/tilequeue/issues/21

Depends on https://github.com/mapzen/TileStache/pull/58


rmarianski:

We might want to create a concept of a dynamically generated or "virtual" layer. The problem with how things stand with the above is that `tileserver` uses the list of configured layers in `queries.yaml` to determine whether a request for a particular url is valid. Because this layer is dynamically generated, the user will receive a 404 when asking for just `landuse_labels`. Conversely, when asking for only the `landuse` layer, `landuse_labels` also appears in the response.

I was thinking of adding a `virtual` or `is_virtual` boolean to the layer config that will get ignored during data fetch but will mean that it will get picked up as a valid layer when asked for specifically. It's not as simple as that though, because just asking for `landuse_labels` wouldn't work in this case. We'd have to know that it implied inclusion of `landuse` during data fetch, and then it would get pruned from the response.

Thoughts?

zerebubuth:
I think we might be able to implement this more simply. In mapzen/tileserver#3 and mapzen/tilequeue#30, you added support for serving tiles by filtering the saved "all" tile. I think we could extend that to this: always generate all the layers, then filter them as a post-process if that's what the user asked for.

zerebubuth:
:+1: looks good to me.

rmarianski:
That's a good point, but it will implicitly mean that we can never have layers that are outside the `all` layer. We don't have this case currently, but it's possible that we might in the future with some experimental layers.

I also just realized that the pull request to serve dynamic layer lists more efficiently would suffer from this already because it assumes the layer is in `all` anyway (https://github.com/mapzen/tileserver/pull/3). Maybe the thing to do then is to try and keep things simple and continue making this assumption until we can't?

zerebubuth:
:+1: for keeping things simple. When we get to having extra experimental layers I think we'll know better how we want to trade off lower latency of generating a partial tile versus the simplicity of just having one generation process.

nvkelso:
If we start packing the result into the `src` layer instead of a new `src-labels` layer, how does that change things? I'd rather avoid a proliferation of new `src-labels` layers, if possible.

rmarianski:
@nvkelso I was starting to think it would just be a slight change, but some questions get raised as to how the geometry for the point would get encoded. The formats use different coordinate systems and encoding strategies for geometries, and I'm wondering how this would fit into that larger scheme across all formats. It would also likely require some custom decoding routines for that property to recognize it as a geometry as well. @bcamper thoughts?

rmarianski:
I created a ticket for the `tileserver` simplification work here: https://github.com/mapzen/tileserver/issues/4

rmarianski:
Ah, I misunderstood what was meant by packing into the src layer. We can introduce additional features that have the same properties in the existing layer, they would just have a point geometry. We can also include a new property, like `label_position`=`yes`.

rmarianski:
@zerebubuth, @nvkelso Look ok now?

nvkelso:
LGTM

Process question: This will be a backwards breaking change. How do we notify users? Do we pack the new labels into the `landuse` layer & the `landuse-labels` layer for  couple releases and then fully remove that layer?

nvkelso:
Related: https://github.com/tangrams/tangram-docs/pull/45#issuecomment-144111034

zerebubuth:
:+1: looks good to me.

> Process question: This will be a backwards breaking change. How do we notify users? Do we pack the new labels into the landuse layer & the landuse-labels layer for couple releases and then fully remove that layer?

If this is a "major" release (i.e: `v1.0.0`, or `v1.3.0` if you don't like the dreaded point-oh-point-oh) then can we keep the old tiles around in S3 (just not updated) for a month or so, and send an email around to anyone with a tile key to ask "please upgrade"?

bcamper:
Can you summarize what data/layers changed here?

nvkelso:
@bcamper Rob added the ability to easily calculate the (polygon) label centroid for any layer. 

The question is:

1. Kill `landuse-labels` now, while moving those features into the `landuse` layer (with a new `label_placement=yes` property.
2. Keep `landuse-labels` for a few months to allow transition & backwards compatibility, while also moving those features into the `landuse` layer (with a new `label_placement=yes` property.

I'm leaning towards option 2.

(The overall architecture change allows us to add water, building & etc label positions at the same time, those would go directly into existing layers.)

bcamper:
So ultimately the `landuse` layer will have polygons for filling and points
for labels (marked with an explicit property), and other layers will follow
the same pattern, yes?

I don't have a strong opinion. Option 2 wastes some space, but there are
not very many landuse areas in any single tile.

nvkelso:
Going with option 2...

PR #264, opened 2015-09-29
fixes #260 by reverting to earlier NE + OSM union

### LINKED ISSUE #260: REGRESSION: country & state tags suddenly no longer available at z0-z8
I think my stylesheet did not change but suddenly countrynames no longer appear at z0 - z3. They do appear at z4 and higher. Something changed on you side!?

nvkelso:
had inforced transition in the v0.3.0, but that had bad consequences.

### DIFF
--- a/queries/places.jinja2
+++ b/queries/places.jinja2
@@ -1,4 +1,3 @@
-{% if zoom < 9 %}
 SELECT
     name,
     featurecla AS kind,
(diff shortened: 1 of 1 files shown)


nvkelso:
@zerebubuth can you review, please? This was a `UNION` before, but I noticed we've been switching to `UNION ALL` on some other queries.

zerebubuth:
This should be `UNION ALL` as well, as the data to the left and the right will never be the same (left has source=NE, right has source=OSM).

PR #265, opened 2015-09-29
Tweak POI and landuse zoom filters

zerebubuth:
Change the zoom / area filter for a bunch of different features based on @nvkelso's feedback. There's probably still a whole bunch of tweaking to do, but this is a step in the right direction.

Connects to #195. See the parent issue for screenshots.


nvkelso:
@rmarianski Can you review, please?

rmarianski:
:+1: 

We're adding the min zoom to the label feature. I've got some pull requests open that will auto-generate the label features though (issue https://github.com/mapzen/tilequeue/issues/21). @zerebubuth Should the min zoom calculation become a part of that post process step?

zerebubuth:
> Should the min zoom calculation become a part of that post process step?

I'm not sure. We'd have to be careful about making sure the `min_zoom` calculation in the post-processing step matched the queries. Otherwise we'd end up missing stuff, or having stuff with `min_zoom=N` in tiles at zoom `N-1` or something.

rmarianski:
@zerebubuth Can we move it to the landuse layer? If we wanted, we could trim that property out in the post process transform except for label features.

zerebubuth:
Sure. If we add it to the landuse, and the calculation is the same for the label and the polygon (or at least the min of the two) then we can always strip stuff out later. I'm just concerned that we'll end up with "missing" features and end up bouncing between the two implementations; one in the database and the other in Python.

nvkelso:
> If we add it to the landuse, and the calculation is the same for the label and the polygon (or at least the min of the two) then we can always strip stuff out later.

Yeah. Sometimes there will be less labels than parent features, but never the other way around.

zerebubuth:

rmarianski:
:+1:

PR #271, opened 2015-09-30
Reimplement admin boundaries processing in Python

zerebubuth:
Connects to #210. Requires mapzen/TileStache#60.

Changes to the queries and data sources, bringing in a buffered land polygon based on NE 50m land. This is used to clip the admin boundaries and assign a `maritime_boundary=yes` property on those segments outside of the buffered region.

It's necessary to use a lower resolution coastline so that the intersection operations aren't time-consuming. Buffering is necessary to avoid weird artefacts at or near coastlines where the boundary runs along it or is close to parallel with it. The buffer I've used is about 3.8km (2nm), which seems to be small enough to exclude most of the close-to-shore boundaries and large enough to avoid shore artefacts. There are some, though. For screenshots, see #210.


rmarianski:
:+1:

PR #272, opened 2015-09-30
Add explicit min zoom column on landuse

zerebubuth:
Re-using and renaming the old landuse labels function. This means we can tweak the function in future, not the query. Also adds `min_zoom` as a property which is exported and can be used for client-side priority and ordering.


zerebubuth:

rmarianski:
:+1:

Issue #226, opened 2015-09-21
Add explicate building labels


nvkelso:
Like we do for Landuse labels now, we need a canonical label centroid for buildings.

This is to address the multiple duplicate building labels we're seeing today. Here's an example from @bcamper in NYC:

![screen shot 2015-09-17 at 6 50 49 pm](https://cloud.githubusercontent.com/assets/853051/9995925/8161448a-6039-11e5-87df-08f00e4ad259.png)

bcamper:
Turns out one of those is actually an additional POI label (note one is not italic)... that's also something tricky to keep in mind here. Do we try at all to "de-dupe" cases where a POI appears in building polygon of the same name? That could be problematic as well since there are legit use cases for both.

nvkelso:
Success! 

**Before:**
<img width="967" alt="screen shot 2015-10-01 at 18 00 37" src="https://cloud.githubusercontent.com/assets/853051/10237455/797ef7d2-6866-11e5-852c-b22080622de3.png">


**After:**
<img width="983" alt="screen shot 2015-10-01 at 18 00 54" src="https://cloud.githubusercontent.com/assets/853051/10237457/7d0ed05c-6866-11e5-971d-23595273e8a4.png">

PR #273, opened 2015-09-30
add post_process step to generate building labels

nvkelso:
References #226.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -168,7 +168,8 @@ post_process:
     source_layer: landuse
     label_property_name: label_position
     label_property_value: "yes"
-  TileStache.Goodies.VecTiles.transform.generate_label_features:
-    source_layer: buildings
-    label_property_name: label_position
-    label_property_value: "yes"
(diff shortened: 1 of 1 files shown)


queries.yaml
```diff
@@ -168,3 +168,7 @@ post_process:
     source_layer: landuse
     label_property_name: label_position
     label_property_value: "yes"
+  TileStache.Goodies.VecTiles.transform.generate_label_features:
```
zerebubuth: Unfortunately, `post_process` is currently a `dict`, so we can't have the same key multiple times. There's no good reason for it to be that way, it was just the easiest way at the beginning. I'm not sure whether this would require a code change - needs to be looked into.

rmarianski: Ah, nice catch.

I'd vote for making it into a list instead. I haven't checked the code, but it seems like that should be a relatively easy change to make.

rmarianski: Added some prs to make the conversion:

* https://github.com/mapzen/tilequeue/pull/32
* https://github.com/mapzen/vector-datasource/pull/275

Issue #221, opened 2015-09-17
Add explicate water area labels


nvkelso:
Just like we do for landuse-labels where we generate the centroid of the **feature** (versus the chunk that's in the tile – trying to label the chunks results in bizarre labeling today).

![screen shot 2015-09-17 at 09 43 46](https://cloud.githubusercontent.com/assets/853051/9939727/aa9d6ebc-5d20-11e5-9c94-049bc497dc64.png)

Image above from: #17/37.76963/-122.47628

Eventually we should do something fancier where we create a net of label points across the feature (just as we should for landuse labels).

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/226

nvkelso:
Better! 

Looks like the `representative point` is required to force the `label_position` to be in the feature it's labeling, though. Here the lake label is labeling some unnamed reservoir :(

<img width="397" alt="screen shot 2015-10-01 at 18 02 45" src="https://cloud.githubusercontent.com/assets/853051/10237480/e0bf9a78-6866-11e5-9d25-afc9126aa77b.png">

nvkelso:
Over to Rob to update to `representative point`.

PR #274, opened 2015-09-30
#221, add post_process step to generate water (polygon) labels

nvkelso:
Connects to #221.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -168,3 +168,7 @@ post_process:
     source_layer: landuse
     label_property_name: label_position
     label_property_value: "yes"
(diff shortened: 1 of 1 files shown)


queries.yaml
```diff
@@ -168,3 +168,7 @@ post_process:
     source_layer: landuse
     label_property_name: label_position
     label_property_value: "yes"
+  TileStache.Goodies.VecTiles.transform.generate_label_features:
```
nvkelso: Same deal @zerebubuth points out in #273 that we're not currently setup to run for more than 1 layer.

PR #279, opened 2015-10-01
Move runway lines up to zoom 9.

### LINKED ISSUE #278: show airport runway lines at zoom 9
They are coming in as minor roads, which generally don't show until z11 now.

To support Eraser Map style.

zerebubuth:
Fixes #278.

278-runway-z9-fix


nvkelso:
LGTM

Issue #187, opened 2015-09-10
Add ferry lines to roads layer


nvkelso:
Routes that go from land to water to land again show gaps in the map today for the ferry part, let's fill those in.

Read up:
https://wiki.openstreetmap.org/wiki/Tag:route%3Dferry

Sounds like it has the same general structure as roads with the highway tag but are route=ferry for the primary distinction. 

Let's brainstorm how we want these kinds to end up in tiles (we don't want them to default render as roads).

zerebubuth:
Perhaps these should end up in the transit layer, rather than the roads layer? They'd end up with `kind=ferry` by default, it looks like.

nvkelso:
We [pack](https://github.com/mapzen/vector-datasource/blob/master/data/functions.sql#L228) `railway` in the roads layer now and a good number of ferries are more part of the road network than the public transit network, so we already have the "not road things drawing sometimes as roads" in a default styling problem. 

Reading up on the wiki, it looks like ferry=* should be included (at least for a minimum set of popular values, like we currently do for roads), and with the high-road logic extended to set `kind=ferry`.

Default OSM styling shows these coming in very early (8 or 9 I remember)? Should we do that or throttle it based on the ferry=* value (like we do for roads)?

zerebubuth:
After mapzen/TileStache#64 and #281, it looks like this:

## Zoom 9
![187-sf-z9](https://cloud.githubusercontent.com/assets/271360/10234815/4e057648-688f-11e5-872d-e855904a0d9f.png)

## Zoom 10
![187-sf-z10](https://cloud.githubusercontent.com/assets/271360/10234819/509bd8b6-688f-11e5-925f-42e2a4000b9d.png)

## Zoom 11
![187-sf-z11](https://cloud.githubusercontent.com/assets/271360/10234821/53014104-688f-11e5-9775-84f8352ff7d7.png)

zerebubuth:
@nvkelso could you review, please?

nvkelso:
I dig it :)

Considerations:

1. I mentioned in the PR but we probably need a way to show select ferries
at zoom 8 (when OSM data starts coming in) as there are ferries from
Natural Earth rendered that would otherwise hide at 8, then show again at
9. See further discussion there.

2. San Francisco has a decent amount of ferries, but New York has way more.
And Athens (and Greece in general) have even more. The length buckets
should be tested in those places too.

zerebubuth:
[...]

PR #281, opened 2015-10-01
Add ferry routes

zerebubuth:
Add ferry routes to the roads layer query and level index. A couple of potentially interesting extra properties are exposed:

* `ferry`: A small proportion of ferry routes have this set to `primary`, `secondary`, etc... mirroring the road network. Unfortunately, not very many do, and none in the Bay Area or New York City.
* `motor_vehicle`: When `yes`, it's a car ferry (the default seems to be passenger-only). This might be interesting or important to expose in the cartography.

Although the `ferry` property turned out to be less useful than originally hoped, it looks like the ferry route's length is a decent proxy for how important it is, as it doesn't tend to get segmentised like the road network. So I'm using that to tell what's "important", although the values may need a little tweaking.

Connects to #187. Requires mapzen/TileStache#64.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -221,11 +221,13 @@ DECLARE
 BEGIN
   RETURN (
     CASE
-      WHEN way_length > 51356 THEN  9
-      WHEN way_length > 17215 THEN 10
-      WHEN way_length >  3400 THEN 11
-      WHEN way_length >  1700 THEN 12
-      ELSE                         13
+      -- this is about when the way is >= 2px in length
+      WHEN way_length > 1223 THEN  8
+      WHEN way_length >  611 THEN  9
+      WHEN way_length >  306 THEN 10
+      WHEN way_length >  153 THEN 11
+      WHEN way_length >   76 THEN 12
+      ELSE                        13
     END);
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -214,7 +214,23 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
 
-CREATE OR REPLACE FUNCTION mz_calculate_road_level(highway_val text, railway_val text, aeroway_val text)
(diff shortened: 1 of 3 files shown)


rmarianski:
:+1:

data/functions.sql
```diff
@@ -214,7 +214,23 @@ BEGIN
...
 
-CREATE OR REPLACE FUNCTION mz_calculate_road_level(highway_val text, railway_val text, aeroway_val text)
+CREATE OR REPLACE FUNCTION mz_calculate_ferry_level(way geometry)
+RETURNS SMALLINT AS $$
+DECLARE
+  way_length real := st_length(way);
+BEGIN
+  RETURN (
+    CASE
+      WHEN way_length > 51356 THEN  9
```
nvkelso: We should show **some** ferries at zoom 8 (when OSM starts showing) as there are a few ferries in Natural Earth we show and only including them starting at zoom 9 would introduce a gap. The ferry between the North and South Islands in New Zealand is a good example. Should this be length of way based? Or curable in the data so if `ferry=primary` then promote it up from 9 to 8? There aren't very many of these, so a little data edits in OSM paired with a commit doesn't bother me.

zerebubuth: I've changed the cutoffs in 3b66d2b to be based on the rendered size, 2px seemed like a decent level. This shows all the longer ferry routes at zoom 8 around SF, see comment below for updated screenshots.

The trouble with `ferry=primary` is that it's only set on about 100 objects at the moment, so really it's a non-standard niche tag. I don't think it's appropriate to support it or add it to OSM just so that we can render it; that's the worst kind of "tagging for the renderer". Vandalising OSM paired with a commit bothers _me_. If we want curated data with no "on the ground" basis, then we should be pulling that in from somewhere else or curating it ourselves.

Alternatively, we should think about what _physical_ attributes make a ferry route important, and I think it's closely related to "network importance", i.e: what does it do to the connectivity of the routing graph to remove it? How much extra delay, or how much of the route network is severed?

PR #285, opened 2015-10-02
Fix layer attribute on buildings

zerebubuth:
The layer tag is promoted to an explicit column by osm2pgsql and this means that it removes it from the tags hstore. Changed the query to select it from the column instead.

Also added a new filter for parsing the layer as a number, as it is much more useful to be able to do numeric comparisons with it.

Requires mapzen/TileStache#65. Connects to #194.


rmarianski:
:+1:

Issue #284, opened 2015-10-02
Shouldn't see state boundary around indian reservation


nvkelso:
8/40.982/-122.195

<img width="555" alt="screen shot 2015-10-01 at 22 44 03" src="https://cloud.githubusercontent.com/assets/853051/10240215/ffad53b8-688d-11e5-8614-4f86afc3d414.png">

zerebubuth:
Related to #214.

zerebubuth:
Should be fixed by mapzen/TileStache#66 and #286.

nvkelso:
Verified.

PR #286, opened 2015-10-02
Set kind to `aboriginal_lands` for first nations state boundaries.

zerebubuth:
Connects to #284.

Selects the `boundary:type` type to disambiguate between `admin_level=4` boundaries [like this](http://www.openstreetmap.org/way/174550914) and the ones we're used to seeing. The kind filter is modified in mapzen/TileStache#66 to use this.


rmarianski:
:+1:

PR #301, opened 2015-10-12
Update indexes to speed up queries

rmarianski:
This adds in an index on boundaries that should improve those queries. Additionally, it also stores the road and transit levels in columns, which should speed up the queries when the filter needs to be rechecked.

@zerebubuth: Could you review please?


zerebubuth:
:+1: looks good to me.

Issue #302, opened 2015-10-12
Investigate reducing the data payload from the database


rmarianski:
We're retrieving the raw geometries from the database for all features. We will want to revisit this for certain types of features, like boundaries. Note that it could be that another layer is responsible for the payload size here, I'm just guessing that it's boundaries.

zerebubuth:
There are two layers where we're going geometry matching requiring original, non-simplified geometry for it to work; boundaries and water polygons. Thinking about it a little more, the water ones would probably be okay with simplified boundaries, but it might increase the incidence of those "dashing" artefacts near overlapping polygons.

For the boundaries; at the moment, we just filter on `boundary='administrative'`. Might it be worth reducing that to `boundary='administrative' and admin_level in ('2', '4')` for the general case, allowing `'8'` for `zoom >` whatever zoom we want county boundaries for?

zerebubuth:
Also, `buffered_land` is a set of very large polygons. We should try chopping that up into a series of unioned blocks so that inland intersections don't pull in the whole coastline.

nvkelso:
Cane do more clipping (versus simplification)?

PR #304, opened 2015-10-13
Reduce boundary layer payload size.

zerebubuth:
Orient admin polygons, take their boundary and clip the result in the query instead of later in Python. This reduces the amount of data that we send over the network for boundaries. Also, take only boundaries that we actually want to display (i.e: no county boundaries at zoom < 10). Finally, clip the buffered land polygons to the query box so that we're not sending the whole land boundary over the network. This would be better done up-front, to avoid the processing cost, but we can do that as a separate thing later.

Connects to #302, requires mapzen/TileStache#71 and mapzen/tilequeue#33.

For tile 8/40/98, the boundaries layer:
* Originally sent approximately 2,991,176 bytes over the wire.
* When reduced to only `admin_level`s 2 & 4 (at this zoom), approx. 1,658,088 bytes (45% fewer).
* When the boundaries were oriented and clipped on the server, approx. 374,446 bytes (77% fewer).
* When the buffered land was clipped on the server too, approx. 26,331 bytes (93% fewer).

For reference, when the `admin_level` filter was relaxed, but the clipping was kept, this went back up to 889,979 bytes (3280% more).

I've tested this on a few tiles, and it looks okay visually for some data on my machine, but it would be better to test in dev - especially that the orientation is still correct.

### DIFF
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -62,7 +62,11 @@ SELECT
   NULL AS maritime_boundary,
   admin_level,
   tags->'boundary:type' AS boundary_type,
-  {% filter geometry %}way{% endfilter %} AS __geometry__,
+  -- note that we force the RHR, which makes outers clockwise,
+  -- then take the boundary, then reverse and intersect with
+  -- the query bbox. this is needed because Shapely expects a
+  -- correctly-oriented outer to run counter-clockwise.
+  {% filter geometry %}{{ bounds|bbox_intersection('st_reverse(st_boundary(st_forcerhr(way)))') }}{% endfilter %} AS __geometry__,
   osm_id AS __id__,
   %#tags AS tags
 
@@ -71,6 +75,7 @@ FROM planet_osm_polygon
 WHERE
   {{ bounds|bbox_filter('way') }}
   AND boundary='administrative'
+  AND admin_level IN ('2', '4' {% if zoom >= 10 %}, '8'{% endif %})
 
 UNION ALL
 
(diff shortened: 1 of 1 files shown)


nvkelso:
This needs to also land with a stable label position.

nvkelso:
Thanks for the breakdown of savings per part, btw :)

zerebubuth:
> This needs to also land with a stable label position.

Sorry, I don't understand, could you explain a bit more what you mean, please? As far as I can tell, we aren't generating point labels from boundaries. And I don't think we have any way to control the linear label position, do we?

nvkelso:
Yeah, you're right. The labels for `country` & `state` are coming from the OSM point data straight (not generated). 

For the other layer's we're generating the `label_position` as a post processing step since v0.4.0. Do you think the fix (clipping) for admin boundaries here needs to be applied to `landuse` polygons, too? If so, we'd need a solution that has stable and unique/de `label_position` points.

zerebubuth:
The same tile had the following (approximate) sizes for the other layers:

* roads: 1.29 MB
* water: 662 kB
* landuse: 412 kB
* transit: 189 kB
* boundaries: 25.7 kB
* earth: 10.2 kB
* places: 4.83 kB

It's likely that clipping could have some beneficial effect on the size of the water and landuse layers, but we'd have to pre-generate the centroids and (as far as I can tell) we can't do "representative point" on the server. Going by size, the next thing to look at might well be the roads layer; this is a zoom 8 tile and we might not need all of the minor roads at this zoom.

rmarianski:
:+1:

nvkelso:
Regarding minor roads: 
* https://github.com/mapzen/vector-datasource/issues/305

PR #306, opened 2015-10-14
Fixed invalid Antarctica polygon in buffered land.

zerebubuth:
The Antarctica polygon was self-intersecting near a part which goes below the -85 degree extent of EPSG 3785. This was causing PostGIS/GEOS to throw TopologyExceptions. I've fixed up the polygon so that it passes a local IsValid check for all polygons, and I can locally generate the tiles which were failing.


nvkelso:
Good catch! Is there an example tile that was failing?

zerebubuth:
Yeah, @rmarianski sent me 8/32/209 as an example of a failing tile. If the errors are confined to the `buffered_land` layer and the Antarctica polygon, then we should be able to regenerate them pretty quickly. I'm pretty sure we don't pre-generate many zooms deep around that part of the world.

rmarianski:
:+1:

Issue #201, opened 2015-09-11
Landuse should exclude buildings source


nvkelso:
In working on https://github.com/mapzen/vector-datasource/issues/186 we've discovered we're duplicating buildings with certain tags into the landuse layer as polygons. We shouldn't be doing that (the building should only appear in the building layer as a building feature, a flattened version of it shouldn't be copied to the landuse layers).

nvkelso:
I suspect this is the root cause of too many university icons showing up in clusters where OSM has `building=yes` and `amenity=university`, but we drop `building=yes` and end up with `amenity=university` and an area, which confuses styling down stream.

#16/37.7243/-122.4790

<img width="633" alt="screen shot 2015-10-01 at 21 44 38" src="https://cloud.githubusercontent.com/assets/853051/10239649/a8cc1c4e-6885-11e5-8ecc-638ff2422d2d.png">

nvkelso:
It's also a little weird to see `kind:university` and `landuse_kind:university`! At least for building labels, maybe for others, does it make sense to drop the feature's `kind` property if it's the same as the `landuse_kind`? 

See:

* http://vector.dev.mapzen.com/osm/all/16/10471/25342.topojson?api_key=vector-tiles-HqUVidw

zerebubuth:
There are a few things we could try for this:

1. Dropping the feature's `kind` when it `==landuse_kind`, as you suggest.
2. Don't include buildings in landuse (i.e: if `building=*` then it's a building, not a landuse polygon).
3. Apply some de-duping filters, e.g: #267.

The first option seems pretty quick, so let's start with that.

zerebubuth:
It sounds like we need to separate out three different things; landuse, building and POI status. Any OSM feature could have a mixture of all three. I've tried to collect together a set of reasonably plausible examples below, along with what we might want them to be exported as in the output.

1. [a building with no name](http://www.openstreetmap.org/way/243511759) has `building=yes`. This should result in a building polygon in the buildings layer with `kind=building`?
2. [a building with a name](http://www.openstreetmap.org/way/69707522) has `building=yes` and `name=*`. This should result in a building polygon in the buildings layer with `kind=building` and a label in the buildings layer with `kind=building`?
3. [a building with an address](http://www.openstreetmap.org/way/69707523) has `building=yes`, `addr:housenumber=*`. This should result in a building polygon in the buildings layer with `kind=building` and an address point in the buildings layer with `kind=addr` (or `kind=address`?).
[...]

PR #307, opened 2015-10-14
Sort out generated features for landuse, buildings & POIs

zerebubuth:
* Added town hall POIs.
* Updated building selection criteria - turns out there's a lot of `building=no` things.
* Separated buildings, landuse and POIs so that the name should come through in only one of these layers. The polygon geometry is only duplicated between landuse and buildings when generating a landuse label, after which the landuse polygon is discarded. This gets rid of many duplicate labels.
* Added config to generate address points for (and from) buildings and drop landuse polygons which are duplicating buildings.

Connects to #201.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -161,6 +161,7 @@ BEGIN
       WHEN historic_val = 'landmark'         THEN LEAST(zoom + 1.76, 15)
       WHEN leisure_val  = 'marina'           THEN LEAST(zoom + 3.45, 17)
       WHEN amenity_val  = 'place_of_worship' THEN LEAST(2 * zoom - 9.55, 17)
+      WHEN amenity_val  = 'townhall'         THEN LEAST(zoom + 1.85, 16)
       WHEN (barrier_val IN ('gate')
             OR craft_val IN ('sawmill')
             OR highway_val IN ('gate', 'mini_roundabout')
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -269,7 +269,11 @@ CREATE OR REPLACE FUNCTION mz_calculate_is_building_or_part(
     building_val text, buildingpart_val text)
 RETURNS BOOLEAN AS $$
 BEGIN
-    RETURN (building_val IS NOT NULL OR buildingpart_val IS NOT NULL);
+    -- there are 12,000 uses of building=no, so we ought to take that into
+    -- account when figuring out if something is a building or not. also,
+    -- returning "kind=no" is a bit weird.
+    RETURN ((building_val IS NOT NULL AND building_val <> 'no') OR
+            (buildingpart_val IS NOT NULL AND buildingpart_val <> 'no'));
 END;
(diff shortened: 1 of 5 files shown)


zerebubuth:
I should have said: Requires mapzen/TileStache#72 also.

queries/buildings.jinja2
```diff
@@ -1,13 +1,20 @@
...
+    -- building.
+    CASE WHEN mz_poi_min_zoom IS NULL AND
+              (mz_is_landuse IS NULL OR mz_is_landuse = FALSE)
+         THEN name
+         ELSE "addr:housename"
+    END AS name,
     way_area::bigint AS area,
     osm_id AS __id__,
-    "building:part",
+    'openstreetmap.org' AS source,
```
nvkelso: +1

data/functions.sql
```diff
@@ -161,6 +161,7 @@ BEGIN
       WHEN historic_val = 'landmark'         THEN LEAST(zoom + 1.76, 15)
       WHEN leisure_val  = 'marina'           THEN LEAST(zoom + 3.45, 17)
       WHEN amenity_val  = 'place_of_worship' THEN LEAST(2 * zoom - 9.55, 17)
+      WHEN amenity_val  = 'townhall'         THEN LEAST(zoom + 1.85, 16)
```
nvkelso: Good add!

data/functions.sql
```diff
@@ -269,7 +270,11 @@ CREATE OR REPLACE FUNCTION mz_calculate_is_building_or_part(
     building_val text, buildingpart_val text)
 RETURNS BOOLEAN AS $$
 BEGIN
-    RETURN (building_val IS NOT NULL OR buildingpart_val IS NOT NULL);
+    -- there are 12,000 uses of building=no, so we ought to take that into
+    -- account when figuring out if something is a building or not. also,
+    -- returning "kind=no" is a bit weird.
+    RETURN ((building_val IS NOT NULL AND building_val <> 'no') OR
+            (buildingpart_val IS NOT NULL AND buildingpart_val <> 'no'));
```
rmarianski: Nice catch!

queries/buildings.jinja2
rmarianski: For buildings the source will always be 'openstreetmap.org`, no? Is this for consistency with other layers? Or would a post process step create features that have a different source or none at all?

zerebubuth: I don't know. I _thought_ I was adding it for compatibility with the other layers, but it seems that we don't include `source` for the boundaries, earth, POIs or transit layers either. Transit and POIs are currently OSM-only, but earth and boundaries do mix data sources...

@nvkelso Should we be annotating the source for _all_ features, or just features in layers which aren't single-source?

nvkelso: In the not too distant future we may also offer Polygon City building data.
So we can either start doing this now, or then.

rmarianski:
:+1:

Issue #311, opened 2015-10-15
Update tilejson tiles url to contain api key


rmarianski:
It's more clear to update the tiles url to reflect that an api key should get passed through. Refs https://github.com/mapzen/vector-datasource/issues/310

g12n:
Great! I was quite confused about that one. Mayebe you could add the URL to rev documentation. I just remembered it because I had used it before.

nvkelso:
@g12n We made a change on the `dev` server to accommodate this. Can you try that with your config and let us know if it works? 

```
http://vector.dev.mapzen.com/osm/tilejson/mapbox.json
```

Note: When I try it in Mapbox Studio 0.2.7 it doesn't work for me, but your mileage may vary?

nvkelso:
From the Mapbox issue (651 inline above), it sounds like we're hitting a low default timeout limit in Mapbox Studio Classic because we're testing against our `dev` setup which (generally) slowly generates tiles on request instead of having them ready before request.

We should pre-generate a region of tiles and test again against that region. 

(When I retested just now I'm getting a `Server returned HTTP 503` in Mapbox Studio Classic.)

nvkelso:
Verified it's all working now :)

PR #312, opened 2015-10-15
Add api_key to tilejson tiles url

rmarianski:
Refs #311

Could this have an adverse effect on any existing tools that might read the url?


zerebubuth:
I tried looking through the "Mapbox Studio Classic" sources for information on whether this would work and how it interpolates its API keys, but I couldn't find anything that would suggest it would work for a normal URL (as opposed to a "mapbox://" one).

Having said that, :+1: let's try it and see if it breaks anything. Hopefully not!

Issue #267, opened 2015-09-29
Filter out duplicate POIs


zerebubuth:
Filtering out duplicate POIs is best done as a step near the end of processing, as we can use earlier filtering steps to discard attributes (zoom appropriate) which aren't important. This means we can merge more POIs, as there's less "different" information on them with fewer properties.

nvkelso:
This is probably why I'm seeing 2 to 3 label_positions for some churches, sports stadiums & etc. The one without area is often stacking on the one with area making debug fun.

Example:

* http://vector.mapzen.com/osm/all/16/10481/25328.topojson?api_key=vector-tiles-HqUVidw

nvkelso:
Food for thought: 

**Lincoln Park Golf Course** is a multi polygon in SF.  We probably want to allow more than one icon here, but only at later zooms (like z15+, or if the mutlipolygon parts are far apart from each other)?

https://www.openstreetmap.org/relation/2142043
#15/37.7828/-122.4939

nvkelso:
Piers in SF: http://mapzen-tiles-prod-us-east-20151007.s3.amazonaws.com/osm/all/19/83894/202606.topojson?api_key=vector-tiles-HqUVidw

nvkelso:
Church: http://mapzen-tiles-prod-us-east-20151007.s3.amazonaws.com/osm/all/20/167692/405409.topojson?api_key=vector-tiles-HqUVidw

zerebubuth:
> Church: http://mapzen-tiles-prod-us-east-20151007.s3.amazonaws.com/osm/all/20/167692/405409.topojson?api_key=vector-tiles-HqUVidw

Looks like this is fixed with the latest code:

![267-saint-pauls-church](https://cloud.githubusercontent.com/assets/271360/10524197/330637b4-7376-11e5-94e3-6cca855dc03a.png)

Does that look okay?

zerebubuth:
> Piers in SF: http://mapzen-tiles-prod-us-east-20151007.s3.amazonaws.com/osm/all/19/83894/202606.topojson?api_key=vector-tiles-HqUVidw

Just checking - what you want is for Pier 1, Pier 3 and Pier 7 to all merge together? Or was it just the building + POI label which should have been fixed by the landuse/building/POI cleanup that just got merged. Let me know - I wouldn't want to go merging things which shouldn't be merged!

[...]

PR #313, opened 2015-10-17
Filter out duplicate POIs and labels.

zerebubuth:
Uses the post-process filter defined in mapzen/TileStache#74 to remove duplicates in the landuse, landuse_labels, buildings and POIs layers.

Connects to #267.


rmarianski:
:+1:

PR #319, opened 2015-10-20
Add landuse boundary partial indexes

rmarianski:
I noticed some slow queries, and these new indexes should speed them up.


zerebubuth:
:+1: nice catch!

Issue #268, opened 2015-09-29
Too many subway stations visible at zoom 13


nvkelso:
At zoom 13 there's a crazy number of subway stations covering the map in NYC and other saturated transit markets (lucky them).

We should thin those out, only including significant stations. Since we're already using OSM area (landuse, building area) as a proxy... this probably that means 2+ lines coming together, with more lines winning over less lines (3 lines wins over 2 lines at a transfer station).

zerebubuth:
I've changed the scene file section for train stations to this, locally:

```yaml
                train-station:
                    filter: { kind: [station, train-staion] }
                    draw: { icons: { sprite: train-station } }
                    low-priority:
                      filter: { rank: { min: 9 } }
                      draw: { icons: { size: [[13, 8px], [15, 12px]] } }
```

Which means it's drawing a smaller icon for the higher ranked stations. This is the effect on NYC:

### Before, z13

![new_york_z13_before_2](https://cloud.githubusercontent.com/assets/271360/10640939/f46fea52-780e-11e5-9b31-9c7d7cce7b39.jpg)

### After, z13

![new_york_z13_after_2](https://cloud.githubusercontent.com/assets/271360/10640951/fff991a2-780e-11e5-95b1-2bb5fb188432.jpg)

### Before, z14

![new_york_z14_before_2](https://cloud.githubusercontent.com/assets/271360/10640961/08b04dcc-780f-11e5-8be1-de7249402246.jpg)

### After, z14

![new_york_z14_after_2](https://cloud.githubusercontent.com/assets/271360/10640971/0c4fbc42-780f-11e5-922e-ce9b228e9402.jpg)

nvkelso:
@zerebubuth Looks good enough to ship & refine. 

But let's call this new property `tile_rank` instead of `rank` to be more inline with `scalerank` and `labelrank` used on other features in other layers (like `places`). 

This new rank value is only valid for the feature in the current tile (the same feature could have a different rank in the next less detailed zoom or the next more detailed zoom), and we should emphasize that in the property name. All the newer properties we've added use `_` to link words (like `state_capital`). 

At some point we should go back and standardize all the old properties to this convention, but that's another issue for another day (and a breaking tiles 1.0.0 task).

zerebubuth:
Updated to `tile_rank` in #328.

nvkelso:
Turns out the PR was landed Friday and is on dev :)

[...]

PR #321, opened 2015-10-21
Too many subway stations

zerebubuth:
In some places with dense public transit infrastructure (e.g: NYC) we are displaying too many stations at zoom 13/14 and need some way to select which are the important ones.

This PR adds a query for pulling out the lines which a station is connected to by looking at the relations and ways tables. This isn't exactly fast, but there aren't very many stations in the database (fewer than 5,000 in North America), so it shouldn't have too much impact if we only run it for stations.

This PR also runs a couple of new post-processing functions to merge and rank station POIs to try and get close to some useful measure of the importance of a station - which is often an extended object with many branches, platforms and lines.

Connects to #268, requires mapzen/TileStache#75.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -463,3 +463,77 @@ BEGIN
     END;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+
+-- calculate the list of refs (or names) of subway lines that a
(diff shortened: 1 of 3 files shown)


rmarianski:
:+1:

Issue #147, opened 2015-08-26
Source neighborhoods from WOF / Quattroshapes


nvkelso:
The OSM neighborhoods are... still a work in progress and weird shit appears in NYC and other cities.

bcamper:
+++++++++++++++++++++++++

nvkelso:
Seems like a struck a nerve (in a good way)!

mjcunningham:
:+1:

nvkelso:
source: whosonfirst

rmarianski:
@nvkelso will we want a layer with just neighborhood name and label position? Or will we need the actual geometries of the neighborhoods?

nvkelso:
For this issue, we put WOF neighborhoods into 'places' layer as points alone, in addition to the OSM points, and source them to whosonfirst.

rmarianski:
At what zoom do we want to start including the WOF neighborhoods?

nvkelso:
For now, the same zoom as for OSM neighborhoods: zoom 12+. 

Since WOF includes area and relative importance, we might later throttle that so some (smaller, less important) are visible only at later zooms. But that's later optimization.

nvkelso:
Looking good. Turns out some neighborhood like things are hiding in WOF Macrohoods – I've added that work in https://github.com/mapzen/vector-datasource/issues/337.

nvkelso:
@bcamper and @mjcunningham and @sensescape Please review this in `dev`. You'll need to add `source: whosonfirst` to any neighborhood style layer to filter out the OSM features.

rmarianski:
It seems like there are only 2 rows for [macrohoods](https://github.com/whosonfirst/whosonfirst-data/blob/master/meta/wof-macrohood-latest.csv). Am I looking at the wrong place?

nvkelso:
That’s the right thing. It needs to be data gardened.

On Mon, Oct 26, 2015 at 10:12 AM, Robert Marianski <notifications@github.com

nvkelso:
Browsing around, comparing and contrasting it's clear that sometimes WOF is better, sometimes OSM is better. There are some obvious funk in WOF that is different than the obvious OSM funk. The question is do we wait for the obvious funk to resolve itself before enabling this or push it out as a way to force more WOF edits.

nvkelso:
[...]

PR #325, opened 2015-10-22
Add WOF neighbourhoods to `places` layer

rmarianski:
Refs #147. Related to https://github.com/mapzen/tilequeue/pull/36.

I spelled it "neighbourhoods" throughout to be consistent with OSM and WOF. This is also the `kind` that I used for these features.

I made the source value `whosonfirst`. Sometimes we use a url and sometimes we don't. Should we try to make it more consistent throughout all layers?

@nvkelso, @zerebubuth: could you review please?

### DIFF
--- a/data/perform-sql-updates.sh
+++ b/data/perform-sql-updates.sh
@@ -11,6 +11,7 @@ psql $@ -f apply-updates-non-planet-tables.sql &
 psql $@ -f apply-planet_osm_polygon.sql &
 psql $@ -f apply-planet_osm_line.sql &
 psql $@ -f apply-planet_osm_point.sql &
+psql $@ -f wof-schema.sql &
 wait
 echo "done."
 
--- a/data/wof-schema.sql
+++ b/data/wof-schema.sql
@@ -1,0 +1,8 @@
+CREATE TABLE wof_neighbourhood (
+  wof_id BIGINT UNIQUE,
+  name TEXT NOT NULL
+);
+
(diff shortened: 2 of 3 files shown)


nvkelso:
+1 on `'neighbourhood` AS kind and `whosonfirst` AS source. Eventually, yes (and I think it should drop the .com on some of the others).

It looks like you're referencing the `label_position`, can you confirm you're importing that from the WOF metafile instead of the basic geom?

rmarianski:
> It looks like you're referencing the label_position, can you confirm you're importing that from the WOF metafile instead of the basic geom? (Or link to a related PR that shows how that's setup.)

Sure, it's related to https://github.com/mapzen/tilequeue/pull/36. More specifically, the parsing happens [here](https://github.com/mapzen/tilequeue/blob/wof-neighborhoods/tilequeue/wof.py#L55-L61). It pulls in the `lbl_latitude`, `lbl_longitude` columns of the the [meta file](https://github.com/whosonfirst/whosonfirst-data/blob/master/meta/wof-neighborhood-latest.csv).

I've also updated the description here to reference the other pr (and vice versa).

nvkelso:
Cool :)

zerebubuth:
Do we want to run `wof-schema.sql` as part of `perform-sql-updates.sh`?

rmarianski:
Great catch! I added that in: https://github.com/mapzen/vector-datasource/commit/6de65fac8d1b40dc6123de03f1912e65b9b8090d

It reminds me that we'll also want to include a step in the getting started instructions to also import the wof data initially.

zerebubuth:
Cool, :+1: looks great now.

Issue #303, opened 2015-10-12
Water polygons sometimes missing boundary line


nvkelso:
Lake Merced in SF shows a water boundary line around one of the two lakes, but both lakes should have a boundary. 

Might be related to MultiPolygons (the missing boundary is from a MultiPolygon)?

![screen shot 2015-10-12 at 15 35 50](https://cloud.githubusercontent.com/assets/853051/10440799/41265fbc-70f7-11e5-914a-0a2f7cfead9c.png)

![screen shot 2015-10-12 at 15 36 42](https://cloud.githubusercontent.com/assets/853051/10440800/41277226-70f7-11e5-8531-201fa56686d4.png)

![screen shot 2015-10-12 at 15 36 51](https://cloud.githubusercontent.com/assets/853051/10440798/410d885c-70f7-11e5-9a6e-1545b8b5be1e.png)

nvkelso:
Seeing similar lack of water boundary features south of Washington DC around: 12/38.6837/-77.1095. Most look to be tagged `riverbank` and not part of relations.

zerebubuth:
## Lake Merced

This appears to be happening because the geometry is duplicated as part of the import: Osm2pgsql usually deleted "superseded" member geometries of relations, but will keep a constituent member geometry of a relation in the database if it has tags which aren't present on the relation. That's happening here, as the largest part of Lake Merced is tagged with a `wikipedia` tag.

I have a patch which fixes this by unioning all geometries together before separating each feature's boundaries. Unfortunately, it's about 3x slower than the existing implementation and I can't think of a way to speed it up, as most of the time is spent in the `unary_union` operation. Given that the exterior boundaries is already one of the slowest operations we perform on a tile, this means it could make the whole process about 2-3x slower.

Alternatives such as simply excluding relations (or non-relations) won't work, as water polygons are sometimes mapped as either, and constituent parts are generally superseded. What's the trade-off between getting it right and getting it fast?

@rmarianski can you think of anything I've missed / any other ideas worth trying?

## South of DC

@nvkelso, I can't find the area you mention - I've checked along from Pohick Bay to Gravelly Point - I can't see any missing boundaries. Could you point out a particular feature that's missing that I can check, please?

rmarianski:
[...]

PR #326, opened 2015-10-23
Enable snapping tolerance for exterior boundaries post-processor.

zerebubuth:
Requires mapzen/TileStache#76, connects to #303.

Config change to enable snapping coordinates to grid before deriving the boundary.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -160,6 +160,7 @@ post_process:
         id: true
         source: true
         boundary: "yes"
+      snap_tolerance: 0.125
   - fn: TileStache.Goodies.VecTiles.transform.overlap
     params:
       base_layer: buildings


queries.yaml
```diff
@@ -160,6 +160,7 @@ post_process:
         id: true
         source: true
         boundary: "yes"
+      snap_tolerance: 0.125
```
nvkelso: Is this in tile (pixel/point) coordinates space or?

zerebubuth: It's in coordinate space, in this case Mercator "meters".

nvkelso: What zoom might a static tolerance start causing us problems, 19 or 20?

zerebubuth: Should be okay across all zooms, since neither the calculation nor the tolerance depends on the zoom.

zerebubuth: I should add, 0.125 Mercator "meters" is <1px until zoom 21. I certainly didn't see any visual artefacts from this when I checked down to z19. If we want to fix it for >z21, then we'll need to take another approach - snapping with a smaller distance leaves the gap intact.

rmarianski:
:+1:

Issue #308, opened 2015-10-15
Railways need hierarchy


nvkelso:
What's the main line, service line, siding, etc?

See Hoboken terminal in NYC area.

#16/40.7357/-74.0346
<img width="888" alt="screen shot 2015-10-15 at 02 36 33" src="https://cloud.githubusercontent.com/assets/853051/10507388/b4b511fe-72e5-11e5-9761-b1482a4cffab.png">

zerebubuth:
After #327, and using the following scene file snippet:

```yaml
        rail:                                                                                        
            filter: { kind: railway }                                                                
            draw:                                                                                    
                lines:                                                                               
                    color: grey                                                                      
                    width: [[14,1px],[16,2px],[18,5m]]                                               
            service:                                                                                 
              filter: { service: true }                                                              
              draw: { lines: { visible: false } }                                                    
```

Then it looks like this:

![hoboken_rail_terminal_z16](https://cloud.githubusercontent.com/assets/271360/10700915/dc677b4a-79b8-11e5-8388-2f0a3864799b.jpg)

The common values for `service` when `railway=rail` appear to be `yard`, `spur` and `siding`. Additionally, `highway=service` uses the `driveway`, `parking_aisle`, `alley`, `parking` and `drive-through` values pretty commonly, so this'll also be helpful in adding some hierarchy to lower levels of the road network. See [taginfo](http://taginfo.openstreetmap.org/keys/service#values) for all the values for `service=*`.

nvkelso:
I like adding [`service`](https://wiki.openstreetmap.org/wiki/Key%3Aservice) as an attribute in the tiles (like we currently do for `highway` and `railway`. 

But this needs to be paired with changes to the High Road classifier logic, too. Since that affects both rail and roads let's track it in a new issue: https://github.com/mapzen/vector-datasource/issues/332.

nvkelso:
Verified. 

We should also consider grouping these `minor_rail` into their own kind, filed #345 to track that.

PR #327, opened 2015-10-23
Add service tag to roads query

zerebubuth:
This gives a second level of information for highway and railway features, and allows us to distinguish some kinds of secondary railways.

Connects to #308.


rmarianski:
:+1:

PR #328, opened 2015-10-23
Call the rank property tile_rank instead.

zerebubuth:
As per https://github.com/mapzen/vector-datasource/issues/268#issuecomment-150382161.

Connects to #268.


rmarianski:
:+1:

Issue #188, opened 2015-09-10
Add city wall lines


nvkelso:
Old cities have old walls, let's show them on the map:

Background reading:

https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcity_wall

nvkelso:
The `boundaries` layer seems to make the most sense for this.

zerebubuth:
After #329, and with the following scene file snippet:

```yaml
        city_walls:
          filter: { kind: city_wall }
          draw:
            lines:
              interactive: true
              order: 8
              color: '#ff0000'
              width: 4px
```

Around [George's Island](https://www.openstreetmap.org/#map=18/44.64117/-63.55951), I get:

## Before

![georges_island_z18_before](https://cloud.githubusercontent.com/assets/271360/10703299/f86d8970-79c6-11e5-912f-2ef0ca6eb348.jpg)

## After

![georges_island_z18_after](https://cloud.githubusercontent.com/assets/271360/10703303/fe3d99bc-79c6-11e5-89df-664f234ea999.jpg)

There are some other features visible in the OSM carto map, which are tagged as `barrier=retaining_wall`. Let me know if you want them included too.

nvkelso:
Might as well pick those up at the same time, but seems like their default zoom should be later than city walls?

nvkelso:
Retaining walls and paths seem like similar level of detail. We show Paths starting at zoom 15, let's match that.

nvkelso:
@zerebubuth what zoom do city walls start coming in? Zoom 10+ seems like a good starting point.

zerebubuth:
City walls start coming in at z14, which matches OSM Carto. Having them come in earlier is definitely possible, although z10 sounds a bit early to me. For comparison z10 is when we start to include county boundaries and secondary roads in the tiles. @nvkelso: What's the appropriate zoom to start bringing them in?

zerebubuth:
Screenshot after 9fc86df in #329. The thinner red lines are the retaining walls.

![georges_island_z18_after_2](https://cloud.githubusercontent.com/assets/271360/10728821/1995d90a-7bdd-11e5-9778-ad1b85adfba0.jpg)

nvkelso:
They should be available for styling by **zoom 12** (that's when we show minor roads, and these play into that hierarchy). We might only show them in style at zoom 14, but generally our practice is to include features a few zooms earlier than we need them.

zerebubuth:
[...]

PR #329, opened 2015-10-23
Add city walls from z14 onwards.

zerebubuth:
Pulls in both `barrier=city_wall` and `historic=citywalls` from zoom 14 onwards, same as OSM carto.

Connects to #188.

### DIFF
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -104,7 +104,9 @@ SELECT
   NULL AS maritime_boundary,
   NULL AS admin_level,
   NULL AS boundary_type,
-  'city_wall' AS kind,
+  CASE WHEN barrier='retaining_wall' THEN 'retaining_wall'
+       ELSE 'city_wall'
+  END AS kind,
   {% filter geometry %}way{% endfilter %} AS __geometry__,
   osm_id AS __id__,
   %#tags AS tags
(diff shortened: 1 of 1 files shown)


queries/boundaries.jinja2
```diff
@@ -84,6 +85,7 @@ SELECT
   'yes' AS maritime_boundary,
   NULL AS admin_level,
   NULL AS boundary_type,
+  NULL AS kind,
```
rmarianski: I'm not familiar with the data, does it make sense to fold the `boundary_type` into the `kind` here?

zerebubuth: We only want to fold in a [specific value](https://github.com/mapzen/TileStache/blob/integration-1/TileStache/Goodies/VecTiles/transform.py#L378) from `boundary:type` into `kind`. We could put it in a `CASE WHEN ... END AS kind` statement, but since it's already implemented in Python I figured it eas just easier to keep the queries the way they are for the moment. What do you think?

nvkelso: As Matt points out we're doing the [kind assignment](https://github.com/mapzen/TileStache/blob/integration-1/TileStache/Goodies/VecTiles/transform.py#L361-L366) in Python and not in the SQL query. That mostly seems to be working for existing boundary features (like country, state, etc), and if it needs refinement let's file a different issue for it.

rmarianski: Sounds good. :+1:

zerebubuth:
@rmarianski, just checking, is this okay to merge?

rmarianski:
Yup, merge away! :zap:

Issue #212, opened 2015-09-15
filter buildings by volume (and area)


meetar:
< zoom 15, many prominent buildings in downtown NYC are filtered out because they have relatively small footprints, eg One World Trade – maybe we could use building height to filter them by volume, either instead of or in addition to area, to keep some of these around.

bcamper:
Great catch, this transition has been bothering me but it didn't click as to why.

nvkelso:
Nice! 

If building has height data to calculate volume, filter on volume. Else still filter on building (footprint) area.

@meetar and @bcamper what volumes do you recommend per zoom?

meetar:
After some tests, I'd suggest these minimum volumes:
z13: 300000
z14: 200000
z15: 100000

nvkelso:
Are the volumes naive footprint area * height?

meetar:
Yep, I was going to dunk each building into a water tank to measure the displacement but somebody else is using the 3D printer today :D

nvkelso:
Ha!

nvkelso:
We should filter what goes into tiles at the above zooms based on the calculated volume.

zerebubuth:
## Zoom 13

### Before

![nyc_before_z13](https://cloud.githubusercontent.com/assets/271360/10738464/7da51f78-7c0f-11e5-9a9d-8cbc4b0be1ce.jpg)

### After

![nyc_after_z13](https://cloud.githubusercontent.com/assets/271360/10738467/8176dc90-7c0f-11e5-8c6f-048975354c2d.jpg)

## Zoom 14

### Before

![nyc_before_z14](https://cloud.githubusercontent.com/assets/271360/10738471/84ef0852-7c0f-11e5-90b6-36c34f1d9c87.jpg)

### After

![nyc_after_z14](https://cloud.githubusercontent.com/assets/271360/10738473/875a961a-7c0f-11e5-80a5-b4c53ae6286a.jpg)

## Zoom 15

### Before

![nyc_before_z15](https://cloud.githubusercontent.com/assets/271360/10738474/8b21807e-7c0f-11e5-9f6b-36447c3c3fc3.jpg)

### After

![nyc_after_z15](https://cloud.githubusercontent.com/assets/271360/10738480/8e30c392-7c0f-11e5-9225-ea0d17d639ab.jpg)

nvkelso:
LGTM, thanks! Let's take it to `dev` so we can QA wider sample of locations (Midtown NYC etc).

[...]

PR #340, opened 2015-10-26
Filter buildings by volume, if possible.

zerebubuth:
Complexify the building filter logic and extract some of it to a separate function for maintainability. This allows us to do an approximate filter on building volume, without proper tag parsing. That can be done in the transform stage, in Python, so that the client can do a proper, final filter before display.

Connects to #212.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -592,3 +592,46 @@ BEGIN
     );
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+-- returns true if the text is numeric and can be cast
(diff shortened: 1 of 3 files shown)


rmarianski:
If we're doing the filter here, will we also need it in python too? Will the python do a more thorough or accurate check and this is meant to drop the obvious outliers?

zerebubuth:
Requires mapzen/TileStache#81 since ec1d6e3.

rmarianski:
:+1:

PR #350, opened 2015-10-27
Updates for transit routes

zerebubuth:
Update query function to include light rail, tram and main line rail routes. Split the `ref`, as this seems to be sometimes comma-delimited. Renamed `subway_lines` to `transit_routes`, as it's collecting more than just subway lines now.

Note that the name of the function changed, so the migration should run `functions.sql` and then:

```SQL
DROP FUNCTION mz_calculate_subway_lines(bigint);
```

Connects to #268.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -534,12 +534,12 @@ END;
 $$ LANGUAGE plpgsql IMMUTABLE;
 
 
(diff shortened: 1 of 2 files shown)


rmarianski:
:+1:

Issue #333, opened 2015-10-24
why is mz_is_building: false in tiles?


nvkelso:
I thought that was only an internal server processing step, not intended for export?

Example:

http://vector.dev.mapzen.com/osm/all/15/5235/12671.topojson?api_key=vector-tiles-HqUVidw

nvkelso:
eg: 

```
landuse_labels: {
type: "GeometryCollection",
geometries: [
{
type: "Point",
properties: {
label_placement: "yes",
kind: "grass",
name: "Maloney Field",
area: 16376,
mz_is_building: false,
min_zoom: 10.2564,
source: "openstreetmap.org",
sport: "baseball",
id: -4792969,
osm_relation: true
},
coordinates: [
413,
774
]
},
```

zerebubuth:
After mapzen/TileStache#84 and #351, now I'm not getting that internal property any more:

```json
  "landuse_labels": {
    "type": "GeometryCollection",
    "geometries": [
      {
        "type": "Point",
        "properties": {
          "label_placement": "yes",
          "kind": "grass",
          "name": "Maloney Field",
          "area": 16376,
          "source": "openstreetmap.org",
          "min_zoom": 10.2564,
          "sport": "baseball",
          "id": -4792969,
          "osm_relation": true
        },
        "coordinates": [
          413,
          774
        ]
      },
```

nvkelso:
Verified.

PR #351, opened 2015-10-27
Drop internal mz_is_building key

zerebubuth:
It is useful only for later processing of polygons, not labels.

Requires mapzen/TileStache#84. Connects to #333.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -189,12 +189,14 @@ post_process:
       source_layer: landuse
       label_property_name: label_placement
       label_property_value: "yes"
+      drop_keys: [ mz_is_building ]
   - fn: TileStache.Goodies.VecTiles.transform.generate_label_features
     params:
       source_layer: landuse
       new_layer_name: landuse_labels
       label_property_name: label_placement
       label_property_value: "yes"
+      drop_keys: [ mz_is_building ]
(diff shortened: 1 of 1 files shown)


rmarianski:
:+1:

queries.yaml
```diff
@@ -189,12 +189,14 @@ post_process:
...
       label_property_name: label_placement
       label_property_value: "yes"
+      drop_keys: [ mz_is_building ]
   - fn: TileStache.Goodies.VecTiles.transform.generate_label_features
     params:
       source_layer: landuse
       new_layer_name: landuse_labels
       label_property_name: label_placement
       label_property_value: "yes"
+      drop_keys: [ mz_is_building ]
```
nvkelso: Is this supposed to be `drop_keys: [ mz_is_landuse]`?

nvkelso: At any rate, it looks to produce the right results, but I wonder if legibility could be improved?

zerebubuth: D'oh. That's just me typo'ing the title of the PR, it should have been `mz_is_building`, same as the parent issue #333. As far as I know, we don't have a property called `mz_is_landuse`, but if you see something like that in the output, we can now easily clean it out :smile:

Issue #266, opened 2015-09-29
Density filter for POIs / labels


zerebubuth:
As a post-process filter, we should be able to control the density of POIs and labels, preventing a whole bunch of them from rendering on top of one another. Client-side clipping will help, but it would be better to not send all of this data to begin with.

The issue arises because many buildings and other features are legitimately clustered, and each may have its own label / icon. However, once zoomed out these become all mashed together. The fix has to be something that's "zoom sensitive" as it doesn't make sense to merge these POIs at all zooms.

One possible implementation is to divide up the tile into NxN blocks (possibly even one 1x1) and rank the POIs within each block, possibly by type, taking only the most prominent and/or collapsing several of the same type into each other. For example, when binning mountain peaks, order by elevation and take only the highest within the block.

zerebubuth:
We kind-of have this with the `keep_n_features` filter implemented as part of #268. We should evaluate the fix for #268 when it hits dev and come up with some specific examples to work on.

zerebubuth:
Example: Lombard Motor Inn, see [these](https://github.com/mapzen/vector-datasource/issues/320#issuecomment-151364692) [comments](https://github.com/mapzen/vector-datasource/issues/320#issuecomment-151461566) for more details.

zerebubuth:
After mapzen/TileStache#85 and #354:

### Before

![lombard_motor_inn_z18_before](https://cloud.githubusercontent.com/assets/271360/10797176/519475ce-7d99-11e5-8d51-4400838892fa.jpg)

### After

Only one "Lombard Motor Inn"!

![lombard_motor_inn_z18_after](https://cloud.githubusercontent.com/assets/271360/10797190/5abeb9b6-7d99-11e5-8b35-7d399f4ce144.jpg)

nvkelso:
Super cool!

zerebubuth:
I'm seeing this for the Everett Middle School / Sanchez Elementary School examples now:

![everett_middle_school_z17_after](https://cloud.githubusercontent.com/assets/271360/10800099/213f5b32-7da8-11e5-8412-243a5cb0bc21.jpg)

nvkelso:
Great! Let's QA this in dev now :)

nvkelso:
Huge improvement overall :)

One problem so far, though...

Looks like we're missing the `California Academy of Sciences` in SF (`kind=museum`) in dev versus prod.

* #17/37.76983/-122.46657
[...]

PR #354, opened 2015-10-28
Filter duplicates by name across the POIs, buildings and landuse labels.

zerebubuth:
Adds configuration for mapzen/TileStache#85.

Connects to #266.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -251,3 +251,9 @@ post_process:
       items_matching:
         kind: station
       rank_key: tile_rank
+  - fn: TileStache.Goodies.VecTiles.transform.remove_duplicate_features
+    params:
+      source_layers: [pois, buildings, landuse_labels]
+      property_keys: [name]
+      geometry_types: [Point]
+      min_distance: 32.0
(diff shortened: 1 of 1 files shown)


queries.yaml
```diff
@@ -251,3 +251,9 @@ post_process:
       items_matching:
         kind: station
       rank_key: tile_rank
+  - fn: TileStache.Goodies.VecTiles.transform.remove_duplicate_features
+    params:
+      source_layers: [pois, buildings, landuse_labels]
```
nvkelso: add landuse?

zerebubuth: Because we duplicate landuse labels into both the landuse and landuse_labels layers, then this filter would end up removing one or the other layer's labels entirely. If you want to continue having labels in both landuse and landuse_layers, then we should copy them as a later step.

I forget why we are duplicating this information anyway?

nvkelso: > I forget why we are duplicating this information anyway?

When we started packing label positions into their source layer, we had to also continue supporting the legacy `landuse-labels` layer so developers wouldn't see regressions. But in the release notes we said we'd remove `landuse-labels` after a few releases. 

That sounds more like a breaking v1.0 change to me, so in the mean time we should copy them in a later step?

zerebubuth: Oh, yeah... okay. I'll add that too.

rmarianski:
:+1:

Issue #361, opened 2015-10-30
Sql error with mz_building_filter


rmarianski:
Error fetching: 13/4447/2879 - Traceback (most recent call last): |   File "/usr/local/lib/python2.7/dist-packages/tilequeue/worker.py", line 132, in __call__    fetch_data = self.fetcher(coord) |   File "/usr/local/lib/python2.7/dist-packages/tilequeue/query.py", line 187, in __call__    raise async_exception | DataError: invalid input syntax for integer: "2.5"CONTEXT:  PL/pgSQL function mz_building_filter(text,text,double precision,double precision,double precision) line 3 at RETURN

nvkelso:
@rmarianski  isn't seeing this error in the logs anymore.

PR #363, opened 2015-10-30
It's possible to have fractional levels, who knew?

zerebubuth:
Rather than cast to `bigint`, we can just cast to `float`, since we know that will work. Fractional levels aren't really a big problem, as we do some transforms and compare to another float anyway.

Connects to #361.


rmarianski:
:+1:

bcamper:
Welcome to the 7.5 floor
https://www.youtube.com/watch?v=xT104QX31QA

PR #365, opened 2015-11-02
Update WOF table schema

rmarianski:
Updates the WOF neighbourhood table schema to add the new required columns.

Connects with #147 


zerebubuth:
:+1: looks good to me.

Issue #376, opened 2015-11-03
Transit stations missing routes


zerebubuth:
In San Francisco, Bart lines are missing from the transit_lines because the data in OSM isn't there to support it. Same for London and Paris.

zerebubuth:
NOTE: _Some_ data is present in SF, but it's mostly on Muni / Caltrain stations. Looks like someone needs to edit the BART stations!

```
   osm_id   |  network  |           name           |    routes     
------------+-----------+--------------------------+---------------
 2150077206 | BART      | 16th Street Mission      | {}
 1723738813 | Muni      | Embarcadero              | {L,J,K,N,M,T}
  313885839 |           | West Portal Muni         | {L,K,M,T}
  297863037 |           | Van Ness Muni            | {L,J,K,N,M,T}
 1564479547 |           | West Portal Muni         | {L,K,M,T}
  301506011 |           | Church St Muni           | {L,K,M,T}
 2160213364 | Caltrain  | 22nd Street              | {100s,200s}
  297863017 |           | Castro Muni              | {L,K,M,T}
 2150077208 | BART      | 24th Street Mission      | {}
 3760504178 |           | Taraval St & Sunset Blvd | {}
 2150077198 | BART;Muni | Embarcadero              | {}
 2150077201 | BART;Muni | Montgomery Street        | {}
 3396509241 | Muni      | Cable Car Loading Area   | {}
 2150077204 |           | Powell Street            | {}
 2150077196 | BART;Muni | Civic Center / UN Plaza  | {}
  301506348 |           | Forest Hill Muni         | {}
```

nvkelso:
Seems like this should go into `data-tasks` repo instead?

PR #377, opened 2015-11-03
Fix transit routes in London & Paris

zerebubuth:
It seems that `public_transport=stop_position` is a widely-used variant for the node representing the stop, but it was previously only looking for `public_transport=stop` (and a couple of other tags). This was preventing it from picking up any transit routes for many stops.

Before this change, in London there were 209 stops without any routes (of 533 total). In Paris, 101 empty of 766 total. In Berlin, 44 empty of 289 total. In Chicago, 44 empty of 77 total. In NYC, 47 empty of 663 total. (Note that these areas are what was in the bounding box I chose, not any meaningful boundary of the city, so don't read too much into them).

After this change, In London 186 empty of 533 total. In Paris, 54 empty of 766 total. In Berlin, 39 empty of 289 total. In Chicago, 11 empty of 77 total. In NYC, 44 empty of 663 total.

The biggest gains were in Paris and Chicago, although some of the 23 improved stations in London were quite prominent (Euston, Paddington). After investigating a few of the stations in London still without `transit_routes`, it seems like (as with BART stations in SF) the issue is that there's no data to connect the railway station node to the route relations which give us the transit information.

Connects to #376.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -577,7 +577,7 @@ BEGIN
     ON r.node_id = n.id
     WHERE (
       mz_rel_get_tag(n.tags, 'railway') IN ('station', 'stop') OR
-      mz_rel_get_tag(n.tags, 'public_transport') = 'stop')
+      mz_rel_get_tag(n.tags, 'public_transport') IN ('stop', 'stop_position'))
     -- manually re-include the original station node, in case it's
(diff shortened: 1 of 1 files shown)


rmarianski:
:+1:

PR #380, opened 2015-11-03
Update WOF places layer

rmarianski:
* Depends on https://github.com/mapzen/vector-datasource/pull/379 (first commit in this pr is shared with that pr)
* Connects with https://github.com/mapzen/TileStache/pull/91
* Refs:
  - https://github.com/mapzen/vector-datasource/issues/337
  - https://github.com/mapzen/vector-datasource/issues/338
  - https://github.com/mapzen/vector-datasource/issues/359

This pr boils down to:

* `kind_tile_rank` is used as the property that can be used in the stylesheet to determine importance, and is present on WOF features.
* `n_photos` and `area` are also included in the properties for WOF features, which may be useful for testing in the interim. Both can be easily removed.
* WOF neighbourhoods are included at z11 now.
* OSM neighbourhoods are no longer included.

Implementation notes:

* The area column is pre-computed in the process that inserts the WOF data into the database. It seemed like a better approach to just query for these values rather than include a separate column in the query to return back another geometry (the label point counts as the geometry we use for display), with which we would only compute the area anyway. If we need to start doing other things with the raw geometry during processing, I think we can revisit this decision. Otherwise I think relatively speaking we'll pay for the IO without much gain. Alternatively we could just call `ST_Area` at query time, but because we'll be querying much more than writing, this seemed like a reasonable trade-off.
* I added a "lookup" table from placetype code -> string. Placetypes are stored as integers as an optimization, and they represent whether the feature is a microhood, macrohood, or neighbourhood. We could also just have a `case` statement in the query instead of doing this join at query time, or we could have a transform take care of it, which might centralize this logic more. Not sure if that would actually save any time though, but it might be better to maintain that way.
* There are additional changes forthcoming, like adding `min_zoom` that we will start picking up from WOF, but I figured that we might be better off tackling those in a separate pr.

@zerebubuth, @nvkelso: could you review please?

### DIFF
--- a/data/wof-schema.sql
+++ b/data/wof-schema.sql
@@ -4,6 +4,7 @@ CREATE TABLE wof_neighbourhood (
   name TEXT NOT NULL,
   hash TEXT NOT NULL,
   n_photos INTEGER,
+  area BIGINT,
   label_position geometry(Point, 900913) NOT NULL,
   geometry geometry(Geometry, 900913) NOT NULL
 );
--- a/data/wof-schema.sql
+++ b/data/wof-schema.sql
@@ -12,3 +12,13 @@ CREATE TABLE wof_neighbourhood (
 CREATE INDEX wof_neighbourhood_wof_id_index ON wof_neighbourhood(wof_id);
 CREATE INDEX wof_neighbourhood_label_position_index ON wof_neighbourhood USING GIST(label_position);
 CREATE INDEX wof_neighbourhood_geometry_index ON wof_neighbourhood USING GIST(geometry);
+
+CREATE TABLE wof_neighbourhood_placetype (
+  placetype_code SMALLINT UNIQUE NOT NULL,
+  placetype_string text NOT NULL
+);
+
+INSERT INTO wof_neighbourhood_placetype VALUES
+  (1, 'neighbourhood'),
+  (2, 'microhood'),
+  (3, 'macrohood');
--- a/data/wof-schema.sql
+++ b/data/wof-schema.sql
@@ -1,6 +1,16 @@
+CREATE TABLE wof_neighbourhood_placetype (
+  placetype_code SMALLINT PRIMARY KEY,
+  placetype_string text NOT NULL
+);
+
+INSERT INTO wof_neighbourhood_placetype VALUES
+  (1, 'neighbourhood'),
+  (2, 'microhood'),
+  (3, 'macrohood');
+
 CREATE TABLE wof_neighbourhood (
-  wof_id BIGINT UNIQUE NOT NULL,
-  placetype SMALLINT NOT NULL,
+  wof_id BIGINT PRIMARY KEY,
+  placetype SMALLINT NOT NULL REFERENCES wof_neighbourhood_placetype(placetype_code),
   name TEXT NOT NULL,
   hash TEXT NOT NULL,
   n_photos INTEGER,
@@ -9,16 +19,4 @@ CREATE TABLE wof_neighbourhood (
   geometry geometry(Geometry, 900913) NOT NULL
 );
 
-CREATE INDEX wof_neighbourhood_wof_id_index ON wof_neighbourhood(wof_id);
 CREATE INDEX wof_neighbourhood_label_position_index ON wof_neighbourhood USING GIST(label_position);
-CREATE INDEX wof_neighbourhood_geometry_index ON wof_neighbourhood USING GIST(geometry);
-
-CREATE TABLE wof_neighbourhood_placetype (
-  placetype_code SMALLINT UNIQUE NOT NULL,
-  placetype_string text NOT NULL
-);
-
-INSERT INTO wof_neighbourhood_placetype VALUES
-  (1, 'neighbourhood'),
-  (2, 'microhood'),
-  (3, 'macrohood');
--- a/queries.yaml
+++ b/queries.yaml
@@ -246,13 +246,10 @@ post_process:
         kind: [neighbourhood, microhood, macrohood]
       rank_key: kind_tile_rank
       start_zoom: 11
-  # not sure if we want to drop n_photos or not, might be useful for
-  # testing in the interim
-  # - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
+  # - fn: TileStache.Goodies.VecTiles.transform.drop_properties
   #   params:
   #     source_layer: places
-  #     property_name: n_photos
-  #     drop_property: true
+  #     properties: [mz_n_photos]
   - fn: TileStache.Goodies.VecTiles.transform.remove_duplicate_features
     params:
       source_layers: [pois, landuse, buildings]
--- a/queries.yaml
+++ b/queries.yaml
@@ -239,6 +239,20 @@ post_process:
       items_matching:
         kind: station
       rank_key: tile_rank
+  - fn: TileStache.Goodies.VecTiles.transform.rank_features
+    params:
+      source_layer: places
+      items_matching:
+        kind: [neighbourhood, microhood, macrohood]
+      rank_key: kind_tile_rank
+      start_zoom: 11
+  # not sure if we want to drop n_photos or not, might be useful for
+  # testing in the interim
+  # - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
+  #   params:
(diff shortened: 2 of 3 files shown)


data/wof-schema.sql
```diff
@@ -11,3 +12,13 @@ CREATE TABLE wof_neighbourhood (
...
 CREATE INDEX wof_neighbourhood_label_position_index ON wof_neighbourhood USING GIST(label_position);
 CREATE INDEX wof_neighbourhood_geometry_index ON wof_neighbourhood USING GIST(geometry);
+
+CREATE TABLE wof_neighbourhood_placetype (
+  placetype_code SMALLINT UNIQUE NOT NULL,
+  placetype_string text NOT NULL
+);
+
+INSERT INTO wof_neighbourhood_placetype VALUES
+  (1, 'neighbourhood'),
```
nvkelso: What happens when we import **all** of WOF place types? Should we shove them all into the same table or different tables?

queries/places.jinja2
```diff
@@ -49,17 +53,17 @@ WHERE
...
     AND {{ bounds|bbox_filter('way') }}
 
-{% if zoom >= 12 %}
+{% if zoom >= 11 %}
 UNION ALL
 
 SELECT
     name,
-    'neighbourhood' AS kind,
+    wnp.placetype_string AS kind,
```
nvkelso: what is `wnp`? Well known place type?

queries/places.jinja2
```diff
@@ -70,9 +74,13 @@ SELECT
     NULL AS population,
     NULL AS capital,
     NULL AS state_capital,
+    n_photos,
```
nvkelso: I don't know what `n` means. Should we say `n_photos AS photos,`?

queries/places.jinja2
```diff
@@ -70,9 +74,13 @@ SELECT
...
+    area::bigint AS area,
     NULL AS tags
 
-FROM wof_neighbourhood
+FROM wof_neighbourhood wn
+
+INNER JOIN wof_neighbourhood_placetype wnp ON wn.placetype = wnp.placetype_code
```
nvkelso: Oh, okay.

data/wof-schema.sql
```diff
@@ -11,3 +12,13 @@ CREATE TABLE wof_neighbourhood (
 CREATE INDEX wof_neighbourhood_wof_id_index ON wof_neighbourhood(wof_id);
 CREATE INDEX wof_neighbourhood_label_position_index ON wof_neighbourhood USING GIST(label_position);
 CREATE INDEX wof_neighbourhood_geometry_index ON wof_neighbourhood USING GIST(geometry);
+
+CREATE TABLE wof_neighbourhood_placetype (
+  placetype_code SMALLINT UNIQUE NOT NULL,
```
zerebubuth: Can we spell `UNIQUE NOT NULL` as `PRIMARY KEY`?

zerebubuth: Do we also want to make `wof_neighbourhood`'s `placetype SMALLINT NOT NULL REFERENCES wof_neighbourhood_placetype(placetype_code)` so that Postgres will check the FK relationship for us? It's of marginal benefit, given the small size of the table, but might catch an error later down the line.

queries.yaml
```diff
@@ -239,6 +239,20 @@ post_process:
...
+  - fn: TileStache.Goodies.VecTiles.transform.rank_features
+    params:
+      source_layer: places
+      items_matching:
+        kind: [neighbourhood, microhood, macrohood]
+      rank_key: kind_tile_rank
+      start_zoom: 11
+  # not sure if we want to drop n_photos or not, might be useful for
+  # testing in the interim
+  # - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
```
zerebubuth: This is a bit confusing; it's dropping the property, but not the feature, because `n_photos` is never identically `True`. Do we need a separate function for just dropping properties, or perhaps just improve `drop_features_where` to be clearer about what's going on? Perhaps that's a separate issue anyway...

queries/places.jinja2
zerebubuth: `n` is "number of", usually, and that seems to fit in this situation. If we're not planning to expose this publicly (except perhaps during development) then perhaps we should put an `mz_` prefix on it?

data/wof-schema.sql
rmarianski: I'd probably have it be driven off of which layer it gets shown in. If it's not in `places`, then I'd start creating other tables for it. Or consider a more general structure for it if that makes more sense.

queries.yaml
rmarianski: Yes, what you describe was my intent. I could create a separate function to make it clearer that it was only dropping properties.

queries/places.jinja2
rmarianski: I was just taking the prefixes from the table name, ie `w`of_`n`eighbourhood_`p`lacetype. In this case I could have just used `placetype_string` because there were no collisions, but it was the only column from the other table so I figured that this might call it out.

Should that prefix be dropped? Updated to be something else? Or just use the whole other table name?

queries/places.jinja2
rmarianski: Yup, it was meant to be "number of". It will be exposed currently because I'm not dropping it, but it's easy to drop it downstream if that's what we wanted. I figured I would include it since we're including area, at least in the interim for testing purposes.

I like the idea of adding an `mz_` prefix to it if we're going to drop it from the output.

If we want to keep it, is it better to call this `num_photos` instead? Or just `photos` like you mentioned?

data/wof-schema.sql
rmarianski: Good ideas. I will add a commit to implement.

rmarianski: @zerebubuth should I make the `wof_id` in the `wof_neighbourhood` table also be a `primary key`? It's also set to `unique not null` at the moment.

zerebubuth: Is there any reason we'd want to insert multiple records with the same `wof_id` into the database - perhaps to de-multi multipolygons?

queries/places.jinja2
nvkelso: the `kind_tile_rank` property should already allow clients to sort / collide / filter neighborhood features, and is based on `n_photos`. Let's only expose this to verify during dev, but remove before taking it to prod. It's not useful to general consumers (but `transit_lines` is in the other issue).

data/wof-schema.sql
rmarianski: I'm leaning towards guarding against duplicates for the time being until we need them. Locally, the fact that it's required to be unique caught a data problem. I'm betting having a neighbourhood multipolygon is a rare thing anyway.

zerebubuth: Cool, sounds like `PRIMARY KEY` is the right thing to do then.

data/wof-schema.sql
nvkelso: Makes sense, carry on.

data/wof-schema.sql
nvkelso: There are cases now in the data where we do have neighborhoods with a MultiPolygon (where there is one record, but the geom has two distinct, non-contiguous area geoms).

queries/places.jinja2
nvkelso: How about `wof_np` instead?

data/wof-schema.sql
rmarianski: Good to know. I'm assuming it still has a single label position though?

queries/places.jinja2
rmarianski: That works.

data/wof-schema.sql
nvkelso: Yes, only a single label position. 

In 95% of these cases, it's funky data that should be cleaned up (the other part of the MultiPolygon is probably some other neighborhood).

rmarianski:
I think I addressed all the comments now. Ok to merge in now?

The new commits depend on these prs:

* https://github.com/mapzen/TileStache/pull/93
* https://github.com/mapzen/TileStache/pull/94

nvkelso:
I dig it, let's take these to `dev`.

Issue #346, opened 2015-10-27
how to filter an underground station 'building'


musicformellons:
This should be possible now, but this does not work:

    "id": "buildings",
    "type": "fill",
    "source": "osm",
    "source-layer": "buildings",
    "filter": ["any", ["!=", "level", "-1"], ["!=", "location", "underground"]],

So how to do it?

nvkelso:
@musicformellons Can you provide more background, please? What styling language are you writing this in, is it Hoverboard?

musicformellons:
@nvkelso Sure! This was a fragment of my Mapbox GL Style Reference (version 8). A working example (just please fill in [SECRETMAPBOXTOKEN] and [SECRETMAPZENKEY]) is here:
https://gist.github.com/musicformellons/e9d56e9cbef52ae36fd0

The rectangular building in the middle (when using the start coordinates and zoom) is an underground station level -1 which still shows although I tried filtering it in the "id": "buildings" - layer. So how can it be filtered?

zerebubuth:
Ah, looks like we have a typo in our documentation (#388). Thanks for reporting this, as we wouldn't have spotted that otherwise!

The [Mapbox GL docs](https://www.mapbox.com/mapbox-gl-style-spec/#filter) (note that you'll have to scroll all the way to the bottom to get to the filter section, see mapbox/mapbox-gl-style-spec#375 for why) says "The binary and set operators implement strictly-typed comparisons", so `-1` needs to be a number in the filter. Also, `layer`s other than `-1` are possible, so we should account for that with something like this:

```
"filter": ["none", ["<", "layer", 0], ["==", "location", "underground"]],
```

I changed the filter to a "none" because I think `layer` is optional, and not specified for most ground-level buildings, so requiring `layer >= 0` would have dropped those erroneously. However, anything with `layer < 0` seems likely to be underground. I haven't actually tested that filter, so apologies if it still doesn't work, but hopefully it's helpful.

musicformellons:
Works now! Great, thanks!

PR #388, opened 2015-11-05
Correct typo level -> layer

zerebubuth:
Fixed docs - tags added in #245 were 'location' and 'layer', not 'level'. Refs #346.


nvkelso:
Good catch. I've also updated the release notes for v0.4: https://github.com/mapzen/vector-datasource/releases/tag/v0.4.0.

PR #341, opened 2015-10-26
Add aerialways to roads layer.

zerebubuth:
Adds aerialways to the roads query. This means altering the roads level, again. The migration I used was to update the functions, then run:

```SQL
DROP TRIGGER mz_trigger_line ON planet_osm_line;

UPDATE planet_osm_line SET
  mz_road_level = mz_calculate_road_level(highway, railway, aeroway, route, service, aerialway, way)
  WHERE
    aerialway IS NOT NULL;
    
CREATE OR REPLACE FUNCTION mz_trigger_function_line()
RETURNS TRIGGER AS $$
BEGIN
    NEW.mz_road_level := mz_calculate_road_level(NEW.highway, NEW.railway, NEW.aeroway, NEW.route, NEW.service, NEW.aerialway, NEW.way);
    NEW.mz_transit_level := mz_calculate_transit_level(NEW.route);
    RETURN NEW;
END;
$$ LANGUAGE plpgsql VOLATILE;

CREATE TRIGGER mz_trigger_line BEFORE INSERT OR UPDATE ON planet_osm_line FOR EACH ROW EXECUTE PROCEDURE mz_trigger_function_line();

DROP FUNCTION mz_calculate_road_level(text, text, text, text, text, geometry);
```

Requires mapzen/TileStache/pull/82. Connects to #189.

@rmarianski could you review, please?

PR #401, opened 2015-11-10
Add aerialway argument to mz_calculate_road_level in line table update trigger

moskvax:
This makes the trigger declaration consistent with the arity change to mz_calculate_road_level in #341. I've also added "'s to the column references for the sake of stylistic consistency.


rmarianski:
Nice catch. Thanks!

Issue #412, opened 2015-11-26
Missing service roads that are parking_aisles


nvkelso:
At the San Francisco Zoo and other landuse AOIs with parking lots, we're missing the service roads that are tagged `service=parking_aisles`. These were in v0.4 tiles, maybe this regressed as part of the zoom range refactor?

http://www.openstreetmap.org/#map=18/37.73372/-122.50607

For example, this way is missing: http://www.openstreetmap.org/way/344535940

<img width="1163" alt="screen shot 2015-11-26 at 00 28 53" src="https://cloud.githubusercontent.com/assets/853051/11418465/d2026224-93d4-11e5-8e71-16832708dbf3.png">

zerebubuth:
D'oh, there was a typo. This should be fixed when #413 is rolled out.

nvkelso:
Verified in prod.

PR #413, opened 2015-11-26
Fixed a typo in 'parking_aisle'

zerebubuth:
The zoom selection statement had a space where it should have had an underscore. I checked the other values, and they seem to be correct (`drive-through` is most commonly used with a hyphen).

Connects to #412.

NOTE: This needs a migration. Something like `UPDATE planet_osm_line SET mz_road_level = mz_calculate_road_level("highway", "railway", "aeroway", "route", "service", "aerialway", "way") WHERE highway='service' AND service='parking_aisle'` should work.


rmarianski:
:+1:

PR #416, opened 2015-11-30
Configure transform to make population an integer

rmarianski:
Refs: mapzen/vector-datasource#230
Connects with: https://github.com/mapzen/TileStache/pull/97

@zerebubuth: could you review please?


zerebubuth:
:+1:

Issue #342, opened 2015-10-26
Add piste routes to road layer


nvkelso:
We have paths, we're adding aerialway line features for the ski chair lifts and such, but we don't have the downhill, x-country & etc ski routes (e.g.: `route=piste`) that everyone goes to the mountain for in the first place. 

Core tags:

- [ ] piste:type > piste_type
- [ ] piste:difficulty > piste_difficulty
- [ ] piste:grooming > piste_grooming

Other tags that might be interesting:

- [ ] piste:name - sometimes the piste route name is different than the basic feature name?
- [ ] piste:abandoned - if it's abandoned, don't export it in tiles?
- [ ] ski=designated or ski=yes as appropriate if tagged on a path or other way not marked piste
- [ ] snowshoe=designated or showshoe=yes as appropriate if tagged on a path or other way not marked piste
- [ ] distance=number - official length of the complete route, useful for text labeling
- [ ] symbol=piste markings - textual description, useful for symbol labeling

We already include these types of tags on road layer features:

- [ ] name=* name of the piste (if it has one)
- [ ] ref=number reference code for the piste
- [ ] operator=* organisation that is responsible for the provision and maintenance of the route
- [ ] network=* name of the network of routes of which this one is part
- [ ] colour=* - colour normally associated with the route when presented on a map

See:

* http://wiki.openstreetmap.org/wiki/Piste_Maps
* http://wiki.openstreetmap.org/wiki/Proposed_features/Tag:route%3Dpiste
* http://wiki.openstreetmap.org/wiki/WikiProject_Piste_Maps
* http://wiki.openstreetmap.org/wiki/OpenPisteMap
* http://www.opensnowmap.org

zerebubuth:
@nvkelso One thing I've found; we're not catching pistes which are not "interesting" in some other way - most appear to be highways (paths, usually) or have names, but some, such as [way 365603700](http://www.openstreetmap.org/way/365603700) don't. Currently we're missing them from the rendering database, and it would take some fiddling to include them. Can I assume it's okay to go ahead with the knowledge that we'll be missing some (about 35%) of these ways, or do we need to address that first?

nvkelso:
@zerebubuth If it's ≤ day's work then add them in now, else make a new issue (and relate it to this one) and we'll tackle adding in those extra ones in a future release. Consider the total time to include ops time to deploy this to prod, not just dev the feature.

zerebubuth:
[...]

PR #422, opened 2015-12-04
Add piste routes

zerebubuth:
Adds piste routes. These are queried out as a separate set of geometries, as this seems to make the most sense when there are items in the database (including a few tertiary roads!) which are tagged as both roads and pistes. Rather than pass the difficulty of handling things which are both to the client / style, we can duplicate the feature and then hopefully 99% of styles will "just work" where there's a road and a piste in the same place.

This needs a small migration:

```SQL
BEGIN;

CREATE INDEX planet_osm_line_piste_geom_index ON planet_osm_line USING gist(way) WHERE tags ? 'piste:type';

COMMIT;
```

Connects to #342 and requires mapzen/TileStache#98.

### DIFF
--- a/data/apply-planet_osm_line.sql
+++ b/data/apply-planet_osm_line.sql
@@ -9,7 +9,7 @@ ALTER TABLE planet_osm_line ADD COLUMN mz_road_level SMALLINT;
 ALTER TABLE planet_osm_line ADD COLUMN mz_transit_level SMALLINT;
 
 UPDATE planet_osm_line
-  SET mz_road_level = mz_calculate_road_level(highway, railway, aeroway, route, service, aerialway, tags->'piste:type', way)
+  SET mz_road_level = mz_calculate_road_level(highway, railway, aeroway, route, service, aerialway, way)
   WHERE mz_calculate_road_level(highway, railway, aeroway, route, service, aerialway, tags->'piste:type', way) IS NOT NULL;
 
 UPDATE planet_osm_line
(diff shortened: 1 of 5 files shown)


data/apply-planet_osm_line.sql
```diff
@@ -33,6 +33,8 @@ CREATE INDEX planet_osm_line_transit_geom_15_index ON planet_osm_line USING gist
 
 CREATE INDEX planet_osm_line_waterway_geom_index ON planet_osm_line USING gist(way) WHERE waterway IN ('canal', 'river', 'stream', 'dam', 'ditch', 'drain');
 
+CREATE INDEX planet_osm_line_piste_geom_index ON planet_osm_line USING gist(way) WHERE tags ? 'piste:type';
```
rmarianski: Should we make the index here a partial on the geometry that matches the query filter exactly?

rmarianski:
:+1:

data/apply-planet_osm_line.sql
zerebubuth: I did that initially, and then worried that the list of `piste:type` that I pulled from the top 10 on [taginfo](http://taginfo.openstreetmap.org/keys/piste:type#values) was incomplete. Checking on taginfo, it seems like the top 10 encompasses 95% or more of the data, so including all those values wouldn't bloat the index too much.

Do you think it should be more precise?

rmarianski: Whoops, I read this too quickly and didn't realize that this already was a partial on the geometry and not just on `piste:type`. I think it's fine the way it is. :+1:

Issue #160, opened 2015-09-02
Add POIs for road exits


nvkelso:
These are in highway:motorway_junction
and the ref stores the exit number.

nvkelso:
cc @sensescape

nvkelso:
These will need a special type of road shield, cc @bcamper and @meetar.

zerebubuth:
With an appropriate icon and entry in the scene file:

```yaml
        exits:
          filter: { highway: motorway_junction }
          draw:
            icons:
              size: 21px
              sprite: wings-joint
              interactive: true
              visible: true
            text:
              visible: true
              interactive: true
              text_source: ref
              font:
                fill: *text_fill
                size: 12px
                stroke: { color: *text_stroke, width: 4 }
```

It looks like this:

![us_101_exit_433-after](https://cloud.githubusercontent.com/assets/271360/11637375/daac27d2-9d18-11e5-8fc2-976b6f455f5b.jpg)

zerebubuth:
With mapzen/TileStache#100, the `kind` is now `exit` rather than `minor_road` as in the screenshot above.

nvkelso:
Is “wings-joint” a nod to self-driving / flying cars?! :)

zerebubuth:
Sure, let's say that :wink:

The truth is far more prosaic; I was scrolling down the list of available icons and thought "Oh, I wonder what that one looks like."

nvkelso:
Should this go into the `pois` layer instead of the `roads` layer? 

Setting the `kind=exit` seems okay, as long as the `highway=motorway_junction` isn't used for other things besides exits, and we preserve that tag in the tile data?

zerebubuth:
With wing joints at zoom 12:

![san-jose-z12](https://cloud.githubusercontent.com/assets/271360/11698006/7d63b908-9eb4-11e5-8bec-bfd061d894b9.jpg)

There is a wing joint at every one of those exits, right?

nvkelso:
Hopes KFC!

nvkelso:
I don't see these in `dev` yet.

zerebubuth:
Argh, sorry. Looks like another casualty of the migration mistake corrected in #436. We'll get it in the migration later today.

nvkelso:
I see these in tiles now: http://vector.dev.mapzen.com/osm/all/12/655/1583.topojson?api_key=vector-tiles-HqUVidw.

[...]

PR #426, opened 2015-12-07
Add highway exit points to the roads layer.

zerebubuth:
This adds the points and an index over them.

**Note** that these are the first points to be added to the road layer, so might cause issues for anyone who isn't expecting them.

Also requires a migration to create the index: `CREATE INDEX planet_osm_point_motorway_junction_index ON planet_osm_point USING gist(way) WHERE highway='motorway_junction';`

Connects to #160.


rmarianski:
:+1:

Issue #244, opened 2015-09-25
Add platforms to transit layer


nvkelso:
As referenced in https://github.com/mapzen/vector-datasource/issues/194:

BART platforms should really be marked `railway:platform` and be included in the `transit` layer.

See how to do it right in Moscow:
http://www.openstreetmap.org/#map=17/55.75196/37.60592

<img width="673" alt="screen shot 2015-09-24 at 22 26 20" src="https://cloud.githubusercontent.com/assets/853051/10093728/88e1c354-630b-11e5-9b90-2d7b6748b0bf.png">

zerebubuth:
After #429, and with a wholly appropriately festive style:

```yaml
    platforms:
      data: { source: osm, layer: transit }
      filter: { kind: platform }
      draw:
        lines:
          color: '#ff0000'
          width: 5m
          order: 10
        polygons:
          color: '#ff0000'
          order: 10
```

## Glen Park station

Before:
![glen-park-z18-before](https://cloud.githubusercontent.com/assets/271360/11658665/ad4c1510-9dbb-11e5-851c-8c4148a46f82.jpg)

After:
![glen-park-z18-after](https://cloud.githubusercontent.com/assets/271360/11658672/b216a4b6-9dbb-11e5-97bc-15d82055c6c0.jpg)

(Drawing under the buildings seemed like the right thing to do here, and is the default anyway. But worth noting that the OSM map draws platforms, even underground ones, above buildings. Of course, this is under the control of the style, but I thought it was worth mentioning in case we look at a place where we're expecting to see platforms but they're rendered underneath a building.)

## Atlantic Center station

Before:
![atlantic-center-z18-before](https://cloud.githubusercontent.com/assets/271360/11658675/b6fcb696-9dbb-11e5-89b3-2cf4fa5bb090.jpg)

After:
![atlantic-center-z18-after](https://cloud.githubusercontent.com/assets/271360/11658678/ba9253ce-9dbb-11e5-9ac3-980aa97d933e.jpg)

nvkelso:
@zerebubuth What zoom would these start showing up with your PR, zoom 16? 

Since these are in the transit layer (and not more general landuse or roads layers) I'd like to see them start at zoom 14. This will allow for constructing transit maps that highlight platforms on the zoomed out map, can be very effective visual technique.

nvkelso:
Verified, including starting at zoom 14.

PR #429, opened 2015-12-08
Added platforms to the transit layer.

zerebubuth:
Note that platforms can be either lines or polygons, and both are about equally common (2,279 lines, 1,612 polygons in North America) so will need handling in the style.

Needs a migration to add indexes. I figure the indexes are worthwhile because platforms are pretty rare, so scanning a whole block of line & polygon features would mean a lot more work per-query.

```SQL
CREATE INDEX planet_osm_line_railway_platform_index ON planet_osm_line USING gist(way) WHERE railway='platform';
CREATE INDEX planet_osm_polygon_railway_platform_index ON planet_osm_polygon USING gist(way) WHERE railway='platform';
```

Connects to #244.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -137,7 +137,7 @@ layers:
   transit:
     template: transit.jinja2
     start_zoom: 6
-    geometry_types: [LineString, MultiLineString]
+    geometry_types: [LineString, MultiLineString, Polygon, MultiPolygon]
     transform:
       - TileStache.Goodies.VecTiles.transform.tags_create_dict
       - TileStache.Goodies.VecTiles.transform.tags_name_i18n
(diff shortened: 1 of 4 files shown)


queries.yaml
```diff
@@ -145,6 +145,7 @@ layers:
       - TileStache.Goodies.VecTiles.transform.add_id_to_properties
       - TileStache.Goodies.VecTiles.transform.detect_osm_relation
       - TileStache.Goodies.VecTiles.transform.route_name
+      - TileStache.Goodies.VecTiles.transform.parse_layer_as_float
```
rmarianski: Nice re-use! :zap:

queries/transit.jinja2
```diff
@@ -23,3 +24,47 @@ FROM planet_osm_line
...
+FROM (
+  SELECT osm_id, way, name,ref, operator, route, railway, layer, tags
+  FROM planet_osm_line
+  WHERE
+    {{ bounds|bbox_filter('way') }}
+    AND railway='platform'
+
+  UNION ALL
+
+  SELECT osm_id, way, name,ref, operator, route, railway, layer, tags
```
rmarianski: Nitpicking on spacing for `name,ref` here and above :trollface:

rmarianski:
:+1:

queries/transit.jinja2
zerebubuth: Yup, well spotted, thanks! Fixed in caf55dd.

Issue #344, opened 2015-10-26
Add winter sports related POIs


nvkelso:
Pairing with https://github.com/mapzen/vector-datasource/issues/342 and https://github.com/mapzen/vector-datasource/issues/343, we should add a few more POIs for winter sports.

- [ ] Sled rental (rental=sled or amenity=sled_rental)
- [ ] Ski rental (rental=ski or amenity=ski_rental)
- [ ] Ski school (amenity=ski_school)
- [ ] Mountain rescue and ski patrol (amenity=mountain_rescue)
- [ ] Bobsled track (sport=bobsled)
- [ ] Ski playgrounds (leisure=ski_playground)

Any others seem interesting? Add them to the list :)

zerebubuth:
`sport=bobsled` doesn't appear to be in the database at all ([taginfo](http://taginfo.openstreetmap.org/tags/sport=bobsled)), but there are 8 instances of `sport=luge`, so we can use that instead. The sled rental tags only have a single use apiece, which makes me think they're not well-used. However, there are 1,603 uses of `man_made=snow_cannon` and 1,485 for `man_made=snow_fence`.

We're going to have the same problem for `rental` keys as we had for #423 - since that key isn't in `osm2pgsql.style` we aren't guaranteed to get all of them in the database. We should get most, though, as rental shops probably have names.

nvkelso:
Luge instead of Bobsled sounds fine. Let's skip the sled rental. Good catch on the snow cannon and snow fence!

File related issue for the rental keys?

zerebubuth:
Screenshots for a couple of winter sports POI types:

## Ski shop in Big Bear:

Scene file snippet:

```yaml
                ski-rental-or-shop:
                  filter:
                    any:
                      - kind: [ski, ski_rental]
                      - { kind: sports, sport: skiing }
                  draw: { icons: { sprite: marijuana-dispensary } }
```

![big-bear-z18-after](https://cloud.githubusercontent.com/assets/271360/11693359/7108bb76-9e9c-11e5-9f96-08008be9551e.jpg)

## Ski jumping "racetracks" in Lake Placid

Scene file snippet:

```yaml
        racetrack:
          filter: { kind: racetrack }
          draw:
            lines:
              interactive: true
              order: 5
              cap: round
              join: round
              color: '#cc3333'
              width: [[13,3px],[15, 4px], [16, 5px], [18, 10m], [19, 15m]]
```

[...]

PR #432, opened 2015-12-09
Added winter sports POIs and lines.

zerebubuth:
Adds features for ski shops, rental places and schools, as well as snow cannon POIs. Also adds support for racetracks (requires mapzen/TileStache#102) and snow fences.

Requires a migration so large, it has its own [gist](https://gist.github.com/zerebubuth/729fc099de7401ea86cd).

Connects to #344.


rmarianski:
:+1:

PR #434, opened 2015-12-09
Move highway exits to POIs layer

zerebubuth:
Move highway exit points to the POIs layer, as they were the only points in the roads layer and it's lower risk of breaking someone else's map to have them in POIs.

Migration, which rolls back part of the migration for #426, is [here](https://gist.github.com/zerebubuth/f8eacf4ed65ea00589fa).

Connects to #160.


zerebubuth:
I should note, this also changes the min zoom for the exit POIs to 12.

rmarianski:
:+1:

PR #435, opened 2015-12-09
Platforms in transit layer should start at zoom 14.

zerebubuth:
Connects to #244.


zerebubuth:
Note, this doesn't need a migration other than the one already in to add an index.

rmarianski:
:+1:

Issue #369, opened 2015-11-03
Start showing (subway) stations at zoom 12 instead of 13


nvkelso:
With the changes in https://github.com/mapzen/vector-datasource/issues/368, we have enough signal to consider showing subway stations at zoom 12 (instead of zoom 13+).

zerebubuth:
A selection of stations are now visible at zoom 12, following the changes in #436:

![new-york-city-z12](https://cloud.githubusercontent.com/assets/271360/11716283/9d84b5b6-9f40-11e5-990a-cfa07312a6cf.jpg)

nvkelso:
Looking at this zoom 12 tile in NYC (Wall Street area), I don't see stations? Nor anywhere else on the?
http://vector.dev.mapzen.com/osm/all/12/1205/1540.topojson

nvkelso:
WFM in SF, NYC, and Paris.

PR #436, opened 2015-12-10
Show subway stations starting at zoom 12

zerebubuth:
Start showing all subway stations at zoom 12 in the tile data. The style then uses a cut-off based on the `kind_tile_rank` to choose which are visible.

Note that this PR includes a bunch of commits related to adding `rental` to the parameters of `mz_calculate_poi_level`, which should have been done as part of #432, but looks like it was missed out from that. This also [needs a migration](https://gist.github.com/zerebubuth/6b9e0fd7b7cdba5de117).

Connects to #369.


rmarianski:
:+1:

Issue #398, opened 2015-11-07
Export flight codes on airports


nvkelso:
For example, `SFO` for "San Francisco International Airport". This will allow labeling at mid zooms where the full name is too long / distracting.

nvkelso:
(Not the longer airport codes, the ones for commercial airports.)

zerebubuth:
After #438, mapzen/TileStache#103 and changing the `text_source` for the `early` airport rule to `iata` instead of `ref`, it now shows:

![sfo-z11](https://cloud.githubusercontent.com/assets/271360/11721685/f0f4f726-9f5c-11e5-8f58-c52ebd4daa9a.jpg)

zerebubuth:
Not sure why Oakland's "OAK" label isn't showing, but it definitely has `iata: OAK` as part of the props, and "NUQ" (a.k.a Moffett Federal Airfield) is showing near Mountain View.

nvkelso:
I remember OAK is on a tile boundary and JS Tangram (WebGL) has issues with
placing text along those lines.

nvkelso:
Verified in SF Bay Area, NYC, Paris.

PR #438, opened 2015-12-10
Run IATA code processing over POIs to include airports.

zerebubuth:
Requires mapzen/TileStache#103. Connects to #398.


rmarianski:
:+1:

Issue #367, opened 2015-11-03
Missing landuse=military AOIs


nvkelso:
The Old Soldiers Home in DC (15/38.9403/-77.0123) is `landuse=military` in OSM but it's missing from tiles:

http://vector.dev.mapzen.com/osm/all/17/37496/50121.topojson?api_key=vector-tiles-HqUVidw
http://vector.mapzen.com/osm/all/17/37496/50121.topojson?api_key=vector-tiles-HqUVidw

Looking at the queries it seems to be selecting correctly? Where else in the processing could it be dropped?

<img width="915" alt="screen shot 2015-11-02 at 21 13 37" src="https://cloud.githubusercontent.com/assets/853051/10901671/83bd1ee0-81a7-11e5-89fa-d1a8b12cfbae.png">

nvkelso:
Related the the missing `school` labels (https://github.com/mapzen/vector-datasource/issues/417), we're missing a whole swath of military features that are just points:

https://taginfo.openstreetmap.org/keys/military

![screen shot 2015-12-04 at 09 46 57](https://cloud.githubusercontent.com/assets/853051/11596788/1af554ee-9a6c-11e5-9fb6-598e92e44239.png)

zerebubuth:
Before:

![us-soldiers-and-airmens-home-z15-before](https://cloud.githubusercontent.com/assets/271360/11722317/fa7cb542-9f5f-11e5-8322-12f6accceb51.jpg)

After:

![us-soldiers-and-airmens-home-z15-after](https://cloud.githubusercontent.com/assets/271360/11722323/00a6b454-9f60-11e5-823c-0991e5fc5bd9.jpg)

nvkelso:
Verified in SF Bay Area and Washington DC areas.

PR #439, opened 2015-12-10
Fix missing military landuse areas.

zerebubuth:
Added missing comma. Not sure how that ever worked...

Note: a [migration](https://gist.github.com/zerebubuth/a9f344be44a5f98acaf7) is necessary.

Connects to #367.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -10,7 +10,7 @@ BEGIN
         landuse_val IN ('park', 'forest', 'residential', 'retail', 'commercial',
                         'industrial', 'railway', 'cemetery', 'grass', 'farmyard',
                         'farm', 'farmland', 'wood', 'meadow', 'village_green',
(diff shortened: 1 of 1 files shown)


rmarianski:
Nice catch! Is it possible that it was concatenating the two strings together? I see that the migration is updating both military and rural just in case.

:+1:

zerebubuth:
Yeah, it looks like it was concatenating. But this syntax only appears to work at the end of a line... It seems very strange to me:

```
mapzen=# select 'foo''bar';
 ?column? 
----------
 foo'bar
(1 row)

mapzen=# select 'foo' 'bar';
ERROR:  syntax error at or near "'bar'"
LINE 1: select 'foo' 'bar';
                     ^
mapzen=# select 'foo'
mapzen-#  'bar';
 ?column? 
----------
 foobar
(1 row)
```

Issue #382, opened 2015-11-04
Add manmade=pier to roads layer


nvkelso:
**Piers** (that are only line ways, not area landuse relations) are missing from tiles around Woodley Island in Eureka, California (16/40.80763/-124.16253) and Sausalito, California (16/37.8665/-122.4888).

* https://wiki.openstreetmap.org/wiki/Tag%3Aman_made%3Dpier
* https://taginfo.openstreetmap.org/tags/?key=man_made&value=pier&filter=ways#map
* https://www.openstreetmap.org/#map=16/37.8680/-122.4958

<img width="848" alt="screen shot 2015-11-03 at 23 41 48" src="https://cloud.githubusercontent.com/assets/853051/10932300/96060764-8284-11e5-81de-aee9c0484df8.png">

They should probably end up as `kind=path` with `manmade=pier` attributes in High Road, with z-order and min_zoom similar to `highway=pedestrian`.

zerebubuth:
After the above PRs, it looks like this:

## Sausalito

Before:
![sausalito-z16-before](https://cloud.githubusercontent.com/assets/271360/11750260/5120659a-a02a-11e5-9034-0ec0e1bdab28.jpg)

After
![sausalito-z16-after](https://cloud.githubusercontent.com/assets/271360/11750265/53f51b6c-a02a-11e5-8f46-94662c0eafeb.jpg)

## Woodley Island

Before
![woodley-island-z16-before](https://cloud.githubusercontent.com/assets/271360/11750266/579feae4-a02a-11e5-94f7-697f36f595c8.jpg)

After
![woodley-island-z16-after](https://cloud.githubusercontent.com/assets/271360/11750267/5a989822-a02a-11e5-9707-900ae6d40d77.jpg)

zerebubuth:
Note: the added bit in the scene file:

```yaml
            pier:
              filter: { man_made: pier }
              draw:
                lines:
                  color: *grey3
                  width: [[13, 0px], [14, 1.5px], [16, 5px], [18, 10m]]
                  outline:
                    color: *grey3
```

nvkelso:
Fantastic!

nvkelso:
Confused why I only see these sparsely at zoom 15 (and not 16, 17, 18...): 

#15/37.8651/-122.4919

zerebubuth:
[...]

PR #441, opened 2015-12-11
Add piers to roads layer.

zerebubuth:
Return pier lines in the roads layer. Polygons are already returned in landuse.

Needs a [migration](https://gist.github.com/zerebubuth/639e127f2eeae99597fe).

Requires mapzen/TileStache#104, connects to #382.

@rmarianski, could you review, please?


rmarianski:
:+1:

PR #451, opened 2015-12-16
Add first attempt to automate running migrations.

zerebubuth:
There's a shell script, `data/migrations/run_migrations.sh`, which will run the `functions.sql` and `triggers.sql` and, if both exited cleanly, run every `data/migrations/*.sql` in parallel. By convention, we might want to keep one file per table, plus additional ones for anything "global" like dropping old function signatures.

The whole thing is supposed to be idempotent, and I can run it against my local database repeatedly okay.

### DIFF
--- a/data/migrations/run_migrations.sh
+++ b/data/migrations/run_migrations.sh
@@ -24,5 +24,5 @@ wait
 
 # re-enable triggers
 for table in planet_osm_point planet_osm_line planet_osm_polygon; do
-    psql -c "ALTER TABLE ${table} DISABLE TRIGGER USER" $*
+    psql -c "ALTER TABLE ${table} ENABLE TRIGGER USER" $*
 done
--- a/data/migrations/run_migrations.sh
+++ b/data/migrations/run_migrations.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+migration_dir=${0%/*}
+
+# first run functions and triggers, bailing if either of these fail, as they
+# are required by later steps.
+psql --set ON_ERROR_STOP=1 -f "${migration_dir}/../functions.sql" $*
+if [ $? -ne 0 ]; then echo "Installing new functions failed.">&2; exit 1; fi
+psql --set ON_ERROR_STOP=1 -f "${migration_dir}/../triggers.sql" $*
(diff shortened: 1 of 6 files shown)


data/migrations/run_migrations.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ $? -ne 0 ]; then echo "Installing new triggers failed.">&2; exit 1; fi
+
+# then disable triggers
+for table in planet_osm_point planet_osm_line planet_osm_polygon; do
+    psql -c "ALTER TABLE ${table} DISABLE TRIGGER USER" $*
+done
+
+# run updates in parallel. note that we don't bail here, as we want to
+# re-enable the triggers regardless of whether we failed or not.
+for sql in ${migration_dir}/*.sql; do
```
rmarianski: Do we want to specify the files that we want to run? Or maybe put all the files for a release in a new directory? I'm imagining that we'll end up having more than one release over time, and we wouldn't want to run the migrations for the previous ones.

zerebubuth: I figured that we could only migrate to the "current" release anyway, because we only keep single copies of `functions.sql` and `triggers.sql`. We could keep older versions of these files, but I figure if I need to migrate to `v0.5.3` then I can always checkout that tag.

Therefore, if we can only migrate to the "current" release, we don't need to version the migration files and we should remove them all in the commit following the release tag by which point they're only of historical interest.

What do you think?

rmarianski: Ah, great point. I'm thinking that we should make it a part of the release process to truncate these files in a commit after the tag has been created.

rmarianski:
:+1:

nvkelso:
Rob can you add this new workflow to the wiki please?

rmarianski:
Here's my first crack at it: https://github.com/mapzen/vector-datasource/wiki/Mapzen-Vector-Tile-Service#tracking-releases

Feel free to review and edit.

nvkelso:
Nice! Workflow wise, when a release has been tagged, should we immediately followup with a commit to clear the migration config on master?

rmarianski:
Yea, I was thinking that too. :+1:

PR #452, opened 2015-12-16
Skip "cleanup" files.

zerebubuth:
As @rmarianski spotted, we sometimes need to delay part of the migration (e.g: dropping functions used in queries) until after the code has been deployed. This allows that by skipping any file in the migrations directory with "cleanup" as part of the name.

### DIFF
--- a/data/migrations/run_migrations.sh
+++ b/data/migrations/run_migrations.sh
@@ -17,7 +17,11 @@ done
 # run updates in parallel. note that we don't bail here, as we want to
 # re-enable the triggers regardless of whether we failed or not.
 for sql in ${migration_dir}/*.sql; do
-    psql -f "$sql" $* &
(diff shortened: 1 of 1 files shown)


data/migrations/run_migrations.sh
```diff
@@ -17,7 +17,11 @@ done
 # run updates in parallel. note that we don't bail here, as we want to
 # re-enable the triggers regardless of whether we failed or not.
 for sql in ${migration_dir}/*.sql; do
-    psql -f "$sql" $* &
+    if [[ $sql = *cleanup*.sql ]]; then
```
rmarianski: I'm not sure, do we have to quote the `*`s here?

zerebubuth: No, because we're using bash's globbing - if we quoted we'd be comparing against a file called `"*cleanup*.sql"`

Issue #443, opened 2015-12-14
Add swimming pools in with water


nvkelso:
We're not picking up `leisure=swimming_pool` as a water like feature:
https://www.openstreetmap.org/way/189470798

<img width="609" alt="screen shot 2015-12-13 at 21 52 25" src="https://cloud.githubusercontent.com/assets/853051/11773814/e501a156-a1e3-11e5-9741-2a5aefdd6985.png">

nvkelso:
Looks like there are two ways of tagging swimming pools (also: `amenity=swimming_pool`):
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dswimming_pool

zerebubuth:
Before:

![cupertino-z18-before](https://cloud.githubusercontent.com/assets/271360/12117138/7a01a97a-b3b6-11e5-9f97-ac2760b3afbd.jpg)

After:

![cupertino-z18-after](https://cloud.githubusercontent.com/assets/271360/12117141/7f788284-b3b6-11e5-8399-a8578dd3c437.jpg)

nvkelso:
Preview images look good.

nvkelso:
@zerebubuth I see this swimming pool at zoom 19 tiles, but not zoom 18 or zoom 17. Can you confirm you see the same pattern?

* z18 good:  http://vector.dev.mapzen.com/osm/all/19/84425/203482.topojson
* z18 missing? http://vector.dev.mapzen.com/osm/all/18/42212/101741.topojson
* z17 missing? http://vector.dev.mapzen.com/osm/all/17/21106/50870.topojson

zerebubuth:
Same issue as before: I forgot to [update fastly](https://github.com/mapzen/fastly-deployer/pull/9) after the last dev release.

nvkelso:
Verified, pool party!

PR #456, opened 2015-12-21
Add swimming_pool

rmarianski:
Refs #443


data/functions.sql
```diff
@@ -371,13 +371,16 @@ END;
...
-    waterway_val text, natural_val text, landuse_val text)
+    amenity_val text, landuse_val text, leisure_val text,
+    natural_val text, waterway_val text)
 RETURNS BOOLEAN AS $$
 BEGIN
     RETURN (
         waterway_val IN ('riverbank', 'dock')
      OR natural_val IN ('water')
      OR landuse_val IN ('basin', 'reservoir')
+     OR amenity_val='swimming_pool'
```
nvkelso: We usually say `OR amenity_val IN ('swimming_pool')`, is there a performance reason not to do that here?

nvkelso:
Before / after images?

data/functions.sql
zerebubuth: I don't think there's a significant performance difference.

nvkelso:
LGTM, let's merge this.

Issue #374, opened 2015-11-03
Add electronics shops to POIs.


zerebubuth:
> Best Buy in San Francisco is was kind:electronics but is now showing up as kind:building, a regression.
> 
> http://vector.mapzen.com/osm/all/17/20966/50665.topojson?api_key=vector-tiles-HqUVidw
> http://vector.dev.mapzen.com/osm/all/17/20966/50665.topojson?api_key=vector-tiles-HqUVidw
> 
> This is also impacting it's styled zoom visibility because the scene file hides generic building names until later zooms, but shows labels with a kind set (to something other than building) at earlier zooms.

`shop=electronics` is missing from the POI selection list, it should be added. @nvkelso, what's an appropriate zoom level for electronics shops with an area of 6595 to show up?

nvkelso:
Zoom 15 seems good. 
If that's too overwhelming, we could try 16.

zerebubuth:
# Best Buy in Sausalito

Note that the name is "Best Buy (closed)", which isn't right - if it's closed then it ought to either be resurveyed as the correct shop or marked as no longer a shop (possibly with `old_shop` / `old_name` as appropriate).

This is (or was) the largest electronics store I could find mapped in the Bay Area - it's in the tiles up to zoom 12, but only displays from 15 because of the area restriction in the style.

![sausalito-best-buy-z15-after](https://cloud.githubusercontent.com/assets/271360/12115964/23a58306-b3ae-11e5-8134-3ed340a3f6c2.jpg)

# Radio Shack in San Francisco

This displays at z15 because it's a point, and therefore doesn't get area-filtered, but it's probably a pretty small shop.

![sf-radio-shack-z15-after](https://cloud.githubusercontent.com/assets/271360/12116021/82d6120a-b3ae-11e5-9703-3a4c375829d7.jpg)

# Apple Store in San Francisco

This is mapped as an area, but it's such a small shop (compared with an out-of-town Best Buy, for example) that the area filter in the style means it shows up at z17. It's in the tiles from z15, though.

![sf-apple-store-z17-after](https://cloud.githubusercontent.com/assets/271360/12116040/c4cbef18-b3ae-11e5-8385-1b4fa613e911.jpg)

@nvkelso what do you think? I reckon the data is in the tiles way too early, and it would be better to bring it in by a couple of zoom levels (first data in the tiles at z14, then all displayed by z17).

nvkelso:
[...]

PR #459, opened 2015-12-21
Add electronics shops

rmarianski:
Refs #374


nvkelso:
@rmarianski Can you add before / after screenshots please? At the same zoom (15) and if these are sometimes visible earlier, from the earlier zoom too.

PR #464, opened 2016-01-04
First attempt at some tests.

zerebubuth:
These are really 'release tests' rather than unit tests, as they require a lot of data to be loaded in the database first (at the moment, it requires North America). A better test suite would load some minimal static _fixture_ data first.

But this is a nice first step, and we can refine it later.

These first few tests are items we added in v0.6.0 which should be, but currently aren't, being displayed on production. The tests pass against dev, though.


nvkelso:
Reads well, and so good to have these! Give me a day or two to verify it locally.

zerebubuth:
In the latest commit, I changed the config format a bit, adding a feature so that we can test against the tile proxy server while prod is generating. If you've already generated a config file, you'll need to update it to replace the URL with a dict with a `url` key and optional `use_all_layers` key. The latter should be set to `true` when using the tile proxy, e.g:

```yaml
proxy:
  url: http://localhost:8080/osm/all/%(z)d/%(x)d/%(y)d.json
  use_all_layers: true
```

bcamper:
An important step! :+1:

README.md
```diff
@@ -248,3 +248,27 @@ See https://github.com/mapzen/mapboxgl-vector-tiles for a sample map of Mapzen v
...
 
+## Tests
+
+There is a suite of tests which can be run against a tile server. You will need to install the [AppDirs](https://pypi.python.org/pypi/appdirs) Python package. First, set up a configuration file telling the test suite which server URL to use. The location of the configuration file will vary depending on your platform. To find out, either run the test suite and read the error message, or read the docs for [AppDirs](https://pypi.python.org/pypi/appdirs). This file should be a YAML dictionary of name to URL pattern pairs. For example:
+
+```yaml
+---
+local:
+  url: http://localhost:8080/%(layer)s/%(z)d/%(x)d/%(y)d.json
+  use_all_layers: false
```
nvkelso: Explain this in text after the code segment. (Looks to be an optimization for the tests which mostly declare what layers to look in.)

zerebubuth: Yeah, I guess it pretty much is an optimisation, although it came "backwards" if that makes any sense. I've updated the README now, and hopefully the documentation makes better sense.

zerebubuth: @nvkelso does everything look good to merge?

README.md
```diff
@@ -248,3 +248,48 @@ See https://github.com/mapzen/mapboxgl-vector-tiles for a sample map of Mapzen v
...
+```
+
+Each top-level key, in this example `local` requires a `url` sub-key which gives a pattern for creating valid tile URLs. There are four variables available, `layer`, `z`, `x` and `y`, which indicate the layer within the tile, zoom level and x, y coordinate respectively. The result should be a valid GeoJSON file without layers. If the data source is not able to filter layers, for example if you're reading the layered-GeoJSON directly from storage (i.e: S3), and do not use the `layer` parameter then you should set the optional `use_all_layers` parameter to `true`. By default, if left unspecified, it's false.
+
+For example, to read against a local server [proxying](https://github.com/mapzen/tile-hash-proxy) from storage such as S3, you might need a config something like this:
+
+```yaml
+---
+local:
```
nvkelso: Recommend removing this here as the example is about proxy only.

```
local:
  url: http://localhost:8080/%(layer)s/%(z)d/%(x)d/%(y)d.json
```

README.md
```diff
@@ -248,3 +248,48 @@ See https://github.com/mapzen/mapboxgl-vector-tiles for a sample map of Mapzen v
 
 See https://github.com/mapzen/hoverboard for more information. This repository has samples using each input format (see geojson.html, topojson.html, or index.html for .mvt). The Leaflet tile layer is defined in example.js, where you specify the Mapzen vector tile layers you want to render and the colors and symbols used to draw them.
 
+## Tests
+
+There is a suite of tests which can be run against a tile server. You will need to install the [AppDirs](https://pypi.python.org/pypi/appdirs) Python package. First, set up a configuration file telling the test suite which server URL to use. The location of the configuration file will vary depending on your platform. To find out, either run the test suite and read the error message, or read the docs for [AppDirs](https://pypi.python.org/pypi/appdirs). This file should be a YAML dictionary of name to URL pattern pairs. For example:
```
nvkelso: Can we provide a sample YAML config file in the source code in the `test` dir or the root dir?

eg:

```
prod:
  url: http://vector.mapzen.com/osm/%(layer)s/%(z)d/%(x)d/%(y)d.json
dev:
  url: http://vector.dev.mapzen.com/osm/%(layer)s/%(z)d/%(x)d/%(y)d.json
local:
  url: http://localhost:8080/%(layer)s/%(z)d/%(x)d/%(y)d.json
proxy:
  url: http://localhost:8081/all/%(z)d/%(x)d/%(y)d.json
  use_all_layers: true
```

nvkelso:
Please add `test.log` to the `.gitignore` list.

nvkelso:
I ran this against `dev` and `prod` and got results within range for both, good!

nvkelso:
I think this is fine, but noting it here because it's different than the other pattern: 

```
[  18/22] FAIL: 'test/443-swimming-pools.py'
  File "test.py", line 137, in run_test
    'features_in_tile_layer': features_in_tile_layer
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 240, in run_path
    return _run_module_code(code, init_globals, run_name, path_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 82, in _run_module_code
    mod_name, mod_fname, mod_loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "test/443-swimming-pools.py", line 9, in <module>
    { 'kind': 'swimming_pool' })
  File "test.py", line 128, in assert_has_feature
    % properties
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "test.py", line 112, in features_in_tile_layer
    yield features
  File "test.py", line 121, in assert_has_feature
    raise Exception, "Layer %r empty" % layer
Exception: Tile 'http://vector.mapzen.com/osm/water/16/19273/24652.json': Layer 'water' empty
```

I had been expecting: 

`Did not find feature including properties {'kind': 'swimming_pool'}`.

nvkelso:
Other than that, looks good.

zerebubuth:
> I had been expecting: Did not find feature including properties

Hmm... okay. I thought that having a more specific error message would be better. When debugging, it's helpful for me to know if there were _any_ features at all. Would a better message be "Did not find feature including properties ... (because layer 'blah' was empty)" instead?

README.md
zerebubuth: Yes, good idea. Added in 697a019.

nvkelso:
LGTM

nvkelso:
> Would a better message be "Did not find feature including properties ... (because layer 'blah' was empty)" instead?

Yes.

Issue #465, opened 2016-01-05
Add Gym / Fitness POIs


zerebubuth:
There are [many ways](http://wiki.openstreetmap.org/wiki/Gym_/_Fitness_centre) of tagging gyms and fitness-related POIs, although a short-list would be (thanks @IndyHurt!):

* Most commonly-used tags:
  * `leisure=sports_centre` + (`sport=fitness` or `sport=gym`) (about 1,827 uses)
  * `leisure=fitness_centre` (about 1,890 uses)
  * `leisure=fitness_station` (about 2,932 uses)
  * `amenity=gym` (about 1,034 uses - although apparently considered "deprecated" due to different meanings of "gym" in different languages).
* Other tags, less used (all < 10 uses)
  * `amenity=fitness_center`
  * `amenity=fitness_centre`
  * `amenity=fitness`
  * `leisure=fitness_center`
  * `leisure=fitness`
  * `leisure=gym`

We should certainly support all the commonly-used tags, although it's debatable whether it's worth supporting the the less-used ones and it may help drive convergence if we don't support them.

zerebubuth:
After the above two commits, fitness related POIs are now visible (e.g: near #19/37.76508/-122.46659):

Before:

![sunset-gym-sf-z19-before](https://cloud.githubusercontent.com/assets/271360/12118331/c3f0a868-b3bd-11e5-8fbf-4cc442d66e8a.jpg)

After:

![sunset-gym-sf-z19-after](https://cloud.githubusercontent.com/assets/271360/12118334/c709f752-b3bd-11e5-8543-e7f487c9f5c6.jpg)

nvkelso:
I see these appearing in the `pois` layer, good! I'm not sure why I can't get them to render in Eraser Map or Skin & Bones, though. Can you include the scene file snippet you're using, please?

But reviewing this issue with @IndyHurt for her Fitness post, it looks like there's a significant number of buildings that are tagged with these properties that we should also pull in for this issue, too. For instance:

* https://www.openstreetmap.org/way/264528016

zerebubuth:
Ah, for this one, I don't think I changed the scene file. There is already an entry for fitness:

```yaml
                fitness:
                    filter: { kind: [gym, fitness, fitness_center] }
                    draw:   { icons: { sprite: fitness } }
```

Which is why I normalised to `kind: fitness` rather than any other value. I don't think that style file changes should be necessary.

[...]

PR #466, opened 2016-01-05
Select fitness-related POIs.

zerebubuth:
This selects fitness-related POIs and normalises them to `kind=fitness`.

Connects to #465. Requires mapzen/TileStache#110.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -178,6 +178,10 @@ BEGIN
       WHEN amenity_val  = 'ski_rental'       THEN LEAST(zoom + 1.27, 17)
       WHEN amenity_val  = 'ski_school'       THEN LEAST(zoom + 2.30, 15)
       WHEN man_made_val = 'snow_cannon'      THEN LEAST(zoom + 4.90, 18)
+      WHEN leisure_val  = 'sports_centre'    THEN LEAST(zoom + 3.98, 17)
+      WHEN leisure_val  = 'fitness_centre'   THEN LEAST(zoom + 3.98, 17)
+      WHEN leisure_val  = 'fitness_station'  THEN LEAST(zoom + 3.98, 17)
+      WHEN amenity_val  = 'gym'              THEN LEAST(zoom + 3.98, 17)
       WHEN highway_val  = 'motorway_junction' THEN 12
       WHEN (barrier_val IN ('gate')
             OR craft_val IN ('sawmill')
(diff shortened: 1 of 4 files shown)


queries.yaml
```diff
@@ -112,6 +112,7 @@ layers:
     transform:
       - TileStache.Goodies.VecTiles.transform.tags_create_dict
       - TileStache.Goodies.VecTiles.transform.add_iata_code_to_airports
+      - TileStache.Goodies.VecTiles.transform.normalize_leisure_kind
```
nvkelso: If we're normalizing them all to `fitness` should this be `normalize_fitness_kind` instead?

data/functions.sql
```diff
@@ -178,6 +178,10 @@ BEGIN
       WHEN amenity_val  = 'ski_rental'       THEN LEAST(zoom + 1.27, 17)
       WHEN amenity_val  = 'ski_school'       THEN LEAST(zoom + 2.30, 15)
       WHEN man_made_val = 'snow_cannon'      THEN LEAST(zoom + 4.90, 18)
+      WHEN leisure_val  = 'sports_centre'    THEN LEAST(zoom + 3.98, 17)
+      WHEN leisure_val  = 'fitness_centre'   THEN LEAST(zoom + 3.98, 17)
+      WHEN leisure_val  = 'fitness_station'  THEN LEAST(zoom + 3.98, 17)
+      WHEN amenity_val  = 'gym'              THEN LEAST(zoom + 3.98, 17)
```
nvkelso: Should `gym` be normalized to `fitness` as well?

queries.yaml
zerebubuth: Depends on whether we want to write another function the next time we normalise another leisure kind or extend the existing one. I don't have a strong feeling either way - do you prefer `normalize_fitness_kind`?

data/functions.sql
zerebubuth: Yes, `amenity=gym` should become `kind=gym` when coalesced in the query, and the transform should normalise that to `kind=fitness` [on this line](https://github.com/mapzen/TileStache/pull/110/files#diff-52de7615e270521ccb11d3a0fe443d33R2713).

nvkelso: Looks good, tnx.

queries.yaml
nvkelso: General is better, leave as is.

nvkelso:
Ready to merge.

Issue #358, opened 2015-10-29
At low and mid-zooms, merge "same" road segments


nvkelso:
We're sending over a lot of chunky data in tiles at low- and mid- zooms which results in larger file sizes, longer time to render, and makes drawing pretty labels difficult (lots of colliding labels on small segments instead of spaced labels over longer runs.

For example, zoom 8:

![screen shot 2015-10-29 at 5 43 05 pm](https://cloud.githubusercontent.com/assets/853051/10833035/485f245a-7e4c-11e5-85ae-d241775188f4.png)

```
roads:
    data: { source: osm }
    filter: { not: { kind: [rail, ferry] } }
    draw:
        lines:
            order: 2
            # color: '#777'
            color: Style.color.randomColor()
            width: 2px
```
Download full [scene.txt](https://github.com/mapzen/vector-datasource/files/23796/scene.txt) file (rename txt to yaml) »

/cc @bcamper, @sensescape, @zerebubuth, discuss!

zerebubuth:
Part of this is probably made worse by the landuse cutting, and #314 should address that. The other part of this is likely to be changes in status, e.g: network membership, name, tunnel, bridge, layer, and any of the other tags we're collecting (operator, type, colour, ascent, descent, etc...). We'll only be able to merge these if we're able to discard / ignore some of these tags. So...

@nvkelso Q: At which zoom do we no longer draw / care about changes in `tunnel`, `bridge`, layer, etc... status?

zerebubuth:
For example, here's a roughly similar area with the roads coloured by a hash of their name, ref, kind and landuse_kind. There is a lot of improvement, because we're (effectively) dropping almost all the rest of the attributes. This picture represents an "upper bound" of what we can realistically do without putting a lot of effort into generalisation techniques (which would be awesome cool, though).

![nyc_z8](https://cloud.githubusercontent.com/assets/271360/10854164/ebed94fe-7f31-11e5-949e-c2b69a1b5d1c.jpg)

Here's the [scene.txt](https://github.com/mapzen/vector-datasource/files/24446/scene.txt). Note that the hash function is [from this SO answer](http://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery#answer-7616484).

bcamper:
[...]

PR #468, opened 2016-01-07
Merge road segments

zerebubuth:
This configures the dropping of some properties and the merging of some road segments.

The configuration here is to drop bridge, tunnel and oneway properties for paths below z15 and other kinds of road and rail below 13. This is based on the [style](https://github.com/nvkelso/tangram-skin-and-bones/blob/gh-pages/scene.yaml#L2649-L2657) which seems to mostly not show bridges or tunnels below those zoom levels (and doesn't show oneway at all?), but this should be checked.

It is also configured to drop name, ref and network properties below certain zooms configured per-kind. Again, this was based on a reading of the style, and should be checked.

The improvements in feature count and tile size were based on this configuration, and could be improved further by dropping more properties later, or some improvement could be sacrificed in return for keeping those properties into lower zooms.

Connects to #358. Requires mapzen/TileStache#111.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -271,3 +271,8 @@ post_process:
     params:
       source_layer: roads
       start_zoom: 9
+  - fn: TileStache.Goodies.VecTiles.transform.merge_features
+    params:
+      source_layer: roads
+      start_zoom: 8
+      end_zoom: 16
--- a/queries.yaml
+++ b/queries.yaml
@@ -273,6 +273,26 @@ post_process:
     params:
       source_layer: roads
       start_zoom: 9
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
+      properties: [is_bridge, is_tunnel, oneway]
+      where: >-
+        (kind == 'path' and zoom < 15) or
+        (kind in ['minor_road', 'major_road', 'highway', 'rail'] and zoom < 13)
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 16
+      properties: [name, ref, network]
+      where: >-
+        (kind == 'path' and zoom < 17) or
+        (kind in ['minor_road', 'rail'] and zoom < 15) or
+        (kind == 'major_road' and zoom < 12) or
+        (kind == 'highway' and zoom < 7)
   - fn: TileStache.Goodies.VecTiles.transform.merge_features
     params:
       source_layer: roads
--- a/queries.yaml
+++ b/queries.yaml
@@ -278,10 +278,17 @@ post_process:
       source_layer: roads
       start_zoom: 0
       end_zoom: 14
-      properties: [is_bridge, is_tunnel, oneway]
+      properties: [is_bridge, is_tunnel]
       where: >-
         (kind == 'path' and zoom < 15) or
         (kind in ['minor_road', 'major_road', 'highway', 'rail'] and zoom < 13)
+  # drop all oneway tags below zoom 14 (i.e: zoom >= 15) for all types.
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
+      properties: [oneway]
   - fn: TileStache.Goodies.VecTiles.transform.drop_properties
     params:
       source_layer: roads
@@ -289,10 +296,21 @@ post_process:
       end_zoom: 16
       properties: [name, ref, network]
       where: >-
-        (kind == 'path' and zoom < 17) or
-        (kind in ['minor_road', 'rail'] and zoom < 15) or
-        (kind == 'major_road' and zoom < 12) or
+        (kind == 'path' and zoom < 16) or
+        (kind == 'rail' and zoom < 15) or
+        (kind == 'minor_road' and zoom < 14) or
+        (kind == 'major_road' and zoom < 11) or
         (kind == 'highway' and zoom < 7)
+  # this is a patch to get rid of name, but keep ref & network, for highways
+  # when zoom < 11.
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 10
+      end_zoom: 7
+      properties: [name]
+      where: >-
+        kind == 'highway'
   - fn: TileStache.Goodies.VecTiles.transform.merge_features
     params:
       source_layer: roads
--- a/queries.yaml
+++ b/queries.yaml
@@ -306,8 +306,8 @@ post_process:
   - fn: TileStache.Goodies.VecTiles.transform.drop_properties
     params:
       source_layer: roads
-      start_zoom: 10
-      end_zoom: 7
+      start_zoom: 7
+      end_zoom: 10
       properties: [name]
       where: >-
         kind == 'highway'


queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
...
+        (kind == 'path' and zoom < 15) or
+        (kind in ['minor_road', 'major_road', 'highway', 'rail'] and zoom < 13)
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 16
+      properties: [name, ref, network]
+      where: >-
+        (kind == 'path' and zoom < 17) or
```
nvkelso: Paths should be label-able at zoom 16, suggest changing this to `zoom < 16` instead.

queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
...
+        (kind in ['minor_road', 'rail'] and zoom < 15) or
```
nvkelso: Only showing minor road labels at z15+ is a factor of Tangram not having great labeling (yet), we should aim to show these starting at zoom 14 (which is more comparable with other mapping systems), suggest revising this to `zoom < 14` (at least for `minor_road`, `rail` could stay `zoom < 15`).

queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
...
+        (kind == 'major_road' and zoom < 12) or
```
nvkelso: We currently label major_road features at zoom 11 (which is comparable with other mapping systems), so this needs to be `zoom < 11`.

queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
...
+        (kind == 'highway' and zoom < 7)
```
nvkelso: This is good as-is, but optionally:

Generally we don't label with `name` less than zoom 11, so you could also consider dropping `name` at zooms less than 11 from highways, that would probably result in a greater amount of line merging (a good thing).

queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
     params:
       source_layer: roads
       start_zoom: 9
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
+      properties: [is_bridge, is_tunnel, oneway]
```
nvkelso: These properties are used to calculate the sort key z-order of road features with respect to one another. It's okay to drop them early, as long as the sort key is calculated **before** they are dropped?

queries.yaml
bcamper: I'm all for more merging :)  Do we think others might have use of those
names though? There are cases like http://github.com/tangrams/explorer but
that one is pretty specific.

queries.yaml
zerebubuth: Yes, the sort key is calculated before this. Although that might defeat the purpose... given that the different sort keys will prevent them merging. So perhaps we need to be running this before the sort order calculation (or modify the sort order calculation)?

queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
...
+      where: >-
+        (kind == 'path' and zoom < 15) or
+        (kind in ['minor_road', 'major_road', 'highway', 'rail'] and zoom < 13)
```
nvkelso: Generally other web maps only show `oneway` starting at zoom 15 (because symbols showing the one way need to fit between street grid intervals). But `is_bridge`, `is_tunnel` is used currently starting at zoom 13 (to change the outline road casing to sort above the road inline). Should the be split apart (similar to highway and `name`, `ref` below)?

queries.yaml
```diff
@@ -273,3 +273,28 @@ post_process:
...
       source_layer: roads
       start_zoom: 9
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
+      properties: [is_bridge, is_tunnel, oneway]
+      where: >-
+        (kind == 'path' and zoom < 15) or
```
nvkelso: Style starts z-sorting and generally caring about these attributes starting at zoom 15 (where paths sort above highways if they do that IRL), so this seems fine.

queries.yaml
nvkelso: Explorer should explore what's in tiles. If it wants raw OSM, it should connect directly with overpass-turbo or similar APIs.

queries.yaml
nvkelso: Should we put this on `dev` (without addressing this) so we can compare with prod and then revise if necessary?

queries.yaml
bcamper: Or we might want to have a heavier weight tile set (like NEAT tiles but
consumable in simpler vector tile formats) eventually.

Agree it's purpose is to explore the tiles vs. all of OSM (since they
aren't that already), I think the important distinction may be a case where
more data is useful for *interactivity* (display on hover, click, etc.) vs.
just considering all initially visible labels. I think that dynamic
cartography is worth taking into account, though definitely don't want to
tilt so far that we lose the merging benefits for dominant use cases.

queries.yaml
bcamper: I'm eager to start trying it out!! :)

queries.yaml
nvkelso: Our vector tiles would still include the `ref` at lower zooms (even if the name is dropped), and only a tiny number of highways _don't_ include a `ref`, and those can usually be worked around with a custom `ref` that indicates a custom shield (I'm thinking of some roads in Texas and in Canada).

At some point interactivity involves asking things of the server by feature ID, can't put everything in vector tiles.

bcamper: Certainly the tiles can't serve all interactivity needs. But it's a big leap from baking in the tiles to having to go to a server (because it's a separate app to create/host etc.). Anyway I just think it's a consideration, in this case sounds like `name` isn't the issue and the important info (`ref`) is captured.

zerebubuth: I think we've already crossed the Rubicon for having a fully general tile set - dropping information in return for fewer / more compact features seems desirable given that perhaps 99% of the requests that we'll serve might come from our app or someone's app displaying something closely related to tangram-skin-and-bones.

_However_, I think there's a need for a way to do more customized tilesets, if only so that we can generate version N and N+1 simultaneously and go back and forth between them at will. Infrastructure that means that small changes are small and we can do multiple deploys per week, rather than having a 3-day turn-around time for a prod deploy. Not wanting to sound like a broken record, but I think the foundation of that is NEAT tiles.

zerebubuth: I think what I meant to say was: Rather than a 3-day turn-around time for prod deploy, regardless of how small the change is. (Which isn't quite true, as we can patch and selectively render an intersecting subset of the TOI, but that's a much more manual process.)

zerebubuth:
I've made the changes to the `queries.yaml` configuration, and the comparison to the previous is:

## Feature Counts

There's a huge jump at 11, but that's because it was going from particularly low to particularly high. It's still way, way better than the 2,558 it was before any merging.

```
 Z   Before ->    After (percent)
 6      214 ->      214 (  0.0%)
 7      218 ->      218 (  0.0%)
 8      216 ->      216 (  0.0%)
 9      199 ->      199 (  0.0%)
10      134 ->      134 (  0.0%)
11       97 ->      570 (487.5%)
12      297 ->      297 (  0.0%)
13      189 ->      159 (-16.1%)
14      115 ->      178 ( 55.2%)
15       77 ->       77 (  0.0%)
16       22 ->       23 (  3.8%)
17       23 ->       23 (  0.0%)
18        9 ->        9 (  0.0%)
```

## Uncompressed Sizes

Size is bigger too with this last change, but still way better than the 880,505 bytes it was before this PR.

```
 Z   Before ->    After (percent)
 6  1907106 ->  1907126 (  0.0%)
 7   820762 ->   820765 (  0.0%)
 8   557706 ->   557622 ( -0.0%)
 9   857379 ->   857360 ( -0.0%)
10   584272 ->   584297 (  0.0%)
11   391521 ->   469302 ( 19.9%)
12   334758 ->   334763 (  0.0%)
13   305935 ->   300733 ( -1.7%)
14   450352 ->   458875 (  1.9%)
15   122791 ->   122791 (  0.0%)
16    46835 ->    46974 (  0.3%)
17    25747 ->    25747 (  0.0%)
18    10932 ->    10932 (  0.0%)
```

## Time to read

The change here is pretty much lost in the noise.

```
 Z   Before ->    After (percent)
 6     15.6 ->     15.5 ( -1.0%)
 7      7.8 ->      7.7 ( -0.3%)
 8     16.7 ->     16.6 ( -0.5%)
 9     21.5 ->     21.4 ( -0.3%)
10     10.0 ->     10.1 (  0.2%)
11      6.0 ->      6.1 (  2.5%)
12      4.9 ->      4.8 ( -1.4%)
13      3.4 ->      3.4 ( -0.5%)
14      4.1 ->      4.1 ( -0.2%)
15      1.3 ->      1.3 (  0.2%)
16      0.6 ->      0.6 (  1.7%)
17      0.4 ->      0.4 ( -5.3%)
18      0.3 ->      0.3 (  4.0%)
```

nvkelso:
Other than the one highway comment, looks good. Looking forward to previewing this on dev!

queries.yaml
```diff
@@ -273,3 +273,46 @@ post_process:
...
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 16
+      properties: [name, ref, network]
+      where: >-
+        (kind == 'path' and zoom < 16) or
+        (kind == 'rail' and zoom < 15) or
+        (kind == 'minor_road' and zoom < 14) or
+        (kind == 'major_road' and zoom < 11) or
```
nvkelso: This line is causing a style regression for Eraser Map scene – turns out we're using the presence of a ref as a proxy for regional road significance.

zerebubuth: Okay. At which zooms & for which kinds is it safe to drop `ref` - or should it never be dropped at all? I think we do want to drop it sometimes, as having it on _all_ roads would seriously impair the ability of the merge routine to make any merges.

PR #481, opened 2016-01-12
Don't import historic WOF neighbourhoods

zerebubuth:
Migration to add `inception`, `cessation`, `is_visible` columns to WOF neighbourhoods. Update to queries to use the visibility information.

Requires and is required by mapzen/tilequeue#63. Connects to mapzen/tilequeue#59.

### DIFF
--- a/queries/places.jinja2
+++ b/queries/places.jinja2
@@ -92,6 +92,7 @@ FROM wof_neighbourhood wof_n
 INNER JOIN wof_neighbourhood_placetype wof_np ON wof_n.placetype = wof_np.placetype_code
 
 WHERE
+  wof_n.is_visible AND
   {{ bounds|bbox_filter('label_position') }} AND
   {{ zoom + 1 }} >= min_zoom AND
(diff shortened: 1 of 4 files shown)


queries/places.jinja2
```diff
@@ -92,6 +92,7 @@ FROM wof_neighbourhood wof_n
 INNER JOIN wof_neighbourhood_placetype wof_np ON wof_n.placetype = wof_np.placetype_code
 
 WHERE
+  wof_n.is_visible AND
```
nvkelso: Where do we strip out `is_visible` so it is not exported in tiles?

nvkelso:
Other than the `is_visible` question above, LGTM

queries/places.jinja2
zerebubuth: We don't need to - `is_visible` isn't one of the `SELECT`ed columns, so doesn't get promoted to a property in the tile. It's not present outside of the database.

Issue #197, opened 2015-09-11
Clip buildings layer with larger region


rmarianski:
There are certain features in the buildings layer that can span multiple tiles at high zooms, which is problematic for rendering. Eg the [highline in nyc](https://www.openstreetmap.org/way/37054313)

Instead of clipping on the tile boundary, we can extend the clip region a tile's worth in every direction, ie a 3x3 grid.

nvkelso:
More from the Trello thread:

Here's another view of the High Light in OSM (relation this time):

* http://www.openstreetmap.org/edit?relation=175905#map=20/40.74585/-74.00578

At zoom 18 the High Line geometry goes everywhere:

http://tangrams.github.io/tangram/#mapzen,40.744275885977395,-74.00397802140873,18.010624999999994

nvkelso:
Talking with @bcamper more about this today, it sounds like this is **not** a problem in Tangram ES because of implementation details, but we should still fix this for Tangram WebGL (but lower priority). Moving this to Compilation 2.1 milestone to load balance.

nvkelso:
Turns out this is an issue for Tangram after all, pulling into v0.7 release.

bcamper:
Thanks!

bcamper:
We just haven't been noticing it in JS, but yeah, still an issue. I had a proof of concept that fixed this client-side, but it's an awful lot of work to put every tile through for the very rare cases where it is actually needed, so I think doing it upfront on the server is the way to go.

zerebubuth:
Unfortunately, this isn't without visual side effects... :crying_cat_face: 

![high-line-z18-bugs](https://cloud.githubusercontent.com/assets/271360/12303231/1645bc66-ba21-11e5-89c6-8995817ddce9.jpg)

@bcamper, is there anything we can do to mitigate this?

bcamper:
Hmm, interesting issue, this is happening because we exclude edges that fall on the boundary of the **current** tile, but these are actually lines that while still falling on *a* tile boundary, it's over a full tile away from the one being rendered :)

Discussing with @blair1618, there are a couple things we'd like to look at:

- Can we make these tiles available on a dev server somewhere so we can try them out? While possibly a hacky solution, we could exclude edges falling on *any* tile boundary (not just the "current" one).
[...]

PR #482, opened 2016-01-13
Enable padded clipping for buildings layer.

zerebubuth:
Connects to #197. Requires mapzen/tilequeue#64.


nvkelso:
LGTM

Issue #489, opened 2016-01-15
Don't drop ref for all road kinds & zooms


zerebubuth:
From #358, plus [context](https://github.com/mapzen/vector-datasource/pull/468#discussion_r49804593).

Dropping `ref` at some zooms is causing a style regression for Eraser Map scene – turns out we're using the presence of a `ref` as a proxy for major_road regional significance, so we'll need to keep that (but can still drop name and network).

Needs info: which zooms / feature kinds is it okay to drop `ref` for?

nvkelso:
From #358: `ref` needs to be present on `major_road` the same range of zooms as for `highway`.

So something like this?

**from**
```
        (kind == 'major_road' and zoom < 11) or
        (kind == 'highway' and zoom < 7)
```

**to**
```
        (kind == 'major_road' and zoom < 7) or
        (kind == 'highway' and zoom < 7)
```

But then then the `name` and `network` tags should still be dropped from `(kind == 'major_road' and zoom < 11)`, so new transform added?

bcamper:
If we're just using the *presence* of a `ref` as a proxy for regional significance, we could accomplish that with just a boolean, and allow roads with different `ref`s to merge, if that's a concern. I suspect that would be an over-optimization for our current style case vs. generally more useful to retain the full `ref`, but if we didn't think it was necessary before and the only reason we're keeping it is to signify something else, there are other cases to accomplish that.

nvkelso:
No, we should keep the `ref` (not replace it with a boolean).

Most of these `major_road`s should labels with shields (once that support lands in Tangram), so we don't want to remove `ref` entirely from `major_road` features. I could see possibly removing the `ref` at zoom 8 instead of 7 used for highways, but too early to say / different issue.

nvkelso:
Refs are back and Eraser Map and Skin & Bones are looking like their old selves.

PR #494, opened 2016-01-15
Don't drop ref for major_roads between zooms 7 and 10

zerebubuth:
But still drop name and network.

Connects to #489.


nvkelso:
LGTM

Issue #490, opened 2016-01-15
Clip to 3x tile boundaries, not 1x


zerebubuth:
Context: #197.

Previous PR, #482, clipped to the tile boundary, but now we want to clip to 3x tile boundary instead.

nvkelso:
LGTM, see #197 for more detail.

PR #495, opened 2016-01-15
Clip buildings to 3x tile bounds, not 1x.

zerebubuth:
Note that the tests didn't need updating, since I forgot to update them for 1x, they're still calibrated for 3x.

Connects to #490.


nvkelso:
LGTM

PR #497, opened 2016-01-16
Move WOF table creation to start of updates script

zerebubuth:
So that the WOF tables are present when the functions are defined. Also adds the `is_visible` column, which was missing from the previous PR.


nvkelso:
LGTM

Issue #440, opened 2015-12-10
Fix up zoos & related tourism=attraction's (pois/landuse)


nvkelso:
`tourism:attraction` is used in **zoos** and other "tourist" recreation areas to tag points and polygons. It's often combined with `attraction=*` to provide more fidelity (that it's an animal & etc). Inside, zoos, the `zoo=*` tag is also used to indicate animal enclosures and `natural=*` to indicate if the landcover is grass, sand, etc. 

Whitelist **tourism** value for pois and landuse, and for POIs default to zoom 17 unless otherwise noted:

- [x] attraction
- [x] artwork
- [x] theme_park - **zoom 13** _likely candidate for variable zoom range_
- [x] wilderness_hut - **zoom 15**
- [x] hanami (flower viewing in Japan)
- [x] resort - **zoom 14** _likely candidate for variable zoom range_
- [x] trail_riding_station
- [x] aquarium - **zoom 14** _likely candidate for variable zoom range_
- [x] winery - **zoom 14** _likely candidate for variable zoom range_

Whitelist these **attraction** values for pois and landuse:

- [x] animal
- [x] water_slide
- [x] roller_coaster
- [x] summer_toboggan
- [x] carousel
- [x] amusement_ride
- [x] maze

We're not picking up landuse layer **natural** values of:

- [ ] sand
- [x] grass

We should whitelist the following landuse **zoo** tags:

- [x] enclosure
- [x] petting_zoo
- [x] aviary
- [x] wildlife_park

And we're missing **barrier** to wrap around the enclosures: 

- [x] fence (starting zoom 17)

Examples:
* Not-zoo point: https://www.openstreetmap.org/node/381538609
* Zoo map: https://www.openstreetmap.org/#map=18/37.73272/-122.50292
* Zoo polygon: https://www.openstreetmap.org/way/382818435

zerebubuth:
The keys `attraction` and `zoo` are not included in our osm2pgsql style file, so we may be missing a few (most will probably have names, and so be included). Created #485 to track it, but we'll probably do okay without a reload for now.

zerebubuth:
For each of these values, should their value be in the `kind` property? For example, [relation 5695997](http://www.openstreetmap.org/relation/5695997) has `attraction=animal`, `name=Asian Rhino`, `natural=sand`, `tourism=attraction` and `zoo=enclosure`.

What properties should this give on the final output polygon:

[...]

PR #499, opened 2016-01-19
Add attraction and zoo special cases

zerebubuth:
And tests for the same.

Connects to #440.

@nvkelso - to save time, I'd like to put this in dev immediately. I'd still appreciate your review.

### DIFF
--- a/data/apply-planet_osm_point.sql
+++ b/data/apply-planet_osm_point.sql
@@ -9,8 +9,8 @@ ALTER TABLE planet_osm_point ADD COLUMN mz_poi_min_zoom REAL;
 -- the coalesce here is just an optimisation, as the poi level
 -- will always be NULL if all of the arguments are NULL.
 UPDATE planet_osm_point SET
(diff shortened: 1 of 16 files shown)


zerebubuth:
Note: Requires mapzen/TileStache/pull/112.

nvkelso:
:+1: Overall code here looks good! Will verify visually once tiles and run tests.

But I wonder if we are picking up landuse layer value of `sand`? Looks like we were already picking up `grass` (https://github.com/mapzen/vector-datasource/blob/a959d47140dab2a7dad2023541535c98fd3e3945/data/functions.sql#L11). If `natural=sand` is set on an zoo enclosure, do we pick it up as an attribute, but not export a landuse polygon when it's `grass` alone?

zerebubuth:
Yup, we pick up the [surface and natural](https://github.com/mapzen/vector-datasource/blob/a959d47140dab2a7dad2023541535c98fd3e3945/queries/landuse.jinja2#L89-L90) tags on everything we're exporting. These properties should be present on the enclosures and other objects, and it would be worth adding [a test for that](https://github.com/mapzen/vector-datasource/issues/500).

Issue #507, opened 2016-01-22
Use points and nodes when exploring transit relations


zerebubuth:
The existing code joins to the `planet_osm_nodes` table, which by default is where `osm2pgsql` stores "raw" nodes before they've been transformed into points or used to build lines or polygons. However, there's another option, called "flat nodes", where these are stored on a file on disk instead.

In production, we use flat nodes. On my local machine, I don't.

All the queries which join to `planet_osm_nodes` need to join to `planet_osm_point` instead. I think all the relevant information is there, so we shouldn't lose any functionality.

zerebubuth:
With #508, I'm now seeing stuff like this:

## Camden Station, Baltimore

Camden station links via nodes only to the route relations, same as Euston station in London, UK. This has been fixed by using the `planet_osm_point` table instead of `planet_osm_nodes`. (Near 17/39.28238/-76.61943)

Before:

![camden-baltimore-z17](https://cloud.githubusercontent.com/assets/271360/12515140/cc23cfe2-c11e-11e5-8c33-e47cdb4c0f02.jpg)

After:

![camden-baltimore-after-z17](https://cloud.githubusercontent.com/assets/271360/12515144/cf661796-c11e-11e5-9c22-de7ac210d90a.jpg)

**Note**: The weirdness in the routes array seems to be down to there being a '>' character in the route ref / name for [the relation](http://www.openstreetmap.org/relation/1403277).

## 30th Street, Philadelphia

30th Street station is a building, and so doesn't have a node ID. The function has been extended to handle that case too. (Near 17/39.95561/-75.18177)

Before:

![30th-street-z16](https://cloud.githubusercontent.com/assets/271360/12515168/eeb23b3e-c11e-11e5-9444-c410613dd982.jpg)

After:

![30th-street-after-z16](https://cloud.githubusercontent.com/assets/271360/12515172/f3fd9dd6-c11e-11e5-802e-9351ecc954b6.jpg)

nvkelso:
LGTM

PR #508, opened 2016-01-22
Fix stations for flat nodes

zerebubuth:
Fixes a bug (sort of) which was stopping us finding all the relevant routes when running on a prod database with a flat nodes file rather than a `planet_osm_nodes` table. Also, adds a test for railway station routes. Also extends the function so that it works with ways (i.e: railway buildings) as well as just nodes/points.

Connects to #507.

@nvkelso, could you review please?

### DIFF
--- a/queries/pois.jinja2
+++ b/queries/pois.jinja2
@@ -30,7 +30,7 @@ FROM (
     -- so we only want to calculate it when we actually need the result.
     CASE
       WHEN railway='station' AND osm_id > 0
-        THEN mz_calculate_transit_routes(osm_id)
+        THEN mz_calculate_transit_routes(osm_id, NULL)
       ELSE NULL::text[]
     END AS transit_routes,
     tags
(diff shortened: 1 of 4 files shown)


queries/pois.jinja2
```diff
@@ -51,7 +51,13 @@ UNION ALL
     aerialway, aeroway, amenity, barrier, craft, highway, historic, leisure, lock, man_made, "natural", office, power, railway, shop, tourism, waterway,
     tags->'attraction' AS attraction,
     tags->'zoo' AS zoo,
-    NULL AS transit_routes,
+    -- note: the mz_calculate_transit_routes function is pretty expensive,
```
nvkelso: good perf refactor!

zerebubuth: Previously it was always `NULL`, so if anything the performance will get worse. But probably only a little bit, since this function isn't _that_ expensive and stations are pretty rare features.

test/507-routes-via-stop-positions.py
```diff
@@ -0,0 +1,59 @@
...
+                found = True
+                routes = props.get('transit_routes', list())
+
+                for r in expected_routes:
+                    count = 0
+                    for route in routes:
+                        if r in route:
+                            count = count + 1
+
+                    if count == 0:
```
nvkelso: Suggest also warning if there are LESS than the given number of routes or resulting `kind_tile_rank` – these are important stations should have a great tile rank, and failing that they won't be prominent on the map resulting in a regression.

nvkelso:
Otherwise LGTM.

test/507-routes-via-stop-positions.py
zerebubuth: Good idea, added in 9f7fdea.

Issue #400, opened 2015-11-09
Add bay water feature label positions


nvkelso:
Into the `water` layer.

* natural=bay, e.g. a A bay, a large body of water partially enclosed by land but with a wide mouth. (Other languages).	

See also:

* http://wiki.openstreetmap.org/wiki/Tag:natural%3Dbay

nvkelso:
Target layer: `water`.

nvkelso:
Also, `straight`?

nvkelso:
Load balancing to v0.8

rmarianski:
Did you mean `natural=strait`? http://taginfo.openstreetmap.org/tags/natural=strait

nvkelso:
Yep, that's the one.

Referenced on the same page, the following natural features are also candidates:

* natural=fjord
* natural=bay
* natural=fishing_bank
* natural=sound

* natural=peninsula
* natural=headland
* natural=cape
* natural=isthmus

rmarianski:
There's only [12 sounds](http://taginfo.openstreetmap.org/tags/natural=sound) and [4 fishing banks](http://taginfo.openstreetmap.org/tags/natural=fishing_bank). Should we still include those anyway?

nvkelso:
Skip

rmarianski:
Before:

![bay-before](https://cloud.githubusercontent.com/assets/53770/12661339/08de3ec8-c5e7-11e5-9390-ce97aa2cfa3b.png)

After:

![bay-after](https://cloud.githubusercontent.com/assets/53770/12661340/0fc496e2-c5e7-11e5-829f-82df972d2604.png)

nvkelso:
@rmarianski How were we picking up the blue water area before? Was it tagged with both water and bay? If we already have a polygon for these, is there a way to only export the label position points (not the full polygon geoms)?

zerebubuth:
@nvkelso, there's a [shoreline](http://www.openstreetmap.org/way/23793211) as well as the [bay polygon](http://www.openstreetmap.org/way/43950409) (rather annoyingly, they don't match up along the shore), which means we're picking up the blue water polygon from the water / coastline polygons from [OpenStreetMapData](http://openstreetmapdata.com/data/coast), and potentially now a mostly-overlapping polygon.

[...]

PR #515, opened 2016-01-28
Add bay water features

rmarianski:
Refs #400

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -237,11 +237,12 @@ post_process:
       property_keys: [name, kind]
       geometry_types: [Point]
       min_distance: 256.0
-  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
+  - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
     params:
       source_layer: landuse
-      properties: [mz_is_building]
       start_zoom: 13
+      where: >-
+        properties.get('mz_is_building') is True
   - fn: TileStache.Goodies.VecTiles.transform.normalize_and_merge_duplicate_stations
     params:
       source_layer: pois
--- a/queries.yaml
+++ b/queries.yaml
@@ -210,6 +210,12 @@ post_process:
       source_layer: water
       label_property_name: label_placement
       label_property_value: "yes"
+  - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
+    params:
+      source_layer: water
+      start_zoom: 9
+      where: >-
+        'label_placement' not in properties and kind in ('bay', 'strait', 'fjord')
   - fn: TileStache.Goodies.VecTiles.transform.generate_address_points
     params:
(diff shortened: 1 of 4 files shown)


rmarianski:
Just remembered that I should probably look to add a test and upgrade the migration for the next release before you review.

zerebubuth:
I forget - do we need a migration to do a fake update for `natural IN ('bay', 'strait', 'fjord')` to make the index pick up these new values, or do we need to rebuild the index?

rmarianski:
In the past I think we've just been rebuilding the entire index by creating a new one and then in a transaction dropping the old and renaming the new.

A fake update sounds like it might be more targeted though. What would that look like? `update planet_osm_polygon set "natural"="natural" where "natural" in ('bay', 'strait', 'fjord')`?

zerebubuth:
Yeah, I _think_ something like that would work, but I'm not 100% sure. Would be a good idea to test it (perhaps on dev) to make sure that it does what we expect.

nvkelso:
BRB: Add migration and test case.

rmarianski:
Depends on https://github.com/mapzen/TileStache/pull/113

@zerebubuth could you review now please?

The targeted update for the index didn't work for me locally, so I added a migration to make a new index.

I transitioned how the `mz_is_building` property is being removed. The previous method was really specific, and that didn't seem necessary to me. Let me know if I missed something here though.

It turns out that I had to add more flexibility to the transform in order to check if `label_placement` was in the list of properties. I put in an additional `properties` variable in the eval scope to point to the original dict to accomplish that.

test/400-bay-water.py
```diff
@@ -0,0 +1,3 @@
+assert_has_feature(
```
nvkelso: Please add 2 additional tests for `strait` and `fjord`.

queries.yaml
```diff
@@ -231,12 +237,11 @@ post_process:
       property_keys: [name, kind]
       geometry_types: [Point]
       min_distance: 256.0
-  - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
+  - fn: TileStache.Goodies.VecTiles.transform.drop_properties
```
zerebubuth: This filter was supposed to drop any building polygons from the `landuse` layer, but will now just drop the `mz_is_building` property and keep the feature. Perhaps this should be changed to?

```yaml
  - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
    params:
      source_layer: landuse
      where: >-
        properties.get('mz_is_building') == True
```

rmarianski: Nice catch! I thought the intent actually was to only drop the property. I'll go ahead and correct this.

rmarianski: Corrected in d6fbfd8.

zerebubuth:
:+1: looks great!

PR #525, opened 2016-02-12
Move docs into main project repo

nvkelso:
This will make it easier to maintain the docs, and sync them with mapzen.com/documentation.

This commit also adds new pages for:

* api-keys-and-rate-limits.md
* attribution.md
* data-sources.md

Those topics are common questions we get via support, and adding these doc sections brings us to doc parity with other projects like Pelias. 

The next step after this PR is merged is to update the doc generator to use this new source, and needs to be coordinated now. 

Separately, the docs are outdated (in the layers.md file in particular new layers and kind values are missing) and this PR will not address that now.


nvkelso:
I'll also update the main README.md as part of this PR.

nvkelso:
Working with @louh, we think adding a new tag to this repo (e.g. v0.7.1) is the best approach (versus linking to a SHA).

zerebubuth:
Definitely, as we'd only want the docs to update when we released a (versioned) tileset to production - could get confusing otherwise!

nvkelso:
Connects to https://github.com/mapzen/tile-tasks/issues/48.

nvkelso:
Ready for testing (to cut a tag and update doc generator).

Issue #502, opened 2016-01-21
Water boundaries can be very slow


zerebubuth:
For example, a tile like `10/280/364` can spend >15 minutes computing the `exterior_boundaries` post-processing filter. This is generally because these polygons are incredibly complex - many fractal lakes with islands in them - and large.

The "real fix" is probably to split these polygons up using a grid. The exterior boundaries code should erase this seam. However, this would require upstream code changes, or something pretty horrible in a trigger.

An easier fix, but slightly less efficient, might be to clip the water polygons to a buffered tile before performing `exterior_boundaries`. This should, at least, reduce the zoom level at which we have these problems.

nvkelso:
Testing water areas around here /#11/46.3931/-84.1985, I do see these generate slowly but surely.

nvkelso:
Hmm, I am getting some gateway timeout errors, though:

* Tangram v0.5.1: worker 1: tile load error(s) for osm/11/583/741/11: Error: Request error with a status of Gateway Time-out
* Tangram v0.5.1: worker 0: tile load error(s) for osm/11/585/741/11: Error: Request error with a status of Gateway Time-out

nvkelso:
Also:

* Tangram v0.5.1: worker 0: tile load error(s) for osm/17/10498/22491/16: Error: Request error with a status of Service Unavailable
* http://vector.dev.mapzen.com/osm/all/16/10498/22492.topojson Failed to load resource: the server responded with a status of 429 (Too Many Requests)

nvkelso:
If I let the dev server _breath_ a few minutes, everything resolves itself. 

Here's another test area: 

* 12/46.1704/-83.9585

nvkelso:
@zerebubuth Can you look at the logs on dev to see how long those tiles took or if there are any red flags associated with this change? I think this everything above was "expected" and "better than before", but let's do one more cross check, please.

zerebubuth:
I've had a look, but I can't tell.

As far as I can see, we're not logging the time to generate each tile from tileserver. I definitely see the period where the tiles were timing out (nginx logs a "499" code for those when your browser closes the connection), and there are warnings like this from GEOS:

```
2016-02-26_00:51:25.71079 Hole lies outside shell at or near point -9333810.4199999999 5788928.2400000002
2016-02-26_00:51:34.36113 Hole lies outside shell at or near point -9326563.1600000001 5796103.4100000001
[...]

PR #535, opened 2016-02-19
Water boundary tests.

zerebubuth:
Added tests to ensure that water boundaries are present where we want them. This tests 3 locations around North America where there are water boundaries to ensure that we aren't putting water boundaries in the tile between two water polygons - only at the boundary between water and not-water.

Connects to #502.

### DIFF
--- a/test.py
+++ b/test.py
@@ -115,27 +115,64 @@ def features_in_tile_layer(z, x, y, layer):
         raise Exception, "Tile %r: %s" % (url, e.message), sys.exc_info()[2]
 
 
+def count_matching(features, properties):
+    """
+    Returns a tuple containing the total number of features in the argument
+    `features` and the number of features matching the properties.
+    """
+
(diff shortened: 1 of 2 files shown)


test.py
```diff
@@ -115,27 +115,64 @@ def features_in_tile_layer(z, x, y, layer):
...
+    """
+
+    num_features = len(features)
+    num_matching = 0
+
+    for f in features:
+        if match_properties(f['properties'], properties):
+            num_matching += 1
+
+    return (num_features, num_matching)
```
rmarianski: don't need the parens here, but you can keep em if you like em!

rmarianski:
:+1:

test.py
zerebubuth: Oh, cool. I didn't realise that. I find this more readable, but perhaps that's just unfamiliarity with Python.

Issue #478, opened 2016-01-08
Adjust tile queries to allow for z16 metatiles


nvkelso:
Specifically, we pack some stuff only into z17 and z18 tiles now. If we meta tile, then those need to be included in z16 (and perhaps marked that their `min_zoom` is only valid at z17, z18).

EG, in POIs:

```
             WHEN (aeroway_val IN ('gate')
                   OR amenity_val IN (
                 'atm', 'bicycle_rental', 'bicycle_parking', 'bus_stop',
                 'drinking_water', 'emergency_phone',
                 'parking', 'post_box', 'telephone', 'theatre',
                 'toilets')
                   OR highway_val IN ('bus_stop', 'traffic_signals')
                   OR historic_val IN ('memorial')
                   OR leisure_val IN ('playground', 'slipway')
                   OR man_made_val IN ('mast', 'water_tower')
                   OR office_val IN ('accountant', 'administrative', 'advertising_agency',
                'architect', 'association', 'company', 'consulting', 'educational_institution',
                'employment_agency', 'estate_agent', 'financial', 'foundation', 'government',
                'insurance', 'it', 'lawyer', 'newspaper', 'ngo', 'notary', 'physician',
                'political_party', 'religion', 'research', 'tax_advisor', 'telecommunication',
                'therapist', 'travel_agent', 'yes')
                   OR shop_val IN ('bakery', 'bicycle', 'books', 'butcher', 'car', 'car_repair',
                                 'clothes', 'computer', 'convenience',
                                 'doityourself', 'dry_cleaning', 'fashion', 'florist', 'gift',
                                 'greengrocer', 'hairdresser', 'jewelry', 'mobile_phone',
                                 'optician', 'pet')
                   OR tourism_val IN ('bed_and_breakfast', 'chalet', 'guest_house',
                                    'hostel', 'hotel', 'motel')
                   OR railway_val IN ('subway_entrance')) THEN 17
             WHEN (amenity_val IN ('bench', 'waste_basket')) THEN 18
```

nvkelso:
Needs to be coordinated with scene file changes (to suppress these from display at zoom 16/17 as appropriate)

rmarianski:
@nvkelso I discussed the implementation with @zerebubuth a little bit, and a good way forward would be to start including _all_ pois in z16 tiles. For pois that don't have a `min_zoom`, we could set that to z18, or maybe flag that some other way?

zerebubuth:
[...]

PR #539, opened 2016-02-22
Query updates to metatile at z16

rmarianski:
Connects to #478.

### DIFF
--- a/queries/pois.jinja2
+++ b/queries/pois.jinja2
@@ -2,6 +2,7 @@ SELECT
     name,
     {% filter geometry %}way{% endfilter %} AS __geometry__,
     osm_id AS __id__,
+    min_zoom AS min_zoom,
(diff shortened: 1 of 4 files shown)


queries/pois.jinja2
```diff
@@ -2,6 +2,7 @@ SELECT
     name,
     {% filter geometry %}way{% endfilter %} AS __geometry__,
     osm_id AS __id__,
+    min_zoom AS min_zoom,
```
zerebubuth: Nitpicking: This doesn't need the `AS`?

zerebubuth:
:+1:

queries/pois.jinja2
rmarianski: Nice catch! I just mindlessly put that in :)

Fixed with c775922.

Issue #526, opened 2016-02-12
Add new POIs for Women's Day


nvkelso:
To support International Women's Day, let's add several new features (some of these are buildings, some are POIs, some might be on landuse polygons).

Mapzen vector tile change proposal:

- [x] `amenity=social_facility`, where the kind is set to value of `social_facility=*`: http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dsocial_facility - zoom 17, this needs to be tied in with a TileStache transform change.
- [x] `amenity=clinic`: http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dclinic - zoom 17
- [x] `amenity=doctors`: http://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddoctors - zoom 17
- [x] `amenity=kindergarten`: http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dkindergarten - zoom 17
- [x] `amenity=childcare`: http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dchildcare - zoom 17
- [x] `emergency=phone`:  http://wiki.openstreetmap.org/wiki/Tag:emergency%3Dphone - zoom 18

/cc @awright

nvkelso:
I used iD to set social_facility type on an old persons home this weekend, and the tag is used was different.

nvkelso:
http://www.openstreetmap.org/way/398030613

zerebubuth:
Do you mean that it has used `social_facility:for=senior` rather than `social_facility=assisted_living` (or `group_home` perhaps)? That seems to be a separate dimension to the tagging - what it is, and who it's for.

Wasn't it possible to add the other tag as well? If not, perhaps we should ask whoever maintains the iD tag presets if they can add / modify?

nvkelso:
Agree our `kind` should be based on `social_facility` and ignore `social_facility:for`.

In iD it's always possible to edit the raw tags, but this is what auto completed so I suspect biasing the data collection.

nvkelso:
Looks like we assigning min_zoom to the `social_facility` features before we split them out into their own kinds. That works, but is pretty confusing. 

Another argument in favor of refactoring how all that code works in https://github.com/mapzen/vector-datasource/issues/532.

zerebubuth:
[...]

PR #542, opened 2016-02-23
Add inclusive POIs.

zerebubuth:
Adds many new POI types, see #526 for details. Adds tests for these too.

Connects to #526.


rmarianski:
:+1:

Issue #520, opened 2016-02-11
Big box stores should be included earlier in tiles


nvkelso:
If the feature area is large, then we should include these at zoom 14 if super huge and 15 if moderately large. Currently we only start showing these at zoom 16. 

This is related to an Eraser Map issue: https://github.com/tangrams/eraser-map/issues/128.

* `shop=department_store`
* `shop=supermarket`
* `shop=hardware`
* `shop=doityourself`
* `shop=trade`

Generally speaking shop=electronics is also a big box store type, and it's currently getting the below, which might be a good starting point to modify (at least it's working for z15, no modification for this existing zoom range requested):

```
      WHEN shop_val     = 'electronics'      THEN LEAST(zoom + 3.30, 17)
```

NOTE: Adding missing hardware store variants is also covered in: https://github.com/mapzen/vector-datasource/issues/519.

Some example sizes and expected tile zoom value:

* z14: https://www.openstreetmap.org/way/152722810 and https://www.openstreetmap.org/way/219072560 and https://www.openstreetmap.org/way/344057345 and https://www.openstreetmap.org/way/259001359.
* z15: https://www.openstreetmap.org/way/194906343 and https://www.openstreetmap.org/relation/3585039 (bigger urban grocery stores)

zerebubuth:
After #544, and after turning off the size filter in the `scene.yaml`:

## Before

![sf-z14-before](https://cloud.githubusercontent.com/assets/271360/13300990/a2434cd0-db3a-11e5-9cdc-5dd223def00f.jpg)

## After

Note that there are _two_ Targets in that tile, and we merge similar POIs within a 1-tile radius, so the Target mentioned in the issue description isn't there, but its equally colossal twin is, and I assume one Target per tile is enough for anyone.

![sf-z14-after](https://cloud.githubusercontent.com/assets/271360/13300992/a47d6418-db3a-11e5-93b4-9efbafb5f43e.jpg)

zerebubuth:
I should have said: That's at 14/37.6709/-122.4691

nvkelso:
LGTM

PR #544, opened 2016-02-24
Make larger 'big box' stores appear at lower zooms.

zerebubuth:
Most of these were pulled up from a flat zoom level. Added a test at the zoom mentioned in #520. Note that one of the Targets has two `id`s because we merge points with the same name within a 1-tile radius, and those are too close together. So one will get merged away.

Connects to #520.


rmarianski:
:+1:

PR #547, opened 2016-03-01
Fix tests for limited dataset, sort key changes

zerebubuth:
The sort key changes in mapzen/TileStache#116 broke the winter sports test.

The bay / strait / fjord data came from all over the world, but we test with the North America extract. I couldn't find a fjord, but found a suitable strait.


rmarianski:
I added those tests from outside North America, and didn't think there were any straits there. Thanks for adapting them.

:+1:

PR #553, opened 2016-03-03
move parking topside, and remove dup park

nvkelso:
Connects to https://github.com/mapzen/vector-datasource/issues/551.


zerebubuth:
Looks like the following values disappeared during the change:

* breakwater (`man_made=`)
* bridge (`man_made=`)
* cutline (`man_made=`)
* dike (`man_made=`)
* groyne (`man_made=`)
* land (`natural=`)
* national_park (`boundary=`)
* pier (`man_made=`)

Was that intentional?

nvkelso:
That was not intentional, thanks for the catch (added back in e6e539d). I should have operated on the CSV alone, instead I attempted to recreate it from the source OpenOffice doc.

zerebubuth:
:+1: looks good to me.

Issue #552, opened 2016-03-03
water boundary missing sort_key


nvkelso:
http://vector.dev.mapzen.com/osm/all/19/83900/202617.topojson

```
"properties":
{
"area":
3005021696,
"boundary":
"yes",
"id":
3324,
"kind":
"ocean",
"source":
"openstreetmapdata.com"
},
```

But other features in this tile do include the sort_key (including in the water layer).

PR #554, opened 2016-03-03
Fix missing sort key on water boundaries

zerebubuth:
Adds a new test, and updates the docs.

The issue was that the `boundary: yes` property is added by the `exterior_boundaries` post-process, but that was being run after the sort key was added.

Connects to #552.


rmarianski:
Nice catch! :zap:

nvkelso:
LGTM. Thanks for updating the docs, too!

PR #559, opened 2016-03-04
fixes #555 landuse funk

### LINKED ISSUE #555: Landuse sort_key funk
* **What did you see?** Wood and other natural landuse are sorting above parks, a regression over v0.7 tiles.
* **What did you expect?** It may actually be better this way, but it's different than before so noting it here. Might require design work?
* **What map location are you having problems with?** East of San Francisco, example tile: http://vector.dev.mapzen.com/osm/all/11/329/791.topojson
* **Screenshot?**

landuse-v0 8-beta-park-woods

nvkelso:
Connects to #555.


zerebubuth:
LGTM

Issue #556, opened 2016-03-03
sort_key for roads at zooms < 15 being adjusted for brunnels, levels


nvkelso:
* **What did you see?** Roads are sorting **with** their bridge, tunnel, and level offsets at zooms less than 15.
* **What did you expect to see?** This is a regression over v0.7 tiles. Brunnel and level offsets are only applied at zoom 15+, we want to continue that behavior. 
* **What map location are you having problems with?** #11/37.8881/-122.2778 and http://vector.dev.mapzen.com/osm/all/11/327/790.topojson
* **Screenshot?** 

![roads-funky](https://cloud.githubusercontent.com/assets/853051/13510146/1f7c3d92-e144-11e5-9274-b21900334da7.gif)

`S 23rd St.` should have `is_bridge:yes`, and has a level of 1. ([view on osm.org](http://www.openstreetmap.org/way/23073282))

```
{
"highway":
"secondary",
"id":
23073282,
"kind":
"major_road",
"name":
"S 23rd St.",
"sort_key":
443,
"source":
"openstreetmap"
},
"type":
"MultiLineString"
```

nvkelso:
We are still correctly dropping the `is_bridge` and `is_tunnel` properties at lower zoom, though :)

**dropped:**
* http://vector.dev.mapzen.com/osm/all/11/327/790.topojson

**included:**
* http://vector.dev.mapzen.com/osm/all/17/20990/50594.topojson

PR #560, opened 2016-03-04
restrict road brunnel, level sorting zoom>=15

nvkelso:
Connects to #556 to restrict road brunnel, level sorting zoom>=15 by updating the `roads.csv` to include a new **zoom** column.

Requires:

* https://github.com/mapzen/TileStache/pull/121


zerebubuth:
We drop the `is_bridge` and `is_tunnel` on most road types when `zoom < 13` in a [later filter](https://github.com/mapzen/vector-datasource/blob/7cf21343090c1895ac5d60a6d5ccbf1464403096/queries.yaml#L354-L362). This means that there are 3 different behaviours:

1. Zoom < 13, no `is_bridge` or `is_tunnel`, and sort_key does not depend on them.
2. Zoom is 13 or 14. `is_bridge` and `is_tunnel` are present, but don't affect sort key.
3. Zoom >= 15, `is_bridge` and `is_tunnel` are present and affect sort key.

I guess I don't really see the point of the "middle" bit where the properties are present, but unused. Are we missing the opportunity for merging if they're genuinely unused for styling? Or are we creating the possibility of z-fighting between the bridge and any road that would have gone under it?

nvkelso:
Are you proposing to drop `is_bridge` and `is_tunnel` in case, 3 (zooms 13, 14) in that later filter? I think we need to keep them – else big tunnels in both NYC and SF would regress on styling. I'm mixed about indicating `is_bridge`, but I don't see a compelling reason to drop now (and if we indicate tunnel, we should probably indicate bridge).

nvkelso:
Styling = transparent line colors in Tangram in this case.

nvkelso:
Zoom >= 15 sort_key offsets is the old behavior from v0.7 logic that we regressed on but need to add back.

zerebubuth:
If we need `is_bridge` and `is_tunnel` at 13 & 14 for big tunnels in SF/NYC, then surely we should also be applying a `sort_key` which takes that into account at those zooms?

nvkelso:
No, we only want to sort zoom >= 15. We need this change to prevent a regression from the current behavior in v0.7. We pick up the sort_key dance from a deliberate design decision in High Road. It's convention in most map providers to do this zoom split, and we need it to make our maps sort properly.

rmarianski:
Is it the case that you want to still style differently for roads/tunnels at z13/14 but not have that impact their ordering?

nvkelso:
Yeah, it seems a little weird, but makes the design hold together better. Otherwise minor roads (like 23rd St in the connecting issue) sort above the highways breaking them up visually.

nvkelso:
> Are we missing the opportunity for merging if they're genuinely unused for styling? 

The tunnels are currently used for styling. The bridges aren't in any of our house styles. If we wanted to drop that from zooms 13 and 14, let's look at it in a separate issue.

> are we creating the possibility of z-fighting between the bridge and any road that would have gone under it?

I don't think so. The road classes would still be separated with different sort keys and there isn't any special styling of bridges at these zooms. I can see a small chance of z-fighting, but our recommendation would be to not style that way at zooms 13 and 14.

rmarianski:
I think we should move the zoom >=15 rules above the general ones. These rules are first match wins.

We might want to have that last catch all be a little more general, ie all `*`. It's a little pedantic in this case though :)

zerebubuth:
> We need this change to prevent a regression from the current behavior in v0.7.

Yes, I realise that the behaviour wasn't what it was before. I'm trying to figure out if the behaviour before was intentional, or whether it was just a result of randomness. In other words: are we just trying to make this release a bug-for-bug compatible with the previous one, or is there an opportunity to make the behaviour make more sense?

> We pick up the sort_key dance from a deliberate design decision in High Road. It's convention in most map providers to do this zoom split, and we need it to make our maps sort properly.

Okay, so just checking I've understood this correctly:

1. At zoom < 13: merging is more important than styling tunnels, so we drop the bridge and tunnel properties.
2. At zoom = 13, 14: merging is less of a concern, but road class is more important than showing bridges (perhaps of lower class) above the roads they go above.
3. At zoom > 14: showing things in "physical" order is now more important, so bridges should go over the roads they go over.

Is that the behaviour we want?

nvkelso:
Yep, that's what we want. Thanks for checking!

rmarianski:
:+1:

Issue #557, opened 2016-03-04
Some large buildings at zoom 16 are missing


nvkelso:
* **What did you see?** Some large building features are missing at zoom 16, but present in zoom 17+. Neighbouring buildings of smaller size are included in tiles, though.
* **What did you expect to see?** These buildings should be included in tiles.
* **Complication:** On prod in v0.7 tiles, this problem corrects itself at zoom 17+, but in dev with v0.8 tiles the problem persists forever because we are starting to meta-tile at zoom 16, so that tile is cut up to form z17+ results, instead of fresh query. 
* **What map location are you having problems with?** Two San Francisco examples below.

**Best Tile**
* OSM id=`103383621`
* https://www.openstreetmap.org/way/103383621#map=18/37.73706/-122.40615
* http://vector.mapzen.com/osm/all/16/10484/25339.topojson?api_key=vector-tiles-HqUVidw
* http://vector.mapzen.com/osm/all/17/20969/50679.topojson?api_key=vector-tiles-HqUVidw
<img width="669" alt="screen shot 2016-03-03 at 16 14 46" src="https://cloud.githubusercontent.com/assets/853051/13514204/b612efd2-e15b-11e5-870b-eb0a319719d0.png">

**Miraloma School**
* OSM id=`338881092`
* https://www.openstreetmap.org/way/338881092#map=18/37.73843/-122.44924
* http://vector.mapzen.com/osm/all/16/10476/25339.topojson?api_key=vector-tiles-HqUVidw
* http://vector.mapzen.com/osm/all/18/41906/101357.topojson?api_key=vector-tiles-HqUVidw
<img width="539" alt="screen shot 2016-03-03 at 16 14 32" src="https://cloud.githubusercontent.com/assets/853051/13514205/b62b23f4-e15b-11e5-8e3a-3ba75d352477.png">

Thanks @burritojustice for helping isolate this issue!

nvkelso:
If this is a quick fix, let's get it into v0.8, else come back to it in v0.9.

nvkelso:
Verified 2 cases.

PR #562, opened 2016-03-04
Show all buildings at z16

rmarianski:
Connects to #557.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -741,6 +741,11 @@ CREATE OR REPLACE FUNCTION mz_building_filter(
 RETURNS BOOLEAN AS $$
 BEGIN
   RETURN CASE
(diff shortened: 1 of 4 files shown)


zerebubuth:
There's something else going on here. I've got some tests which I don't understand why they're failing... I've pushed what I've got at https://github.com/mapzen/vector-datasource/tree/557-missing-buildings. Not sure why the test for [this small building](http://www.openstreetmap.org/way/278410540) is failing... It seems to pass all the filters and get discarded by the formatting step?!

zerebubuth:
:+1: looks good to me!

Do we need to add something in the docs, or is `min_zoom` already documented?

nvkelso:
Cool.

PR #563, opened 2016-03-04
Add tests for missing buildings.

zerebubuth:
Adds a `min_zoom` parameter for buildings, too.

Requires mapzen/tilequeue#78.

### DIFF
--- a/queries/buildings.jinja2
+++ b/queries/buildings.jinja2
@@ -29,6 +29,12 @@ SELECT
     "roof:orientation" AS roof_orientation,
     "addr:housenumber" AS addr_housenumber,
     "addr:street" AS addr_street,
+{% if zoom >= 16 %}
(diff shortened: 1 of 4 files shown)


queries/buildings.jinja2
```diff
@@ -82,6 +86,9 @@ SELECT
     NULL AS roof_orientation,
     "addr:housenumber" AS addr_housenumber,
     "addr:street" AS addr_street,
+{% if zoom >= 16 %}
+    16 AS min_zoom,
```
rmarianski: Good call on adding `min_zoom`. Would we want this to be 17 to match the previous behavior, before z16 metatiling?

rmarianski:
:+1: 

Nice catches on all the other places.

queries/buildings.jinja2
zerebubuth: Yup, sounds good. Updated in dc503d8.

Issue #558, opened 2016-03-04
Do we need default tunnel, default bridge sort_key values? Pistes?


nvkelso:
Double checking the new roads.cvs spreadsheet, it looks like we're missing rows for bridge and tunnel fallbacks (for when there is not a more specific match). Is that deliberate?

```
layer::int,kind,highway,type,railway,aeroway,service,aerialway,is_bridge,is_tunnel,sort_key
-,*,*,*,*,*,*,*,-,yes,305
-,*,*,*,*,*,*,*,yes,-,405
```

Are there any roads appearing in 355 (the default catch all)? If so, we should probably put pistes and paths below them? 
```
-,piste;path,*,*,*,*,*,*,*,*,354
```

Moving the gondolas up above everything besides layers is good optimization :)

rmarianski:
Whoops, didn't mean to change assignment here, sorry.

zerebubuth:
Please, take it :smile:

PR #564, opened 2016-03-04
Default brunnel sort keys

zerebubuth:
Fix sort key test.

Add sort keys for paths, pistes and defaults.

Connects to #558.


zerebubuth:
As part of this, I changed the non `>=15` to have layer `*` rather than layer `-` so that z14 tiles with a layer set still match one of the "fallback" rules at the bottom.

rmarianski:
:+1:

Issue #368, opened 2015-11-03
Don't included "disused" (subway) stations


nvkelso:
Old South Ferry in New York is marked `disused=yes`

https://www.openstreetmap.org/node/2086974744

We Shouldn't include it in our selections, or if we do we need to mark it "closed" somehow.

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/268

nvkelso:
No longer see `Old South Ferry` station in NYC. :+1:

PR #573, opened 2016-03-07
Remove disused railway stations

zerebubuth:
Removes all POIs tagged with `disused=*` other than `disused=no`.

Connects to #368.


rmarianski:
:+1:

PR #578, opened 2016-03-08
Wrote workaround that makes sure populated places are imported

batje:
- [ ] Update tests // Should not be necessary
- [ ] Update docs // Should not be necessary


rmarianski:
Thanks for the pr! We're tracking subsequent work here in https://github.com/mapzen/vector-datasource/issues/582.

Issue #483, opened 2016-01-13
Suppress "closed" feature label_position until zoom 17+


nvkelso:
#18/37.87298/-122.51169

The Best Buy in this mall has closed, so someone noted that by adding " (Closed)" to the name. Kinda makes sense for recently closed stores, but less so for stores that have been closed more than 1 year?

So zoom 15 is too earlier to see these closed features:
![screen shot 2016-01-13 at 09 41 33](https://cloud.githubusercontent.com/assets/853051/12302419/e3cf0acc-b9d9-11e5-92fa-4cd982228a55.png)

Does make some sense to show them at zoom 17+ (here at zoom 18):
![screen shot 2016-01-13 at 09 41 12](https://cloud.githubusercontent.com/assets/853051/12302420/e3dcd1c0-b9d9-11e5-8c91-c2c9ba4962e0.png)

nvkelso:
Related: #291

rmarianski:
I started on the implementation of this in sql, but it seems that the discussion in https://github.com/mapzen/vector-datasource/issues/291 suggests that we might want to do it in a transform due to tag rewriting.

I also noticed some buildings that have the suffix `(closed)` as well, eg http://www.openstreetmap.org/way/395475802.

rmarianski:
@nvkelso sounds like we would want to just update the zooms here too then, and not alter any of the tags. Do we want to do the same for the `buildings` layer?

nvkelso:
> sounds like we would want to just update the zooms here too then, and not alter any of the tags. 

Three options:

1. Don't alter the tags.
2. Do the same as the ` (historical)` features in the other issue where we'd sub out so `kind:closed`.
3. Same as 2, but throttled by time. Do we know when it was marked closed? If it's recently closed, keep the original `kind=*`, but if it's a year+ closed then transform to `kind:closed`?

> Do we want to do the same for the buildings layer?

Yes

rmarianski:
How can we tell if it was closed recently? Is there a way of encoding that in a separate tag? I searched the osm wiki quickly but the closest thing I found was for [opening hours](http://wiki.openstreetmap.org/wiki/Key:opening_hours).

My vote would be for option 2 to be consistent with the `(historical)` features, and do the exact same thing for the `buildings` layer.

nvkelso:
:+1:

nvkelso:
Verified:

[...]

Issue #291, opened 2015-10-03
Suppress "historical" feature label_position until zoom 17+


nvkelso:
#15/37.7697/-122.1435
https://www.openstreetmap.org/edit?#map=16/37.7693/-122.1452
https://www.openstreetmap.org/relation/317369

Sometimes these are points, sometimes they're landuse AOIs. We should still show the AOI, but the label shouldn't show until zoom 17+ (was z15+, revised recommendation 13 Jan 2016).

<img width="305" alt="screen shot 2015-10-03 at 15 21 06" src="https://cloud.githubusercontent.com/assets/853051/10265372/abd754a4-69e2-11e5-8108-7dd85eef517c.png">

<img width="477" alt="screen shot 2015-10-03 at 15 20 56" src="https://cloud.githubusercontent.com/assets/853051/10265374/abf6f566-69e2-11e5-96d0-ae0a5a855cda.png">

nvkelso:
Related: #267.

zerebubuth:
A couple of questions:

1. How do we want to identify these? Looking in the North America extract, there are 49,437 points and 146 polygons with `name ILIKE '% (historical)'`. Do we want to suppress all of those, or only some of them? See below for more detail.
2. The example given, "[Naval Hospital Oakland](https://en.wikipedia.org/wiki/Naval_Hospital_Oakland)" seems like it's not just historical, but in fact no longer a hospital at all. Would this instance perhaps be better dealt with by removing `amenity=hospital` from the source data? If so, is that something we could or should apply more generally?

### Historical stuff

For the polygons, the top counts of `kind` (plus "building" when it's a building but no matching kind - usually means it has a `disused:amenity` tag or similar - [e.g](http://www.openstreetmap.org/way/346638999)):

```
        kind         | num 
---------------------+-----
 building            |  49
 school              |  41
 place_of_worship    |  14
 water               |  13
```

And the top values for points:

```
          kind           |  num  
-------------------------+-------
 school                  | 35068
 post_office             |  6262
 place_of_worship        |  6186
 hospital                |   371
 grave_yard              |   191
 park                    |   178
```

The reason there are so many more points than polygons is that many of these features came from the GNIS import and have never been touched or verified.

nvkelso:
Your frequency list match what I've seen on the map. 

[...]

PR #579, opened 2016-03-08
Filter on parenthetical properties

zerebubuth:
Enable suppression of pois and buildings where they end with '(closed)' or '(historical)'.

Requires mapzen/TileStache#128. Connects to #291 and connects to #483.


rmarianski:
:+1:

Issue #501, opened 2016-01-20
Surprising number of rail features at low zooms


nvkelso:
I'd expect to see one or two here, but not this many! 
http://vector.mapzen.com/osm/all/7/20/48.topojson

nvkelso:
```
transit: {
type: "GeometryCollection",
geometries: [
{
type: "LineString",
properties: {
kind: "train",
route_name: "Skunk Train",
osm_relation: true,
id: -4710759
},
arcs: [
8
]
},
{
type: "LineString",
properties: {
kind: "railway",
route_name: "California Western Railroad",
osm_relation: true,
id: -4710758
},
arcs: [
9
]
},
{
type: "LineString",
properties: {
operator: "Northwestern Pacific Railroad",
kind: "railway",
route_name: "NWP Mainline",
osm_relation: true,
id: -4520158
},
arcs: [
10
]
},
{
type: "LineString",
properties: {
operator: "Northwestern Pacific Railroad",
kind: "railway",
route_name: "NWP Mainline",
osm_relation: true,
id: -4520158
},
arcs: [
11
]
},
{
type: "LineString",
properties: {
operator: "Northwestern Pacific Railroad",
kind: "railway",
route_name: "NWP Mainline",
osm_relation: true,
id: -4520158
},
arcs: [
12
]
},
{
type: "LineString",
properties: {
operator: "Northwestern Pacific Railroad",
kind: "railway",
route_name: "NWP Mainline",
osm_relation: true,
id: -4520158
},
arcs: [
13
]
},
{
type: "LineString",
properties: {
kind: "railway",
route_name: "West Line",
osm_relation: true,
id: -4508022
},
arcs: [
14
]
},
{
type: "LineString",
properties: {
operator: "Union Pacific Railroad",
kind: "railway",
route_name: "West Valley",
osm_relation: true,
id: -4498171
},
arcs: [
15
]
},
{
type: "LineString",
properties: {
operator: "Union Pacific Railroad",
kind: "railway",
route_name: "West Valley",
osm_relation: true,
id: -4498171
},
arcs: [
16
]
},
{
type: "LineString",
properties: {
kind: "railway",
route_name: "Roseville Subdivision",
osm_relation: true,
id: -4489569
},
arcs: [
17
]
},
{
type: "LineString",
properties: {
kind: "railway",
route_name: "Roseville Subdivision",
osm_relation: true,
id: -4489569
},
arcs: [
18
]
},
{
type: "LineString",
properties: {
[...]

PR #585, opened 2016-03-09
Drop physical railways from transit.

zerebubuth:
Drop physical railways from `transit` layer. Merge geometry of remaining rail services / lines.

Connects to #501.


rmarianski:
:+1:

PR #591, opened 2016-03-11
Add functions of row type for points and polygons tables.

zerebubuth:
Add a kind of polymorphism so that these functions can be called on whole rows with, e.g: `select mz_calculate_min_zoom_landuse(p.*) from planet_osm_point p`.


rmarianski:
:zap:

Issue #506, opened 2016-01-22
Adjust tile rank for transit stations per railway type


nvkelso:
In London the `kind_tile_rank` sometimes biases towards lower quality service features like light rail & trams to the detriment of larger stations serving "heavy" rail. The same thing is sometimes true in New York (comparing Manhattan to New Jersey when suddenly many light rail appear when only heavy rail appear in Manhattan). There are probably two factors at work here: tile versus meta tile, and railway type. This issue is focused only on railway type.

Since OSM doesn't have trip frequency level data, the route relations act as a good proxy, but the fact that something has more physical infrastructure investment should also act as a signal that there are probably more trips passing thru a station.

On railway type, from @zerebubuth:

```
If we want to count subway routes separately from overground routes, 
then we can do that. Many (although not all) route relations are tagged 
to say whether they're surface, subway, light rail or tram lines. It just 
seems like this makes comparing stations harder (i.e: which is more 
important, the one with 4 surface lines and 2 subway lines, or the one 
with 2 surface lines, 1 light rail line and 3 tram lines?).
```

Proposed solution is to determine the "preferred" / "best" / "dominant" transit type for a station, and to also sort based on that as a tie breaker, maybe weight it, add zooms to the mix?

* heavy rail station A with 2 lines, light rail station with 2 lines.

not

* light rail station A with 2 lines, heavy rail station with 2 lines.

London
12/51.5093/-0.0760

<img width="1025" alt="screen shot 2016-01-21 at 16 12 18" src="https://cloud.githubusercontent.com/assets/853051/12498766/ce08f7b2-c059-11e5-8743-5acdd1060bd6.png">

zerebubuth:
One possible way to do it, which feels like it might be quite intuitive, is to take:

1. The number of mainline rail routes, up to a maximum of 9 and multiply by 100.
2. The number of subway plus light rail routes, up to a total maximum of 9 and multiply by 10.
3. The number of tram / other routes, up to a maximum of 9.

Sum these, and we'll get a score for each station between 0 and 999 (it would be interesting to see if any station in the world gets 999). That can then be the value used for sorting to produce the final rank.

Alternatively, if we think that something with 9 light rail routes should be more important that a single mainline route, then we could add them together after scaling, say `6 * mainline + 2 * light + others` or something to that effect.

Finally, a few observations about London:

1. What's that "tear" near West Silvertown near the river to the east?
[...]

Issue #590, opened 2016-03-11
Add `osm_site_relation: true` on transit stations


nvkelso:
For debugging transit, it's useful to know if a transit station POI is part of a site relation (or a site group relation). 

We learn that attribute as we calculate the `tile_kind_rank`, let's export that on the feature.

nvkelso:
Verified on:

* https://www.openstreetmap.org/relation/224211
* https://www.openstreetmap.org/relation/4637816

<img width="438" alt="screen shot 2016-03-24 at 00 11 30" src="https://cloud.githubusercontent.com/assets/853051/14010730/2b0473de-f155-11e5-8dbd-64d8f96b2f8d.png">

<img width="689" alt="screen shot 2016-03-24 at 00 12 11" src="https://cloud.githubusercontent.com/assets/853051/14010729/2ae8a5e6-f155-11e5-8361-4825dc4b9048.png">

PR #607, opened 2016-03-15
Adjust tile rank for stations

zerebubuth:
Adjust tile rank for stations. This:

1. Breaks out the different types of routes going through a station.
2. Calculates a numeric score for the station prioritising stations which have mainline rail, subway and light rail. A bonus is awarded for being an interchange between mainline rail and subway or light rail.
3. Sorts on this new score instead of the number of routes.
4. Adds a `root_relation_id` for linking together features in the same public transport or site relation.

Part of this is now that `area` is no longer a meaningful metric to be deciding at what zoom stations come in. For now, they're queried from the database starting at zoom 10, and then dropped by the last stages of the post-processor after having been scored. This (hopefully) will mean that only the important stations show for crowded, metropolitan areas but also that single-line mainline stations in rural areas will also be available at low zooms. It remains to be seen what the impact of pulling so many stations at low zooms will be - but stations are fairly rare features...

Connects to #506.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -152,7 +152,7 @@ BEGIN
       WHEN natural_val IN ('peak', 'volcano')
         THEN 11 -- these are generally point features
       WHEN railway_val IN ('station')
-        THEN LEAST(zoom + 0.38, 12)
+        THEN 10
       WHEN tourism_val = 'zoo'
         THEN LEAST(zoom + 3.00, 13)
       WHEN (natural_val IN ('spring')
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -536,7 +536,7 @@ BEGIN
     WHERE parts && ARRAY[way_id]
       AND parts[way_off+1:rel_off] && ARRAY[way_id]));
 END;
-$$ LANGUAGE plpgsql IMMUTABLE;
+$$ LANGUAGE plpgsql STABLE;
 
 -- adds the prefix onto every key in an hstore value
 CREATE OR REPLACE FUNCTION mz_hstore_add_prefix(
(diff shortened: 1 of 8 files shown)


zerebubuth:
Refs #590 as well for the root relation ID addition.

data/functions.sql
```diff
@@ -152,7 +152,7 @@ BEGIN
       WHEN natural_val IN ('peak', 'volcano')
         THEN 11 -- these are generally point features
       WHEN railway_val IN ('station')
-        THEN LEAST(zoom + 0.38, 12)
+        THEN 10
```
rmarianski: I'll incorporate this change into the csv branch I've been working on.

data/functions.sql
```diff
@@ -536,7 +536,7 @@ BEGIN
     WHERE parts && ARRAY[way_id]
       AND parts[way_off+1:rel_off] && ARRAY[way_id]));
 END;
-$$ LANGUAGE plpgsql IMMUTABLE;
+$$ LANGUAGE plpgsql STABLE;
```
rmarianski: Nice catch!

data/functions.sql
```diff
@@ -640,21 +640,94 @@ END;
...
+RETURNS text AS $$
+DECLARE
+  route_name text;
+BEGIN
+  SELECT TRIM(BOTH FROM "name") INTO route_name
+    FROM (
+      SELECT UNNEST(string_to_array(COALESCE(
+        -- prefer ref as it's less likely to contain the destination name
+        tags->'ref',
+        tags->'name'
```
rmarianski: Will we have `name` set here on `tags`?

data/functions.sql
```diff
@@ -640,21 +640,94 @@ END;
...
+
+    -- the list of distinct refs (or names if refs aren't present) on each
+    -- type of route.
+    train_routes      text[],
+    subway_routes     text[],
+    light_rail_routes text[],
+    tram_routes       text[],
+    railway_routes    text[]
+  );
+END$$;
```
rmarianski: `sql_level++` :zap:

data/functions.sql
```diff
@@ -668,63 +741,200 @@ DECLARE
...
+          r.id = ANY(path)
+        FROM
+          planet_osm_rels r JOIN downward_search s
+        ON
+          r.id = ANY(s.parts[s.rel_off+1:array_upper(s.parts,1)])
+        WHERE
+          mz_is_interesting_transit_relation(hstore(r.tags)) AND
+          NOT cycle
+    ) SELECT id FROM downward_search WHERE NOT cycle);
+
```
rmarianski: `sql_level++`

data/functions.sql
zerebubuth: Not necessarily, but if `tags->'name'` doesn't exist, then it will just return NULL. The whole query should return NULL and set `route_name` to NULL:

```SQL
select trim(both from "name") from (
  select unnest(string_to_array(coalesce(null::text, null::text), ',')) as "name"
) x;
```

Returns 0 rows.

queries/pois.jinja2
```diff
@@ -7,7 +7,12 @@ SELECT
     cuisine,
     religion,
     sport,
-    transit_routes,
+    (transit_routes).score AS mz_transit_score,
+    (transit_routes).root_relation_id AS mz_transit_root_relation_id,
+    (transit_routes).train_routes AS train_routes,
+    (transit_routes).subway_routes AS subway_routes,
+    (transit_routes).light_rail_routes AS light_rail_routes,
+    (transit_routes).tram_routes AS tram_routes,
```
rmarianski: I'll make sure to incorporate this into the other branch too.

data/functions.sql
zerebubuth: Yeah, especially with the `WITH RECURSIVE` CTEs below... that was a whole new thing, and full of interesting restrictions.

data/functions.sql
zerebubuth: I think this whole function is a good argument in favour of NEAT :wink:

rmarianski:
:+1: 

The relation walk sounds expensive. I take it we actually need to do that to get the best results here though. Should we consider pre-calculating this with a trigger? Or are the actual number of stations relatively low so we won't feel the impact here?

zerebubuth:
> The relation walk sounds expensive. I take it we actually need to do that to get the best results here though. Should we consider pre-calculating this with a trigger? Or are the actual number of stations relatively low so we won't feel the impact here?

I'm hoping for the number of stations is relatively low. But I can do a whole station (King's X) in 28ms, and every station in London (414 of them) in 4632ms... So that makes me hope that it's not as expensive as it might appear - I think there generally aren't that many steps in the recursion.

If we notice a significant slow-down from this, then we should consider storing the result in a column and updating using a trigger. The difficulty there will be that the trigger will also need to fire if any of the relations found during the sweep changes, which probably means having another column for "dependencies" or something like that.

rmarianski:
> The difficulty there will be that the trigger will also need to fire if any of the relations found during the sweep changes, which probably means having another column for "dependencies" or something like that.

Ah good point, I hadn't considered that. Sounds good the way it is.

data/functions.sql
rmarianski: I meant to ask if we would ever have name set here. Or is it because it's coming from the rels table and that always has everything in there?

zerebubuth: Oh, sorry, I misunderstood.

Yeah, the schema of the `_rels` table means it hasn't got any columns from tags, there's just a `tags` hstore with the tags in. I'm not sure if it's everything or whether `osm2pgsql` already dropped `note` and the other stuff it drops, but it's almost everything...

Issue #580, opened 2016-03-08
Use csv matching functionality to assign layer kinds


rmarianski:
Creating `kind`s for layers is usually implemented as a lookup. Now that we have the csv matching functionality, we should consider transitioning the `kind` assignment to a csv too.

nvkelso:
Looking great, I don't see any regressions and much simpler to read and maintain :)

PR #610, opened 2016-03-16
Generate sql functions for kind and min_zoom

rmarianski:
Connects to #580.


data/apply-planet_osm_polygon.sql
```diff
@@ -13,30 +13,26 @@ CREATE INDEX planet_osm_polygon_admin_level_geom_index ON planet_osm_polygon USI
...
-CREATE INDEX planet_osm_polygon_landuse_boundary_geom_8_index ON planet_osm_polygon USING gist(way) WHERE mz_is_landuse=TRUE AND (boundary IN ('national_park', 'protected_area') OR leisure='nature_reserve') AND mz_landuse_min_zoom <= 8;
+CREATE INDEX planet_osm_polygon_landuse_boundary_geom_4_index ON planet_osm_polygon USING gist(way) WHERE mz_landuse_min_zoom IS NOT NULL AND (boundary IN ('national_park', 'protected_area') OR leisure='nature_reserve') AND mz_landuse_min_zoom <= 4;
+CREATE INDEX planet_osm_polygon_landuse_boundary_geom_6_index ON planet_osm_polygon USING gist(way) WHERE mz_landuse_min_zoom IS NOT NULL AND (boundary IN ('national_park', 'protected_area') OR leisure='nature_reserve') AND mz_landuse_min_zoom <= 6;
+CREATE INDEX planet_osm_polygon_landuse_boundary_geom_8_index ON planet_osm_polygon USING gist(way) WHERE mz_landuse_min_zoom IS NOT NULL AND (boundary IN ('national_park', 'protected_area') OR leisure='nature_reserve') AND mz_landuse_min_zoom <= 8;
```
zerebubuth: Do we need both `mz_landuse_min_zoom IS NOT NULL` and `mz_landuse_min_zoom <= N` in the lines above? As I understand it, `NULL <= N` will also be `NULL`, which is falsey in the where clause.

zerebubuth:
Should we also drop `mz_calculate_landuse_min_zoom` from `data/functions.sql`?

queries/buildings.jinja2
```diff
@@ -4,8 +4,7 @@ SELECT
     -- name tag, if it has been set. likewise for landuse areas
     -- where we might want to label the landuse rather than the
     -- building.
-    CASE WHEN mz_poi_min_zoom IS NULL AND
-              (mz_is_landuse IS NULL OR mz_is_landuse = FALSE)
+    CASE WHEN mz_poi_min_zoom IS NULL
```
zerebubuth: I think this still needs clause for landuse to make sure that the names on landuse or POI polygons go to the landuse or POI label rather than the building label. Should it be `CASE WHEN mz_poi_min_zoom IS NULL AND mz_landuse_min_zoom IS NULL`?

data/migrations/v0.9.0-cleanup.sql
```diff
@@ -2,3 +2,25 @@
 DROP FUNCTION IF EXISTS mz_calculate_poi_level(
   text, text, text, text, text, text, text, text, text, text, text, text, text,
   text, text, text, text, text, hstore, real);
+
```
zerebubuth: Should we also drop the 3 landuse-related functions here? `mz_calculate_is_landuse`, `mz_calculate_landuse_kind` & `mz_calculate_landuse_min_zoom`?

data/apply-planet_osm_polygon.sql
rmarianski: Nice catch! Fixed in 186cb46.

data/migrations/v0.9.0-cleanup.sql
rmarianski: Nice catch! Fixed in 19c4028.

queries/buildings.jinja2
rmarianski: Thanks for spotting this! Fixed in c7d2fae.

rmarianski:
@zerebubuth thanks for spotting all those :zap: ! I pushed over fixed.

data/migrations/v0.9.0-cleanup.sql
```diff
@@ -1,7 +1,12 @@
--- old version of the POI level calculation without the 'disused' parameter.
+-- drop no longer used functions, supplanted by generated functions
 DROP FUNCTION IF EXISTS mz_calculate_poi_level(
   text, text, text, text, text, text, text, text, text, text, text, text, text,
   text, text, text, text, text, hstore, real);
+DROP FUNCTION IF EXISTS mz_calculate_landuse_min_zoom(text, text, text, text, text, text, text, text, text, text, real);
+DROP FUNCTION IF EXISTS mz_calculate_landuse_kind(text, text, text, text, text, text, text, man_text, text, text);
```
zerebubuth: I think the regexp replace might have needed an underscore for `man_made` in this and the next 3 lines?

zerebubuth:
Other than the `man_text` thing in 19c4028, everything else looks great! :+1:

data/migrations/v0.9.0-cleanup.sql
rmarianski: Whoops! Fixed in f93b845.

PR #612, opened 2016-03-16
Fix transit stations and migration

zerebubuth:
Fix the migration so that instead of trying to replace an existing function, which would cause problems with the return type being unexpected in existing concurrent queries, a new function with a new name is created instead. This means we can just drop the old one once the migration and deploy are complete.

Also fixed a small bug - the new function supports relations, so the `osm_id > 0` check was obsolete.


rmarianski:
:+1:

Issue #588, opened 2016-03-10
Add funicular & monorail to transit layer


nvkelso:
Paris has the [Montmartre Funicular](https://en.wikipedia.org/wiki/Montmartre_Funicular) which is part of the public transit network. Looking at Wikipedia, including funiculars in the public transit network is fairly common ([list](https://en.wikipedia.org/wiki/List_of_funicular_railways)). 

Looking at TagInfo [values](https://taginfo.openstreetmap.org/keys/route#values), around 72 of these are in OSM today, versus ~541 light rail lines.

* Related national rail zoom changes: #471
* Related subway & light rail / Tram zoom changes: #472

**In the mz_calculate_transit_level SQL function:**

- [x] Add new zoom block for `funicular` set to 12
- [x] Add new zoom block for `monorail` set to 12

OSM view of Montmartre Funicular:

* http://indyhurt.github.io/te-transit/map/#16/48.8846/2.3434
* <img width="801" alt="screen shot 2016-03-11 at 12 46 21" src="https://cloud.githubusercontent.com/assets/853051/13715052/597edd94-e787-11e5-93a8-c1551f542bf9.png">

nvkelso:
Also `monorail` for AirTrain at SFO airport.

nvkelso:
Here's the `monorail` at SFO airport at zoom 15 (shows up at z12 as requested):

<img width="770" alt="screen shot 2016-03-23 at 23 59 56" src="https://cloud.githubusercontent.com/assets/853051/14010566/842e1c00-f153-11e5-9f91-ed77750b267c.png">

And the `funicular` in Pittsburgh at zoom 14 (shows up at z12 as requested):

<img width="497" alt="screen shot 2016-03-24 at 00 03 35" src="https://cloud.githubusercontent.com/assets/853051/14010608/e2f6f57c-f153-11e5-9a28-94906c672c1e.png">

The funicular in Paris wasn't tagged right, I've fixed that in OSM.

PR #614, opened 2016-03-17
Add funicular and monorail

rmarianski:

Unfortunately, I couldn't find any route=funicular features in north america so I didn't add a test for it.

### DIFF
--- a/spreadsheets/transit.csv
+++ b/spreadsheets/transit.csv
@@ -1,5 +1,7 @@
 kind,sort_key
-tram,415
-light_rail,415
+light_rail,417
+tram,416
+funicular,415
+monorail,414
 platform,385
 *,275
--- a/spreadsheets/transit.csv
+++ b/spreadsheets/transit.csv
@@ -1,7 +1,7 @@
 kind,sort_key
-light_rail,417
-tram,416
-funicular,415
-monorail,414
+light_rail,418
+tram,417
+funicular,416
+monorail,415
 platform,385
 *,275
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -173,6 +173,7 @@ BEGIN
              WHEN route_val = 'train' THEN 6
              WHEN route_val = 'subway' THEN 8
              WHEN route_val IN ('light_rail', 'tram') THEN 9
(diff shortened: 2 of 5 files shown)


rmarianski:
Connects to #588

zerebubuth:
Does it need a sort key adding in the transit CSV? Also, the new values should be mentioned in the docs.

rmarianski:
Good call on the docs. Updated in a206a2a.

And thanks for catching the `sort_key` too! @nvkelso should these new features be at the same level as `tram` and `light_rail`? ie `415`?

nvkelso:
I've been to these two Pittsburgh "inclines", which were in OSM but missing their relation:

* Duquesne Incline: [OSM relation](http://www.openstreetmap.org/relation/6060405) and [Wikipedia](https://en.wikipedia.org/wiki/Duquesne_Incline)
* Monongahela Incline: [OSM relation](http://www.openstreetmap.org/relation/6060404) and [Wikipedia](https://en.wikipedia.org/wiki/Monongahela_Incline)

nvkelso:
@rmarianski We should probably break them out:

```
kind,sort_key
light_rail,417
tram,416
funicular,415
```

rmarianski:
Updated in 953074f. I added an explicit `sort_key` for `monorail` too, just below `funicular`.

rmarianski:
Thanks for updating those features in osm! Added a test for funicular in 09e0073.

spreadsheets/transit.csv
```diff
@@ -1,5 +1,7 @@
 kind,sort_key
-tram,415
-light_rail,415
+light_rail,417
+tram,416
+funicular,415
+monorail,414
```
nvkelso: This range should start at 415, there's only a couple slots at 414 and below before there are collisions with bridges. The room to grow is up towards 425. So monorail 415, funicular 416, and on up.

zerebubuth:
:+1: looks great now!

spreadsheets/transit.csv
rmarianski: Updated in f0090ef.

nvkelso:
LGTM

Issue #471, opened 2016-01-08
Categorize regional & national public transit trains


nvkelso:
In the `transit` layer, we just tossed in all `route=train` features. Let's characterize if it's a regional train or not so the styling and zoom ranges can be distinct.

* Related subway & light rail / Tram zoom changes: https://github.com/mapzen/vector-datasource/issues/472
* Related to funicular zoom changes: https://github.com/mapzen/vector-datasource/issues/588

**UPDATE:** Once we add the `service` property from the route relation, we can make the following selection: `service IN ('high_speed', 'long_distance', 'international')`. That set should have a **min_zoom = 5**, while the remainder should remain **min_zoom = 6**.

- [x] Include `service` in `transit.jinja` layer selection SQL
- [x] Update minimum zoom for "national" rail to 5 in the `mz_calculate_transit_level` SQL function.
- [x] Do we need to update the GIST index in `apply-planet_osm_line.sql`?

* Connects with?: https://github.com/mapzen/vector-datasource/issues/469.

nvkelso:
minimum zoom of national trains should change to 5, and regional trains should stay at 6

nvkelso:
Related service tag discussion on OSM transit relations: https://github.com/mapzen/vector-datasource/issues/548#issuecomment-193766298

nvkelso:
From the discussion in #548:

Says @zerebubuth:

The transit layer currently contains the route / "line" relations, which cover the whole length of the route / named service / "line". The `service` attribute might be different from what you expect, especially if you're looking for short segments to drop. From my North America extract, here's a table of the count of features with distinct `route` and `service` values:

```
   route    |    service    |  num  
------------+---------------+-------
 train      | regional      | 13853
 train      | high_speed    |  6768
 train      | long_distance |  3092
 train      |               |  2347
 train      | commuter      |   218
 train      | international |    68
 train      | car           |    23
 train      | tourism       |     3
```

99% of the features are covered by `service IN ('regional', 'high_speed', 'long_distance') OR service IS NULL`.

nvkelso:
Looks like there is a similar tag to service called [passenger](http://wiki.openstreetmap.org/wiki/Key:passenger). TagInfo [says](http://taginfo.openstreetmap.org/keys/?key=passenger#values), but those look to be rarely applied to route_relations. 

One example is [Amtrak California Zephyr](https://www.openstreetmap.org/edit?relation=905830#map=19/39.00518/-120.98230).

* `passenger=national`
[...]

PR #616, opened 2016-03-17
Categorize regional/national public transit trains

rmarianski:
Connects to #471

@nvkelso should we make a more specific sort_key entry for these national/regional trains compared to the other trains (without this service tag)?

I made a small update to the docs about when train features start appearing, but I didn't list the specific tag combinations for it. Could certainly update it to be more explicit if we want to do that.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -439,7 +439,7 @@ Railway values in this layer include: `rail`, `tram`, `light_rail`, `narrow_gaug
 * Layer name: `transit`
 * Geometry types: `line`, `polygon`
 
-Transit line features from OpenStreetMap start appearing at zoom 5+ for regional and national trains. Then `subway`,`light_rail`, and `tram` are added at zoom 10+. Platform polygons are added zoom 14+.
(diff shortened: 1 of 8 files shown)


zerebubuth:
:+1: looks good to me.

docs/layers.md
```diff
@@ -439,7 +439,7 @@ Railway values in this layer include: `rail`, `tram`, `light_rail`, `narrow_gaug
 * Layer name: `transit`
 * Geometry types: `line`, `polygon`
 
-Transit line features from OpenStreetMap start appearing at zoom 6+ for basic rail kind of `train`. Then `subway`,`light_rail`, and `tram` are added at zoom 10+. Platform polygons are added zoom 14+.
+Transit line features from OpenStreetMap start appearing at zoom 5+ for regional and national trains. Then `subway`,`light_rail`, and `tram` are added at zoom 10+. Platform polygons are added zoom 14+.
```
nvkelso: "zoom 5 for national trains, with regional trains added at zoom 6."

nvkelso:
Let's leave the sort_key alone at this time. (If we need to adjust it, we can do that as part of https://github.com/mapzen/vector-datasource/issues/353.)

docs/layers.md
rmarianski: Updated in 47f7f47.

PR #620, opened 2016-03-18
Drop mz_is_building property from landuse layer

rmarianski:
Connects to #333


rmarianski:
Updated issue: https://github.com/mapzen/vector-datasource/issues/619

nvkelso:
LGTM

zerebubuth:
Follow-up for later in #622 - let's exclude everything starting `mz_` so that we don't have to keep watching for these?

Issue #611, opened 2016-03-16
Add an is_bus tag to roads


nvkelso:
For OSM bus route relations, tag the roads with `is_bus` if one or more relations pass thru that way.

nvkelso:
Only at zoom 11+. 
Don't calculate or drop at zooms 0-10.

nvkelso:
If there's a better name than `is_bus`, by all means :)

nvkelso:
This is so awesome!

**1) Looks like we're missing a few "bus" roads when they are `trolley bus`?** For our purpose, we should pretend like `trolley_bus` are in the same class as basic `bus`. 

![bus_funk](https://cloud.githubusercontent.com/assets/853051/14009852/141286be-f14d-11e5-80eb-ab7c2bb4d54b.gif)

Some examples that are missing in ours but present in the default OSM.org transport style:

* SF route 24: http://www.openstreetmap.org/relation/2996736
* SF route 33: http://www.openstreetmap.org/relation/2097813
* SF route 41: http://www.openstreetmap.org/relation/3413068

Correctly "missing" from the earlier example GIF because it's a service road:

* SF route 38: http://www.openstreetmap.org/relation/3443930

**2) For now let's _not_ calculate the attribute at zoom 11 (only 12+).** We only add minor roads at zoom 12, so the network looks incomplete at zoom 11. 

A later improvement could be to include minor roads (and even service roads) earlier at zoom 11 – if they were `is_bus_route:true only`, but that's out of scope for this release.

<img width="651" alt="screen shot 2016-03-23 at 14 24 03" src="https://cloud.githubusercontent.com/assets/853051/14001434/0e2986c0-f103-11e5-8039-587ce996c3dd.png">

<img width="595" alt="screen shot 2016-03-23 at 14 24 12" src="https://cloud.githubusercontent.com/assets/853051/14001433/0e0f5ee4-f103-11e5-853d-ff3c98eb0ee8.png">

nvkelso:
Zoom 12 change verified.

I'm still not seeing trolley_bus roads marked `is_bus_route`, though?

nvkelso:
This is generally working, with final fix in #629.

PR #624, opened 2016-03-21
Add bus route property.

zerebubuth:
Adds a new property `is_bus_route` on roads to indicate if they are members of bus route relations.

Connects to #611.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -688,3 +688,21 @@ UPDATE wof_neighbourhood a
     b.new_visible <> a.is_visible
   RETURNING a.wof_id;
 $$ LANGUAGE sql VOLATILE;
+
+-- returns TRUE if the given way ID (osm_id) is part of a bus route relation,
+-- or NULL otherwise.
+CREATE OR REPLACE FUNCTION mz_calculate_is_bus_route(osm_id BIGINT)
+RETURNS BOOLEAN AS $$
+BEGIN
+  IF EXISTS(
+    SELECT 1 FROM planet_osm_rels
+    WHERE
+      parts && ARRAY[osm_id] AND
+      parts[way_off+1:rel_off] && ARRAY[osm_id] AND
+      hstore(tags)->'type' = 'route' AND
+      hstore(tags)->'route' = 'bus') THEN
+    RETURN TRUE;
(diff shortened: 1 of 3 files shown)


zerebubuth:
I should mention performance quickly:

Using the test suite as a baseline, it ran in 5m42.33s, 5m42.06s and 5m43.19s before I made this modification and 5m44.08s, 5m41.13s and 5m41.99s. There's no significant difference between the run-times, which suggests that the performance impact could be small. We'll know more when we do a dev push.

rmarianski:
Thanks for reporting those numbers. Not sure how representative the tests are necessarily, but it's a great sign that it wasn't significantly longer. I agree that we should just wait and see on dev.

data/functions.sql
```diff
@@ -688,3 +688,21 @@ UPDATE wof_neighbourhood a
...
+CREATE OR REPLACE FUNCTION mz_calculate_is_bus_route(osm_id BIGINT)
+RETURNS BOOLEAN AS $$
+BEGIN
+  IF EXISTS(
+    SELECT 1 FROM planet_osm_rels
+    WHERE
+      parts && ARRAY[osm_id] AND
+      parts[way_off+1:rel_off] && ARRAY[osm_id] AND
+      hstore(tags)->'type' = 'route' AND
+      hstore(tags)->'route' = 'bus') THEN
```
nvkelso: We should capture both `bus` or `trolleybus`.

See also:

* http://wiki.openstreetmap.org/wiki/Tag:route%3Dtrolleybus

queries/roads.jinja2
```diff
@@ -68,6 +68,14 @@ SELECT
     leisure,
     sport,
     man_made,
+{% if zoom >= 11 %}
```
nvkelso: Since we don't have minor roads at zoom 11, let's make this >= 12 for this release. Otherwise it looks broken at zoom 11.

data/functions.sql
zerebubuth: Tracking in #629.

queries/roads.jinja2
zerebubuth: Tracking in #630.

PR #625, opened 2016-03-21
nvkelso/docs overlay underlay

nvkelso:
* provide 4 basic overlay, underlay `order` ranges
* recommend which ranges should be used for overlay and underlay.
* explain how the ranges sit with respect to UI elements and map labels.


nvkelso:
Connects to: https://github.com/mapzen/vector-datasource/issues/586.

Issue #469, opened 2016-01-08
Add new transit layer features


nvkelso:
To support more advanced transit applications, we need to augment the existing set of transit attributes in the `transit` layer.

Add additional line attributes (zoom 11+ only):

* from
* to
* forward
* backward
* alternate

Add nodes for:

* railway=halt (zoom 11)
* railway=tram_stop (zoom 11)
* highway=platform* (zoom 15)
* public_transport=platform* (zoom 15)
* railway=platform* (zoom 15)
* railway=stop* (zoom 11)

`*` The last 4 stop types depend on additional tags such as rail=yes (and should exclude bus=yes, trolleybus=yes).

Stations: 

* public_transport=stop_area (zoom 15)
* site=stop_area (zoom 15)

**Add node attributes:**

This should be in the `transit` layer and the `pois` layer for existing stations there.

* [uic_ref](http://wiki.openstreetmap.org/wiki/Key:uic_ref) - like IATA codes for airports.

References:

* http://wiki.openstreetmap.org/wiki/Relation:route#Public_transport_routes
* http://wiki.openstreetmap.org/wiki/WikiProject_Metro_systems
* http://wiki.openstreetmap.org/wiki/Public_transport

Example applications:

* http://www.openbusmap.org/?zoom=10&lat=37.85457&lon=-121.97473&layers=BT
* http://wiki.openstreetmap.org/wiki/%C3%96pnvkarte
* http://demo.3liz.com/osmtransport/
* http://openptmap.org/
* http://wiki.openstreetmap.org/wiki/Openptmap

nvkelso:
Some or most of the nodes listed above should end up in the `pois` layer (not the `transit` layer).

nvkelso:
In the nodes list above, the following two are already covered in https://github.com/mapzen/vector-datasource/issues/587:

* railway=halt (zoom 11)
* railway=tram_stop (zoom 11)

nvkelso:
Over to Matt to list what's been done, what's outstanding. Then remove assignment.

zerebubuth:
## Add additional line attributes (zoom 11+ only):

* from
* to
* forward
* backward
* alternate

## Add nodes for:

* `highway=platform` (zoom 15)
* `public_transport=platform` (zoom 15)
* `railway=platform` (zoom 15) **These are already included for lines and polygons**
* `railway=stop` (zoom 11)

[...]

PR #626, opened 2016-03-22
Add new transit features

rmarianski:
Connects to #469

This is still a work in progress, but I wanted to share what I had thus far in case you noticed something already. I was originally going to just move ahead with trying to get it to work with manual sql functions, but now I think it'll actually be easier to transition to a csv for transit as well because the kind isn't being set properly. The logic is the same as with the pois layer, so it should be easy enough to just copy over the relevant csv rows for transit and query those for lines/polygons.

### DIFF
--- a/data/migrations/v0.9.0-polygons.sql
+++ b/data/migrations/v0.9.0-polygons.sql
@@ -4,6 +4,13 @@ UPDATE planet_osm_polygon SET
     "disused" <> 'no' OR
     "railway" = 'station';
 
+ALTER TABLE planet_osm_polygon ADD COLUMN mz_transit_level REAL;
+
+UPDATE planet_osm_polygon SET
+  mz_transit_level = mz_calculate_min_zoom_transit(planet_osm_polygon.*)
+  WHERE
+    mz_calculate_min_zoom_transit(planet_osm_polygon.*) IS NOT NULL;
+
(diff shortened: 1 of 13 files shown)


data/migrations/v0.9.0-polygons.sql
```diff
@@ -4,6 +4,13 @@ UPDATE planet_osm_polygon SET
     "disused" <> 'no' OR
     "railway" = 'station';
 
+UPDATE planet_osm_polygon SET
```
zerebubuth: Does this need to have the same `ALTER TABLE` as `apply-planet_osm_polygon.sql` to add the column?

rmarianski: Nice catch! Updated in ae52cae.

rmarianski:
I transitioned to using the a csv file for `transit` as well as mentioned in the description and rebased the commits.

Something I just realized is that the score calculation is now only done for the `pois` layer. Is that something that we'll want to do in the `transit` layer too? If so, across both line and polygon tables?

nvkelso:
CSV is legible, nice work!

I believe we only calculate the score for `tile_kind_rank` on `kind=station` features. We might want to extend that in the future to more kinds, but out of scope for this release I think.

We should only ever calculate that for POI layer content (not transit layer content which should just be lines and polygons).

zerebubuth:
:+1: looks good. :zap: on the macro for the transit SQL - that looks great.

Issue #629, opened 2016-03-24
Add trolleybus routes


zerebubuth:
Refs [this comment](https://github.com/mapzen/vector-datasource/pull/624/files/9f2f64412f28e237cee719dde09030544379a1fe#r57278348) for context.

Add trolleybus routes to the `is_bus_route` property.

@nvkelso, should trolleybuses be counted as a type of bus (and trigger `is_bus_route: true`) or should we have a separate property for `is_trolleybus_route`?

zerebubuth:
Confusement: [This comment](https://github.com/mapzen/vector-datasource/issues/630#issuecomment-200825966) should have been on _this_ ticket.

nvkelso:
Verified:

<img width="631" alt="screen shot 2016-03-24 at 15 55 27" src="https://cloud.githubusercontent.com/assets/853051/14033600/e75af52a-f1d8-11e5-8700-c2970060eda3.png">

PR #633, opened 2016-03-24
Make trolleybus a kind of bus.

zerebubuth:
Make trolleybus a kind of bus for the purposes of the `is_bus_route` property.

Connects to #629.


rmarianski:
:+1:

PR #634, opened 2016-03-24
Update public_transport funk

rmarianski:
Connects to #469


nvkelso:
LGTM

spreadsheets/kind/transit.csv
```diff
@@ -6,9 +6,9 @@ highway,public_transport,railway,route,service,tags->rail,tags->site,min_zoom,ki
 *,*,*,funicular;monorail,*,*,*,${12},${route}
 *,*,halt;stop;tram_stop,*,*,*,*,${13},${railway}
 platform,*,*,*,*,*,*,${15},platform
-*,platform,*,*,*,yes,*,${15},platform
+*,platform,*,*,*,yes,*,${17},platform
```
zerebubuth: Should the kind here be `bus_stop`?

zerebubuth:
I think the changes to POIs would make it turn something that was plain `public_transport=platform` into a railway platform, where it's missing the positive `tags->bus='yes'`? Should that be `tags->rail != 'yes'` instead?

Also, tests would be good for figuring out what the effect of these would be, i.e: including one for the various different [common cases](https://github.com/mapzen/vector-datasource/issues/469#issuecomment-199357595).

spreadsheets/kind/transit.csv
rmarianski: Nice catch, fixed in 08b3202.

spreadsheets/kind/transit.csv
```diff
@@ -1,14 +1,15 @@
...
-*,*,*,*,*,*,stop_area,${15},stop_area
-bus_stop,*,*,*,*,*,*,${17},bus_stop
+highway,public_transport,railway,route,service,tags->light_rail,tags->rail,tags->site,min_zoom,kind
+*,*,*,train,high_speed;long_distance;international,*,*,*,${5},${route}
+*,*,*,train,*,*,*,*,${6},train
+*,*,*,subway,*,*,*,*,${8},subway
+*,*,*,light_rail;tram,*,*,*,*,${9},${route}
+*,*,*,funicular;monorail,*,*,*,*,${12},${route}
+*,*,halt;stop;tram_stop,*,*,*,*,*,${13},${railway}
+platform,*,*,*,*,*,*,*,${15},platform
```
zerebubuth: Shouldn't `highway=platform` also be a `bus_stop` at `z >= 17`?

rmarianski: Updated in 01cce9a.

rmarianski:
@zerebubuth I added a few tests, do you think it looks good enough to merge now?

nvkelso:
Looking good. Let's get this on dev.

rmarianski:
I'll merge this in, we can make any adjustments afterwards if need be.

PR #638, opened 2016-03-24
Resolve test failures

rmarianski:
These updates allow all the tests to pass, except for the slow water boundaries one.


nvkelso:
LGTM

zerebubuth:
Did we intend to move bus stops from the `transit` to `pois` layer?

Otherwise, LGTM.

rmarianski:
I hope so! The separation I've been using is that all nodes/points go to `pois`, and lines/polygons go to `transit`.

PR #642, opened 2016-03-29
Print test coordinates instead of running tests

rmarianski:
I needed to generate a list of tile coordinates, and this was my first hack at it. @zerebubuth, is something like this what you had in mind? If so, I could try and clean this up a little bit.

It might also be worth automating a little bit more so that we can pipe this into a separate tilequeue command or script that would automatically add these to the toi and enqueue them.

### DIFF
--- a/test.py
+++ b/test.py
@@ -244,6 +244,22 @@ def assert_no_matching_feature(z, x, y, layer, properties):
                 "%r" % (properties, layer, feature['properties'])
 
 
+def print_coord(z, x, y, *ignored):
+    print '%d/%d/%d' % (z, x, y)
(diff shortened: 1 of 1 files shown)


zerebubuth:
:zap: Yeah, that's great! I tried it out locally, and it looks like it needed a bit of a tweak to get all the coords for tests like `test/507-routes-via-stop-positions.py` which call `features_in_tile_layer` directly in a loop. I've committed that in a3ce3c2.

:+1:

rmarianski:
Closed in favor of https://github.com/mapzen/vector-datasource/pull/644

PR #648, opened 2016-03-30
Normalize ne kinds and admin_levels

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/517

I'm including admin_level as an integer. It looks like the admin_level that we include from osm will be returned back as a string though. Should we also ensure that the admin_level value is always an integer?

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -108,13 +108,11 @@ Combination of OpenStreetMap administrative boundaries (zoom >= 8), Natural Eart
 
 ![image](images/mapzen-vector-tile-docs-barriers.png)
 
-**Gotchas:** Boundary `kind` values are not yet normalized between OpenStreetMap and Natural Earth. See Boundary kind values (line) gotchas section below for more detail.
-
 **Boundary properties (common):**
 
 * `name`
 * `id`
-* `kind`: mapping of OpenStreetMap's `admin_level` int values to strings like `country` and `state`, plus `aboriginal_lands` boundary type, and inclusive of some barrier and man_made tags: `city_wall` (zoom 12+), `retaining_wall`, `snow_fence` (zoom 15+), and `fence` (zoom 16+ only). Also includes raw Natural Earth values.
+* `kind`: mapping of OpenStreetMap's `admin_level` int values to strings like `country` and `state`, plus `aboriginal_lands` boundary type, and inclusive of some barrier and man_made tags: `city_wall` (zoom 12+), `retaining_wall`, `snow_fence` (zoom 15+), and `fence` (zoom 16+ only).
 * `sort_key`: a suggestion for which order to draw features. The value is an integer where smaller numbers suggest that features should be "behind" features with larger numbers.
 
 **Boundary properties (common optional):**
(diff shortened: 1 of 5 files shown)


nvkelso:
Agree the OSM admin_level should always be an integer (or dropped).

nvkelso:
We should also filter out (not export in tiles): `Admin-1 statistical boundary` and `Admin-1 statistical meta bounds`.

zerebubuth:
Was this already the documented behaviour, or do we need to update the docs too?

rmarianski:
> Agree the OSM admin_level should always be an integer (or dropped).

Transform added in pr https://github.com/mapzen/TileStache/pull/136. Configured in aa2694a.

> We should also filter out (not export in tiles): Admin-1 statistical boundary and Admin-1 statistical meta bounds.

The only `kind`s coming from ne will now be `country` or `state`. I also pruned the `sort_key` spreadsheet to remove the now unused ne `kind`s in 232afcf.

> Was this already the documented behaviour, or do we need to update the docs too?

Nice catch. Updated in 4d9ccb0.

zerebubuth:
:+1:

nvkelso:
@rmarianski Sorry for the confusion. Only the two "statistical" boundaries should be filtered out – the others should remain, but go lowercase with _ between words (and parentheticals removed) so they're more in line with all the rest of our "kind" values.

rmarianski:
Here's the list that we were using from NE:

```
1st Order Admin Lines
Admin-1 boundary
Admin-1 region boundary
Admin-1 statistical boundary
Admin-1 statistical meta bounds
Disputed (please verify)
Indefinite (please verify)
Indeterminant frontier
International boundary (verify)
Lease limit
Line of control (please verify)
Overlay limit
```

IIUC now, we'd filter out these features:

```
Admin-1 statistical boundary
Admin-1 statistical meta bounds
```

And use these values as the new kinds (I assumed lower case was better):

```
1st_order_admin_lines
admin-1_boundary
admin-1_region_boundary
disputed
indefinite
indeterminant_frontier
international_boundary
lease_limit
line_of_control
overlay_limit
```

But wouldn't it be easier to have the `kind` values match up exactly with osm, and for this to turn into more of a "subkind"?

nvkelso:
admin-1_boundary > state
admin-1_region_boundary > macroregion
disputed
indefinite
indeterminant_frontier > indeterminate
international_boundary > country
lease_limit
line_of_control
overlay_limit

This one is weird, but let's map to:

1st_order_admin_lines > state

nvkelso:
> But wouldn't it be easier to have the kind values match up exactly with osm, and for this to turn into more of a "subkind"?

Good point. But while many of these are generally `kind:country`, major political incidents can be caused by rendering them in the same line pattern as they are disputed boundaries in some way or another. I could see making those ones something like `kind:country_other` grouping the following, listing them as a `subkind` (there must be a better word for this!):

* disputed
* indefinite
* indeterminant_frontier > indeterminate
* lease_limit
* line_of_control
* overlay_limit

rmarianski:
The work for this pr has been done in ad4e11b8.

Issue #148, opened 2015-08-26
Move sea and ocean labels in places to water layer


nvkelso:
I'm seeing seas in the "places" layer, but most of the water labels are in water.

nvkelso:
From:
https://github.com/mapzen/vector-datasource/blob/master/queries/places.jinja2#L46

```
        {% if zoom >= 3 %}, 'sea'{% endif %}
```

nvkelso:
Also `ocean`

nvkelso:
Related: #400.

nvkelso:
Load balancing to v0.8

nvkelso:
**Gotchas:**

* ocean and Ocean are equivelant, both should be used in filters.
* reservoir and Reservoir are equivalent, both should be used in filters.
* The (capitalized) Natural Earth values don’t seem to be coming thru, but should be, instead they’re being grouped into lake, playa, and ocean.
* Alkaline Lake and Playa only exist in Natural Earth. Zooming in, your feature may disappear. Beware the desert around Great Salt Lake in Utah!
* kinds should probably only be lowercase

nvkelso:
Working for as expected.

PR #649, opened 2016-03-31
Move ocean and sea points to the water layer

zerebubuth:
Also gives them a `label_placement=yes` tag.

Connects to #148.


rmarianski:
:+1:

PR #678, opened 2016-04-04
Add whitewater pois

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/599

This adds the whitewater features to the pois layer. The issue referenced also requests `portage_way` features added to the `roads` layer, which this pr doesn't address. The way we manage adding to the `roads` layer will most likely change with https://github.com/mapzen/vector-datasource/issues/646. Can we merge this pr in first, and then once we have csv's for the other layers add in the `portage_way`s to the `roads` layer?


nvkelso:
Yep

zerebubuth:
:+1:

PR #682, opened 2016-04-06
Remove deprecated tile format

louh:

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,14 @@ More information is available about how to [use the vector tile service](docs/us
 
 ##### Formats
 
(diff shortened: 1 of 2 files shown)


rmarianski:
:+1:

nvkelso:
This is a tricky one. The documentation we want to show on the mapzen.com/documentation site should not reference this, since we no longer support it in the service. But our software still supports it.

Agree on removing it here, but we should then enumerate the larger format list in the project [README](https://github.com/mapzen/vector-datasource/blob/master/README.md) (instead of linking to the docs).

nvkelso:
* update service docs (remove OpenScienceMap format)
* update project docs (still include OpenScienceMap format)

Issue #646, opened 2016-03-30
Calculate kind using YAML for remaining layers


nvkelso:
We're calculating kind via CSV for POIs, landuse, transit layers. Extend that to the other layers.

nvkelso:
@zerebubuth Is there remaining work here? I think `earth.yaml` and `roads.yaml` are outstanding?

- * boundaries
    - [x] **kind** via `yaml`
    - [x] **sort_key** via `csv`
- * buildings.jinja2
    - [x] **kind** via `yaml`
    - [x] **sort_key** via `csv`
- * earth.jinja2
    - [ ] **kind** via `yaml` – There was a YAML as part of the cliffs PR, but it was [removed](https://github.com/mapzen/vector-datasource/pull/695/commits/3c0d934ae20b459c23a8551c130762c9509f3fb0) since it needed fixes (base > earth) and wasn't wired up to be used. Rescue from there?
    - [x] **sort_key** via `csv`
- * landuse.jinja2
    - [x] **kind** via `yaml`
    - [x] **sort_key** via `csv`
- * places.jinja2
    - [x] **kind** via `yaml`
    - * **sort_key** via `csv`. Not relevant since point features don't need sort_key.
- * pois.jinja2
    - [x] **kind** via `yaml`
    - * **sort_key** via `csv`. Not relevant since point features don't need sort_key.
- * roads.jinja2
    - [ ] **kind** via `yaml`. Outstanding?
    - [x] **sort_key** via `csv`
- * transit.jinja2
    - [x] **kind** via `yaml`
    - [x] **sort_key** via `csv`
- * water.jinja2
    - [x] **kind** via `yaml`
    - [x] **sort_key** via `csv`

zerebubuth:
@nvkelso - hitting a corner case: http://www.openstreetmap.org/way/308154534 is tagged both `highway=service` _and_ `whitewater=portage_way`. Should it be output as a `kind: minor_road, highway: service` or `kind: portage_way`?

nvkelso:
Let’s go with portage_way

nvkelso:
Roads is done, still need `earth`?

nvkelso:
Rescue earth YAML, update `create-sql-functions.py` and `sql.jinja2`.

nvkelso:
Everything seems to be working now :)

PR #683, opened 2016-04-06
Use CSV for places.

zerebubuth:
Use CSV to set place kinds and min zooms, except for WOF.

Connects to #646.


rmarianski:
:+1:

PR #686, opened 2016-04-07
Update sql zoom/kind generation to use yaml

rmarianski:
This converts the min_zoom and kind calculations from being specified in csv's to yaml.


zerebubuth:
:+1:

PR #694, opened 2016-04-08
Use YAML for buildings properties & min zoom calculation.

zerebubuth:
Connects to #646.


rmarianski:
:+1:

PR #695, opened 2016-04-08
Nvkelso/cliff arete ridge valley

nvkelso:
* add earth layer selections for cliff, arete, ridge, and valley
* allow earth layer to contain line features
* add earth.yaml since we have multiple kinds now
* add sort_key values for cliff, arete, and rename base to earth
* add new Postgres index for select natural features
* [tests] add tests for cliff, arete, ridge, valley
* [docs] update docs for earth layer for kind change & listing, additons of new kinds, and label_placement details.


nvkelso:
Preview images:

cliff:
z13_cliff

arete and valley:
z13_arete_valley

ridge:
z13_ridge

nvkelso:
Connects with: https://github.com/mapzen/vector-datasource/issues/601.

nvkelso:
Feels like we should have a length property here, or some other way to indicate relative importance.

rmarianski:
Did you intend to include the earth.yaml file in the pr? It's not set up for the earth layer yet, and will probably just be confusing if included now. I think that we should convert to yaml in a subsequent pr.

> Feels like we should have a length property here, or some other way to indicate relative importance.

Do we want to selectively filter certain features out based on the zoom level, ie only show features at lower zoom levels that have a certain minimum length? Or do we want to just include a new property in the results so we can rank the features or style them differently?

nvkelso:
> Did you intend to include the earth.yaml file in the pr? It's not set up for the earth layer yet, and will probably just be confusing if included now. I think that we should convert to yaml in a subsequent pr.

**Removed.** Is `yaml/earth.yaml` setup correctly but just won't be exercised yet without another config change somewhere else? 

> Do we want to selectively filter certain features out based on the zoom level, ie only show features at lower zoom levels that have a certain minimum length? Or do we want to just include a new property in the results so we can rank the features or style them differently?

In the tests `Lyell Canyon` is a huge (long) feature, but in the ridge example image some of those are relatively small. At zoom 13 we're more or less showing "all" features, so it's less of a concern than if we were starting at say zoom 11.

But if we wanted to set priority for label collisions in Tangram based on importance where to features are nearby we'd need something like `length` to calculate this on, and we do `area` right now for landuse polygons. (We wound't want to start including area on the `earth` features since it's somewhat meaningless after the import process.)

Let's get what we have on dev now, experiment with it on the map, and then made revisions in a separate issue if needed.

rmarianski:
> Removed. Is yaml/earth.yaml setup correctly but just won't be exercised yet without another config change somewhere else?

It should be grouped under a "filters" key, but the details of the rules look good to me. What'll need to happen to hook it up is to add a new column to the corresponding tables to persist the min zoom, and update the sql generation to emit the appropriate functions.

> But if we wanted to set priority for label collisions in Tangram based on importance where to features are nearby we'd need something like length to calculate this on, and we do area right now for landuse polygons.

Cool, makes sense. Thanks for clarifying.

rmarianski:
:+1:

PR #699, opened 2016-04-11
Add whitewater=portage_way to roads

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/599

Depends on https://github.com/mapzen/TileStache/pull/140

### DIFF
--- a/test/599-whitewater.py
+++ b/test/599-whitewater.py
@@ -2,3 +2,8 @@
 assert_has_feature(
     16, 34003, 23060, 'pois',
     { 'kind': 'put_in' })
+
+# way 308154534
+assert_has_feature(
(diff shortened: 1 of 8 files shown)


nvkelso:
:+1:

test/599-whitewater.py
```diff
@@ -1,3 +1,9 @@
+# node 2356117215
```
zerebubuth: This node is in Europe, and so won't be in any North America extracts we use to test with. Suggest http://www.openstreetmap.org/node/3134398100 instead?

rmarianski: Good call, updated in 210b24b.

zerebubuth:
:+1:

Issue #704, opened 2016-04-12
Many null values for buildings in v0.10 dev tiles


nvkelso:
Null properties like this should not be exported in tiles:

* `building_levels: null`
* `roof_material: null`

See:

* http://vector.dev.mapzen.com/osm/all/16/10481/25319.topojson

```
buildings: {
type: "GeometryCollection",
geometries: [
{
type: "Polygon",
properties: {
building_levels: null,
kind: "building",
roof_material: null,
name: "Main Prison",
area: 4693,
sort_key: 475,
landuse_kind: "national_park",
addr_street: null,
roof_shape: null,
roof_height: null,
building_part: null,
addr_housenumber: null,
source: "openstreetmap.org",
roof_orientation: null,
roof_color: null,
id: 128245373,
building_min_levels: null,
location: null
},
arcs: [
[
5
]
]
},
{
type: "Point",
properties: {
label_placement: "yes",
kind: "building",
roof_material: null,
name: "Main Prison",
building_levels: null,
sort_key: 475,
landuse_kind: "national_park",
addr_street: null,
area: 4693,
roof_shape: null,
id: 128245373,
building_part: null,
addr_housenumber: null,
source: "openstreetmap.org",
roof_height: null,
roof_color: null,
roof_orientation: null,
building_min_levels: null,
location: null
},
coordinates: [
625,
894
]
```

rmarianski:
This will be resolved with https://github.com/mapzen/tilequeue/pull/84

nvkelso:
Fixed on dev.

PR #707, opened 2016-04-13
Add test for null values in property values for buildings.

zerebubuth:
Test for mapzen/tilequeue#84.

Connects to #704.


rmarianski:
:+1:

Issue #593, opened 2016-03-12
Show highway=path at zoom 13


nvkelso:
Thinking about an outdoor style, the convention is to show path features at **zoom 13**. 

(We're currently showing them at zoom 14.)

nvkelso:
Verified.

PR #708, opened 2016-04-13
Fix trigger update performance

zerebubuth:
Instead of using array subset operations, which Postgres doesn't appear to be able to use the btree index for, use the `ANY` keyword, which does appear to use the index. The `WHERE` clauses aren't quite so easy to read, but it results in much better query plans against my local database.

Connects to #593.


rmarianski:
:+1:

PR #712, opened 2016-04-13
Track path route rel updates in pending table

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/711

### DIFF
--- a/data/migrations/v0.10.0-schema-prefunction.sql
+++ b/data/migrations/v0.10.0-schema-prefunction.sql
@@ -50,3 +50,15 @@ BEGIN
 END$$;
 
 DROP FUNCTION tmp_add_col(regclass, text);
+
+DO $$
+BEGIN
+ IF NOT EXISTS (
+   SELECT 1 FROM pg_class WHERE relname='mz_pending_path_major_route'
+   ) THEN
+
+   CREATE TABLE mz_pending_path_major_route (
+     osm_id BIGINT NOT NULL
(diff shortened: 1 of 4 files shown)


rmarianski:
The new table name `mz_pending_path_major_route` was just the first thing that popped into my head, and if you think of a better one I'm happy to use that instead. I'm also assuming that it makes sense to have a separate, specific table for this, and that if we need to do something similar in the future we can re-evaluate.

I plan on submitting another set of prs to update the chef recipes to contain the update sql. In terms of maintenance though, we'll have to remember to update the chef recipe as the function signature changes. What do you think of making an adapter function in the vector-datasource repo that just takes the row itself and will call the calculate function with the appropriate signature? Perhaps this might already be in progress with the change to convert the roads layer to yaml?

zerebubuth:
:+1:

The (upcoming) roads YAML conversion calls `mz_calculate_path_major_route`, so I think it might be enough for the update SQL to just say something like `update planet_osm_line set mz_road_level = NULL where osm_id in (select distinct osm_id from mz_pending_path_major_route)`. Wouldn't that mean the trigger fires and sets the road level to its intended value?

rmarianski:
Yup, it would fire and update road level value. That seems a little implicit, but it would probably be the path of least maintenance worry going forward. It would also mean that all the other triggers would fire too, but i don't see this ever being an issue. I'll take that approach.

Issue #677, opened 2016-04-04
Add waterfall features


nvkelso:
These are most often POIs, but sometimes as areas. If it's an area, let's only generate a label_placement.

- [x] [waterway=waterfall](http://wiki.openstreetmap.org/wiki/Tag:waterway%3Dwaterfall), [pois](https://www.openstreetmap.org/node/2389658224), zoom 14 – including the height=* property! Needs to win over `tourism=attraction` as it's often double tagged. Around 15000 pois, 2100 other, including large areas for Niagra Falls. Water [label_placement](https://www.openstreetmap.org/way/56539663) only? https://www.openstreetmap.org/way/56539663

nvkelso:
Into the `water` layer.

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/715

nvkelso:
Including these into the `pois` layer until #715 settles out. Added some logic for height and areas.

zerebubuth:
Should be done once #720 lands?

nvkelso:
Looking good!

<img width="1355" alt="screen shot 2016-04-30 at 09 41 04" src="https://cloud.githubusercontent.com/assets/853051/14934646/b8e18af0-0eb7-11e6-904d-cc6ea83b4ab4.png">

<img width="1071" alt="screen shot 2016-04-30 at 09 40 34" src="https://cloud.githubusercontent.com/assets/853051/14934645/b896817c-0eb7-11e6-963a-d3be37bc70ca.png">

PR #720, opened 2016-04-15
add waterfalls

nvkelso:
Connects with #677.

* add waterfall pois, with zoom variation per height / area
* [docs] add waterfall to list of poi kinds
* [test] add waterfall tests
* [migration] add waterfall migration for point and polygon


yaml/pois.yaml
```diff
@@ -172,6 +172,14 @@ filters:
     min_zoom: 17
     output:
       kind: {col: tags->zoo}
+  - filter: {waterway: waterfall}
+    min_zoom: CASE WHEN to_number(tags->'height', '9999.99') < 10 THEN 15
```
zerebubuth: The `height` is a regular column, and isn't (on my DB at least) in `tags`.

Also, using `to_number` like this ignores the case where height has a unit. We handle units in [the Python transform](https://github.com/mapzen/TileStache/blob/integration-1/TileStache/Goodies/VecTiles/transform.py#L43-L87), but that's not available at this point in the process. According to my local DB, about 1/3rd of the cases where `height` is present it's in feet. We have 3 options:
1. Ignore units (current behaviour), which will make waterfalls in feet appear more important than they really are.
2. Select all waterfalls at zoom 12, and cull the smaller ones in a Python transform. Waterfalls are pretty rare features, so I don't expect this would measurably affect performance.
3. Write a SQL function to get the height in meters, effectively porting the Python one.

Which of those options is best?

yaml/pois.yaml
```diff
@@ -172,6 +172,14 @@ filters:
...
+      WHEN to_number(tags->'height', '9999.99') >= 10 and to_number(tags->'height', '9999.99') < 50 THEN 14
```
zerebubuth: The `CASE WHEN ... WHEN ...` acts like an `else if`, so there's no need to make the cases mutually exclusive - this would be more readable if it were just `CASE WHEN <10 THEN ... WHEN <50 THEN ...` etc...

yaml/pois.yaml
```diff
@@ -172,6 +172,14 @@ filters:
     min_zoom: 17
     output:
       kind: {col: tags->zoo}
+  - filter: {waterway: waterfall}
```
zerebubuth: There's a smaller number, but still significant ([1,578 according to taginfo](http://taginfo.openstreetmap.org/tags/natural=waterfall#overview)), of `natural=waterfall`. Are those worth including too?

nvkelso: Considering Nevada Falls, one of the most important waterfalls in Yosemite NP is tagged `natural=waterfall`, we should include them. Good catch!

nvkelso: Haha, that one is a duplicate. But none the less.

nvkelso: Resolved in aba2e1657648146e3bd949490ec17ba9ba01adfd.

yaml/pois.yaml
nvkelso: Fixed in 90cce5c7ba12d96c2ef67bdd6b91f34662907223.

yaml/pois.yaml
nvkelso: Hmm, `height` is only in `tags` on my local machine?

Agree simple `to_number` SQL isn't the best to capture variable null, meter, and feet `height` units.  We'll have the similar problem for peak heights?

Should there be a post processing step that converts the waterfall height to implied meters, and does a cull? The downside is we'd have to maintain the height logic cutoffs in two places, so only select all waterfalls starting at zoom 12 in the YAML here, and move the zoom logic to the new transform, after the value is cast into meters?

zerebubuth: Re: `height` - when you imported your database, did you use the stock osm2pgsql `default.style` or did you use [the version in vector-datasource](https://github.com/mapzen/vector-datasource/blob/master/osm2pgsql.style#L98) which includes `height` as part of the table?

Re: logic in two places - I agree, that's why I think the best option is to have the logic in the YAML file only. Otherwise we'll update that logic at some point in the future and wonder why it's having no effect and have to go chasing through the other layers of the stack to debug it.

zerebubuth: I had to add the stuff in https://github.com/mapzen/vector-datasource/commit/6924d7a4236ecd347d87a02ed4ee2f0ea376eaed to get the tests to pass. The `extra_columns` thing is just so that the processor knows we're using the `height` column in the query, as it doesn't actually parse the query.

I didn't want to merge it in here directly, but the diff is a bit of a mess because I had to merge master to get a working install. Let me know what you think.

nvkelso: +1 for porting the python logic to SQL so everything is more legible (both for waterfalls and peaks).

nvkelso: I've incorporated your stuff in https://github.com/mapzen/vector-datasource/commit/6924d7a4236ecd347d87a02ed4ee2f0ea376eaed into this branch via 366a79c4ab316807ebfae1cc9a68eaf9810bf495 and 5d55d4e525fcd45db4b1df884d425c65ce1027f9. It's also rebased against master now.

nvkelso:
To recap the above inline comments:

* Port current python logic to SQL to convert height values to meters.
* Refactor the waterfall height case statement to take advantage of this new value.
* We should also export the height for labeling on the map.

nvkelso:
Exporting the height in efb5d16aed2aaa20906fbf8ee5f61bfe522c9dfa.

nvkelso:
Over to @zerebubuth for the Python to SQL conversion.

zerebubuth:
LGTM

nvkelso:
Woot

Issue #668, opened 2016-04-04
Add intermittent=* to water layer


nvkelso:
Add intermittent=* to water layer query, around 1,000,000:

* http://wiki.openstreetmap.org/wiki/Key:intermittent

nvkelso:
Working :)

<img width="963" alt="screen shot 2016-04-30 at 07 47 43" src="https://cloud.githubusercontent.com/assets/853051/14934145/dcfc6140-0ea7-11e6-9265-04e3b94f8051.png">

PR #721, opened 2016-04-15
add intermittent water flag

nvkelso:
Connects to: #668.

* add intermittent to water selections
* [doc] add intermittent note to water layer
* [test] add tests for intermittent water

Don't believe a migration is necessary.


zerebubuth:
Would it be possible to use the YAML for this?

nvkelso:
Yes and no. The intermittent value could be present on almost any water features type, so seemed easier to include it at the query level instead.

If we were to do YAML setup, something like this on every one?

```
    output:
      {kind: stream}
      intermittent: {col: intermittent}
```

zerebubuth:
Where there's a set of common features, I've been using the `globals` section to make [an anchor](https://github.com/mapzen/vector-datasource/blob/master/yaml/buildings.yaml#L3-L17) which then gets pulled into [each relevant output](https://github.com/mapzen/vector-datasource/blob/master/yaml/buildings.yaml#L48). This is just one output, so that may or may not be worth it.

The curly braces around the kind aren't needed:

```yaml
    output:
      kind: stream
      intermittent: {col: tags->intermittent}
```

I think intermittent is not a full column, so likely to find it in the tags.

nvkelso:
Nice, thanks for the `globals` tip! I'll rewrite and update the PR.

nvkelso:
@zerebubuth I've updated to the YAML syntax, but I get a few errors :(

I've pushed my latest to this branch. Note you recommended `intermittent: {col: tags->intermittent}` but in my DB it's a top-level attribute so I've done `intermittent: {col: intermittent}` instead. 

1) When updating the functions, using `$ python create-sql-functions.py | psql osm`, I get errors:

```
CREATE FUNCTION
ERROR:  parameter name "intermittent" used more than once
ERROR:  parameter name "intermittent" used more than once
CREATE FUNCTION
```

2) Similarly, tiles now 500 for the tests:

```
127.0.0.1 - - [19/Apr/2016 00:49:11] "GET /water/16/12368/26272.json HTTP/1.1" 500 -
  File "/Users/nvkelso/git-repos/tileserver/tileserver/__init__.py", line 188, in __call__
    response = self.handle_request(request)
  File "/Users/nvkelso/git-repos/tileserver/tileserver/__init__.py", line 275, in handle_request
    feature_data_all = self.data_fetcher(coord, self.layer_config.all_layers)
  File "/Library/Python/2.7/site-packages/tilequeue/query.py", line 195, in __call__
    raise async_exception
ProgrammingError: function mz_calculate_output_water_(unknown, unknown, text, text, text, text, text, text, text, text, unknown, hstore) does not exist
LINE 1: SELECT mz_calculate_output_water_(
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:  SELECT mz_calculate_output_water_(
        NULL,
        NULL,
        row."amenity",
        row."intermittent",
        row."landuse",
        row."leisure",
        row."name",
        row."natural",
        row."place",
        row."waterway",
        NULL,
        row.tags)
CONTEXT:  PL/pgSQL function mz_calculate_output_water(planet_osm_polygon) line 5 at RETURN
```

zerebubuth:
The issue was that you were including the `intermittent` column in `water_standard_properties_osm` in filters which had `table: ne` set. Since features come from only one table, it's not possible to mix properties from one table with another. And, at the moment, the code isn't sophisticated enough to be able to detect that and issue a more appropriate warning.

I fixed that for you in 5e6947b and also merged master, which resolved issues for me with master having some changes to TileStache which weren't reflected in this branch.

nvkelso:
Verified all tests pass, merging. Thanks for the corrections :)

PR #723, opened 2016-04-15
add downloaded shapefiles to gitignore

batje:
the idea is to just download and unpack zipped shapefiles. not to commit them.


zerebubuth:
That's great, thanks!

PR #725, opened 2016-04-15
Convert roads kind and min zoom functions to YAML.

zerebubuth:
Connects to #646.


data/migrations/create-sql-functions.py
```diff
@@ -16,6 +16,8 @@ def format_value(val):
         elif 'col' in val:
             if val['col'].startswith('tags->'):
                 return "tags->'%s'" % val['col'][len('tags->'):]
+            elif val['col'].startswith('$'):
+                return '"%s"' % val['col'][len('$'):]
```
rmarianski: Do we need to do this because of a yaml parsing issue?

data/migrations/create-sql-functions.py
```diff
@@ -64,9 +66,18 @@ def __init__(self, column, value):
...
+                format_value(self.value))
+
+        # tags->'foo' is NULL when the hstore doesn't contain that key, which
+        # makes all the following comparisons NULL, so combine the equals
+        # with an exists check.
+        if self.column.startswith('tags->'):
+            exists_check = "tags ? '%s'" % (self.column[len('tags->'):],)
+            return '(%s) AND (%s)' % (exists_check, equals_check)
+        else:
+            return equals_check
```
rmarianski: Nice catch! :zap:

data/migrations/create-sql-functions.py
```diff
@@ -126,7 +137,10 @@ def __init__(self, column):
         self.column = column
 
     def as_sql(self):
-        return '%s IS NOT NULL' % format_column(self.column)
+        if self.column.startswith('tags->'):
+            return "tags ? '%s'" % (self.column[len('tags->'):],)
+        else:
+            return '%s IS NOT NULL' % format_column(self.column)
```
rmarianski: :zap:

rmarianski:
:+1: Great work on the conversion to yaml :zap:

data/migrations/create-sql-functions.py
zerebubuth: Ah, no. I'm using the `$column` syntax to denote a column which is suppressed from the input, but thinking about it now, that's unnecessary. I've added something in 4857b8c which hopefully makes the usage clearer and removes some unnecessary code branches.

rmarianski: Ah ok, makes sense. :+1:

rmarianski:
Latest changes LGTM. :+1:

PR #728, opened 2016-04-15
fixes #727 for dive_centre order w/r/t scuba_diving

### LINKED ISSUE #727: Priority of diver centre vs. scuba diving shop
The [basic outdoor pois test](https://github.com/mapzen/vector-datasource/blob/master/test/662-basic-outdoor-pois.py#L11-L14) fails because [node 2910259124](http://www.openstreetmap.org/node/2910259124), which is tagged both `amenity=dive_centre` and `shop=scuba_diving`, is expected to be a dive centre, but the filter for [dive center](https://github.com/mapzen/vector-datasource/blob/master/yaml/pois.yaml#L460-L462) comes later than [scuba shop](https://github.com/mapzen/vector-datasource/blob/master/yaml/pois.yaml#L75-L77) and so the scuba shop matches first.

Either the test should be fixed to be `kind: scuba_diving` (and presumably another test for `dive_centre` found), or the priorities in `pois.yaml` need to be altered.

nvkelso:


zerebubuth:
:+1:

rmarianski:
LGTM too. :+1:

PR #733, opened 2016-04-20
Remove semicolon in apply-planet_osm_line.sql

mr:
It causes a syntax error when running perform-sql-updates.sh


zerebubuth:
Ooops! Well spotted, thanks!

PR #734, opened 2016-04-20
Heights in meters

zerebubuth:
Added a function in SQL to convert heights to meters, and use this in the test for waterfall zoom level.

Requires mapzen/TileStache#145 for a second conversion of the `height` property into meters for output in tiles.

Added tests for `height`.

Connects to #677.

### DIFF
--- a/test/677-waterfall.py
+++ b/test/677-waterfall.py
@@ -2,7 +2,7 @@
 # Upper Yosemite Falls, because it's so tall at 550 meters, more than 300 meters
 assert_has_feature(
     12, 687, 1583, 'pois',
-    { 'kind': 'waterfall', 'min_zoom': 12 })
+    { 'kind': 'waterfall', 'min_zoom': 12, 'height': 550 })
 
 #https://www.openstreetmap.org/node/2384221575
 # Middle Yosemite Falls, 206 meters, more than 50 meters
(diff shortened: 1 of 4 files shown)


test/677-waterfall.py
```diff
@@ -52,4 +52,22 @@
...
+# Rainbow Falls - height 150ft = 45m
+assert_has_feature(
+    14, 4416, 6484, 'pois',
+    { 'kind': 'waterfall', 'min_zoom': 14, 'height': 45.72 })
+
+# https://www.openstreetmap.org/node/404574988
+# Toccoa Falls - height 186' = 56.6929m
+assert_has_feature(
+    13, 2199, 3256, 'pois',
+    { 'kind': 'waterfall', 'min_zoom': 13, 'height': 56.6928 })
```
nvkelso: Should we limit heights to ints in the tiles? Or only tenths place? 

In this case it's a foot value that was an int, carrying the original precision thru I'd only expect to see 56 meters.

zerebubuth: I'm not sure about carrying the original precision through, that might be complicated. But it's easy enough to limit them to integers or one decimal place. What's the reason for doing that - space in the tile, readability?

nvkelso: These values (for peaks, waterfalls) will be used to label the height on the map, e.g.:

```
+ Mountain name
  512m
```

```
+ Waterfall name
  25m
```

zerebubuth: Wouldn't you want the ability to label in either meters or feet, depending on the user preference? In which case, the client will need to convert and format as integer anyway? (i.e: `text_source: function() { '' + int(height) + 'm' }`?)

nvkelso: Fair enough.

Issue #592, opened 2016-03-12
Add bicycle map POIs


nvkelso:
**Bicycle POIs:**

- [x] **`shop=bicycle`**
- [x] **min zoom = 15** for `shop=bicycle`
- [x]  **`amenity=bicycle_rental`**
- [x] **min zoom = 16** for bicycle_rental
- [x] add [more properties](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_rental) for bicycle_rental: `network, operator, ref, capacity`
- [x] **`amenity=bicycle_parking`** at zoom 17
- [x] add [more properties](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_parking) for bicycle_parking: `capacity, covered, access, fee, operator, cyclestreets_id, maxstay, surveillance`
- [x] [rcn_ref nodes](https://wiki.openstreetmap.org/wiki/Key:rcn_ref) - **min_zoom = 16** (was 15)
- [x] [cycle_barrier](http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dcycle_barrier) at zoom 17

nvkelso:
We can split off the extra properties (later) from the zoom range changes (now).

nvkelso:
Adjustments to the above recommendations after reviewing the data more:

* If a bicycle_rental is a shop, min zoom = 16
* If a bicycle_rental is bike share station, min zoom = 17
* cycle_barrier at zoom 18
* several new attributes for bicycle_rental were added, but not all. Ref was already present: 
  * network:  {col: tags->network}
  * operator: {col: operator}
  * capacity: {col: tags->capacity}
* several new attributes for bicycle_parking were added, but not all:
  * operator: {col: operator}
  * capacity: {col: tags->capacity}
  * covered: {col: tags->covered}

nvkelso:
@zerebubuth I'm okay calling the `bicycle_guidepost` and `walking_guidepost` something else, more like `bicycle_junction` or `bicycle_network_node`. Use your best judgement.

nvkelso:
Looking good.

PR #736, opened 2016-04-21
add adjust bicycle pois (#592)

nvkelso:
Connects with #592 to add new and adjust existing bicycle related POIs.

* Update POI filters (add, adjust)
* Update tests (adjustments didn't affect any existing tests, add new test coverage for those)
* Update docs (started adding definitions, too)
* Updated migration

WARNING: Two of the new kinds are not present in North America, only Europe for `rcn_ref` and `rwn_ref`.


yaml/pois.yaml
```diff
@@ -355,8 +355,13 @@ filters:
     output:
       kind: {col: tourism}
   - filter:
-      amenity: [atm, bicycle_rental, bicycle_parking, bus_stop, drinking_water, emergency_phone,
-        parking, post_box, telephone, theatre, toilets]
+      amenity: [theatre]
```
nvkelso: Split theatre off to it's own filter as it should really be visible earlier than these "infrastructure" related things, and considered separately from.

yaml/pois.yaml
```diff
@@ -389,7 +394,52 @@ filters:
     min_zoom: 17
     output: {kind: office}
   - filter:
-      shop: [bakery, bicycle, books, butcher, car, car_repair, clothes, computer, convenience,
+      shop: [bicycle]
```
nvkelso: Many bicycle shops sometimes also rent bicycles, that's buried in some service tags. That's beyond the scope of this issue.

nvkelso:
I think this is unrelated TileStache config funk, but one of my tests does fail:

```
127.0.0.1 - - [20/Apr/2016 21:10:09] "GET /pois/16/19298/24633.json HTTP/1.1" 500 -
  File "/Users/nvkelso/git-repos/tileserver/tileserver/__init__.py", line 190, in __call__
    response = self.handle_request(request)
  File "/Users/nvkelso/git-repos/tileserver/tileserver/__init__.py", line 286, in handle_request
    [], [])
  File "/Library/Python/2.7/site-packages/tilequeue-0.8.0-py2.7.egg/tilequeue/process.py", line 399, in process_coord
    padded_bounds, scale, layers_to_format)
  File "/Library/Python/2.7/site-packages/tilequeue-0.8.0-py2.7.egg/tilequeue/process.py", line 318, in _process_feature_layers
    shape, props, feature_id, coord.zoom)
  File "/Library/Python/2.7/site-packages/TileStache/Goodies/VecTiles/server.py", line 45, in transform_fn
    shape, properties, fid = fn(shape, properties, fid, zoom)
  File "/Library/Python/2.7/site-packages/TileStache/Goodies/VecTiles/transform.py", line 3119, in add_state_to_stations
    if kind not in ('station'):
TypeError: 'in <string>' requires string as left operand, not NoneType
```

zerebubuth:
You've only included signposts for regional cycle and walking networks - did you want to add them for international, national and local networks too?

nvkelso:
Good catch, will do.

nvkelso:
* `icn_ref` is not-common  / none-existant. `ncn_ref` is common, but only on ways (not nodes). `lcn_ref` is common (~1,000 nodes), versus over ~15,000 for `rcn_ref` nodes.
* `iwn_ref`, `nwn_ref` are extremely uncommon / none-existant. `lwn_ref` is somewhat common (~200 nodes) in TagInfo primarily but with letters not numbers in the ref value, versus over ~15,000 for `rwn_ref` nodes.

zerebubuth:
`network=icn` on routes is also not-common, non-existent (286 uses / 159 for `icn_ref`) but we're still going to include those, aren't we?

I was talking to @gravitystorm about these, and it seems like `*cn_ref` on nodes indicates a numbered waypoint rather than a guidepost. In other words, it means "this is reference point N" rather than "this way to N" or "this is route N". The use of these seems to be to describe a route as a series of locations, e.g: "go to 26, then 88 then 12", which would seem to require special (or at least different) cartographic consideration than a numbered route.

Apologies if this was already what you intended, but it wasn't clear to me. For me, a guidepost is something with directional signage indicating a destination or numbered route. Perhaps this needs more documentation and/or a change to something clearer, e.g: `bicycle_waypoint`?

nvkelso:
Yep, I'm including all the 4 bicycle guidepost types and 4 walking guidepost types are included in the PR in the [pois.yaml#L441-L496](https://github.com/mapzen/vector-datasource/blob/nvkelso/592-add-adjust-bicycle-pois/yaml/pois.yaml#L441-L496) filters. 

My proposal for calling these `kind:bicycle_guidepost` and `kind:walking_guidepost` stems from many of these features are also tagged `tourism=information` and [`information=guidepost`](http://wiki.openstreetmap.org/wiki/Tag:information%3Dguidepost), and `guidepost` seems to be the OSM vernacular (versus `signpost`). Even if they "guide" to other nodes in the network only (versus named places), like the photos in [this OSM wiki page](http://wiki.openstreetmap.org/wiki/Fietsroutes) and [2](http://wiki.openstreetmap.org/wiki/Key:rcn_ref).

Agree they could use more definition, added that to the docs in 66e0b0727a8406490e2bd0107cea929822164fbb.

nvkelso:
@zerebubuth I think this is good for merging now. Can you please give it a final code review, please?

nvkelso:
Depends on https://github.com/mapzen/TileStache/pull/146.

nvkelso:
Merging to keep things moving while I'm still online. If there are problems, let's sort them out with another PR.

Issue #703, opened 2016-04-12
Move place=continent to the earth layer


nvkelso:
We already proposed to move sea and ocean to the water layer, we should additionally move continent to the earth layer as a label placement.

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/148

nvkelso:
Easy enough to do this setting up the new YAML for earth layer, pulling back to v0.10

nvkelso:
Working

Issue #399, opened 2015-11-09
Add island labels


nvkelso:
**earth layer** label positions:

* place=**island** Identifies the coastline of an island (> 1 km2), also consider place=islet for very small islands. zoom 14 default, area adjusted up to zoom 9
* place=**islet** Identifies an islet (< 1 km2). zoom 16 default, area adjusted up to zoom 15.
* place=**archipelago** {Relation} Identifies the relation of an archipelago, which contains several islands and islets. zoom 14 default, area adjusted up to zoom 9

===

These islands of southern California have names, but we don't include them yet:

z10/33.9060/-119.9944

![screen shot 2015-11-09 at 10 39 54](https://cloud.githubusercontent.com/assets/853051/11042840/507bcb5e-86ce-11e5-99cc-c187fb41f781.png)

Looks like the related tag is:

http://wiki.openstreetmap.org/wiki/Tag:place=island

But sometimes it's hiding on relation parts, fun.

nvkelso:
Looks like these would also be useful:

* place=**archipelago**	Relation	Identifies the relation of an archipelago, which contains several islands and islets.		
* place=**island**	Area	Identifies the coastline of an island (> 1 km2), also consider place=islet for very small islands		
* place=**islet**	Node Area	Identifies an islet (< 1 km2).

nvkelso:
Target layer: `physical-labels` or `earth` or `places`, tbd.

nvkelso:
Load balancing to v0.8

rmarianski:
Is that what's happening in this case? Seems like it's coming through in the tile in the `boundaries` layer. This looks to be the relevant [osm feature](http://www.openstreetmap.org/relation/3636334).

Can certainly add the `place`s you mentioned, just don't think it'll help with this example, unless I'm missing something.

rmarianski:
Err sorry, I see it in the `landuse` layer. An example tile is: [10/171/409](http://vector.mapzen.com/osm/all/10/171/409.topojson).

nvkelso:
Other candidates, from http://taginfo.openstreetmap.org/tags/natural=strait:

* natural=peninsula
* natural=headland
* natural=cape
* natural=isthmus

nvkelso:
That feature is the national park, this is the (generic) island, tagged `place=island`:

* http://www.openstreetmap.org/relation/237600#map=12/34.0184/-119.7252

[...]

PR #737, opened 2016-04-21
add island labels (#399, #646, #703)

nvkelso:
Connects with #399 to add island, islet & other labels. Also refactors how the earth layer is constructed.

* [query] add earth queary JINJA
* [query] hook up schema prefunction for new earth layer
* [query] hook up create sql functions and sql.jinja to new earth layer
* [filter] remove continent place labels from places layer
* [filter] init earth YAML filter, add continent place label to earth layer
* [filter] adjust island, islet zoom ranges
* [apply] hook up data/apply functions, set new fields for new earth layer
* [tests] add full set of island label tests
* [migration] add custom migration for new earth layer features
* [docs] add new kinds to earth layer docs

### DIFF
--- a/data/apply-updates-non-planet-tables.sql
+++ b/data/apply-updates-non-planet-tables.sql
@@ -1,14 +1,21 @@
 DO $$
 BEGIN
 
+UPDATE ne_10m_urban_areas SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
+CREATE INDEX ne_10m_urban_areas_way_area_index ON ne_10m_urban_areas(way_area);
+
+UPDATE ne_50m_urban_areas SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
+CREATE INDEX ne_50m_urban_areas_way_area_index ON ne_50m_urban_areas(way_area);
+
+UPDATE ne_10m_parks_and_protected_lands SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
+CREATE INDEX ne_10m_parks_and_protected_lands_way_area_index ON ne_10m_parks_and_protected_lands(way_area);
+
+
 UPDATE ne_110m_ocean SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
 CREATE INDEX ne_110m_ocean_wayarea_index ON ne_110m_ocean(way_area);
 
 UPDATE ne_110m_lakes SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
 CREATE INDEX ne_110m_lakes_wayarea_index ON ne_110m_lakes(way_area);
-
-UPDATE ne_110m_land SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
-CREATE INDEX ne_110m_land_wayarea_index ON ne_110m_land(way_area);
 
 UPDATE ne_50m_ocean SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
 CREATE INDEX ne_50m_ocean_wayarea_index ON ne_50m_ocean(way_area);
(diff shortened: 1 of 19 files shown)


nvkelso:
@zerebubuth This is still a WIP, but it's pretty close to done. Please have a look around, particularly at migration and query parts. 

I've verified the zooms are working well in the DB for the test cases, but still need to fully run the test via python.

nvkelso:
Also connects to #703 to move the **continent** label placement from `places` layer to `earth` layer.

nvkelso:
Also connects to #646 to move the last layer, `earth`, to the new YAML setup.

queries/earth.jinja2
```diff
@@ -1,78 +1,99 @@
...
-    NULL AS label_placement,
+    '' AS name,
+    sum(way_area)::bigint AS area,
+    mz_calculate_output_earth(t.*) AS mz_properties,
     'naturalearthdata.com' AS source,
     {% filter geometry %}st_union(the_geom){% endfilter %} AS __geometry__,
-    max(gid) AS __id__
+    max(gid) AS __id__,
+    NULL AS label_placement,
+    NULL AS boundary
```
zerebubuth: I think we prefer to do the addition of extra parameters (`label_placement`, `boundary`) via YAML now, to avoid having to do `NULL AS ...` in the other queries.

queries/earth.jinja2
```diff
@@ -1,78 +1,99 @@
...
+    ne_10m_land t
 {% endif %}
 
 WHERE
+    mz_earth_min_zoom <= {{ zoom }} AND
     {{ bounds|bbox_filter('the_geom') }}
+GROUP BY mz_properties
 
-{% else %}
+{% elif 9 <= zoom %}
```
zerebubuth: Did you mean `zoom >= 9`? If so, would it be easier to use `else`? Apologies if not - I'm having trouble following the `if - elsif - endif` stuff through the diff!

yaml/earth.yaml
```diff
@@ -0,0 +1,79 @@
+# synthetic columns are values available to filters and output, but which are
+# generated in the SQL function and not passed in from the tables. in this case,
+# we generate the volume from the building height and way_area, where available.
+synthetic_columns:
+  - test
```
zerebubuth: Does this need to be here?

data/apply-updates-non-planet-tables.sql
```diff
@@ -25,19 +35,26 @@ CREATE INDEX ne_10m_lakes_wayarea_index ON ne_10m_lakes(way_area);
...
+UPDATE ne_50m_land SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
+CREATE INDEX ne_50m_land_wayarea_index ON ne_50m_land(way_area);
+
+UPDATE ne_10m_land SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
+CREATE INDEX ne_10m_land_way_area_index ON ne_10m_land(way_area);
+
+
+UPDATE land_polygons SET way_area=ST_Area(the_geom) WHERE the_geom IS NOT NULL;
+CREATE INDEX land_polygons_wayarea_index ON land_polygons(way_area);
+
```
zerebubuth: Do these also need a `UPDATE table SET mz_earth_min_zoom = mz_calculate_min_zoom_earth(table.*)`?

queries/earth.jinja2
nvkelso: Agree, should just be an `else`.

queries/earth.jinja2
nvkelso: Good point, I'll revise.

(This is loosely based off the `water.jinja2` file, probably similar cleanup over there, but I'd rather not mess with that in this PR.)

yaml/earth.yaml
nvkelso: If I remove the `synthetic_columns` block here, I get `way_area` related errors :/

nvkelso:
I've got it all working (!?) except for these handful of tests which fail:

* Failing tests: https://github.com/mapzen/vector-datasource/pull/737/commits/98cc13da679d6e400346c1c33cd5452a894c7b0b

Similar test do pass, though, and the features exist in the DB and seem to be marked with the expected zoom values.

The `synthetic_columns: - test` bit should also be investigated & resolved. 

Because I'll be traveling the rest of the week, please pick up this PR and see it to completion. Thanks!

zerebubuth:
Tests pass for me. I think all the issues above should be fixed now. @rmarianski could you review this, please?

rmarianski:
:+1:

Issue #596, opened 2016-03-12
Add hiking routes (paths)


nvkelso:
Thinking more about an outdoor style, we should show selection "national" and regional trails like the [John Muir Trail](http://www.openstreetmap.org/relation/1244828), [Pacific Crest Trail](http://www.openstreetmap.org/relation/1225378), and Appalachian Trail earlier than zoom 13. This should probably be a new, optional layer, or the existing roads layer.

**See also:**

* http://wiki.openstreetmap.org/wiki/Hiking
* http://wiki.openstreetmap.org/wiki/Tag:network=nwn?uselang=en-US
* https://wiki.openstreetmap.org/wiki/Tag:network%3Drwn
* http://hiking.waymarkedtrails.org/en/#

**Route types, zooms:**

* **continental**: `network=iwn` at zoom **9**
* **national**: `network=nwn` at zoom **10**
* **regional**: `network=rwn` at zoom **11**
* **local**: `network=lwn`at zoom **12**

**Tags:**

* `name`
* `route` = `kind`
* `network`
* `operator`
* `ref`

**John Muir Trail:** _zoom 10_

_where the [Pacific Crest Trail](http://www.openstreetmap.org/relation/1225378) would be shown at zoom 9, then JMT at zoom 10._

<img width="861" alt="screen shot 2016-03-11 at 16 48 45" src="https://cloud.githubusercontent.com/assets/853051/13719352/29d5a42a-e7a9-11e5-801f-7c63a7f16185.png">

**Appalachian Trail:**  _outside of DC at zoom 10_

<img width="1158" alt="screen shot 2016-03-11 at 17 01 17" src="https://cloud.githubusercontent.com/assets/853051/13719466/fbfd01f4-e7aa-11e5-98df-6993b1118978.png">

**Near Geneva:** _zoom 10_

<img width="1064" alt="screen shot 2016-03-11 at 17 04 08" src="https://cloud.githubusercontent.com/assets/853051/13719495/55256c44-e7ab-11e5-826b-a02bab887d9c.png">

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/pull/681

nvkelso:
The early zoom part of this issue was completed (though in modified form).

No attributes from the relation were projected onto the way, though. EG, `name` was not backfilled, and `network` & `operator` are only present when already attached to the way directly.

nvkelso:
[...]

Issue #647, opened 2016-03-30
Add basic is_bicycle_route, bike lane left, right


nvkelso:
Picking up thread in https://github.com/mapzen/vector-datasource/issues/391 to limit to simple change:

* Boolean for `is_bicycle_route` based on the relations (like `is_bus_route` logic)
* Indicate "best" network value for bicycle route (see https://github.com/mapzen/vector-datasource/issues/647#issuecomment-214572904 below)
* Indicate if bicycle lane left (way property)
* Indicate bicycle lane right (way property)
* Indicate bicycle contraflow (way property)

nvkelso:
See line notes in https://github.com/mapzen/vector-datasource/pull/739 for indicating bicycle route network type (not just a boolean).

* https://github.com/mapzen/vector-datasource/pull/739#discussion_r61013189
* https://github.com/mapzen/vector-datasource/pull/739#discussion_r61013106
* https://github.com/mapzen/vector-datasource/pull/739#discussion_r61013242

nvkelso:
We've added basic attributes (good), but we're missing a calculated `is_bicycle_route: yes`. Do we need that still? We include `bicycle_network` with values in `icn`, `ncn`, `rcn`, and `lcn` which is probably more useful. But sometimes a road might not part of a route relation and miss bicycle_network but still have a cycleway on it. Here's the logic from the scene file we're using for this for debug:

```
    bicycle-roads:
        data: { source: mapzen, layer: roads }
        filter: 
            any:
                - bicycle_network: true
                - cycleway: yes
                - cycleway_left: yes
                - cycleway_right: yes
                - cycleway_both: yes
```

nvkelso:
More complete example:

```
    bicycle-roads:
        data: { source: mapzen, layer: roads }
        filter: 
            any:
                - bicycle_network: true
                - cycleway: true
                  not: { cycleway: no }
                - cycleway_left: true
                  not: { cycleway_left: no }
                - cycleway_right: true
                  not: { cycleway_right: no }
                - cycleway_both: true
                  not: { cycleway_both: no }
```

zerebubuth:
> But sometimes a road might not part of a route relation and miss bicycle_network but still have a cycleway on it.

If a way is not part of a `route=bicycle` relation (or has `?cn_ref`), then it's not a bicycle route - at least as far as we can tell from the data. We shouldn't be implying that any road with a cycle lane on it is a cycle _route_ - those are two separate concepts.

[...]

PR #739, opened 2016-04-21
Add hiking routes

zerebubuth:
This "lifts" more types of roads up to lower zooms when they're members of hiking networks. Membership of hiking networks is shown by a new `hiking_network` property, which takes the value of the "most important" hiking network that a road is a member of (if more than one).

Connects to #596.

NOTE: I ran this migration, but didn't add it to the code, since it was already covered in the existing migration. But it'll need to be run again, if you've already run a partial v0.10 migration.

```SQL
update planet_osm_line
set mz_road_level = mz_calculate_min_zoom_roads(planet_osm_line.*)
where highway in ('residential', 'unclassified', 'road', 'trunk_link', 'primary_link',
  'secondary_link', 'tertiary_link', 'living_street', 'service') OR
  ((tags ? 'whitewater') AND (tags->'whitewater' = 'portage_way'));
```


docs/layers.md
```diff
@@ -796,6 +796,7 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `aeroway`: See kind list below.
 * `ferry`: See kind list below.
 * `highway`: See kind list below.
+* `hiking_network`: Present if the road is part of a hiking network. If so, the value will be one of `iwn` for International Walking Network, `nwn` for National Walking Network, `rwn` for Regional Walking Network, `lwn` for Local Walking Network.
```
nvkelso: Should we say `walking_network` or `hiking_network`?

For cycle network elsewhere, we're saying `bicycle_*` as that's the more common OSM'ism (excepting `cycle_barrier`), so it's okay to have some variation. But flagging for discussion.

yaml/roads.yaml
```diff
@@ -202,25 +208,42 @@ filters:
...
+      <<: *osm_network_from_tags
+      kind: path
+      highway: {col: highway}
+  - filter:
+      highway: [footway, steps]
+      any:
+        - {name: true}
+        - {bicycle: designated}
+        - {foot: designated}
+        - {horse: designated}
```
nvkelso: nice!

yaml/roads.yaml
```diff
@@ -202,25 +208,42 @@ filters:
...
       <<: *osm_network_from_tags
       kind: path
       highway: {col: highway}
   - filter:
-      highway: [path, footway, steps]
-    min_zoom: |
-      COALESCE(mz_calculate_path_major_route(osm_id),
-        CASE
-          WHEN highway = 'path' THEN 13
-          WHEN mz_is_path_named_or_designated(highway, name, bicycle, foot,
```
nvkelso: This logic needs to be carried over.

nvkelso: I made some further changes from this branch in https://github.com/mapzen/vector-datasource/commits/nvkelso/early-path-revisions a couple days ago. I'll take on merging those concepts & updated tests into this PR. Thanks!

docs/layers.md
zerebubuth: I went back and forth on this. On the one hand, it's `route=hiking`, on the other it's `*wn` for Walking Network. Whatever you think is more consistent and easy to understand - it's only a property name.

nvkelso: We're using `walking_network` elsewhere. We should here, too, for consistency.

nvkelso: Resolved in beebc6048aa29a2057e37f7e70bffbfb45873b03.

yaml/roads.yaml
nvkelso: Done, though cycleway got tangled up in it.

data/functions.sql
```diff
@@ -556,10 +556,33 @@ RETURNS BOOLEAN AS $$
 BEGIN
     RETURN (
       tags->'type' = 'route' AND
-      tags->'route' IN ('hiking', 'foot') AND
-      tags->'network' IN ('iwn','nwn','rwn','lwn')
+      tags->'route' IN ('hiking', 'foot', 'bicycle') AND
```
nvkelso: I included cycling networks here (see related https://github.com/mapzen/vector-datasource/issues/647) because their zoom ranges needed adjusting for the paths to have the right balance. 

I don't think any additional zoom work is need for those with respect to #647:

* `CREATE OR REPLACE FUNCTION mz_is_path_major_route_relation(tags hstore)`
* `CREATE OR REPLACE FUNCTION mz_calculate_path_major_route(osm_id BIGINT)`

yaml/roads.yaml
```diff
@@ -31,6 +31,7 @@ global:
       leisure: {col: leisure}
       sport: {col: sport}
       man_made: {col: man_made}
+      walking_network: {expr: "mz_hiking_network(osm_id)"}
```
nvkelso: For #647, we'd add something like,

* `bicycle_network: {expr: "mz_cyling_network(osm_id)"}`

data/functions.sql
```diff
@@ -556,10 +556,33 @@ RETURNS BOOLEAN AS $$
...
+      tags->'network' IN ('iwn','nwn','rwn','lwn','icn','ncn','rcn','lcn')
     );
 END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+-- Looks up whether the given osm_id is a member of any hiking routes
+-- and, if so, returns the network designation of the most important
+-- (highest in hierarchy) of the networks.
+--
+CREATE OR REPLACE FUNCTION mz_hiking_network(osm_id bigint)
```
nvkelso: For cycling networks (see related #647), we'd create a cycling copy of:

* `CREATE OR REPLACE FUNCTION mz_hiking_network(osm_id bigint)` 
* like `mz_cyling_network`

nvkelso:
@zerebubuth I merged in changes from my path polish branch in https://github.com/mapzen/vector-datasource/pull/749. Please give it a once over, making sure tests still work & etc.

Specifically:

* add track & cycleway networks to the mix & update tests.
* stop stripping path names at low zooms (they are like highway names), works in USA, not sure performance in Europe.
* suppress zoom range of uninteresting footways of type sidewalk or crossing.
* export footway value (e.g., sidewalk, crossing) for client-side styling.
* add extra path properties (e.g.: `extra_columns: [bicycle, foot, horse, tracktype, "tags->incline", "tags->trail_visibility", "tags->sac_scale"]`)

zerebubuth:
@rmarianski I think I've got everything in this PR now. Could you review it, please?

rmarianski:
:+1:

test/593-early-path.py
```diff
@@ -8,4 +8,4 @@
 for z, x, y in tiles:
     assert_has_feature(
         z, x, y, 'roads',
-        {'highway': 'path', 'name': true})
```
nvkelso: We need to ensure that these feature have names (not that they are None).

test/593-early-path.py
```diff
@@ -8,4 +8,4 @@
...
\ No newline at end of file
+        {'highway': 'path', 'name': None})
```
nvkelso: We need to ensure that these feature have names (not that they are None).

zerebubuth: Yeah, that's slightly confusing, but having `None` in the properties means that [the key must be present](https://github.com/mapzen/vector-datasource/blob/master/test.py#L52). Having `true` here was an "unknown variable" error (Python calls it `True` with a capital T), but `True` would have checked for a literal (JSON) `true`, not for any value.

nvkelso: 👍

PR #740, opened 2016-04-22
Fix failing tests

zerebubuth:
I've fixed the YAML where possible, or the tests where not, and they all pass for me now. Other sources of failure might include:

1. Running the migration (or not having run it). Make sure you run `psql -f data/functions.sql $YOUR_DB_NAME` and `python data/migrations/create-sql-functions.py | psql $YOUR_DB_NAME`. And then update the failing feature by running `update planet_osm_XXX set mz_YYY_min_zoom = NULL where ZZZ;` for appropriate values of XXX, YYY and ZZZ. Note that setting the min zoom to `NULL` is actually okay, as the trigger will reset it back to the correct value.
2. Lack of data. I've put together [a little script](https://gist.github.com/zerebubuth/3a3e504ebf1973cdf82eb2f05a696682) which helps import data into your local DB. When I got the latest version of all the failing cases, everything passed.

@nvkelso could you take a look and check the tests are what you wanted, please? Particularly Antietam, which is tagged as a national park as well as a battlefield - to get the tests to pass, I had to move the battlefield first in priority, which I'm not sure is what was intended.

NOTE: The migration already covers this, but a shorter version which just works for this PR is:

```SQL
update planet_osm_point set mz_poi_min_zoom = NULL where tourism='picnic_site';
update planet_osm_polygon set mz_poi_min_zoom = NULL where tourism='picnic_site';
update planet_osm_polygon set mz_landuse_min_zoom = NULL where historic='battlefield';
update planet_osm_point set mz_poi_min_zoom = NULL where historic='battlefield';
update planet_osm_line set mz_boundary_min_zoom = NULL where waterway='dam';
```

### DIFF
--- a/test/343-winter-sports-resorts.py
+++ b/test/343-winter-sports-resorts.py
@@ -1,4 +1,4 @@
 assert_has_feature(
     15, 5467, 12531, 'landuse',
     { 'kind': 'winter_sports',
-      'sort_key': 33 })
+      'sort_key': 36 })
--- a/test/663-combo-outdoor-landuse-pois.py
+++ b/test/663-combo-outdoor-landuse-pois.py
@@ -64,13 +64,12 @@ assert_has_feature(
 assert_has_feature(
     15, 9219, 11714, 'pois',
     { 'kind': 'summer_camp', 'min_zoom': 15 })
-    
+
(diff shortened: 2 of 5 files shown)


nvkelso:
Thank you!

PR #749, opened 2016-04-25
rescue early path revisions into primary PR

nvkelso:
- [ ] add track & cycleway networks to the mix & update tests
- [ ] stop stripping path names at low zooms (they are like highway names)
- [ ] suppress zoom range of uninteresting footways of type sidewalk or crossing
- [ ] export footway value
- [ ] add extra path properties

### DIFF
--- a/yaml/roads.yaml
+++ b/yaml/roads.yaml
@@ -236,6 +236,7 @@ filters:
       <<: *osm_network_from_tags
       kind: path
       highway: {col: highway}
+      footway: {col: "tags->footway"}
   - filter:
       highway: [steps]
     min_zoom: |
--- a/yaml/roads.yaml
+++ b/yaml/roads.yaml
@@ -201,7 +201,7 @@ filters:
       kind: minor_road
       highway: {col: highway}
   - filter:
-      highway: [pedestrian, track, cycleway]
+      highway: [pedestrian]
     min_zoom: 13
     table: osm
     output:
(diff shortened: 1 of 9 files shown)


yaml/roads.yaml
```diff
@@ -235,14 +236,26 @@ filters:
         - {tags->ski: designated}
     min_zoom: 'LEAST(mz_calculate_path_major_route(osm_id), 13)'
     table: osm
+    extra_columns: [bicycle, foot, horse, tracktype, "tags->incline", "tags->trail_visibility", "tags->sac_scale"]
```
rmarianski: Were these extra_columns values required?

zerebubuth: Hmm... looks like a mix-up. Perhaps they were supposed to be in the `output`?

nvkelso: Yeah, those should have been in output. Was this corrected later?

(What is `extra_columns` for?)

zerebubuth: Yup, they were moved to `output` in a323875e967b768c38c9a518034e729c1bb787df.

`extra_columns` is for columns which need to be available because they're used in an expression, but not used in output or any filter. This because the script which generates the SQL can't "see" inside the SQL expressions, and therefore has no way of knowing that a column is used there. If the column isn't used anywhere else, then you'll see errors saying that the variable isn't defined.

There _was_ only one place where it was needed [in `pois.yaml`](https://github.com/mapzen/vector-datasource/blob/master/yaml/pois.yaml#L194-L200), but it seems we're outputting `height` now, so the `extra_columns` statement is superfluous now.

nvkelso: 👍

PR #750, opened 2016-04-26
Fix caravan site tests

zerebubuth:
Remove earlier duplicate of caravan site which was causing zoom to be fixed at 16.


rmarianski:
:+1:

Issue #523, opened 2016-02-11
Add elevation to peak features


nvkelso:
Peaks have an elevation value available in OSM, we should include that in tiles for labeling and later calculating tile_kind_rank values. [Example](https://www.openstreetmap.org/node/358807700) in Marin.

We should force it to be an int number, watch for string values.

nvkelso:
Working :)

<img width="238" alt="screen shot 2016-04-30 at 08 08 27" src="https://cloud.githubusercontent.com/assets/853051/14934232/c2a33780-0eaa-11e6-883c-443d725d6514.png">

<img width="752" alt="screen shot 2016-04-30 at 08 07 57" src="https://cloud.githubusercontent.com/assets/853051/14934233/c2c0a73e-0eaa-11e6-9953-db342238186e.png">

PR #751, opened 2016-04-26
Add elevation to peaks

zerebubuth:
Connects to #523. Requires mapzen/TileStache#147

Adds an `elevation` property to peaks and volcanoes, which is the height in meters. There's a small amount of fiddling with the polygons table, as `ele` isn't a column there, but is in the points table.

### DIFF
--- a/data/migrations/sql.jinja2
+++ b/data/migrations/sql.jinja2
@@ -146,8 +146,12 @@ DECLARE
   row ALIAS FOR $1;
 BEGIN
   RETURN mz_calculate_min_zoom_pois_(
-{%- for param in layers['pois']['params'] %}
+{%- for param in layers['pois']['params'] -%}
+{%- if param.key == '"ele"' %}
+        row.tags->'ele',
+{%- else %}
         row.{{param.key}},
+{%- endif -%}
 {%- endfor %}
         row.tags,
         row.way_area);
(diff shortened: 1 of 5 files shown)


data/migrations/sql.jinja2
```diff
@@ -146,8 +146,12 @@ DECLARE
   row ALIAS FOR $1;
 BEGIN
   RETURN mz_calculate_min_zoom_pois_(
-{%- for param in layers['pois']['params'] %}
+{%- for param in layers['pois']['params'] -%}
+{%- if param.key == '"ele"' %}
+        row.tags->'ele',
+{%- else %}
```
rmarianski: Is this so we can specify `ele` directly in the yaml without prefixing with `tags->`?

rmarianski:
:+1:

data/migrations/sql.jinja2
zerebubuth: Yes, for `planet_osm_polygon`. However, `planet_osm_point`, at least on my local DB, has it as a "full" column. This seemed like the simplest (although kinda klidgy) way to avoid having to introduce a distinction between the two tables which are otherwise very similar.

rmarianski: Ah, nice catch. I didn't realize that we had different schemas across tables. :+1:

PR #755, opened 2016-04-26
Remove `yes` value from pois kind values in docs

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/705

While scanning through the possible `pois` kind values, I also took the opportunity to normalize the syntax for consistency.


zerebubuth:
:+1: Looks great!

PR #756, opened 2016-04-26
Update boundaries docs for normalized kind values

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/718

The `type` values that are coming back from the `ne` queries are now _always_ either `country` or `state`. Since the `kind`s have been normalized, should we just drop the `type` property completely? The docs currently don't reflect what's going on, and I'm happy to update them subsequently to what we want to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+v0.10.0
+-------
+* remove `type` property from Natural Earth `boundaries` layer features
+
 v0.9.1
 ------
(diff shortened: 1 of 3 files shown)


zerebubuth:
:+1:

Those values are the ones I see in `boundaries.yaml` - looks good to me.

rmarianski:
Removed `type` from features and docs in 3fabb18.

zerebubuth:
:+1:

PR #758, opened 2016-04-27
Update the polygon poi min_zoom migration

rmarianski:
It looks like the polygon mz_poi_min_zoom migration was missing some
values in the where clause. This updates it to match the where clause in
the point table.


zerebubuth:
:+1:

PR #760, opened 2016-04-27
Add bicycle route details

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/647

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -585,6 +585,38 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql STABLE;
 
+CREATE OR REPLACE FUNCTION mz_cycling_network_(way_tags hstore, osm_id bigint)
+RETURNS text AS $$
+DECLARE
+  networks text[] := ARRAY(
+    SELECT hstore(tags)->'network'
+    FROM planet_osm_rels
+    WHERE parts && ARRAY[osm_id]
+      AND parts[way_off+1:rel_off] && ARRAY[osm_id]
+      AND mz_is_path_major_route_relation(hstore(tags)));
+BEGIN
+  RETURN CASE
+    WHEN networks && ARRAY['icn'] THEN 'icn'
+    WHEN networks && ARRAY['ncn'] THEN 'ncn'
+    WHEN way_tags->'ncn'='yes' OR way_tags ? 'ncn_ref' THEN 'ncn'
+    WHEN networks && ARRAY['rcn'] THEN 'rcn'
+    WHEN way_tags->'rcn'='yes' OR way_tags ? 'rcn_ref' THEN 'rcn'
+    WHEN networks && ARRAY['lcn'] THEN 'lcn'
+    WHEN way_tags->'lcn'='yes' OR way_tags ? 'lcn_ref' THEN 'lcn'
+  END;
+END;
+$$ LANGUAGE plpgsql STABLE;
+
+CREATE OR REPLACE FUNCTION mz_cycling_network(way_tags hstore, osm_id bigint)
+RETURNS text AS $$
+BEGIN
+  RETURN CASE
+    WHEN way_tags->'icn' = 'yes' OR way_tags ? 'icn_ref' THEN 'icn'
+    ELSE mz_cycling_network_(way_tags, osm_id)
+  END;
+END;
+$$ LANGUAGE plpgsql STABLE;
+
 -- returns TRUE if the given way ID (osm_id) is part of a path route relation,
 -- or NULL otherwise.
 -- This function is meant to be called for something that we already know is a path.
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -830,6 +830,11 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 
 * `aerialway`: See kind list below.
 * `aeroway`: See kind list below.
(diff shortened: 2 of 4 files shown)


data/functions.sql
```diff
@@ -585,6 +601,38 @@ BEGIN
...
+  END;
+END;
+$$ LANGUAGE plpgsql STABLE;
+
+CREATE OR REPLACE FUNCTION mz_cycling_network(way_tags hstore, osm_id bigint)
+RETURNS text AS $$
+BEGIN
+  RETURN CASE
+    WHEN way_tags->'icn' = 'yes' OR way_tags ? 'icn_ref' THEN 'icn'
+    ELSE mz_cycling_network_(way_tags, osm_id)
```
zerebubuth: Looks like you could move all the `way_tags` tests from `mz_cycling_network_` above to here. Do we need this logic (about `way_tags`) in `mz_calculate_is_bicycle_route` as well?

yaml/roads.yaml
```diff
@@ -32,6 +31,14 @@ global:
       sport: {col: sport}
       man_made: {col: man_made}
       walking_network: {expr: "mz_hiking_network(osm_id)"}
+      cycle_network: {expr: "mz_cycling_network(tags, osm_id)"}
```
zerebubuth: Nitpick: I think there was some discussion about making it `bicycle_network` rather than `cycle_network`, for consistency or something?

zerebubuth:
Yup, that's looking good so far!

data/functions.sql
rmarianski: This was my attempt to try and avoid the relatively expensive relation query if the way tags were already `icn`. I couldn't figure out a way to inline it all into the case branch.

Turns out that I wasn't using `mz_calculate_is_bicycle_route` anywhere. I removed it in 72d48a1.

yaml/roads.yaml
rmarianski: Good catch! I updated it in 33e87e0.

data/functions.sql
```diff
@@ -585,6 +601,38 @@ BEGIN
...
+DECLARE
+  networks text[] := ARRAY(
+    SELECT hstore(tags)->'network'
+    FROM planet_osm_rels
+    WHERE parts && ARRAY[osm_id]
+      AND parts[way_off+1:rel_off] && ARRAY[osm_id]
+      AND mz_is_path_major_route_relation(hstore(tags)));
+BEGIN
+  RETURN CASE
+    WHEN networks && ARRAY['icn'] THEN 'iwn'
```
zerebubuth: Should that say `icn` insead of `iwn`?

rmarianski: It should, nice catch! :zap: Fixed in ae0b48d.

As you can tell, I don't have tests for the case where we pull the network from relations yet :)

rmarianski:
@zerebubuth how do things look now?

zerebubuth:
:+1:

I was thinking "where's the rest of it?" but that's the power of YAML :zap:!

docs/layers.md
```diff
@@ -839,9 +844,11 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `leisure`: See kind list below.
 * `man_made`: See kind list below.
 * `network`: eg: `US:I` for the United States Interstate network, useful for shields and road selections.
+* `oneway_bicycle`: `oneway:bicycle` tag from feature
 * `oneway`: `yes` or `no`
 * `piste_type`: See kind list below.
 * `railway`: the original OSM railway tag value
+* `segregated`: `segregated` tag from feature.
```
nvkelso: Is `segregated` for bicycle cycleways? If so, please update documentation to clarify that.

yaml/roads.yaml
```diff
@@ -32,6 +31,14 @@ global:
       sport: {col: sport}
       man_made: {col: man_made}
       walking_network: {expr: "mz_hiking_network(osm_id)"}
+      bicycle_network: {expr: "mz_cycling_network(tags, osm_id)"}
+      oneway: {col: oneway}
```
nvkelso: Does the logic here conflict with the TileStache transform logic later?

rmarianski: For the `oneway` column? I only just moved that so both `oneway` and `oneway_bicycle` were next to each other. so this won't change anything wrt to `oneway`.

But, we do have some logic to normalize the `oneway` tag, which we don't do for `oneway:bicycle`. Do we want to do something similar for `oneway:bicycle` too?

nvkelso: Yes, we should normalize & take the line direction into account the same as with `oneway`.

Looks like basic set of attributes as `oneway`: 

* https://taginfo.openstreetmap.org/keys/oneway%3Abicycle#values.

rmarianski: What do we do in the situation where both exist? If they both say reverse, just reverse the geometry once? And otherwise, would we use the `oneway` tag's reversal orientation be what we honor first, and make the `bicycle:oneway` direction match accordingly? (If this doesn't make sense I can enumerate the cases)

nvkelso: Let's ignore this for now (leave the `oneway:bicycle=no`) and pick it back up next release in: https://github.com/mapzen/vector-datasource/issues/782.

Issue #524, opened 2016-02-11
Calculate peak tile_kind_rank values


nvkelso:
Like we do for subway stations now, but based on the elevation values ([example](https://www.openstreetmap.org/node/358807700)) from nearby peaks. Depends on #523 to add those elevation values.

In areas with dense peak coverage, this would allow (client-side) filtering.

nvkelso:
For instance, around Marin:

12/37.9243/-122.6018

<img width="766" alt="screen shot 2016-02-10 at 23 38 37" src="https://cloud.githubusercontent.com/assets/853051/12971384/71f20896-d04f-11e5-9eca-5e2f783e6791.png">

nvkelso:
And in Germany (12/47.5412/12.9766):

<img width="1256" alt="screen shot 2016-04-24 at 11 06 05" src="https://cloud.githubusercontent.com/assets/853051/14769421/b4536ada-0a0c-11e6-957f-ae78b4ceb247.png">

nvkelso:
Working:

<img width="915" alt="screen shot 2016-04-30 at 08 13 23" src="https://cloud.githubusercontent.com/assets/853051/14934258/8d3ebc62-0eab-11e6-8c0a-69fff81d19d8.png">

<img width="1013" alt="screen shot 2016-04-30 at 08 13 32" src="https://cloud.githubusercontent.com/assets/853051/14934256/89dcf1c4-0eab-11e6-8ac4-54e37c53fe8a.png">

PR #761, opened 2016-04-28
Tile kind rank for peaks

zerebubuth:
Adds a `kind_tile_rank` for peaks and volcanos, and limits the number at zoom < 16 to 5 per tile.

Connects to #524. Requires mapzen/TileStache#148.


rmarianski:
:+1:

Issue #731, opened 2016-04-19
Return of the son of zombie: mz_is_building: false


nvkelso:
[tile/12/653/1582](http://vector.dev.mapzen.com/osm/all/12/653/1582.topojson) has `mz_is_building: false` again :(

```
landuse: {
type: "GeometryCollection",
geometries: [
{
type: "Polygon",
properties: {
kind: "national_park",
protect_class: "5",
area: 54302000,
sort_key: 18,
mz_is_building: false,
min_zoom: 4.40875,
source: "openstreetmap.org",
operator: "U.S. National Park Service",
id: -3947724,
osm_relation: true,
name: "Marin Headlands (GGNRA)"
},
arcs: [
[
96
]
]
},
```

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/333

nvkelso:
Confirmed this is also a problem on my local dev build against current master branch.

nvkelso:
I still see the offending property on dev tiles :/

rmarianski:
Nice catch! The test was for the `buildings` layer and it's showing up in the `landuse` layer. I'll update the test and dig into why this is still happening.

nvkelso:
Verified.

PR #762, opened 2016-04-28
Check we don't get mz_is_building.

zerebubuth:
Add a test that we don't get `mz_is_building` on any output building in the example test tile.

I can't reproduce the problem locally, and I don't see it in dev for that tile. But it seems that we should have this test to make sure we don't regress on this again.

Connects to #731.


rmarianski:
:+1:

PR #764, opened 2016-04-28
Correct the sql.jinja2 load path

rmarianski:
Fix issue where template would not get loaded unless
create-sql-functions was run from the cwd


zerebubuth:
:+1:

PR #765, opened 2016-04-28
Update sort key tunnel test because data changed.

zerebubuth:
[TCH](http://www.openstreetmap.org/way/268517418) was updated to layer=-1, which changed the sort key. Updating test to point at [another layer=null trunk tunnel](http://www.openstreetmap.org/way/89912879#map=16/42.3811/-71.0690).

The `set()` of IDs is in there because we seem to be merging roads - are we intending to do that at z16?


rmarianski:
:+1:

PR #770, opened 2016-04-29
Clarify docs for segregated property

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/pull/760/files/a32c8a6ee2a0eab89638d5a13a218fe30731e2ed#r61586659

@nvkelso this look ok? Or did you have something else in mind?

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -848,7 +848,7 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `oneway`: `yes` or `no`
 * `piste_type`: See kind list below.
 * `railway`: the original OSM railway tag value
-* `segregated`: `segregated` tag from feature.
+* `segregated`: Set to `yes` when a path allows both pedestrian and bicycle traffic, but when pedestrian traffic is segregated from bicycle traffic.
 * `service`: See value list below, provided for `railway` and `highway=service` roads.
 * `type`:  Natural Earth roads and ferry
 * `walking_network`: Present if the feature is part of a hiking network. If so, the value will be one of `iwn` for International Walking Network, `nwn` for National Walking Network, `rwn` for Regional Walking Network, `lwn` for Local Walking Network.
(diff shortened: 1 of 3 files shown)


zerebubuth:
Looks like the [most common values](http://taginfo.openstreetmap.org/keys/segregated#values) are `yes` and `no`. Do we strip the `no` values? If so, probably worth just saying something like "Present on paths allowing both pedestrian and bicycle traffic with the value `yes` when pedestrian traffic is segregated from bicycle traffic."

rmarianski:
At the moment we're only just exposing the value on the feature. Should we only include it if its value is `yes`?

nvkelso:
Yeah, agree we should be stripping the `no` value (the property should only be present when it's se to `yes`). That's more in line with what we do for `oneway`, etc.

Something like this for docs? (Riffing on Matt's proposal):

```
Set to `yes` when a path allows both pedestrian and bicycle traffic, but when
pedestrian traffic is segregated from bicycle traffic.
```

Is this property on more `highway` types than `path`?

zerebubuth:
> Is this property on more `highway` types than `path`?

It's currently on all road types, since it could appear anywhere there's a segregated foot/cycle path, including on the sidewalk of another type of road. But in North America, it only appears on these `highway` values:

```
   highway    | num  
--------------+------
 cycleway     | 1556
 path         |  244
 footway      |   84
 service      |   10
 tertiary     |    6
 construction |    5
 pedestrian   |    5
 secondary    |    4
 trunk        |    3
 residential  |    3
 track        |    3
              |    2
 steps        |    1
```

I am intrigued as to where the segregated steps are, though...

nvkelso:
Haha :)

I think it's okay to say "path" in this context (instead of road or way), since we're already grouping (cycleway,  path, footway) into `kind:path` and that's where almost all the tags are.

rmarianski:
Updated the logic in 3cdbc9a, and the docs in 02ce1d5.

nvkelso:
Neat use of expression in the YAML!

zerebubuth:
Oh, I thought we were stripping out all the tags with the value `no` later on in the pipeline... I must have been mistaken.

PR #771, opened 2016-04-29
Update tests to reflect new data

rmarianski:

### DIFF
--- a/test/596-add-hiking-routes.py
+++ b/test/596-add-hiking-routes.py
@@ -32,11 +32,11 @@ assert_has_feature(
     { 'kind': 'minor_road', 'highway': 'residential', 'walking_network': 'nwn' })
 
 # http://www.openstreetmap.org/way/5254587
-# Mount Baker Highway - secondary_link - part of Pacific Northwest Trail (nwn)
+# Mount Baker Highway - secondary - part of Pacific Northwest Trail (nwn)
 # should be visible at z11
 assert_has_feature(
     11, 331, 704, 'roads',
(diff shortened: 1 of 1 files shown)


test/592-add-adjust-bicycle-pois.py
```diff
@@ -22,8 +22,8 @@
 assert_has_feature(
     16, 17238, 25950, 'pois',
     { 'kind': 'bicycle_rental_station', 'min_zoom': 17,
-      'capacity': 19, 'network': 'Bike Chattanooga',
-      'operator': 'Alta Bike Share', 'ref': '5' })
+      'capacity': 15, 'network': 'Bike Chattanooga',
+      'operator': 'Alta Bike Share', 'ref': '24' })
```
zerebubuth: I'm still seeing `capacity=19` and `ref=5` on [the node](http://www.openstreetmap.org/node/1960994139). Is it possible that it didn't get an `mz_poi_min_zoom` set for some reason?

rmarianski: Nice catch. I got confused with [this node](http://www.openstreetmap.org/node/1960994177), which is showing up in the dev tile and looked similar.

rmarianski: Turns out that this was being caused by the `remove_duplicate_features` transform removing it. This test failure should get resolved with: https://github.com/mapzen/vector-datasource/pull/772

rmarianski:
@zerebubuth I removed the commit to update the pois tests.

zerebubuth:
:+1: looks good!

PR #779, opened 2016-05-02
Remove mz_is_building from landuse layer

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/731

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -328,13 +328,13 @@ post_process:
   - fn: TileStache.Goodies.VecTiles.transform.drop_features_where
     params:
       source_layer: landuse
-      start_zoom: 13
+      start_zoom: 6
       where: >-
         properties.get('mz_is_building') is True
   - fn: TileStache.Goodies.VecTiles.transform.drop_properties
     params:
       source_layer: landuse
-      start_zoom: 13
+      start_zoom: 6
       properties: [mz_is_building]
   - fn: TileStache.Goodies.VecTiles.transform.merge_duplicate_stations
     params:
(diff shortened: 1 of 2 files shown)


queries.yaml
```diff
@@ -328,13 +328,13 @@ post_process:
...
       source_layer: landuse
-      start_zoom: 13
+      start_zoom: 6
       where: >-
         properties.get('mz_is_building') is True
   - fn: TileStache.Goodies.VecTiles.transform.drop_properties
     params:
       source_layer: landuse
-      start_zoom: 13
+      start_zoom: 6
```
nvkelso: So anytime we modify the landuse start zoom elsewhere in the codebase we need to update it here? 

Why was it set to 13 before?

Otherwise, LGTM.

rmarianski: > So anytime we modify the landuse start zoom elsewhere in the codebase we need to update it here? 

Yes. More specifically any time we update the starting zoom in the [landuse query](https://github.com/mapzen/vector-datasource/blob/9aa9507ab5489b8613127e0db6c87a5e2d1c8a65/queries/landuse.jinja2#L32) for when we include the mz_is_building property.

> Why was it set to 13 before?

We start including buildings at z13.

rmarianski: > Yes. More specifically any time we update the starting zoom in the landuse query for when we include the mz_is_building property.

I can't find it now, but at some point I think @zerebubuth had mentioned that we might want to just skip these targeted removals, and just have a single transform at the end that removes all `mz_*` properties from all features. That seems more universal and more robust to me.

nvkelso: Agree: that's covered in https://github.com/mapzen/vector-datasource/issues/622.

PR #780, opened 2016-05-02
Drop cycleway=no property

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/774

This only addresses the part of the issue related to cycleway's.

### DIFF
--- a/yaml/roads.yaml
+++ b/yaml/roads.yaml
@@ -36,10 +36,16 @@ global:
       oneway_bicycle: {col: "tags->oneway:bicycle"}
       cycleway:
         expr: |
-          CASE WHEN tags->'cycleway' <> 'no' THEN tags->'cycleway' END
-      cycleway_left: {col: "tags->cycleway:left"}
-      cycleway_right: {col: "tags->cycleway:right"}
-      cycleway_both: {col: "tags->cycleway:both"}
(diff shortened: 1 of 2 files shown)


nvkelso:
Word.

yaml/roads.yaml
```diff
@@ -34,7 +34,9 @@ global:
       bicycle_network: {expr: "mz_cycling_network(tags, osm_id)"}
       oneway: {col: oneway}
       oneway_bicycle: {col: "tags->oneway:bicycle"}
-      cycleway: {col: tags->cycleway}
+      cycleway:
+        expr: |
+          CASE WHEN tags->'cycleway' <> 'no' THEN tags->'cycleway' END
       cycleway_left: {col: "tags->cycleway:left"}
```
nvkelso: Each of the variants `cycleway_left`, `cycleway_right`, `cycleway_both`, needs the `no` logic too.

nvkelso: `none` looks common to as a synonym for `no`.

rmarianski:
Updated in 80f256e. Look ok now?

Also, the numbers for [cycleway:both](http://taginfo.openstreetmap.org/keys/cycleway:both#values) look small. Should we just drop that completely?

nvkelso:
LGTM

PR #784, opened 2016-05-02
Include highway=footway|pedestrian for polys only

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/776

Migration required is:
```sql
UPDATE planet_osm_line
  SET mz_landuse_min_zoom = mz_calculate_min_zoom_landuse(planet_osm_line.*)
  WHERE
    highway IN ('pedestrian', 'footway');
```

### DIFF
--- a/data/migrations/create-sql-functions.py
+++ b/data/migrations/create-sql-functions.py
@@ -192,6 +192,23 @@ class NotRule(object):
         return self.rule.columns()
 
 
+class ExpressionRule(object):
+
+    def __init__(self, column, expr, extra_columns=None):
+        self.column = column
+        self.expr = expr
+        self.extra_columns = extra_columns
+
+    def as_sql(self):
+        return self.expr
+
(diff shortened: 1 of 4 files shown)


nvkelso:
LGTM

rmarianski:
Restricting `aeroway` and `man_made` features to polygons in 8a7dde0.

rmarianski:
Does it hurt to have it though? Happy to remove it but I think it might be easier to understand if we can have the rule be that all `aeroway`s need to be polygons.

nvkelso:
You're right, I don't think having them in there will hurt. But it's also confusing to have them in there when we're pretty sure they aren't guarding against real problems. 

It's not clear to me where these rules originate, besides on the OSM wiki, e.g.:

screen shot 2016-05-02 at 15 35 33

rmarianski:
Updated in 3388ecd.

Issue #773, opened 2016-04-30
Dam lines should only be in boundaries layer


nvkelso:
* **What did you see?** I see duplicate features for dam in the `water` and `boundaries` layers.
* **What did you expect to see?** When it's eligible for both, only include it in the `boundaries` layer.
* **What map location are you having problems with?** 
  * Phoenix area
  * 15/33.5567/-112.0342
  * "Dreamy Draw Dam", id: 31442072
  * http://vector.dev.mapzen.com/osm/all/15/6186/13137.topojson
* **Screenshot?** 

<img width="495" alt="screen shot 2016-04-30 at 07 42 17" src="https://cloud.githubusercontent.com/assets/853051/14934120/2496703c-0ea7-11e6-9222-c3fa69215ca5.png">

nvkelso:
Good news: not a new bug (also in v0.9 prod tiles). But we should still fix it!

nvkelso:
Verified.

PR #785, opened 2016-05-02
connects to #773 to remove dams from water layer, keep them in boundaries

nvkelso:
Updated docs, updated changelog, updated queries.


rmarianski:
:+1:

PR #786, opened 2016-05-03
default footway and steps to 14, unless sidewalk or crossing still at 15

nvkelso:
@rmarianski I think this is right, but I need to run more migrations to get the tests to not fail for unrelated things. Can you run the test on your DB config to verify this change, please? If anything needs updating, please update.


rmarianski:
LGTM and I got the tests in test/596-add-hiking-routes.py to pass locally. :+1:

PR #789, opened 2016-05-03
Fix island label placements.

zerebubuth:
Don't allow polygons with `label_placement:yes` set in output, as Tangram will label these. Instead, generate the label points using the post-processor and delete the original geometry.

For future work: We should probably only be labelling the largest island in a multipolygon with the island label, or at least suppressing the nearby ones until later zooms.

Also: should we have a filter in the output which disallows setting `label_placement:yes` directly from the YAML? Given that the YAML-generated functions can be applied to points, lines and polygons?

Connects to #399. Requires mapzen/TileStache#153.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -298,12 +298,14 @@ post_process:
       source_layer: earth
       label_property_name: label_placement
       label_property_value: "yes"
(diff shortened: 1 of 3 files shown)


rmarianski:
:+1:

PR #793, opened 2016-05-04
Resolve test failure due to deleted node

rmarianski:

### DIFF
--- a/test/663-combo-outdoor-landuse-pois.py
+++ b/test/663-combo-outdoor-landuse-pois.py
@@ -34,14 +34,13 @@ assert_has_feature(
 assert_has_feature(
     16, 19314, 24677, 'pois',
     { 'kind': 'beach_resort', 'min_zoom': 16 })
-  
-#https://www.openstreetmap.org/node/1500943741
-# Needs to appear **before** tourism=hotel
-# Best Western Plus Blue Sea Lodge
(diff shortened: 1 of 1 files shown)


test/663-combo-outdoor-landuse-pois.py
```diff
@@ -34,14 +34,12 @@
 assert_has_feature(
     16, 19314, 24677, 'pois',
     { 'kind': 'beach_resort', 'min_zoom': 16 })
-  
-#https://www.openstreetmap.org/node/1500943741
+
+# Way: Terrasol Beach Resort (381817391) https://www.openstreetmap.org/way/381817391
```
nvkelso: These should be on two different lines:

```
# https://www.openstreetmap.org/way/381817391
# Terrasol Beach Resort
```

nvkelso:
Other than the above comment in https://github.com/mapzen/vector-datasource/pull/793/files#r62083093, LGTM.

test/663-combo-outdoor-landuse-pois.py
rmarianski: I've been putting them all on the same line for other tests :( I'll start using this pattern going forward.

rmarianski:
Updated in 5e64b10.

Issue #797, opened 2016-05-05
Some boundary lines missing from low zooms


nvkelso:
* **What did you see?** Some state boundary lines are missing at low zooms (zooms 7 and less).
* **What did you expect to see?** All boundary lines to show up.
* **What map location are you having problems with?** 7/41.307/-121.102 between California, Oregon, Nevada.
* **Screenshot?** 

**Zoom 7, missing some state boundary lines:**

<img width="1188" alt="screen shot 2016-05-05 at 16 57 23" src="https://cloud.githubusercontent.com/assets/853051/15060736/a4fe36b2-12e2-11e6-9654-e373420e8d86.png">

**Zoom 8, all good:**

<img width="1061" alt="screen shot 2016-05-05 at 16 57 31" src="https://cloud.githubusercontent.com/assets/853051/15060735/a4e4fcba-12e2-11e6-9ad1-def71746f517.png">

nvkelso:
@zerebubuth Can you take a look at this, please? We might need to role a v0.10.1 update as this is a regression over v0.9 prod.

nvkelso:
I made a bad assumption about **Natural Earth** in https://github.com/mapzen/vector-datasource/issues/687 and https://github.com/mapzen/vector-datasource/pull/648#issuecomment-204032004 and https://github.com/mapzen/vector-datasource/issues/517#issuecomment-200467229  :\

The `Admin-1 statistical boundary` and `Admin-1 statistical meta bounds` are inclusive of Admin-1 boundaries, not independent. We need to include them as `kind:state`.

* http://vector.mapzen.com/osm/all/7/21/47.topojson?api_key=vector-tiles-HqUVidw
* http://vector.dev.mapzen.com/osm/all/7/21/47.topojson?api_key=vector-tiles-HqUVidw

nvkelso:
<img width="914" alt="screen shot 2016-05-05 at 17 20 54" src="https://cloud.githubusercontent.com/assets/853051/15061077/e6cc0c42-12e5-11e6-9fc9-38a46db42822.png">
<img width="321" alt="screen shot 2016-05-05 at 17 20 49" src="https://cloud.githubusercontent.com/assets/853051/15061078/e6e54694-12e5-11e6-9bc6-67bda8247cd4.png">

nvkelso:
[...]

PR #798, opened 2016-05-06
Update state boundaries from NE to include statistical boundaries.

zerebubuth:
Include `Admin-1 statistical boundary` and `Admin-1 statistical meta bounds` as types of `state` boundary.

Connects to #797.

This was (somewhat surprisingly) the only migration I needed to run to get the tests to pass. Since this is on top of the v0.10.0 tag, will we tag the pre-merge state as v0.10.1? If so, then it should already be covered by the existing migrations in the repo, but the shorter version is:

```SQL
update ne_10m_admin_1_states_provinces_lines
set mz_boundary_min_zoom = mz_calculate_min_zoom_boundaries(ne_10m_admin_1_states_provinces_lines.*)
where featurecla in ('Admin-1 statistical boundary', 'Admin-1 statistical meta bounds');
```

### DIFF
--- a/test/797-add-missing-boundaries.py
+++ b/test/797-add-missing-boundaries.py
@@ -1,0 +1,10 @@
+# NE data - no OSM elements
+# boundary between NV and CA is _also_ a "statistical" boundary
+assert_has_feature(
+    7, 21, 49, 'boundaries',
+    { 'kind': 'state' })
+
+# boundary between MT and ND is _also_ a "statistical meta" boundary
+assert_has_feature(
+    7, 21, 49, 'boundaries',
+    { 'kind': 'state' })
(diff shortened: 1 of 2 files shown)


rmarianski:
:+1: 

Since other code has landed on master since, should we make a v0.10 release branch off the v0.10.0 tag, and merge this change into there? And then we can merge the release branch into master?

zerebubuth:
ea43400 is a commit on top of v0.10.0, so we could just tag it v0.10.1 if we want to?

rmarianski:
> ea43400 is a commit on top of v0.10.0, so we could just tag it v0.10.1 if we want to?

:+1: This is what I'll do.

Do we want to make a separate pr on top of this commit (ea43400) to include a v0.10.1 changelog entry and tag that? Or do we not bother and just add in a changelog entry on master?

rmarianski:
[Pr](https://github.com/mapzen/vector-datasource/pull/799) for the changelog entry.

PR #800, opened 2016-05-06
Test more than one thing

zerebubuth:
The test _should_ have tested two boundaries with different `featurecla`, but ended up being a copy/paste error. This fixes that.

Connects to #797.


rmarianski:
:+1:

PR #811, opened 2016-05-11
Add corridors, crossing, sidewalk

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/605

The `crossing` and `sidewalk` tags had some varied values, and I opted to just pass those through for now. `crossing` values of `no` are omitted, which seemed like a fair thing to do. `sidewalk`s are always included, whether `no` or `none`, since those might be useful based on the context.

### DIFF
--- a/test/605-corridor.py
+++ b/test/605-corridor.py
@@ -1,0 +1,5 @@
+# Way: The Nave (205644309)
+# http://www.openstreetmap.org/way/205644309
+assert_has_feature(
+    16, 10485, 25332, 'roads',
+    { 'id': 205644309, 'kind': 'path', 'highway': 'corridor' })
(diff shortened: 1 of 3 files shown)


test/605-corridor.py
```diff
@@ -0,0 +1,5 @@
+# Way: The Nave (205644309)
+# http://www.openstreetmap.org/way/205644309
+assert_has_feature(
+    16, 10485, 25332, 'roads',
+    { 'id': 205644309, 'kind': 'path' })
```
nvkelso: We should also test `, 'highway': 'corridor'`.

nvkelso:
Otherwise looks good!

test/605-corridor.py
rmarianski: Good call. Updated in cf88d33.

nvkelso:
💯

PR #813, opened 2016-05-12
Configure area threshold to 1 pixel

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/202
Depends on https://github.com/mapzen/tilequeue/pull/87


zerebubuth:
Isn't the default `1`? Do we need to specify it explicitly?

rmarianski:
Yup, the default is 1. I just figured that putting it in there will make it easier to change in the future without having to look up what the property is called. Happy to leave as is too.

zerebubuth:
Yup, that's a good point! :+1:

Issue #492, opened 2016-01-15
Optimize release tests


nvkelso:
From MattA: `These are really 'release tests' rather than unit tests, as they require a lot of data to be loaded in the database first (at the moment, it requires North America). A better test suite would load some minimal static fixture data first.`

nvkelso:
This work has been accomplished in: https://github.com/mapzen/vector-datasource/pull/464, this is the placeholder issue to link that PR.

nvkelso:
More work will be done in: https://github.com/mapzen/vector-datasource/issues/491.

PR #815, opened 2016-05-12
Add CI support for tests.

zerebubuth:
This is a first pass at adding CI support for tests. It works for me, and may not work for anyone else. It also doesn't have all the bits that would be necessary for using in CI right now. With those caveats out of the way...

This adds a script `scripts/setup_and_run_tests.sh` which:

1. Creates and sets up a new local database (with a temporary name like `vector_datasource_NNNNN`)
2. Reads the comments in the tests to figure out which OSM data is used by the tests.
3. Downloads that from Overpass (defaults to the public server, override with `$OVERPASS_SERVER` to use a local one).
4. Loads that up, plus the Natural Earth data, and some extracts of the water and earth polygons in `test/fixtures/*/*.shp`.
5. Starts up an instance of `tileserver` and runs the tests against it.

This whole process runs, all on my local machine including Overpass, in 140s, which seems reasonable for a CI run.

Still TODO:

* Figure out how to make a CircleCI config for this.
* Run our own Overpass server to be good citizens and avoid getting 429ed by the public instance.
* Add a `requirements.txt` to specify the versions (exact hashes) of `tilequeue` and `TileStache` that we need to pass the tests.
* Figure out how to test the migrations too.

It's WIP, but shouldn't affect any other code - this is functionality in addition to what we already have (or should be!). I wanted to get a PR in so that we could review and figure out next steps.

Connects to #492 (sort of).

/cc @nvkelso.


data/apply-planet_osm_line.sql
```diff
@@ -22,7 +22,7 @@ UPDATE planet_osm_line
   WHERE mz_calculate_min_zoom_water(planet_osm_line.*) IS NOT NULL;
 
 UPDATE planet_osm_line
-  SET mz_water_min_zoom = mz_calculate_min_zoom_boundaries(planet_osm_line.*)
+  SET mz_boundary_min_zoom = mz_calculate_min_zoom_boundaries(planet_osm_line.*)
```
rmarianski: Great catch! :zap:

data/apply-updates-non-planet-tables.sql
```diff
@@ -1,3 +1,93 @@
...
+
+UPDATE ne_10m_admin_0_boundary_lines_land
+  SET mz_boundary_min_zoom = mz_calculate_min_zoom_boundaries(ne_10m_admin_0_boundary_lines_land.*)
+  WHERE mz_calculate_min_zoom_boundaries(ne_10m_admin_0_boundary_lines_land.*) IS NOT NULL;
+
+UPDATE ne_10m_admin_1_states_provinces_lines
+  SET mz_boundary_min_zoom = mz_calculate_min_zoom_boundaries(ne_10m_admin_1_states_provinces_lines.*)
+  WHERE mz_calculate_min_zoom_boundaries(ne_10m_admin_1_states_provinces_lines.*) IS NOT NULL;
+
+
```
rmarianski: Good catch on the move here too.

queries/boundaries.jinja2
```diff
@@ -84,7 +84,7 @@ UNION ALL
 SELECT
   NULL AS name,
   'yes' AS maritime_boundary,
-  HSTORE('') as mz_properties,
+  HSTORE('kind=>"maritime"') as mz_properties,
```
rmarianski: :zap:

test.py
```diff
@@ -75,6 +99,10 @@ def match_properties(actual, expected):
                 if not isinstance(v, exp_v):
                     return False
 
+            elif hasattr(exp_v, '__call__'):
```
rmarianski: You can spell it this way:

    callable(exp_v)

yaml/landuse.yaml
```diff
@@ -323,9 +326,6 @@ filters:
   - filter: {tourism: picnic_site}
     min_zoom: GREATEST(LEAST(zoom, 16), 9)
     output: {kind: picnic_site}
-  - filter: {historic: fort}
-    min_zoom: GREATEST(LEAST(zoom, 16), 9)
-    output: {kind: fort}
```
rmarianski: Does this resolve a test failure?

rmarianski:
:+1:

yaml/landuse.yaml
nvkelso: I recall needing to move `fort` above attraction a few weeks ago as they are sometimes also tagged `attraction`. So moving it is "okay", but confusing why it wasn't already up there.

nvkelso:
Very nice! 

One day when we add tests for road landuse intercuts looks like we'll need to do the same as you've done for the stations (referencing multiple imports for related overlapping landuse features, not just the road).

test.py
zerebubuth: Oh, cool. It wasn't clear to me from the internet whether that was a Python 3 feature. Fixed in 144c7cc.

yaml/landuse.yaml
zerebubuth: I can't find any commits moving it. Is it possible that you pushed them to a branch which never got merged?

PR #819, opened 2016-05-16
Add barrier=toll_booth to pois layer

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/479

The first test, http://www.openstreetmap.org/way/25814380 , also has a `building` tag. It is showing up in the `buildings` layer, and now it will be showing up in the `pois` layer too. It also has a `landuse_kind=park` on it. Is it a problem that it will start showing up in multiple layers?

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -268,6 +268,9 @@ filters:
   - filter: {barrier: gate}
     min_zoom: 15
     output: {kind: gate}
+  - filter: {barrier: toll_booth}
+    min_zoom: 15
+    output: {kind: toll_booth}
   - filter: {craft: sawmill}
     min_zoom: 15
     output: {kind: sawmill}
--- a/data/migrations/v1.0.0-point.sql
+++ b/data/migrations/v1.0.0-point.sql
@@ -1,0 +1,6 @@
+UPDATE
+  planet_osm_point
+  SET mz_poi_min_zoom = mz_calculate_min_zoom_pois(planet_osm_point.*)
+  WHERE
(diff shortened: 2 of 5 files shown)


yaml/pois.yaml
```diff
@@ -265,9 +265,10 @@ filters:
   - filter: {amenity: toilets}
     min_zoom: 18
     output: {kind: toilets}
-  - filter: {barrier: gate}
+  - filter: {barrier: [gate, toll_booth]}
```
nvkelso: While this works, I propose two separate filters for `gate` and `toll_booth` as the gate min_zoom is likely far too early (!?) and should be modified to something like zoom 17, setting both the kind values explicitly for each filter.

nvkelso:
It's normal for us to harvest "icon-able" features from the buildings and put them into the `pois` layer instead. I expect the feature in the `buildings` layer to be a polygon of `kind:roof`, while the feature in the `pois` layer is a point with `kind:tool_booth`. (It's getting `landuse_kind:park` from the landuse intercut.) In this case the POI isn't going to intersect with the road (centroid of the building isn't on the road), but that's a minor problem we'll ignore for now.

yaml/pois.yaml
rmarianski: Updated commit in 5884507.

nvkelso:
LGTM

rmarianski:
Added migration in 0522f11 and updated docs in a2611a2.

nvkelso:
👍

Issue #778, opened 2016-05-01
Boundaries layer osm_relation value should be `yes` not `true`


nvkelso:
* `osm_relation`: `true`, which can also be deduced from negative `id` values.

Our convention is to use `yes`, not `true`.

zerebubuth:
I think the use of `yes` (actually the string `"yes"`) is just historical accident following the normalisation of OSM's use of `yes`, `1` and `true` to all mean positive.

I find it a compelling argument for keeping `true` (not `"true"`, though), that JSON and MVT both natively support a boolean type. Using a boolean rather than a string hopefully makes it clearer to anyone consuming the data that it only has two values, unlike a string, and it can be stored more compactly.

Anywhere we're putting a `"yes"` value in the output tiles and we mean a boolean, I think we ought to convert it to a boolean type.

nvkelso:
I'm good with converting the boolean `yes` values to `true`.

zerebubuth:
I started doing some work on this and hit an annoying bump: `HSTORE` values can only be text, therefore cannot store booleans. This makes it difficult to pull boolean properties with the correct types out of the query.

An alternative way of doing things would be to keep `"yes"` as truth in the queries, and have a universal conversion of `"yes"` to `true` when we're decoding the row and `mz_properties`. The problem with that is that it might introduce mixed type values (e.g: with `true` and `"random other value"`).

Alternatively, we could specify this as a post-processing step for each layer, with the keys to convert to boolean explicitly set. It'a a little annoying (to me, at least) to do this as a final step, meaning we have to keep the `"yes"` values through the rest of the processing pipeline.

@rmarianski what do you think?

rmarianski:
Instead of an hstore, can we use a json value as the result of the properties?

zerebubuth:
@rmarianski - great idea, btw! Thanks for that. I thought I was at a dead end, but JSON(B) made #161 much easier to implement :smile:

nvkelso:
👍

PR #824, opened 2016-05-17
Normalise 'yes' values to true

zerebubuth:
Instead of `"yes"` use `true` where the value can only be `true` or `false`. Generally `false` is implied by omission. Hopefully this makes the intent clearer, as boolean values can only be true or false, whereas string values could potentially have many values.

There are potentially memory and time savings for anyone decoding the tile also, as boolean values are generally compact compared with strings.

Notes:

1. For the moment, I've left `oneway` as-is. It seems that we reverse the linestring for `oneway=reverse`, so this could be a candidate for booleanification too.
2. I've collapsed all the values for `surveillance` down to a boolean. Do we want to expand them again, if so, is it worth re-writing the `yes` value to `unknown`?

Connects to #778.

/cc @nvkelso does this look good - it's a bit of a bigger "breaking change" than the ticket originally specified.


rmarianski:
:+1:

docs/layers.md
```diff
@@ -500,10 +498,10 @@ Features from OpenStreetMap which are tagged `disused=*` for any other value tha
 * `access`: Whether the bicyle parking is for general public use (`yes`, `permissive`, `public`) or for customers only (`customers`) or private use only (`private`, `no`).
 * `capacity`: Approximate number of total bicycle parking spots.
 * `covered`: Is the parking area covered.
-* `fee`: Whether a fee must be paid to use the bicycle parking (`yes`/`no`).
+* `fee`: Whether a fee must be paid to use the bicycle parking (if `true`, the default is `false`).
```
nvkelso: Would we ever export `false` for **fee**?

nvkelso: (If not, edit to be more like the **surveillance** description?)

docs/layers.md
```diff
@@ -839,18 +837,18 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `ferry`: See kind list below.
 * `highway`: See kind list below.
 * `is_bicycle_related`: Present and `true` when road features is a dedicated cycleway, part of an OSM bicycle network route relation, or includes cycleway infrastructure like bike lanes or designed for shared use.
-* `is_bridge`: `yes` or `no`
+* `is_bridge`: `true`, default is `false` if not present.
```
nvkelso: If value is `false` in the data, do we even export this property? Description reads as if it would be.

nvkelso:
I'm fine with expanded scope, great work!

Looking at the updated tests, the tile content is `True` for GeoJSON and TopoJSON. But we now say `true` in the docs. Should that be `True` instead?

What does MVT store the `True` as? Why does the Mapbox documentation say they convert booleans to `1` and `0` values instead of passing thru the boolean?

@bcamper how would the Tangram handle the new boolean values? We've been saying things like `intermittent: yes` and `intermittent: true` in scene files. Does Tangram correctly distinguish between `true` (meaning the property is present with any value) and `True` (the value of the property as a boolean)?

docs/layers.md
zerebubuth: We do export `false` for "fee" - as in [this test case](https://github.com/mapzen/vector-datasource/blob/master/test/592-add-adjust-bicycle-pois.py#L97-L100). I've updated the docs in #825, which hopefully make that clearer.

docs/layers.md
zerebubuth: For `is_bridge`, `is_link` and `is_tunnel` they should never be exported if they're `false`. I've updated the docs to try and make that clearer in #825.

zerebubuth:
The value is `True` for Python, but `true` in JSON and YAML (also and confusingly, `yes`, `True`, `on` and [various other values](http://yaml.org/type/bool.html) in YAML - so we'll stick to JSON). I've tried to be consistent and use the lower-case `true` in the docs, as that should be what's actually present in the (Geo/Topo)JSON.

MVT [can store booleans natively](https://github.com/mapbox/vector-tile-spec/blob/v1.0.0/1.0.0/vector_tile.proto#L24), and I can't find a reference to converting that to `0` or `1`, although it's possible that might be happening in later processing stages. Do you have a link to where in the Mapbox docs it says about booleans, please?

For `intermittent` and the `is_*` road properties, using `true` should behave as it did before - those values should only be present when they're `true`. However, `fee` and `surveillance` can be `true`, `false` or unspecified, so we might have to revert those if we can't match them in the Tangram style.

nvkelso:
Thanks for the background :)

@bcamper needs to weigh in on Tangram side of things to determine if we can use `true` to distinguish the general presence of a key with any value (really testing null, not null of any `key:*`)  versus the feature boolean property value of `key:true`.

I see MVT boolean reference on: https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#layer-reference

```
Boolean fields

Some fields represent a boolean condition; the value may be either true or false. 
To keep the vector tiles compact these fields may be stored as integers, 
where 0 = false and 1 = true.
```

zerebubuth:
I'm not quite sure what to make of that. Possibly it's referring to the fact that protobuf stores booleans as a [wire type 0](https://developers.google.com/protocol-buffers/docs/encoding#structure) varint under the hood? Or perhaps it can be more compact if there are other 0s and 1s in the file, as the layer values dictionary will store only one value for 0/`false` instead of two?

On the list of things which are making our tiles large, it is probably not top of the list, though.

PR #827, opened 2016-05-18
Update json behavior to be 9.3 compatible

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/826


queries/earth.jinja2
```diff
@@ -21,14 +21,14 @@ FROM
 WHERE
     mz_earth_min_zoom <= {{ zoom }} AND
     {{ bounds|bbox_filter('the_geom') }}
-GROUP BY mz_properties
+GROUP BY gid
```
zerebubuth: Is it not possible to group by JSON data types?

rmarianski: Alas no, postgresql complained about it.

If it's safe to assume that we always have a `kind`, would it be better to group on that?

rmarianski: Or maybe we should coerce the json to a string?

zerebubuth: Actually, I'm getting errors like:

```
ProgrammingError: column "t.*" must appear in the GROUP BY clause or be used in an aggregate function
LINE 6:     mz_calculate_json_earth(t.*) AS mz_properties,
                                    ^
```

rmarianski: Hhmm, I missed the updates in the `water` queries and am seeing that too.

rmarianski: I eliminated the `group by`s in 24c9a4e.

zerebubuth:
:+1: Thanks! Let's take a look at it on dev

PR #828, opened 2016-05-19
Dog park was updated

zerebubuth:
Dog park was updated and [turned into an area](http://www.openstreetmap.org/way/417184097)! Test needed updating to match.


nvkelso:
This test is for nodes, we need a new node.

zerebubuth:
Added a node test in 447bb80

nvkelso:
LGTM

PR #835, opened 2016-05-20
Add highway=rest_area|services to pois and landuse

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/480

The `kind` will be set to `rest_area` or `services`. Do we want to normalize this instead?

I used `GREATEST(14, LEAST(zoom, 15))` for the pois `min_zoom`. Sounds about right?

I didn't update the `landuse` `sort_key`. Would it fit in between a `caravan_site` and `recreation_ground`?


nvkelso:
They're more similar to highway exits (junctions), which we have at zoom 12 now. Let's try showing the new `rest_area` and `services` features at POI default of **zoom 11** before them. 

Agree on **normalizing** the names (we do this for fishing areas, for instance) to `service_area`.

In Japan there are amusement parks in service areas (I know, right!?). Recommend landuse **sort_key** below `commercial` and above `college`.

rmarianski:
Updates in 3429fe5.

nvkelso:
Oops, my bad. We need two separate features, only `services` needs to be normalized:

- `service` -> `service_area`
- `rest_area`

yaml/landuse.yaml
```diff
@@ -332,3 +332,7 @@ filters:
   - filter: {barrier: hedge}
     min_zoom: GREATEST(LEAST(zoom, 16), 9)
     output: {kind: hedge}
+  - filter: {highway: [rest_area, services]}
+    min_zoom: GREATEST(LEAST(zoom, 16), 9)
```
nvkelso: We should never see these features before zoom 11, but setting to `9` here allows them to show up earlier. (This is a problem with many of the queries!)

rmarianski:
Updated in b86deed.

nvkelso:
💯

PR #836, opened 2016-05-20
Transition pois transforms to yaml

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/754

These transforms seemed like good candidates for specifying in yaml instead. The others were a little bit more involved or were applied more broadly.


zerebubuth:
:+1: looks good to me.

PR #838, opened 2016-05-20
Limit addresses to points

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/834

Migration in prod will be to create sql functions again, and run:

```sql
UPDATE
  planet_osm_polygon
  SET
    mz_building_min_zoom = mz_calculate_min_zoom_buildings(planet_osm_polygon.*)
  WHERE
    "addr:housenumber" IS NOT NULL
     AND COALESCE(mz_building_min_zoom, 999) <> COALESCE(mz_calculate_min_zoom_buildings(planet_osm_polygon.*), 999);
```

### DIFF
--- a/yaml/buildings.yaml
+++ b/yaml/buildings.yaml
@@ -106,6 +106,9 @@ filters:
   # address points
   - filter:
       "addr:housenumber": true
+      way:
+        expr: |
+          GeometryType(way) = 'POINT'
     min_zoom: 17
     output:
       kind: address
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+v0.10.3
+-------
+* **Release date**: 2016-05-20.
+* Limit addresses to points. See [#834](https://github.com/mapzen/vector-datasource/issues/834).
+
 v0.10.2
 -------
 * **Release date**: 2016-05-10.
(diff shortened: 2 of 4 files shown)


yaml/buildings.yaml
```diff
@@ -106,6 +106,9 @@ filters:
   # address points
   - filter:
       "addr:housenumber": true
+      way:
```
nvkelso: This looks to limit it to address nodes only?

What about adding 2nd filter like:

```
- "addr:housenumber": true
- building: true
```

zerebubuth: We should have already picked up anything with a `building` tag in an earlier filter. And the `building_standard_properties` includes the house number and street, so it should be only the points that we need to filter in at this stage.

zerebubuth:
Looks good to me! And tests, too :smile: :+1:

yaml/buildings.yaml
rmarianski: Do we want to drop addresses unless they are on a building? According to [taginfo](http://taginfo.openstreetmap.org/keys/addr:housenumber#combinations), ~ 60% don't have a `building` tag.

nvkelso: No, we don't want to drop address points if they are not on a building.

nvkelso: I see the post processing transform now, thanks. LGTM.

PR #839, opened 2016-05-20
Remove duplicate kinds in docs

rmarianski:


nvkelso:
👍 

Matches the 3 dups I see in my local file, and matches up with the `kind` audit I'm doing.

Issue #663, opened 2016-04-04
Add outdoor combo landuse & pois


nvkelso:
The following are features with mix of OSM node and way data that need to be added both to landuse and poi layers.

**Format is:**
_LINK to osm wiki, EXAMPLE to construct test, recommended MIN_ZOOM value, and COUNT of features._

- [ ] [leisure=water_park](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dwater_park), similar zooms to theme_park (`"${GREATEST(13, LEAST(zoom + 6.32, 17))}"`). Around 8,000
    - [ ] [poi](https://www.openstreetmap.org/node/2753215890)
    - [ ]  [landuse](https://www.openstreetmap.org/way/31198945)
- [ ] [leisure=beach_resort](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dbeach_resort), same zoom ranges as leisure=resort (`"${GREATEST(14, LEAST(zoom + 5.32, 17))}"`), without even a name. Around 2000.
    - [ ] [POI](https://www.openstreetmap.org/node/2407351205), example [2](https://www.openstreetmap.org/node/3575812477), POI example with [tourism=hotel](https://www.openstreetmap.org/node/1500943741) that should be overriden. Here's a [weird one](https://www.openstreetmap.org/way/359653483) that is natural=beach and not hotel – hould just be `kind:beach`. See [similar](https://www.openstreetmap.org/way/257716817)
    - [ ] [landuse](https://www.openstreetmap.org/way/381817396)
- [ ] [leisure=summer_camp](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dsummer_camp), zoom 15 range up to zoom 13. Around 200.
    - [ ] [pois](https://www.openstreetmap.org/node/4050178586) and [2](https://www.openstreetmap.org/node/3655879348)
    - [ ] [landuse](https://www.openstreetmap.org/way/333764953)
- [ ] [historic=battlefield](http://wiki.openstreetmap.org/wiki/Tag:historic%3Dbattlefield), zoom 14 up to zoom 12? POIs default zoom 17. Around 2000.
    - [ ] [pois](https://www.openstreetmap.org/node/3992988013) and [2](https://www.openstreetmap.org/node/3838356961)
    - [ ] [landuse](https://www.openstreetmap.org/way/231393152)
[...]

Issue #566, opened 2016-03-04
Add natural=tree_row and other landuse line geoms


nvkelso:
Looks like some landuse are stored as [line geoms only](http://wiki.openstreetmap.org/wiki/Key:natural) in OSM:

- `landuse=tree_row`: [taginfo](http://taginfo.openstreetmap.org/tags/natural=tree_row) and [osm wiki](http://wiki.openstreetmap.org/wiki/Tag:natural=tree%20row?uselang=en-US). Example [1](https://www.openstreetmap.org/way/382951737). Around 187 ,725 features.
- `landuse=hedge`: [taginfo](http://wiki.openstreetmap.org/wiki/Tag:barrier%3Dhedge) around 512 features.

/cc @IndyHurt

nvkelso:
This is a nice to have, moving to next release.

nvkelso:
These new features need a custom sort_key value (currently defaulting to 11).

These should work:

* tree_row,229
* hedge,228

nvkelso:
Verified tree_row at 19/41.19816/-8.68721.

nvkelso:
Verified hedge at 19/37.76778/-122.40111.

Issue #418, opened 2015-12-04
WOF neighbourhoods should export localized names


nvkelso:
https://whosonfirst.mapzen.com/spelunker/id/85910293/

This neighbourhood has a default name in Japanese, which is what we're exporting as the name in tiles. But there are other languages, and we should export those, too.

nvkelso:
Moving out of the v0.7 milestone to load balance.

There's also some changes in the works on the WOF side for default name and localized names that might require additional work, so would like to batch with that.

Plus Tangram still doesn't have ability to auto-switch displayed languages, so...

nvkelso:
And again to v0.9

thisisaaronland:
See also: https://github.com/whosonfirst/whosonfirst-data/issues/183

rmarianski:
@zerebubuth what do you think of making a new hstore column in the database to store all of these? That way, we can include it as the tags value in the query, and it should just get included in the output via the usual transform. We'd have to make sure that we used the same i18n scheme as osm when we save the data, or maybe it makes more sense to have the transform be more flexible to support both schemes.

zerebubuth:
Hstore sounds good.

If we start translating between schemes, then it makes it more work for us later on. Reading the [WOF name docs] I'm confused as it keeps referring to country codes, and we use [ISO 639-1 language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)...

I feel it might be easier to just expose the data that WOF has and let whoever's making the map choose what's the appropriate country / language codes for them.

nvkelso:
We need should resolve this inconsistency for customers on the server as part of this issue.

rmarianski:
@dianashk we'd like to be consistent with what you end up doing in search. What do you think?

thisisaaronland:
That the WOF documentation refers to country codes is a bug. The docs _mean_ to indicate that we are using ISO-639-3 language codes. See also:

https://www.w3.org/International/articles/language-tags/#language

I will update the docs, shortly, when I am not using my hands to put food in my mouth...

rmarianski:
Just spoke with @dianashk and @orangejulius, and we'll update to use iso-693-3 codes across the board. We'll normalize the tags from osm to convert the iso-693-1 codes to iso-693-3 codes.

nvkelso:
This is pretty awesome :)

[...]

PR #846, opened 2016-05-25
Fix tests

zerebubuth:
A few tests were failing against the new CI tests:

1. #418 WOF tests needed fixture data with the l10n column to pass.
2. #465 Changes to `fitness` had excluded `amenity=gym` and `fitness_station`.
3. #566, #663 Sort key changes had resulted in a bunch of changes.
4. #647 OSM way 17125595 was part of an ICN network, now changed to NCN, and apparently the subject of [much discussion](https://www.openstreetmap.org/changeset/39469430) as to whether it should be a *CN or MTB route. Instead, I've included an unclassified way which is part of a stable ICN.


test/418-wof-l10n_name.py
```diff
@@ -17,7 +17,7 @@
       'name:spa': type(None) })
 
 # San Francisco (osm city)
-# https://whosonfirst.mapzen.com/data/858/826/41/85882641.geojson
+# http://www.openstreetmap.org/node/26819236
```
rmarianski: Whoops! Sorry forgot to update this.

yaml/pois.yaml
```diff
@@ -163,9 +163,15 @@ filters:
...
-      leisure: [fitness_centre, gym, fitness_station]
+      any:
+        leisure: [fitness_centre, gym]
+        amenity: gym
     min_zoom: LEAST(zoom + 3.98, 17)
     output: {kind: fitness}
+  - filter:
+      leisure: fitness_station
+    min_zoom: LEAST(zoom + 3.98, 17)
+    output: {kind: fitness_station}
```
rmarianski: We don't want `fitness_station` to normalize to `kind=fitness`?

rmarianski:
:+1:

yaml/pois.yaml
zerebubuth: I don't think so... at least, that's not what the tests said!

There's a big difference between a gym and a "push up" station in the park, so I think it makes sense to keep them separate.

PR #848, opened 2016-05-25
Don't drop building features which are also label placements.

zerebubuth:
We used to drop the `mz_is_building` property from label placements during the generation process using `drop_keys`, and then remove all features with `mz_is_building` in a later post-process stage. We recently removed `drop_keys` in favour of a more general process to drop all `mz_*` properties, but this meant the second stage was removing the label placements too.

This changes the filter so that anything marked `label_placement` is not removed.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -342,7 +342,8 @@ post_process:
       source_layer: landuse
       start_zoom: 6
       where: >-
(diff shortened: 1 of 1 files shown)


rmarianski:
:+1:

PR #850, opened 2016-05-31
Ignore l10n for non-OSM & non-WOF names

zerebubuth:
We have l10n conversion functions for WOF and OSM, but leave the variable unset when the source doesn't match either of those. Just to be on the safe side, this changes the behaviour to exit the function when the source doesn't match either WOF or OSM.

Also, fixes up a test case which was failing due to a data change.

Also, minor fixes to clean up from the test running script and ignore Python-related development files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,14 @@ test.log
 
 # editor temporaries
 *~
+
 # downloaded shapefiles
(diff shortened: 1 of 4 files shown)


rmarianski:
:+1:

PR #851, opened 2016-05-31
Add support for configurable pixel sizes

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/810

Included in this pr is a system to assign `class`es to features. Based on earlier conversations, I assume that we want to expose these properties in the tiles themselves. If we don't, we should rename to `mz_class` and they will get dropped. I also just made them integers for now; but they can be changed to whatever is easier to work with. Something else to consider is that the word `class` might not be the smartest choice because it's a keyword in many languages. However, I tried it in a node.js repl and didn't seem to have any problems with it though.

The implementation here re-uses the csv machinery to specify and assign these properties. The idea is that the `mz_min_pixels` spreadsheets will define the rules for the configurable pixel sizes, and a transform is configured that drops any features with areas that don't meet this threshold. The transform assumes that the thresholds will be in pixel values, and calculates the required area for the feature based on the zoom level. Having the capability to assign `class` values could be useful to determine the pixel thresholds.

Note that there aren't any rules specified, but the pieces are in place to honor any subsequent changes.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -339,6 +339,80 @@ post_process:
       source_layer: transit
       target_value_type: int
 
+  # add minimum pixel thresholds per layer
+  # this should run after all the `class`es have been assigned to features
+  # these are configured only for layers that contain polygonal
+  # geometries
+  - fn: vectordatasource.transform.csv_match_properties
+    resources:
+      matcher:
+        type: file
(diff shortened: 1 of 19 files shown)


zerebubuth:
Looks good to me! :+1:

PR #861, opened 2016-06-15
Remove deleted node from integration test

rmarianski:
The subsequent test was exercising the same behavior.


nvkelso:
LGTM

PR #863, opened 2016-06-15
Update the meters per pixel area call

rmarianski:


rmarianski:
@zerebubuth I'll merge this in, but putting in a pr in case you want to review after the fact.

PR #867, opened 2016-06-16
Add transform to merge buildings

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/845

This is a wip still.


rmarianski:
I took a stab at writing a transform to merge building features together. @zerebubuth mind taking a look to see if it matches what you were thinking?

I've only run preliminary tests so far. Ironically, it seems like there's more detail on the map after merging buildings together. I re-calculated the area and volume on the merged value, so perhaps it's much bigger than before now and tangram itself was filtering some features out previously. This might be the wrong thing to do though.

I also just picked quantization strategies based on zoom level which are more aggressive than what we were talking about previously. Happy to change these to whatever makes more sense.

queries.yaml
```diff
@@ -623,6 +623,22 @@ post_process:
...
+  - fn: vectordatasource.transform.merge_building_features
+    params:
+      source_layer: buildings
+      start_zoom: 13
+      end_zoom: 15
+      quantize:
+        13: vectordatasource.transform.quantize_height_round_nearest_10_meters
+        14: vectordatasource.transform.quantize_height_round_nearest_5_meters
+        15: vectordatasource.transform.quantize_height_round_nearest_meter
+      exclude:
```
nvkelso: What's the difference between `exclude` and `ignore` here? 

Reading the src it looks like `ignore` is similar to what we call `remove` elsewhere in the transforms?

nvkelso:
@rmarianski There are some assumptions in our Tangram styles around individual building `area`, `height`, `volume`, `location`, `name`, `kind`, and `landuse_kind`. 

Since the new area of the merged feature is "very big" compared to the smaller pieces pre-merge, the style would end up showing most buildings instead of a selection of buildings. 

To work around this we could modify the tile merge logic (e.g., large buildings with names wouldn't get merged, smaller buildings with names would & etc). Based on the specific merging inputs, I think we can accomplish the same effective design in the scene file to calculating a new importance attribute (similar to tile rank), either a number (e.g.: `1`, `2`, `3`, `4`, and `5`,  or something like `XS`, `S`, `M`, `L`, `XL` and the scene file would need to be updated). Might be interesting to look at assigning `min_zoom` to these, too?

Things that originally appear in zoom 13 could get `XL`.
Things that get added in zoom 14 could get `L`.
Things that get added in zoom 15 could get `M`.
Things that get added in zoom 16 could get `S`.
Things that get added in zoom 17 (the style, see below) could get `XS`.

Here's an example from the outdoor style (which is mostly the same as the other house styles):

```
                    - { $zoom: [13], area: { min: 100000 }, name: true }
                    - { $zoom: [13], area: { min: 100000 }, kind: true }
                    - { $zoom: [13], height: { min: 250 } }
                    - { $zoom: [13], volume: { min: 250000 } }
                    - { $zoom: [14], area: { min: 20000 }, name: true }
                    - { $zoom: [14], area: { min: 20000 }, kind: true }
                    - { $zoom: [14], area: { min: 5000 }, landuse_kind: true }
                    - { $zoom: [14], height: { min: 150 } }
                    - { $zoom: [14], volume: { min: 150000 } }
                    - { $zoom: [15], area: { min: 5000 }, name: true }
                    - { $zoom: [15], area: { min: 5000 }, kind: true }
                    - { $zoom: [15], area: { min: 3000 }, landuse_kind: true }
                    - { $zoom: [15], height: { min: 100 } }
                    - { $zoom: [15], volume: { min: 100000 } }
                    - { $zoom: [16], area: { min: 1000 }, name: true }
                    - { $zoom: [16], area: { min: 1000 }, kind: true }
                    - { $zoom: [16], area: { min: 2000 }, landuse_kind: true }
                    - { $zoom: [16], volume: { min: 50000 } }
                    #- { $zoom: [16], height: { min: 50 } }
                    - { $zoom: { min: 17 } }
                all:
                    - not: { location: underground }
```

bcamper:
There will also be a significant client performance win to boil these down
to pre-calculated size/rank/min_zoom values.

queries.yaml
rmarianski: I always struggle with naming. I meant `exclude` to represent that if a feature had a particular property, it should not be considered for merging. And `ignore` meant to drop those particular properties before considering it. We're `remove`ing it in the sense that those properties will get dropped if we are able to merge, but not we're not absolutely removing those properties if there aren't any collisions.

The current configuration will not consider any feature that has a `name` for merging, and will drop `addr_housenumber` and `addr_street` if it is able to merge multiple features.

rmarianski:
Merging anything with a name is risky, no? Isn't it better to just leave those be?

I think the best way to handle this would be to have a step that ran before merging that assigned this size classification to a new property (which could be zoom level dependent). Then, the merging would inherently respect that when grouping. The area/volume would change, but the classification would remain as a target for styling.

rmarianski:
@nvkelso would this be a good opportunity to start using the [class](https://github.com/mapzen/vector-datasource/pull/851) transforms to add in this ranking?

rmarianski:
Renamed the `ignore` column `drop` in 1c21ea6; that nomenclature seems to be used more in a properties context.

Discussed the `name` issue with @zerebubuth, and dropping it seems to make sense for the purposes of merging. If the building is a poi or landuse, that will take the name anyway. And otherwise, a label point will be generated for it.

@nvkelso , the rules were taken from the scene file, and ported it a `class` property on buildings. This is an integer value because the small/medium/large distinctions won't always apply to the rest of the layers.
Note that the min zoom rules did not change, ie the same buildings that were coming through in the tiles before will continue to do so. To keep the same tangram style behavior, the rules should be:
```yaml
                - { $zoom: [13], class: 1 }
                - { $zoom: [14], class: 2 }
                - { $zoom: [15], class: 3 }
                - { $zoom: [16], class: 4 }
                - { $zoom: { min: 17 } }
```

Alternatively, the min zoom rules can be changed to match, but that would imply fewer buildings. If no styles ever show those buildings though, we might want to consider this change.

Also, do we ever want `location=underground` buildings?

@zerebubuth what do you think?

nvkelso:
Regarding use of `class`, I think that'll be confusing to folks used to CartoCSS and MapboxGL. 

Alternates:

- **rank** - _variations on this used in both Mapbox and Mapzen tiles already (they're probably both getting it from Natural Earth and extending the concept to new features)._ Suggest: `scalerank`.
- **importance**
- **significance**

> Also, do we ever want location=underground buildings?

Only show underground buildings starting at zoom 14, they can be dropped from the earlier zooms. (At zoom 14 we can start to see large transit stations and other underground buildings.)

spreadsheets/class/buildings.csv
```diff
@@ -1 +1,16 @@
-kind,class
+location,area::int,height::float,volume::int,landuse_kind,class
+!underground,>=100000,*,*,*,1
```
nvkelso: Remove underground from this CSV and move that zoom 14+ only logic to the query filter?

nvkelso:
> Discussed the name issue with @zerebubuth, and dropping it seems to make sense for the purposes of merging. If the building is a poi or landuse, that will take the name anyway. And otherwise, a label point will be generated for it.

+1

rmarianski:
I liked using `scalerank` instead of `class` because that was already a concept in ne. Updated in 9529da5.

`location=underground` buildings are excluded at z13 now, and that logic is removed from the scalerank assignment in 0e69e31.

vectordatasource/transform.py
```diff
@@ -2849,6 +2850,126 @@ def merge_features(ctx):
     return layer
 
 
+def quantize_val(val, step):
+    result = int(5 * round(val / float(step)))
```
zerebubuth: Should that `5` be `step`?

vectordatasource/transform.py
```diff
@@ -2849,6 +2850,126 @@ def merge_features(ctx):
...
+        # for merging. area and volume will be re-calculated afterwards
+        props.pop('area', None)
+        props.pop('volume', None)
+
+        if drop:
+            for prop in drop:
+                props.pop(prop, None)
+
+        props.pop('addr_housenumber', None)
+        props.pop('addr_street', None)
```
zerebubuth: These are in the `drop` list as currently configured, and would we want them dropped if not configured?

vectordatasource/transform.py
rmarianski: Doh, yes! Nice catch! :zap:

vectordatasource/transform.py
rmarianski: Probably not. I forgot to update that when I made the list configurable. Good catch!

vectordatasource/transform.py
rmarianski: Updated in b6f069b.

vectordatasource/transform.py
rmarianski: Updated in 73cf3cd.

rmarianski:
Going to merge and deploy to dev.

PR #870, opened 2016-06-21
Update queries/transforms for new padded bounds

rmarianski:
Connects to https://github.com/mapzen/vector-datasource/issues/106
Depends on https://github.com/mapzen/tilequeue/pull/97

The padded bounds are now a dict that are keyed off a geometry type:
point, line, polygon. All queries should now use the type specific for
the geometry they are querying for, and transforms/code should use the
bounds appropriate for their type during processing.


zerebubuth:
LGTM :+1:

Issue #872, opened 2016-06-28
Update tilejson files


rmarianski:
The tilejson metadata files are out of date. We should update them now and ensure that we have all the layers represented.

Longer term we should consider an automated process to auto-generate them so they don't fall out of date.

nvkelso:
Let's split off the automated piece into a separate issue for the backlog & fix the immediate issue now (add missing layers).

nvkelso:
This is on prod now.

PR #873, opened 2016-06-28
Update tilejson layers

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/872

This updates the tilejson files with the immediate obvious changes.

* Add the transit and boundaries layers
* Update the max zoom to 16 for all layers
* Update the earth layer to use `kind` instead of `land`

@nvkelso could you review please? Let me know if there's something else that we should add in at this point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+v0.10.4
+-------
+* **Release date**: 2016-06-28
+* Update tilejson layers. See [#874](https://github.com/tilezen/vector-datasource/issues/872).
+
 v0.10.3
 -------
 * **Release date**: 2016-05-20.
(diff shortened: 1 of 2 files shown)


nvkelso:
Do we need to bump the version number?

> "version" : "1"

nvkelso:
👍 

2 new layers & old layers checkout (and we don't want to include `landuse-labels`).

Good catch on the earth layer kinds!

rmarianski:
> Do we need to bump the version number?

We haven't been updating it in general as changes have come in before, so practically speaking it probably doesn't matter. But technically, we probably should be. What do you think the version should be now?

nvkelso:
1.1

rmarianski:
Updated in 7d576fb.

PR #877, opened 2016-07-07
Update testing documentation

zerebubuth:
Update the documentation for running the tests. It now includes not only information about running the tests against an existing server, but also the "CI style" tests which download and set up the database themselves and the "unit/code" tests which were moved into `vector-datasource` from `Tilestache`.

Also fixed a couple of typos.

@rmarianski, @nvkelso could you review, please?


rmarianski:
:+1:

Issue #852, opened 2016-06-01
Remove landuse-labels layer


nvkelso:
We marked it deprecated a while back. Now that v1.0 is here, let's remove it.

nvkelso:
Verified this is no longer being output.

PR #880, opened 2016-07-07
Remove landuse_labels layer

rmarianski:
I didn't see a way to re-use any of the existing assertions, so I created a new one for this case.


rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/852

zerebubuth:
:+1: looks good to me

Issue #842, opened 2016-05-23
Normalize building layer kind and subkind values


nvkelso:
In preparation for our v1.0 release I've been auditing kind values across all the layers.

Today we look at the value of `building=*` and `building:part=*` in OSM and if it's specified we pass it thru to the `kind` value on building polygons. When it's not specified we default the kind to `building`. 

Proposal:

- Always set `kind` to `building` for buildings and `building` for building parts.
- Set new `kind_detail` value to the OSM value for `building=*` and `building:part=*`.

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/840

nvkelso:
Spot checked a few of these and they are working ⚡

Issue #653, opened 2016-03-31
Unify building parts with building


rmarianski:
In the `buildings` layer, it would be useful to know the building that all the parts were associated with. This can be done by seeing if the `building:part` overlaps a certain amount with any buildings in the
tile.

nvkelso:
This can be tracked directly with the building relation per building part. SF example: [Ferry Building](http://www.openstreetmap.org/relation/6062613).

We added `root_relation_id` for transit stations in v0.9, let's reuse the property & extend the logic for buildings.

rmarianski:
The problem is, at least in nyc, the buildings and their parts aren't always linked with a relation. For example: http://www.openstreetmap.org/way/160967745

zerebubuth:
Yup, the logic required is more similar to #142. We might be able to use the "base" building outline as the "root ID", depends on whether we find that many building parts overlap multiple "base" buildings.

nvkelso:
Related: https://github.com/tilezen/vector-datasource/pull/886

nvkelso:
It's sounding like we should say `root_id` (and rename the transit `root_relation_id` to `root_id`).

nvkelso:
@meetar I remember you were interested in this one? Can you please have a look at `dev` tiles to verify this one this week?

New source url setup to use is:

- `http://tile.dev.mapzen.com/vector/v1/all/{z}/{x}/{y}.mvt`

meetar:
This is great news!

I see some things which make sense here, such as features with a `building_part: yes` tag plus a `root_id` tag pointing to the root feature.

But then I see some others which simply have an `osm_relation: 1` tag instead, such as the Bank of England building at http://tangrams.github.io/explorer/#17.0/51.5143/-0.0890/osm_relation/true:

<img width="571" alt="screen shot 2016-08-26 at 3 38 26 pm" src="https://cloud.githubusercontent.com/assets/459970/18018178/7c7743fe-6ba3-11e6-8762-82e7edc579e1.png">

In our tiles, the inner and outer shapes here seem to be split into two features, correlating to 
https://www.openstreetmap.org/way/4959510 and https://www.openstreetmap.org/relation/553472, but with no `root_id` or other node which would associate them with each other.

Should this kind of situation have been fixed in the changes mentioned above, or is this a separate issue?

rmarianski:
[...]

PR #886, opened 2016-07-08
Normalize building kind and add kind_detail

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/842

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -687,3 +687,202 @@ BEGIN
   RETURN CASE WHEN val IS NULL THEN 'null' ELSE to_json(val) END;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION mz_building_kind_detail(val TEXT)
+RETURNS TEXT AS $$
+DECLARE
+BEGIN
+  RETURN CASE WHEN val IN (
+                  'bangunan',
+                  'building',
+                  'other',
+                  'rumah',
+                  'Rumah',
+                  'Rumah Masyarakat',
+                  'rumah_penduduk',
+                  'true',
+                  'trullo',
+                  'yes'
+              ) THEN NULL
+              WHEN val IN (
+                  'abandoned',
+                  'administrative',
+                  'agricultural',
+                  'airport',
+                  'allotment_house',
+                  'apartments',
+                  'arbour',
+                  'bank',
+                  'barn',
+                  'basilica',
+                  'beach_hut',
+                  'bell_tower',
+                  'boathouse',
+                  'brewery',
+                  'bridge',
+                  'bungalow',
+                  'bunker',
+                  'cabin',
+                  'carport',
+                  'castle',
+                  'cathedral',
+                  'chapel',
+                  'chimney',
+                  'church',
+                  'civic',
+                  'clinic',
+                  'clubhouse',
+                  'collapsed',
+                  'college',
+                  'commercial',
+                  'construction',
+                  'container',
+                  'convent',
+                  'cowshed',
+                  'dam',
+                  'damaged',
+                  'depot',
+                  'destroyed',
+                  'detached',
+                  'disused',
+                  'dormitory',
+                  'duplex',
+                  'factory',
+                  'farm',
+                  'farm_auxiliary',
+                  'fire_station',
+                  'garage',
+                  'garages',
+                  'gazebo',
+                  'ger',
+                  'glasshouse',
+                  'government',
+                  'grandstand',
+                  'greenhouse',
+                  'hangar',
+                  'healthcare',
+                  'hermitage',
+                  'hospital',
+                  'hotel',
+                  'house',
+                  'houseboat',
+                  'hut',
+                  'industrial',
+                  'kindergarten',
+                  'kiosk',
+                  'library',
+                  'mall',
+                  'manor',
+                  'manufacture',
+                  'mobile_home',
+                  'monastery',
+                  'mortuary',
+                  'mosque',
+                  'museum',
+                  'office',
+                  'outbuilding',
+                  'parking',
+                  'pavilion',
+                  'power',
+                  'prison',
+                  'proposed',
+                  'pub',
+                  'public',
+                  'residential',
+                  'restaurant',
+                  'retail',
+                  'roof',
+                  'ruin',
+                  'ruins',
+                  'school',
+                  'semidetached_house',
+                  'service',
+                  'shed',
+                  'shelter',
+                  'shop',
+                  'shrine',
+                  'silo',
+                  'slurry_tank',
+                  'stable',
+                  'stadium',
+                  'static_caravan',
+                  'storage',
+                  'storage_tank',
+                  'store',
+                  'substation',
+                  'summer_cottage',
+                  'summer_house',
+                  'supermarket',
+                  'synagogue',
+                  'tank',
+                  'temple',
+                  'terrace',
+                  'tower',
+                  'train_station',
+                  'transformer_tower',
+                  'transportation',
+                  'university',
+                  'utility',
+                  'veranda',
+                  'warehouse',
+                  'wayside_shrine',
+                  'works'
+              ) THEN val
+              WHEN val = 'barne' THEN 'barn'
+              WHEN val = 'commercial;residential' THEN 'mixed_use'
+              WHEN val = 'constructie' THEN 'construction'
+              WHEN val = 'dwelling_house' THEN 'house'
+              WHEN val = 'education' THEN 'school'
+              WHEN val = 'greenhouse_horticulture' THEN 'greenhouse'
+              WHEN val IN ('apartment', 'flat') THEN 'apartments'
+              WHEN val IN ('houses', 'residences', 'residence', 'perumahan permukiman', 'residentiel1') THEN 'residential'
+              WHEN val IN ('semi_detached', 'semi-detached', 'semi') THEN 'semidetached_house'
+              WHEN val = 'offices' THEN 'office'
+              WHEN val = 'prefab_container' THEN 'container'
+              WHEN val = 'public_building' THEN 'public'
+              WHEN val = 'railway_station' THEN 'train_station'
+              WHEN val = 'roof=permanent' THEN 'roof'
+              WHEN val = 'stables' THEN 'stable'
+              WHEN val = 'static caravan' THEN 'static_caravan'
+              WHEN val = 'station' THEN 'transportation'
+              WHEN val = 'storage tank' THEN 'storage_tank'
+              WHEN val = 'townhome' THEN 'terrace'
+              ELSE NULL END;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION mz_building_part_kind_detail(val TEXT)
+RETURNS TEXT AS $$
+DECLARE
+BEGIN
+  RETURN CASE WHEN val IN ('yes', 'part', 'church:part', 'default') THEN NULL
+              WHEN val IN (
+                'arch',
+                'balcony',
+                'base',
+                'column',
+                'door',
+                'elevator',
+                'entrance',
+                'floor',
+                'hall',
+                'main',
+                'passageway',
+                'pillar',
+                'porch',
+                'ramp',
+                'roof',
+                'room',
+                'steps',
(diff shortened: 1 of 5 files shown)


nvkelso:
LGTM

data/functions.sql
```diff
@@ -687,3 +687,14 @@ BEGIN
...
+RETURNS TEXT AS $$
+DECLARE
+BEGIN
+  RETURN CASE WHEN val = 'yes' THEN NULL
+              WHEN val IN ('house', 'residential', 'garage', 'apartments', 'hut', 'industrial', 'roof', 'shed', 'detached', 'commercial', 'garages', 'terrace', 'school', 'retail', 'farm_auxiliary', 'construction', 'church', 'barn', 'greenhouse', 'service', 'cabin', 'manufacture', 'farm', 'warehouse', 'civic', 'office', 'collapsed', 'university', 'public', 'hangar', 'entrance', 'hospital', 'static_caravan', 'hotel', 'chapel', 'semidetached_house', 'damaged', 'kindergarten', 'transportation', 'train_station', 'dormitory', 'storage_tank', 'trullo', 'stable', 'ger', 'houseboat', 'shop', 'college', 'ruins', 'semi', 'bunker', 'pajaru', 'mosque', 'bungalow', 'slurry_tank', 'supermarket', 'transformer_tower', 'duplex', 'factory', 'cowshed', 'constructie', 'dam', 'agricultural', 'tank', 'shelter', 'dwelling_house', 'temple', 'tower', 'agricultural', 'tank', 'shelter', 'dwelling_house', 'temple', 'tower', 'silo', 'residence', 'pavilion', 'store', 'support', 'allotment_house', 'summer_cottage', 'mobile_home', 'administrative', 'kiosk', 'other', 'boathouse', 'glasshouse', 'storage') THEN val
```
zerebubuth: I wonder if we should exclude the `building=collapsed` and `building=ruins` from inclusion as a `kind:building`, perhaps by excluding them from the definition of `*is_building` in the YAML?

And perhaps also `building=no`, which at the moment would produce `kind:building` but no kind detail, which could be a bit misleading.

zerebubuth: Previously, we would have given a `building=no` a `kind:no`, which might also have been unexpected and confusing.

yaml/buildings.yaml
```diff
@@ -18,6 +18,12 @@ globals:
       roof_orientation: {col: "roof:orientation"}
       addr_housenumber: {col: "addr:housenumber"}
       addr_street: {col: "addr:street"}
+  - &building_kind_detail
+      kind_detail:
+        expr: mz_building_kind_detail(building)
+  - &building_part_kind_detail
+      kind_detail:
+        expr: mz_building_kind_detail("building:part")
```
zerebubuth: What's the overlap between the popular `building:part` values and `building` values? Are there any popular `building:part` values we're missing from the list in `mz_building_kind_detail()`?

data/functions.sql
rmarianski: > And perhaps also building=no, which at the moment would produce kind:building but no kind detail, which could be a bit misleading.

Aren't we already [filtering those out](https://github.com/tilezen/vector-datasource/blob/45af11b772bdd1eec09d4285fbf00883d3334984/yaml/buildings.yaml#L21-L28)? 

> I wonder if we should exclude the building=collapsed and building=ruins from inclusion

Sounds good to me, or even treat them as we do [closed names](https://github.com/tilezen/vector-datasource/issues/483), ie only show them at z17+. @nvkelso what do you think?

yaml/buildings.yaml
rmarianski: Good point, I hadn't considered the different popular values.

Ignoring misspellings, capitalizations, numbers, "y" and "true", here's what's in `building:part` that isn't in `building` in the first 4 pages according to taginfo:

```
apartment
arch
balcony
barracks
base
basement
bell_tower
bridge
bus_station
castle
chimney
church:part
column
corridor
Corridor
default
dome
door
elevator
floor
hall
historic
horizontal
main
museum
offices
parking
part
passageway
pillar
porch
proposed
ramp
room
silos
stairs
stairway
steps
stilobate
technical
tier
vertical
verticalpassage
wall
window
```

Should we just have separate `kind_detail` whitelists for `building` and `building:part`?

data/functions.sql
zerebubuth: > > And perhaps also building=no, which at the moment would produce kind:building but no kind detail, which could be a bit misleading.
>
> Aren't we already filtering those out?

D'oh! You're absolutely right - I hadn't expanded the non-diff parts enough to read that, sorry!

yaml/buildings.yaml
zerebubuth: Some of these (`dome`, `part`, `main`, `floor`) seem like they could be useful for rendering, although I don't know a huge amount about the 2.5D building/part stuff.

Values like `historic` and `proposed` make me a bit nervous - it sounds like they may be applied to buildings which aren't currently present in reality. Might we want to give it a different kind to help make it clearer to someone using the data that it's different and, for example, might overlap with other buildings or features which pre- or post-date the historic / proposed / collapsed / ruined -building?

zerebubuth: Nitpick - there's `Corridor` and `corridor` on the list.

rmarianski: Good catch. While diff'ing, I also realized that I had some dupes in the original building list. Updated in e59ecdc.

data/functions.sql
nvkelso: > I wonder if we should exclude the building=collapsed and building=ruins from inclusion as a kind:building, perhaps by excluding them from the definition of *is_building in the YAML?

If a building is **collapsed** it still has a "footprint" _but we should probably zero out it's height_? I'm thinking of Christchurch earthquake in New Zealand and the HOT work around mapping all those buildings – we'd still want them to show up.

For buildings marked **ruins**, it looks like there is a tiny bit of overlap between [building=ruins](http://wiki.openstreetmap.org/wiki/Tag:building%3Druins) and [historic=ruin](http://wiki.openstreetmap.org/wiki/Tag:historic%3Druins). But they are different concepts. If it were historic I'd expect a POI to be generated, but for these basic buildings I think we'd only generate a label placement and pass thru `ruins` as the `kind_detail`. Ruins can still have height, so I wouldn't mess with that value.

I can see only showing the label position for these at zoom 17+, but the features themselves should be visible like normal based on their area.

yaml/buildings.yaml
nvkelso: I expect our kind details to more closely match what's in our existing kind list for landuse, and maybe a few POI values. This also lines up with what's on the OSM wiki page: http://wiki.openstreetmap.org/wiki/Key:building.

Some of these are sounding more like `roof:shape` which there's regular (every month it seems) interest in adding to tiles though there is currently no Tangram support for it. http://wiki.openstreetmap.org/wiki/Simple_3D_buildings

`proposed` is interesting, we should do what we do for transit stations for that one.

Should building parts have an additional property to store some of these other random values?

bcamper: We already have `roof:shape` in tiles!

http://tangrams.github.io/explorer/#16.5/40.7104/-74.0123/roof_shape/dome

zerebubuth: Hmm.... does that mean someone added `building:part=dome` when they meant `roof:shape=dome`? If so, we should probably ignore the `building:part`. Looks like there's [22 instances](http://taginfo.openstreetmap.org/keys/building%3Apart#values) of `building:part=dome`, so it's likely a typo / mis-tag.

There's so few of these values (other than `yes` & `roof`) that perhaps I was wrong to suggest we include them in the list as well.

zerebubuth:
Sorry - I think my previous comments about `building:part` values were a wild goose chase. Contrary to what I thought, looks like there's not much information there.

:+1: looks good as it is at the moment - no `building:part` changes needed.

rmarianski:
@nvkelso is this ok to merge as is?

There were a couple of related thoughts here on how to update the behavior , but it seems like the tag scenarios appear so infrequently that it might not be worth it. Can we create follow up issues for any specifics that we want to follow up on?

rmarianski:
@nvkelso can you review the list of [values whitelisted](https://github.com/tilezen/vector-datasource/pull/886/files#diff-85235d701f80586d926a4211261dbbc1R697) to appear in `kind_detail`? This list was the first four pages of [taginfo for building](http://taginfo.openstreetmap.org/keys/building#values).

nvkelso:
Generally things 'thing' quoted where your list, unquoted are my additions. But remake using these values.

I'm not 100% sure about the residential stuff. There was also a bunch of Indonesian words that I tried to replace with British English values.

### BUILDING PARTS

Whitelist for `kind_building_part` (new property in addition to `kind` and `kind_details` tags):

**default values:**

_discuss!!!_

yes ('yes')
part ('yes')
church:part ('yes')
default ('yes')

**good unique values:**

balcony
base
column
elevator
entrance
floor
pillar
room
hall
main
passageway
porch
ramp
roof
steps
stilobate
tier
tower
verticalpassage
wall
window
door
arch

**Building part replacements:** 

corridor ('verticalpassage')
Corridor ('verticalpassage')
vertical ('verticalpassage')
vertical_passage ('verticalpassage')
stairs ('steps')
stairway ('steps'

### BUILDING_KIND

**these stand on their own (aren't a kind from landuse/pois)**

'collapsed'
'construction'
'damaged'

abandoned
proposed
destroyed
disused

'residential'
'administrative'
'agricultural'
'allotment_house'
'apartments'
'barn'
'boathouse'
'bungalow'
'bunker'
'cabin'
'chapel'
'church'
'civic'
'college'
'commercial'
'cowshed'
'dam'
'detached'
'dormitory'
'factory'
'farm'
'farm_auxiliary'
'garage'
'garages'
'ger'
'glasshouse'
'greenhouse'
'hangar'
'hospital'
'hotel'
'houseboat'
'hut'
'industrial'
'kindergarten'
'kiosk'
'manufacture'
'mobile_home'
'mosque'
'office'
'pavilion'
'public'
'retail'
'roof'
'ruin'
'ruins'
'school'
'service'
'shed'
'shelter'
'shop'
'silo'
'slurry_tank'
'stable'
'static_caravan'
'storage'
'storage_tank'
'store'
'summer_cottage'
'supermarket'
'tank'
'temple'
'terrace'
'tower'
'train_station'
'transformer_tower'
'transportation'
'university'
'warehouse'

grandstand
carport
veranda
pub
cathedral
wayside_shrine
outbuilding
castle
bridge
beach_hut
utility
government
monastery
gazebo
prison
stadium
container
manor
basilica
restaurant
museum
healthcare
depot
bell_tower
substation
shrine
clubhouse
hermitage
power
summer_house
synagogue
convent
clinic
library
mall
parking
arbour
airport
chimney
bank
mortuary
works
brewery
fire_station

**Building kind replacements:** 

'static caravan' ('static_caravan')
'storage tank' ('storage_tank')
railway_station ('train_station')
stables ('stable')
'perumahan permukiman' ('residential')
barne ('barn')
prefab_container ('container)
public_building ('public')
greenhouse_horticulture ('greenhouse')
semi_detached ('semidetached_house')
flat ('apartments')
houses ('residential')
roof=permanent ('roof')
offices ('office')
residentiel1 ('residential')
semi-detached ('semidetached_house')
residences ('residential')
'residence' ('residential')
apartment ('apartments')
station ('transportation')
'constructie' ('construction')
'semi' ('semidetached_house')
'dwelling_house' ('residential')
'duplex' ('residential')
'semidetached_house' ('residential')
education ('school')
townhome ('residential')
'house' ('residential')

'commercial;residential' ('mixed_use')  -- this is made up


**we'd ignore these:** 

'Rumah Masyarakat' ('yes')
rumah ('yes')
rumah_penduduk ('yes')
Rumah ('yes')
true ('yes')
building ('yes')
bangunan ('yes') - see https://en.wiktionary.org/wiki/bangunan
'trullo' ('yes')
'other' ('yes')

nvkelso:
Complication, when we start showing building parts at zoom 15, some features are tagged as both `building=*` and `building:part=*`. At earlier zooms the "bottom" one should be a `kind:building` and the then at zoom 15+ we'd call it a `building_part`.

The area and volume calculations should be updated once #653 lands, too, so we don't clip of certain building parts zooming in and out.

rmarianski:
Test failures should be resolved in b5f8cdd.

nvkelso:
New version 💯, ready to merge!

PR #888, opened 2016-07-11
Add contributing doc with detailed examples

nvkelso:
Top part is modified from Pelias document, bottom part is me adding a bunch of notes on common workflow.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,13 +108,13 @@ We'll cover the following topics in the next sections:
 - [Choose an issue to work on](CONTRIBUTING.md#1-choose-an-issue-to-work-on)
 - [Do you work in a local branch](CONTRIBUTING.md#2-create-a-new-branch)
 - [Create a new test](CONTRIBUTING.md#3-create-a-new-test)
-- [Edit database &/or query logic](CONTRIBUTING.md#create-a-new-test)
-- [Verify the new logic by running the test](CONTRIBUTING.md#create-a-new-test)
-- [Perform any modifications, if necessary](CONTRIBUTING.md#create-a-new-test)
-- [Specify data migrations](CONTRIBUTING.md#create-a-new-test)
-- [Update documentation](CONTRIBUTING.md#create-a-new-test)
-- [Push your local branch to the server](CONTRIBUTING.md#create-a-new-test)
-- [Submit a Pull Request (PR)](CONTRIBUTING.md#create-a-new-test)
+- [Edit database &/or query logic](CONTRIBUTING.md#4-edit-database-&/or-query-logic)
+- [Verify the new logic by running the test](CONTRIBUTING.md#5-verify-the-new-logic-by-running-the-test)
+- [Perform any modifications, if necessary](CONTRIBUTING.md#6-perform-any-modifications,-as-necessary)
+- [Update data migrations](CONTRIBUTING.md#7-update-data-migrations)
+- [Update documentation](CONTRIBUTING.md#8-update-documentation)
+- [Push your local branch to the server](CONTRIBUTING.md#9-push-your-local-branch-to-the-server)
+- [Submit a Pull Request (PR)](CONTRIBUTING.md#10-submit-a-Pull-Request-PR)
 
 ### 1. Choose an issue to work on
 
@@ -206,7 +206,7 @@ Sometimes you'll find a feature in OverPass that is more recent than your local 
 
     ./test-data-update-osm.sh https://www.openstreetmap.org/node/418185265 osm
 
-### Edit database &/or query logic
+### 4. Edit database &/or query logic
 
 Tk tk tk intro
 
@@ -254,7 +254,7 @@ UPDATE planet_osm_line
     way && ST_Transform(ST_SetSrid(ST_MakeBox2D(ST_MakePoint(-124.03564453125, 36.59788913307022), ST_MakePoint(-117.333984375, 39.06184913429154)), 4326), 900913);
 ```
 
-### Verify the new logic by running the test
+### 5. Verify the new logic by running the test
 
 This step is not necessary if only database properties were changed.
 
@@ -302,11 +302,11 @@ python tileserver/__init__.py config.yaml
 
 Then run your test like in the previous step.
 
-### Perform any modifications, as necessary
+### 6. Perform any modifications, as necessary
 
 Rinse and repeat, rewrite your code.
 
-### Update data migrations
+### 7. Update data migrations
 
 Once you've finished testing your new database logic in the steps above you need to record that that same SQL in modified form in `data/migrations/` to ensure someone with an earlier database can catch up with you. (They are reset for each release.)
 
@@ -374,7 +374,7 @@ UPDATE planet_osm_polygon
   WHERE shop IN ('outdoor');
 ```
 
-### Update documentation
+### 8. Update documentation
 
 Everything good? time to update the docs!
 
@@ -382,7 +382,7 @@ Project documentation is [publicly accessable](https://mapzen.com/documentation/
 
 Tk tk tk
 
-### Push your local branch to the server
+### 9. Push your local branch to the server
 
 First let's commit our changes. Let's confirm which files changed:
 
@@ -422,7 +422,7 @@ NOTE: Your first push for a branch might require additional details:
 
     git push --set-upstream origin olga/713-urban-areas
 
-### Submit a Pull Request (PR)
+### 10. Submit a Pull Request (PR)
 
 Back on Github.com load the project page and notice there's a button suggested you create a PR for your active branch. Press that green button.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,19 @@
+<html lang="en-us">
+  <head>
+    <style>
+		.alert-message {
+			color: #8a6d3b;
+			background-color: #fcf8e3;
+			padding: 15px;
+			margin-bottom: 20px;
+			border: 1px solid #faebcc;
+			border-radius: 4px;
+		}
+    </style>
+  </head>
+
+  <body>
+
 **Tilezen** loves contributions from community members like you! Contributions come in many different shapes and sizes. In this file we provide guidance around two of the most common types of contributions: opening issues and opening pull requests.
 
 # Community Values
@@ -55,6 +71,10 @@ Generally speaking there are three aspects of developing vector tiles.
 - Updating database properties (can be done ahead of time or at runtime)
 - Changing how features are selected from the database (requires TileServer restart)
 
+<div class='alert-message'>
+Yellow call-outs like this are meant to draw your attention to an important idea or distinction you should keep in mind.
+</div>
+
 ### vector-datasource
 
 Map database in Postgres stores data from OpenStreetMap and other projects like Natural Earth and Who's On First.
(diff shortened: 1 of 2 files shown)


nvkelso:
Thanks @irealva for the yellow highlight CSS and design principle!

nvkelso:
Ready for review!

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+git checkout master
+```
+
+Ensure your `master` branch is up-to-date with the server:
+
+```bash
+git pull
+```
+
+Then checkout a new branch by naming it something descriptive like the `yourname/#issue-issue title`.
```
zerebubuth: Extra "the"?

Perhaps this would be clearer if it were separated into two sentences? For example:

> You will need a descriptive name for your new branch, and one way to do that is concatenate your user name, any relevant issue number(s) and a brief description, for example `olga/875-camp-ground-zoom`. Then create a new branch using `git checkout -b branchname`, like so:

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+git checkout -b olga/875-camp-ground-zoom
+```
+
+### 3. Create a new test
+
+Create a new test for the issue in `integration-test` dir. Sometimes it's helpful to look thru the existing tests to find one that is a close match to the pattern and start there.
+
+- Create new test file
+- You'll need a specific OpenStreetMap feature ID to test against
+- You'll need a specific map tile (0/0/0) to test with
```
zerebubuth: Might this make more sense as `z/x/y` rather than `0/0/0`? Would it be clearer to say "You'll need the coordinates (`z/x/y`) of a map tile containing that specific feature"?

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+
+### 6. Perform any modifications, as necessary
+
+Rinse and repeat, rewrite your code. Don't be afraid to ask for help!
+
+### 7. Update data migrations
+
+Once you've finished testing your new database logic in step 4 above you need to record that that same SQL in modified form in `data/migrations/` to ensure someone with an earlier database configuration can catch up with you. (Migrations are reset for each Tilezen release.)
+
+Continuing the `camp_site` example, edit the following in the `data/migrations/v1.0.0-polygon.sql` file:
```
zerebubuth: Should we mention the version number in the name of the file? It's `v1.0.0` at the moment, but that will change. Would it be better than we (remember to) update this file, or put some extra text in here to say that the files are prefixed with the _upcoming_ version?

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+- Updating database properties (can be done ahead of time or at runtime)
+- Changing how features are selected from the database (requires TileServer restart)
+
+<div class='alert-message'>
+Yellow call-outs like this are meant to draw your attention to an important idea or distinction you should keep in mind.
+</div>
+
+Map database in Postgres stores data from OpenStreetMap and other projects like Natural Earth and Who's On First.
+
+When data is loaded, database "triggers" calculate if a feature is included in which layer(s), at what "minimum zoom", and other Mapzen specific "mz" properties.
```
zerebubuth: I think it might be helpful to link "triggers" to https://www.postgresql.org/docs/current/static/triggers.html - and anywhere else we're using a term that's possibly new to contributors. I hope terms like "tile" or "feature" would be well-known, but perhaps there are things which aren't so universally known that people might stumble over?

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+- Actual outcome
+- Screenshots, if applicable
+- Code that exposes the bug, if you have it (such as a failing test or a basic script)
+
+## Submitting issues around vector tile quality
+
+It's important to get feedback about the quality of local tile results. Your local knowledge will make it easier for us to understand the problem. When submitting issues be sure to include:
+
+- Where in the world you were looking (a placename &/or zoom, latitude, longitude)
+- Your tile query
```
rmarianski: Is this the tile coordinate? Or tile url? Knowing the layer can be helpful in some circumstances too.

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+## Project standards overview
+
+Tilezen has several miscellaneous standards:
+
+- we follow [PEP8](https://www.python.org/dev/peps/pep-0008/) coding style for Python and use [Flake8](http://flake8.pycqa.org/en/latest/) to enforce those conventions
+- we *love* tests, [check them out](https://github.com/tilezen/vector-datasource/tree/master/integration-test)
+- we use [CircleCI](https://circleci.com/gh/mapzen/vector-datasource) for continuous integration testing
+- we use [semver](http://semver.org/) for package versioning
+
+All  unit tests in a project will be automatically invoked when you commit to an existing project; make sure they exit successfully!
```
rmarianski: Nitpick on spacing "All[ ][ ] unit".

And it sounds a little more natural (to me) to phrase this as: "make sure they pass".

CONTRIBUTING.md
rmarianski: Agreed on linking to triggers here. It can be hard to find where the pieces are if unfamiliar with the current layout.

About tile and feature, we might want to introduce those terms and define them earlier on, probably as the first part of the overview or maybe a separate section before it?

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+
+When data is loaded, database "triggers" calculate if a feature is included in which layer(s), at what "minimum zoom", and other Mapzen specific "mz" properties.
+
+When modifying the logic below, we'll need to update our Postgres functions, migrate the data, and cut new tiles.
+
+### Changing tile content in the vector-datasource repo
+
+Most content changes (eg: adding a new kind of feature) only require a database modification. Content changes are configured using YAML files which specify which database features are "filtered" and outputed in tiles. The location for these content filters is in the `yaml/` directory, for example: [pois.yaml](yaml/pois.yaml).
+
+Some preexisting feature filters are configured using an older raw SQL format in `jinja` files in the `queries/` directory. This older syntax is still helpful to select multiple feature properties across many kinds of features at once. For example, [pois.jinja](queries/pois.jinja).
```
rmarianski: Spell "pre-existing".

We might want to take a moment first to describe the pipeline in general and how everything fits together more. The yaml files help manage filtering and typical property value rules, especially the feature's `kind`. These get folded into the sql, which is generated via jinja templates. And then we have python processing that occurs per feature and across layers, which is useful for more involved logic.

I wouldn't quite frame it as jinja is old and yaml is new, but more that yaml is for the typical differences between features in a layer, and the jinja is better suited for rules that apply to all features in a layer.
But generally speaking, I'd expect most typical contributions or updates to occur at the yaml level.

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+
+Additionally, the root `queries.yaml` specifies which `jinja` file to use per layer, and also specifies post-processing via Python transforms.
+
+### Serving tiles in the TileServer repo
+
+Listens for API requests on localhost, which are in the format of 0/0/0.ext
+
+When TileServer hears a request it asks Postgres for "the stuff" inside that tile's bounding box, configured via the `queries.yaml` file and `.jinja2` files.
+
+_**NOTE:** A change to one of the query files (jinja) requires TileServer to be restarted so they can be reloaded. But content filter changes (YAML) generally doesn't require restarting TileServer._
```
rmarianski: Updates to the jinja don't require restarting tileserver; they are re-read on request during development. And the dev server will automatically detect any python changes and restart itself too. The only changes that require restarts are those to the configuration files themselves, like queries.yaml.

It's probably worth mentioning that changes to the layer yaml files will require at a minimum reloading the sql functions. This is sufficient if only the `kind` or any output properties have changed. For `min_zoom` updates, affected features will need to be recalculated, probably via a migration. But tileserver doesn't need to get restarted in these circumstances.

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+
+```bash
+python integration-test.py local integration-test/875-camp-grounds-zoom.py
+```
+
+Once it fails, we'll update our logic in step 4 below so it passes.
+
+<div class='alert-message'>So what's happening here? The <code>integration-test.py</code> script is asking TileServer for that specific tile to test with on your `local` machine. But before that runs, we're setting up a temporary database to load the specified OpenStreetMap feature into. Once the tile is received, we run the python based test, in this example that's <code>160-motorway-junctions.py</code> in the <code>integration-test</code> directory.</div>
+
+Not the gory details...
```
rmarianski: "Now" the gory details?

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+FAILED 1 TESTS. For more information, see 'test.log'
+```
+
+You can investigate why the test failed by printing out the full debug:
+
+```bash
+cat test.log
+```
+
+<div class='alert-message'>NOTE: It's best practice to run your own test AND to also confirm that all other tests are still passing before submitting a pull request. It's possible that you might need to run an overall database migration to achive this locally, or you can rely on CircleCI to run all the tests for you in your branch by pushing it to the server.</div>
```
rmarianski: Spelling: "achieve" this locally.

CONTRIBUTING.md
```diff
@@ -0,0 +1,586 @@
...
+
+#### Some tests require tileserver restart
+
+A minority of issues will require updating the `queries.yaml` file. In those cases you'll also need to restart TileServer to reload this file (and related `jinja` files).
```
rmarianski: Just for `queries.yaml`.

okavvada:
Will there actually be "yellow call-outs"? they are currently showing up as NOTES.

:+1:

CONTRIBUTING.md
nvkelso: Fixed in 359cce4fd1ec37f907ae8e86171e46f31de014b0.

CONTRIBUTING.md
nvkelso: Fixed in ace1c76b4c99fa85e25e6a55bb7b026bf09533c5.

CONTRIBUTING.md
nvkelso: Fixed in af86b3626dab5450cc6b513560e0a17ee9070fe1. I added extra text to notify the file name will change based on upcoming release version.

CONTRIBUTING.md
nvkelso: Addressed in e98e7b890fd4cdbb47375bc031025b6f1b12271b, linking triggers and adding COMMON TERMS yellow highlight section.

CONTRIBUTING.md
nvkelso: Fixed in 43683f0903326996c989e650dfdbb59517eb0178.

CONTRIBUTING.md
nvkelso: Fixed in 1c9cd6388b6e65202ae373683f09cc0759ec1012 and 62e003887ffaaa0d3705b3ca4a8af84cddc85b21.

CONTRIBUTING.md
nvkelso: Fixed in 8c3f7288297392f7cdfc2a9225f9db5877fd27da.

CONTRIBUTING.md
nvkelso: Fixed in a5c7bede26ddc4a4ffbe34157d9a2699e121db0c.

CONTRIBUTING.md
nvkelso: Fixed in 2791b9be246b133328a65230a8b9a19f77de5e15.

CONTRIBUTING.md
nvkelso: Fixed in 38c592a31f6d02bd0ccdf6ca3a483ebd12d2e61b.

CONTRIBUTING.md
nvkelso: Fixed in 05a63bdbf5d6fd233d727c812c97fb109b192c68.

nvkelso:
Weeee!

Issue #857, opened 2016-06-09
Move barriers into landuse layer?


nvkelso:
The following kinds are in the boundaries layer, should they move to the landuse layer instead?

- city_wall
- dam
- fence
- retaining_wall
- snow_fence

nvkelso:
Verified on dev.

PR #889, opened 2016-07-12
Olga/857 move barriers to landuse

okavvada:
Connects to #857
Moved `city_wall`, `citywalls`, `dam`, `fence`, `retaining_wall`, `snow_fence` from boundaries.yaml to landuse.yaml.
Created test file to check if they exist in landuse and do not exist in boundaries.

- [X] Update tests
- [X] Update docs


docs/layers.md
```diff
@@ -255,7 +250,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 
 * `name`
 * `id`: osm_id
-* `kind`: combination of the `landuse`, `leisure`, `natural`, `highway`, `aeroway`, `amenity`, `tourism`, `zoo`, `attraction`, `man_made`, `power`, and `boundary` OSM tags, or `urban_area` for Natural Earth features.
+* `kind`: combination of the `landuse`, `leisure`, `natural`, `highway`, `aeroway`, `amenity`, `tourism`, `zoo`, `attraction`, `man_made`, `power`, and `boundary` OSM tags, or `urban_area` for Natural Earth features. Also includes of some barrier and man_made tags: `city_wall` (zoom 12+), `retaining_wall`, `snow_fence` (zoom 15+), and `fence` (zoom 16+ only).
```
zerebubuth: Should `dam` be included here, and in the list below, as well?

yaml/landuse.yaml
```diff
@@ -305,7 +305,18 @@ filters:
...
+          GeometryType(way) IN ('LINESTRING', 'MULTILINESTRING')
+    min_zoom: 12
+    output: {kind: dam}
+  - filter: 
+      waterway: dam
+      way:
+        expr: |
+          GeometryType(way) IN ('POLYGON', 'MULTIPOLYGON')
     min_zoom: GREATEST(LEAST(zoom, 16), 9)
     output: {kind: dam}
```
rmarianski: I mentioned to use 2 rules the other day, but seeing this again with fresh eyes, do you think folding them together and having the condition worked out in the `expr` is clearer?

```yaml
  - filter: {waterway: dam}
    min_zoom: |
      CASE WHEN GeometryType(way) IN ('LINESTRING', 'MULTILINESTRING') THEN 12
           WHEN GeometryType(way) IN ('POLYGON', 'MULTIPOLYGON') THEN GREATEST(LEAST(zoom, 16), 9) END
    output: {kind: dam}
```

nvkelso: Yes, the combined expression clarifies the logic. Please update.

docs/layers.md
nvkelso: Good catch!

zerebubuth:
Looks great! :+1:

rmarianski:
:+1:

rmarianski:
The earlier commits had some integration test failures that look like they might be related. Do those still need to get resolved?

zerebubuth:
Oh, looks like perhaps we're applying this to a table (or subquery) where there's no `way` column?

```
psql:apply-planet_osm_polygon.sql:70: ERROR:  column "way" does not exist
LINE 99:   WHEN GeometryType(way) IN ('LINESTRING', 'MULTILINESTRING'...
```

rmarianski:
We probably need to add a `extra_columns: [way]` on the rule.

okavvada:
```
[ 111/111] PASS: 'integration-test/857-move_barriers_to_landuse.py'
PASSED ALL TESTS.
SUCCESS
```

```
find . -not -path '*/.eggs/*' -not -path '*/integration-test*' -not -path '*/data/*' -name '*.py' | xargs flake8 returned exit code 123
```
Some good news and some bad news. @zerebubuth @rmarianski , help?

queries.yaml
```diff
@@ -177,6 +177,23 @@ post_process:
...
+  - fn: vectordatasource.transform.build_fence
+    params:
+      base_layer: landuse
+      start_zoom: 16
+      prop_transform:
+        kind: true
+        id: true
+        source: true
+        area: true
+        min_zoom: true
```
rmarianski: Is there a difference between this list of properties and the source feature? If not, would it be better to just make a copy of all the source properties instead of specifying them here? That way if we add a property to landuse we don't have to remember to add it in here.

queries.yaml
```diff
@@ -177,6 +177,23 @@ post_process:
...
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: landuse
+      start_zoom: 16
+      #end_zoom: 20
+      properties: [barrier]
+
```
rmarianski: We don't actually need this transform, do we? Looks like we always drop the barrier property as part of the build_fence transform.

nvkelso: The earlier transform doesn't drop the barrier tag, so it needs to be dropped here (and for all features, not just fence features?

These could be on features at zoomed earlier than 16, change here to be 0 or the jinja to add an if macro?

queries.yaml
nvkelso: For the feature we are testing against (zoo enclosure) this results in the amenity tag being dropped so fence only has basic props.

queries.yaml
rmarianski: `barrier = props.pop('barrier', None)` drops the barrier tag.

If we updated the `start_zoom` to 0 or whenever landuse started, then yes it would make a difference. But as it stands now it's configured for z16 anyway.

queries.yaml
rmarianski: Ah, so we are selecting a smaller whitelist anyway on purpose. Sounds good.

rmarianski:
:zap: Nice work getting all the details sorted through here! :+1:

Issue #844, opened 2016-05-23
Cleanup POI kind filters (remove dups, rename, and remove totally)


nvkelso:
The following `kind` filters are specified 2 or more times. We should remove the later instances:

- `amenity=boat_storage`
- `shop=dry_cleaning`
- `leisure=firepit`
- `craft=sawmill`
- `craft=sawmill` (yes, it's in there 3 times!!!)
- `amenity=theatre`
- `amenity=toilets`
- `craft=winery`
- `craft=brewery`

The following `kind` filters specify should specify a different output `kind`:

- `shop=gas` should output `kind:gas_canister` (was `gas`)
- `aeroway=gate` should output `kind:aeroway_gate` (was `gate`)
- `rental=ski` should output `kind:ski_rental` (was `ski`)
- `iwn_ref=TRUE`  should output `kind:walking_junction` with `kind_detail:iwn` (was `kind_detail:rwn`)
- `landuse=wood` should  output `kind:wood_landuse` to differentiate it from the `natural=wood` value. This has caught us up several times (see Stanley Park in Vancouver and at Mt Diablo east of SF). Or maybe we should just normalize it to `kind:park` since that seems to be how it is used? 
- ¿`kind=park` is also ambiguous from the `landuse` or `natural` tags. But in this case I think it's ok to coalesce them?

The following `kind` filters specify a OSM feature that is essentially non-existent and we should remove it: 

- `highway=gate` (we already support `barrier=gate` which is the recommended usage)
- `amenity=picnic_site` (we already support several other ways of specifying this)

nvkelso:
* `landuse=wood` = `kind:wood`
* `natural=wood` = `kind:natural_wood`
* `landuse=forest` = `kind:forest`
* `natural=forest` = `kind:natural_forest`
* `landuse=park` = `kind:park`
* `natural=park` = `kind:natural_park` – this is probably bogus, research and consider blacklisting.

rmarianski:
@nvkelso couple of questions that came up while I was implementing:

1. You mentioned `kind_detail` for the iwn change above. The yaml currently calls it `walking_network`. Should I replace `walking_network` -> `kind_detail` too as part of this pr?

2. For the landuse/natural updates, there are some instances where *both* are set, eg `natural=wood` *and* `landuse=wood`. Which should take precedence?

nvkelso:
[...]

PR #893, opened 2016-07-13
Cleanup poi kind filters

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/844

### DIFF
--- a/integration-test/454-aeroway-gates.py
+++ b/integration-test/454-aeroway-gates.py
@@ -2,4 +2,4 @@
 #https://www.openstreetmap.org/node/656398641
 assert_has_feature(
     16, 10487, 25368, 'pois',
-    { 'kind': 'gate' })
+    { 'kind': 'aeroway_gate' })
--- a/integration-test/674-outdoor-shops.py
+++ b/integration-test/674-outdoor-shops.py
@@ -40,7 +40,7 @@ assert_has_feature(
 #http://www.openstreetmap.org/node/2135237099
 assert_has_feature(
     16, 10558, 25443, 'pois',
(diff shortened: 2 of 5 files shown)


nvkelso:
👍

zerebubuth:
There are a couple of nodes tagged `aeroway=gate, barrier=gate` ([here](http://www.openstreetmap.org/node/2618197593) and [here](http://www.openstreetmap.org/node/877193500)). At the moment, the check for `barrier=gate` comes before `aeroway=gate`, so these will get `kind:gate` (although with an additional `aeroway:` property).

It seems to me that `aeroway=gate` is more "specific"... Should we put that above the `barrier=gate` check and remove the extra `aeroway:` property from the `barrier=gate` check? (I'm not seeing anything that's `barrier=gate` but `aeroway != gate`)

zerebubuth:
Upon closer inspection of the examples - neither are traditional passenger en-planing points. One is a [dock-side gate for Victoria Seaplane Terminal](http://www.openstreetmap.org/node/2618197593) and the other seems to be a [gate on an access taxiway](http://www.openstreetmap.org/node/877193500) to a DLR research base from a ["special" airport](http://www.edmo-airport.de/en) which doesn't seem to do commercial flights.

So perhaps it's better to give both `kind:gate`, although I still think it might be worth taking off the additional `aeroway:` property.

rmarianski:
Good points. I added the first example as a test case, and removed the `aeroway` tag in 0488d5f.

zerebubuth:
:+1: looks good

nvkelso:
👍

Issue #841, opened 2016-05-23
Normalize boundaries layer kind & kind_detail values


nvkelso:
In preparation for our v1.0 release I've been auditing kind values across all the layers.

1) The `boundaries` layer need some normalizing to `region` for administrative subdivisions and `locality` for city like things:

```
tile_layer	tile_kind	tile_kind_detail	osm_key	osm_value	old_tile_kind
boundaries	region	4	admin_level	4	state
boundaries	locality	8	admin_level	8	municipality
```

2) We'll use the new `kind_detail` property to store the admin_level from OSM and that we sometimes calculate. Set the `kind_detail` value of the following to `2` as they are "country" related:

- disputed
- indefinite
- indeterminate
- country
- lease_limit
- line_of_control
- overlay_limit

**Related:** `places` layer kinds https://github.com/mapzen/vector-datasource/issues/840.

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/842

rmarianski:
Are we removing `admin_level` throughout and naming it `kind_detail`?

nvkelso:
Yes, putting the `admin_level` into `kind_detail` and then dropping `admin_level`.

rmarianski:
@nvkelso what should the `kind_detail` be for `macroregion`? `2`?

nvkelso:
Macroregion is between country and region so `3`?

nvkelso:
Verified.

PR #901, opened 2016-07-18
Normalize boundaries kind

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/841

### DIFF
--- a/yaml/boundaries.yaml
+++ b/yaml/boundaries.yaml
@@ -1,75 +1,77 @@
 filters:
+
+  # osm
   - filter:
-      admin_level: ['2', '4']
-      boundary: administrative
-      'tags->boundary:type': aboriginal_lands
+      any:
+        all:
+          boundary: administrative
+          'tags->boundary:type': aboriginal_lands
+        boundary: aboriginal_lands
     min_zoom: 8
-    output: {kind: aboriginal_lands, admin_level: {col: admin_level}}
-    table: osm
-  - filter:
-      admin_level: ['6', '8']
-      boundary: administrative
-      'tags->boundary:type': aboriginal_lands
-    min_zoom: 10
-    output: {kind: aboriginal_lands, admin_level: {col: admin_level}}
+    output:
+      kind: aboriginal_lands
+      kind_detail:
+        expr: CASE WHEN admin_level IN ('2', '4', '6', '8') THEN admin_level END
     table: osm
   - filter: {admin_level: '2', boundary: administrative}
     min_zoom: 8
-    output: {kind: country, admin_level: {col: admin_level}}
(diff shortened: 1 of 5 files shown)


yaml/boundaries.yaml
```diff
@@ -1,75 +1,79 @@
...
-    output: {kind: aboriginal_lands, admin_level: {col: admin_level}}
+    output: {kind: aboriginal_lands, kind_detail: {col: admin_level}}
     table: osm
   - filter:
       admin_level: ['6', '8']
       boundary: administrative
       'tags->boundary:type': aboriginal_lands
     min_zoom: 10
-    output: {kind: aboriginal_lands, admin_level: {col: admin_level}}
+    output: {kind: aboriginal_lands, kind_detail: {col: admin_level}}
```
nvkelso: Is this transformed later on? Should we set it explicitly here to a value like `3`?

yaml/boundaries.yaml
```diff
@@ -1,75 +1,79 @@
 filters:
+
+  # osm
   - filter:
       admin_level: ['2', '4']
       boundary: administrative
       'tags->boundary:type': aboriginal_lands
     min_zoom: 8
-    output: {kind: aboriginal_lands, admin_level: {col: admin_level}}
+    output: {kind: aboriginal_lands, kind_detail: {col: admin_level}}
```
nvkelso: Is this transformed later on? Should we set it explicitly here to a value like `3`?

rmarianski: We used to have a transform in python, but that got removed when we moved to the yaml syntax. I updated this to set the value explicitly to 3 in 4958518.

yaml/boundaries.yaml
```diff
@@ -1,75 +1,79 @@
 filters:
+
+  # osm
   - filter:
       admin_level: ['2', '4']
       boundary: administrative
       'tags->boundary:type': aboriginal_lands
     min_zoom: 8
-    output: {kind: aboriginal_lands, admin_level: {col: admin_level}}
+    output: {kind: aboriginal_lands, kind_detail: '3'}
```
nvkelso: Were we sending thru `2` or `4` or `6` or `8` before depending on the input?

rmarianski: 2 or 4 for the above, and 6 or 8 for the below, although I couldn't find any actual data that matched the below.

rmarianski: Updated to use the other naming scheme for aboriginal lands, namely `boundary=aboriginal_lands` in 17c3774.

nvkelso:
(So merge the two filters, so all are result is min_zoom of 8.)

rmarianski:
Merged in 6c32f2c.

PR #902, opened 2016-07-18
Update test building id

rmarianski:
The test building id we were using for the height test got updated recently.


nvkelso:
LGTM except the one remaining test failure. 

Maybe close this one in favor of https://github.com/tilezen/vector-datasource/pull/903/files which also addresses this same building case?

Issue #875, opened 2016-06-29
Camp grounds (camp_site) with areas should show up early


nvkelso:
* **What did you see?** Only at zoom 16 now for Fallen Leaf Campground near Tahoe.
* **What did you expect to see?** Up to zoom 13 if they are big
* **What map location are you having problems with?** 16.86866/38.92667/-120.05147
* **Screenshot?** 
<img width="821" alt="screen shot 2016-06-29 at 14 53 55" src="https://cloud.githubusercontent.com/assets/853051/16471407/5d603012-3e11-11e6-9b0f-2f80f4fa7bb1.png">

nvkelso:
Good news: Zoom ranges are working!

Bad news: The sort order is defaulting to 11, it should be much higher.

okavvada:
That is because they are not specified in the sort key spreadsheet. What should they be set to?

nvkelso:
Between these to (so it gets 92, and everything maze and above gets +1):

```
maze,*,92
generator,*,91
```

nvkelso:
Woot!

![image](https://cloud.githubusercontent.com/assets/853051/17960716/9b286186-6a5c-11e6-9d9f-62cf91418055.png)

And the node only features (here Angle Island) are still only shown later:

![image](https://cloud.githubusercontent.com/assets/853051/17960754/d488263c-6a5c-11e6-9299-23c26107a054.png)

PR #904, opened 2016-07-19
Olga/875 camp ground zoom

okavvada:
Connects to #875. 

- [x] Update tests
- [x] Update docs

### DIFF
--- a/integration-test/875-camp-grounds-zoom.py
+++ b/integration-test/875-camp-grounds-zoom.py
@@ -1,0 +1,44 @@
+# update landuse to include campground features and fix label zoom
+
+# large campground in landuse zoom 16
+# http://www.openstreetmap.org/way/431725967
+assert_has_feature(
+    16, 10959, 25337, 'landuse',
+    { 'kind': 'camp_site'})
+
+# large campground in point zoom 16
+assert_has_feature(
+    16, 10959, 25337, 'pois',
+    { 'kind': 'camp_site'})
+
+# large campground in landuse zoom 13
+
+assert_has_feature(
+    13, 1369, 3167, 'landuse',
+    { 'kind': 'camp_site'})
+
+# large campground in point zoom 13
+assert_has_feature(
+    13, 1369, 3167, 'pois',
+    { 'kind': 'camp_site'})
+
+# small campground in landuse zoom 16
+# http://www.openstreetmap.org/way/417405356
+assert_has_feature(
+    16, 10471, 25326, 'landuse',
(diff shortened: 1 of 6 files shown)


integration-test/875-camp-grounds-zoom.py
```diff
@@ -0,0 +1,49 @@
...
+# http://www.openstreetmap.org/way/417405356
+assert_has_feature(
+    13, 1308, 3165, 'landuse',
+    { 'kind': 'camp_site'})
+
+# small campground in point zoom 13
+# http://www.openstreetmap.org/way/417405356
+assert_no_matching_feature(
+    13, 1308, 3165, 'pois',
+    { 'kind': 'camp_site'})
```
rmarianski: As of late I've been adding the `id` to all assertions to prevent matches against other features in the tile, which can be confusing on failures.

rmarianski:
:+1:

integration-test/875-camp-grounds-zoom.py
nvkelso: The danger of using `id` is if someone changes the feature substantially the `id` will drift (like with Mt. Whitney a couple weeks ago). As long as the test feature is isolated by itself if should be okay? (Another case for "unit test island"!)

rmarianski: Yea, I think one approach can be better than the other depending on the situation; just that in the past I've gotten confused by the actual failures without ids. In the log for example, the output is usually not going to help, and you'll have to look at the tile. But with the id the error is usually more obvious.

zerebubuth: Possibly we should be changing the tests? It seems like there's two different use-cases:

1. I want to test that _something_ here has some properties. E.g: highways drop bridge/tunnel at some zoom level. I don't care what the highways are: they should all have this behaviour.
2. I want to test that a specific feature is transformed in a specific way.

The former could be done with a tile coordinate and general matching, whereas the second needs the ID and possibly the zoom level - after we've downloaded the feature, we can figure out the tile coords. This would make us robust to movements of features across tile boundaries.

I'm just thinking out aloud here :wink:

okavvada:
So is the executive decision what @zerebubuth said?

integration-test/875-camp-grounds-zoom.py
okavvada: Tracking this with https://github.com/tilezen/vector-datasource/issues/906

okavvada:
I talked to Nathaniel, ready to merge.

Issue #912, opened 2016-07-20
Zoom 17 missing some building parts


mjcunningham:
Go to: http://tangrams.github.io/bubble-wrap/#18.87804/40.74546/-73.99049

Note bump on southern side of the building below the waldy's woodfire pizza label: 

![screen shot 2016-07-20 at 5 09 49 pm](https://cloud.githubusercontent.com/assets/307758/17003217/db76455e-4e9c-11e6-8fc8-69cb5c90609e.png)

Zoom out once: http://tangrams.github.io/bubble-wrap/#17.87804/40.74546/-73.99049

Note that the bump is gone:
![screen shot 2016-07-20 at 5 11 37 pm](https://cloud.githubusercontent.com/assets/307758/17003246/0af30ad8-4e9d-11e6-98b9-8c3e3c6a4927.png)

Zoom out again: http://tangrams.github.io/bubble-wrap/#16.87804/40.74546/-73.99049

The bump has returned. 
![screen shot 2016-07-20 at 5 12 15 pm](https://cloud.githubusercontent.com/assets/307758/17003268/20202a9e-4e9d-11e6-8261-0c8fc7bf6088.png)

This happens on web and on ES

rmarianski:
This the osm way in question? http://www.openstreetmap.org/way/287494678

Possibly related to https://github.com/tilezen/vector-datasource/issues/653? Although, it does show up in the [tile](https://vector.mapzen.com/osm/buildings/16/19298/24632.topojson) with a `min_zoom` of 16, assuming that is indeed the way.

Not sure why the bump would be disappearing at z17 and appearing at z16. I'd expect it to start showing at z17+ if anything. Could there be a rule on the styling side controlling that @nvkelso?

zerebubuth:
I've added a test for whether the data is present in the tiles, which passes. That suggests that @rmarianski is right and client-side filtering is playing a role.

I think it might be in the [extrusion filtering](https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap.yaml#L2492-L2500), as there's a min-height rule at zoom 16, but no min-height rule at 17. This feature has a height of 111, so is included at 16, but the area and volume rules are not triggered at 17.

I think that's what's happening - if I'm understanding the rules correctly!

rmarianski:
And looking at the requests made out to the server, only requests for z16 are made and then they are "overzoomed".

nvkelso:
From that chunk of code:

```
[...]

PR #913, opened 2016-07-21
Add test for missing building part seen in production.

zerebubuth:
Connects to #912.

This just checks that the building data is present in the tiles, which it appears to be across zooms 18, 17 and 16.


rmarianski:
:+1:

PR #920, opened 2016-07-26
add initial changelog for v1.0.0-pre1

nvkelso:
Another set of release notes, another Markdown style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,17 +8,19 @@ v1.0.0-pre1
 
 - **new url scheme**: Mapzen now offers several different types of tiles in vector and raster formats and we combine data from multiple sources. The URL scheme has been updated to reflect this, and emphasize versions. The old URL will continue to work (~1 year), but updates will stop once v1.0.0 is released to production. ([#652](https://github.com/tilezen/vector-datasource/issues/652))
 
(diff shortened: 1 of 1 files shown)


nvkelso:
The **requires** is definitely not correct! (But waiting for updated tags or SHAs on those.)

nvkelso:
Added commit to:

* update URL section
* remove TileStashe requirement, and note removal of same in Internal Changes section
* add buroughs from WOF
* reformat (#issue_number) on end of each list item for see more details links

nvkelso:
@rmarianski what should I mention in the **requires** section for tileserver and tilequeue?

Other than that I think this is ready to be merged?

rmarianski:
We'll need to make new releases for tileserver and tilequeue as well and point to those.

nvkelso:
@rmarianski what SHAs did you use for the v1.0.0-pre1 release on dev? Should we cut new dev releases or point at the SHA?

rmarianski:
Just pointed to the master branches. We should just cut new releases of those repos.

rmarianski:
@nvkelso I put up some prs with changelogs and version bumps for tilequeue and tileserver. If those look good to you, I'll go ahead and make the tags.

Issue #919, opened 2016-07-25
barrier: gates with line geometry missing


okavvada:
* **What did you see?** At the time `"barrier" = "gate"` features of line geometry are missing
* **What did you expect to see?** They should be added in landuse
* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._   
http://www.openstreetmap.org/way/391260223

nvkelso:
Related: https://github.com/tilezen/vector-datasource/issues/820.

nvkelso:
This is an addition (not a breaking change), moving to v1.1

Issue #820, opened 2016-05-16
Modify gate min zoom to 17


nvkelso:
In working on https://github.com/mapzen/vector-datasource/pull/819 for `kind:toll_booth`, we've discovered the min zoom of `kind:gate` POIs to be very high at zoom 15. That should be defaulting to zoom 17 instead as many gates are digitized in fences, not just across roads. 

@zerebubuth How hard would it be to modify the zoom based on what road class the the gate was on?

A gate on a path &/or fence at zoom 17. But a gate on a major road zoom 14 and minor road zoom 15?

nvkelso:
For context, we hide most gates from display in the map because the data is too funky in tiles now.

zerebubuth:
A query like this will do the trick:

```SQL
SELECT
  osm_id,
  MAX(CASE
    WHEN highway IN ('motorway', 'trunk', 'primary', 'motorway_link',
                     'trunk_link', 'primary_link') THEN 3
    WHEN highway IN ('secondary', 'tertiary', 'secondary_link',
                     'tertiary_link') THEN 2
    WHEN highway IN ('residential', 'unclassified') THEN 1
    ELSE 0
  END) AS highway_level
FROM (
  SELECT
    p.osm_id,
    hstore(w.tags)->'highway' AS highway
  FROM planet_osm_point p
  JOIN planet_osm_ways w
  ON ARRAY[p.osm_id] && w.nodes
  WHERE
    p.way && <insert bbox here>
    AND p.barrier='gate'
    AND hstore(w.tags) ? 'highway'
) x
GROUP BY osm_id;
```

This will have a performance impact but hopefully, since gates are relatively rare, it might not be noticeable. The form above is good for experimentation, but it's probably best for maintainability to rewrite it as a function of `osm_id` returning the appropriate highway level (or zoom level) and then just use that as the `min_zoom` expression in the YAML.

rmarianski:
> The form above is good for experimentation, but it's probably best for maintainability to rewrite it as a function of osm_id returning the appropriate highway level (or zoom level) and then just use that as the min_zoom expression in the YAML.

:+1:

okavvada:
Is this referring to `kind:gate` or `kind:toll_booth` ? Usually gates are not digitized on the road, I couldn't find any examples of a gate on a major highway, secondary road. `kind:toll_booth` are digitized on highways. Should the rule apply to `kind:toll_booth`?

nvkelso:
[...]

PR #921, opened 2016-07-26
Olga/820 gates min zoom

okavvada:
Connects to #820 and issue #919.

- [x] Update tests
- [x] Update data migrations
- [x] Update docs

@rmarianski , @zerebubuth and @nvkelso, review please?

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -250,7 +250,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 
 * `name`
 * `id`: osm_id
-* `kind`: combination of the `landuse`, `leisure`, `natural`, `highway`, `aeroway`, `amenity`, `tourism`, `zoo`, `attraction`, `man_made`, `power`, and `boundary` OSM tags, or `urban_area` for Natural Earth features. Also includes of some `barrier` and `waterway` tags: `city_wall` (zoom 12+), `dam` (zoom 12+), `retaining_wall` (zoom 15+), `snow_fence` (zoom 15+), and `fence` (zoom 16+ only).
+* `kind`: combination of the `landuse`, `leisure`, `natural`, `highway`, `aeroway`, `amenity`, `tourism`, `zoo`, `attraction`, `man_made`, `power`, and `boundary` OSM tags, or `urban_area` for Natural Earth features. Also includes of some `barrier` and `waterway` tags: `city_wall` (zoom 12+), `dam` (zoom 12+), `retaining_wall` (zoom 15+), `snow_fence` (zoom 15+), `fence` (zoom 16+ only) and `gate` (zoom 16+ only).
 * `sort_key`: a suggestion for which order to draw features. The value is an integer where smaller numbers suggest that features should be "behind" features with larger numbers.
 * `area`: in square meters (spherical Mercator, no real-world), `polygon` features only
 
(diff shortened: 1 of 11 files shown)


docs/layers.md
```diff
@@ -250,7 +250,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 
 * `name`
 * `id`: osm_id
-* `kind`: combination of the `landuse`, `leisure`, `natural`, `highway`, `aeroway`, `amenity`, `tourism`, `zoo`, `attraction`, `man_made`, `power`, and `boundary` OSM tags, or `urban_area` for Natural Earth features. Also includes of some `barrier` and `waterway` tags: `city_wall` (zoom 12+), `dam` (zoom 12+), `retaining_wall` (zoom 15+), `snow_fence` (zoom 15+), and `fence` (zoom 16+ only).
+* `kind`: combination of the `landuse`, `leisure`, `natural`, `highway`, `aeroway`, `amenity`, `tourism`, `zoo`, `attraction`, `man_made`, `power`, and `boundary` OSM tags, or `urban_area` for Natural Earth features. Also includes of some `barrier` and `waterway` tags: `city_wall` (zoom 12+), `dam` (zoom 12+), `retaining_wall` (zoom 15+), `snow_fence` (zoom 15+), `fence` (zoom 16+ only) and `gate` (zoom 16+ only).
```
zerebubuth: Nitpick: gates are now zoom 14+

yaml/pois.yaml
```diff
@@ -288,7 +288,11 @@ filters:
     min_zoom: 18
     output: {kind: toilets}
   - filter: {barrier: gate}
-    min_zoom: 15
+    min_zoom: CASE WHEN mz_get_highway_level_point(osm_id) = 3 THEN 14
+        WHEN mz_get_highway_level_point(osm_id) = 2 THEN 15
+        WHEN mz_get_highway_level_point(osm_id) = 1 THEN 16
```
zerebubuth: I worry that this might result in running this function 3 times for most instances, and we're expecting the function to be moderately expensive as it joins back to the `planet_osm_ways` table. Would it be okay to re-write this as `17 - COALESCE(mz_get_highway_level_point(osm_id), 0)`? (The `COALESCE` is just there to make sure a number is used, as the function can sometimes return `NULL`).

docs/layers.md
okavvada: In landuse (geometry lines) they only show up as 16+. The point features are the ones that show up in different zoom levels.

zerebubuth: Ooops, you're absolutely right - I should have checked which section this was in!

yaml/pois.yaml
```diff
@@ -288,7 +288,7 @@ filters:
     min_zoom: 18
     output: {kind: toilets}
   - filter: {barrier: gate}
-    min_zoom: 15
+    min_zoom: {17 - COALESCE(mz_get_highway_level_point(osm_id), 0)}
```
nvkelso: Should we return the **min_zoom** value in the function, instead of needing the `17 - ...` logic? What do we do for the trails logic?

nvkelso: More like: 

* **YAML:** https://github.com/tilezen/vector-datasource/blob/16a1ef908f59b9dd2e9b68d1a96733748b37e57d/yaml/roads.yaml#L225
* **functions.sql:** https://github.com/tilezen/vector-datasource/blob/1150ac3c0e5cf61421f49b2d8a7243617f74376f/data/functions.sql#L624

okavvada: Yes, okay

nvkelso:
Once the CI check is fixed, LGTM

Issue #661, opened 2016-04-04
Don't include historic transit stops


nvkelso:
Looks like there's another use case around historic transit features we need to guard against. 

* **What did you see?** A `kind:stop` feature in NYC for a stop that isn't in service, it's marked `historic=yes` at OSM.org. 
* **What did you expect to see?** This feature shouldn't be included in tiles at all. It hasn't had service in a long time. We should only be showing current transit in our tiles.
* **What map location are you having problems with?** [node 3039734895](https://www.openstreetmap.org/node/3039734895) in NYC at 5th Av and 3rd St.
* **Screenshot?** 

<img width="682" alt="screen shot 2016-04-03 at 21 17 59" src="https://cloud.githubusercontent.com/assets/853051/14238103/9ca23456-f9e1-11e5-949f-173f65a3f038.png">

/cc @bcamper

nvkelso:
Moving to next release for load balancing.

okavvada:
* Does this refer to transit stops only or all transit features?
* `railway: stop` is included in both the `transit.yaml` and the `pois.yaml`. Do we want to _not include_ both layers if historic?
* Is it limited to `historic=yes`? or any `historic` tag value? example from [TagInfo](https://taginfo.openstreetmap.org/keys/historic#values):
`historic = yes`, `historic = railway`, `historic = railway_station`

nvkelso:
* Stops & stations only (generally all points, but not rail lines). We've only had reports of points, so let's keep this issue scoped to that.
* Yes, filter it out in both layers.
* Good TagInfo sleuthing, let's black list all 3 of those values.

okavvada:
No instances of historic tags in the transit layer, so it is not included in the filtering process. The filtering exists for the pois layer for features in railway: [stop, halt, tram_stop] with historic: [yes, railway_station].

nvkelso:
Good sleuthing!

nvkelso:
Erased!

![image](https://cloud.githubusercontent.com/assets/853051/17960551/9a9efb36-6a5b-11e6-8c45-d6bd1e538e55.png)

PR #924, opened 2016-07-27
Olga/661 historic transit stops

okavvada:
Connects to #661. 

- [x] Update tests
- [x] Update data migrations
- [x] Update docs

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -668,7 +668,6 @@ filters:
     output: {kind: picnic_site}
   - filter: {highway: services}
     min_zoom: 11
-    
     output:
       kind: service_area
   - filter: {highway: rest_area}
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -668,6 +668,7 @@ filters:
     output: {kind: picnic_site}
   - filter: {highway: services}
     min_zoom: 11
+    
     output:
       kind: service_area
   - filter: {highway: rest_area}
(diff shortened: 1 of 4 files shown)


zerebubuth:
:+1: looks good to me!

yaml/pois.yaml
```diff
@@ -316,8 +316,10 @@ filters:
     output:
       kind: station
       state: {col: tags->state}
-  - filter:
-      railway: [halt, tram_stop, stop]
+  - filter: 
+      all:
```
rmarianski: Don't need this `all` here, it's implied, but you can keep it if you think it makes it more readable.

yaml/pois.yaml
```diff
@@ -316,8 +316,10 @@ filters:
...
+        railway: [halt, stop, tram_stop]
+        any: {historic: [-yes, -railway_stop], historic: false}
```
rmarianski: Does this work as expected? I'm pretty sure that it wouldn't. Actually, I think a couple of things are going on here:

1. It would translate that first check into `historic IN ('-yes', '-railway')` rather than `historic NOT IN ('yes', 'railway')`
2. Because we're using the same key, `historic` twice, one would probably override the second. I'm guessing that just the `historic: false` rule is the one that would actually get applied.

We can spell it like:

```
  - filter:
      railway: [halt, stop, tram_stop]
      any:
        not:
          historic: ['yes', railway_stop]
        historic: false
```

okavvada: Updated it to 
```
- filter: 
      all:
        railway: [halt, stop, tram_stop]
        any: {historic: 'no', historic: false}
```
@rmarianski is this better?

rmarianski:
:+1:

rmarianski:
LGTM :+1:

PR #937, opened 2016-08-02
major upgrade to tilejson

nvkelso:
For all property keys and add value types, short descriptions, common values, and layer geometry types.

It's probably missing a few things, but this should unblock @louh !

@rmarianski Once the PR is landed, please update `dev` with this new config file.


rmarianski:
:+1:

nvkelso:
Connects to: https://github.com/tilezen/vector-datasource/issues/938.

Issue #742, opened 2016-04-22
More predictable layers for landuse label positions (more pois)


nvkelso:
We currently have a random mix of poi like features in the landuse and the pois layer. We should make this more predictable. 

Generally if it is worth having an icon for it, it should be in the `pois` layer. Things with name only (like building names in the building layer) should stay in the landuse layer.

The landuse polygon should always be in the landuse layer. But this is additionally complicated since a few of these should probably move to the `earth` layer (like grass, meadow, land, glacier, etc; see also the road landuse_kind intercut problems), and some potentially to a new `infrastructure` layer (eg apron, footway, pedestrian). And sometimes a natural area is also a protected area, in which case it should stay in the landuse layer. 

### ~50 `landuse` features have a label position in the `landuse` layer only.

_Some / most of these need filters added in the POIs YAML config file._

* `allotments` - move to `earth` layer?
* `apron` - move to `earth` layer?
* `breakwater` - move to `earth` layer?
* `bridge` - candidate for `pois` layer filter
* `cemetery` - candidate for `pois` layer filter
* `commercial` - move to `earth` layer?
* `common` - move to `earth` layer?
* `cutline` - move to `earth` layer?
* `dike` - move to `earth` layer?
* `farm` - candidate for `pois` layer filter
* `farmland` - move to `earth` layer?
* `farmyard` - move to `infrastructure` layer?
* `footway` - move to `infrastructure` layer?
* `forest` - candidate for `pois` layer filter
* `fuel` - candidate for `pois` layer filter
* `garden` - candidate for `pois` layer filter
* `glacier` - candidate for `pois` layer filter
* `golf_course` - candidate for `pois` layer filter
* `grass` - move to `earth` layer?
* `groyne` - move to `earth` layer?
* `industrial` - move to `earth` layer?
* `land` - move to `earth` layer?
* `meadow` - move to `earth` layer?
* `military` - candidate for `pois` layer filter
* `national_park` - candidate for `pois` layer filter
* `nature_reserve` - candidate for `pois` layer filter
* `park` - candidate for `pois` layer filter
* `park or protected land` - this was is an old Natural Earth value that should probably just be `park`
* `pedestrian` - move to `infrastructure` layer?
* `pier` - move to `infrastructure` layer?
* `pitch` - candidate for `pois` layer filter
* `plant` - candidate for `pois` layer filter
* `protected_area` - candidate for `pois` layer filter
* `quarry` - candidate for `pois` layer filter
[...]

PR #939, opened 2016-08-02
Olga/742 predictable layers pois

okavvada:
Connects to #742.

- [x] Update tests
- [x] Update data migrations
- [x] Update docs

Ready for review, @nvkelso , @rmarianski , @zerebubuth


zerebubuth:
Code changes look good!

The original issue talks about moving label positions to the `pois` layer, but nothing was taken out of the `landuse` layer YAML. I think we are making a lot of those label positions from the polygons in the post-process - do we also need to remove (or not generate) those label positions? Could you add a test to make sure that these aren't now in both layers, please?

nvkelso:
@zerebubuth Good catch! 

My intent was to stop generating label positions in the **landuse** layer if the "label" was already in the **pois** layer (keep the landuse polygons, ditch the label placement). Is that already configured or does new logic need to be added (and where)?

There are still a few dozen landuse kinds that aren't in pois, those should get label placements but only starting at maybe zoom 13?

zerebubuth:
I think it _should_ already have the logic to [set the landuse name to NULL if the feature is a POI](https://github.com/tilezen/vector-datasource/blob/dec70254a1c26a0fb9daa32758634be46e5bab29/queries/landuse.jinja2#L14), but it's pretty complicated and I'm not totally sure I understand it. Just to be on the safe side, seems worth adding some tests to assert that there are no matching features (with `label_placement:yes` - or is it `true` now?) in the `landuse` layer.

okavvada:
@zerebubuth , I added the tests for label placement. Can you please check to make sure this is what you were thinking of?

zerebubuth:
Thanks! That looks great! :+1:

nvkelso:
@okavvada the PR looks good now, but let's still test on my local dev config before merging – this way we can see the impact on overall map balance first.

okavvada:
okay @nvkelso

okavvada:
Updated all the min_zoom levels we talked about with @nvkelso. Let me know if its ready to merge

Issue #946, opened 2016-08-05
Remove -W UTF-8 from data import config


nvkelso:
`-W UTF-8` is probably not needed here: https://github.com/tilezen/vector-datasource/blob/master/data/Makefile-import-data.jinja2#L9

But might be version dependent, @zerebubuth to investigate and add details.

zerebubuth:
I checked v2.1.2 (used in Ubuntu trusty) and v2.1.8 on my local machine, neither of which had problems with `-W UTF-8`, and they don't have the same issue with the encoding. However, UTF-8 is the default in all these versions, and appears to have been the default since 2.0.0 at least, possible before.

So it looks like we're safe to remove `-W UTF-8` wherever we find it, as we're unlikely to be running on that old a version of shp2pgsql. I'll put a check for that in, just in case.

nvkelso:
Working.

PR #947, opened 2016-08-05
Omit default shapefile encoding.

zerebubuth:
Omit shapefile encoding when it's `-W UTF-8`, as that has been the default since version 2. Add check for shp2pgsql version. This can cause a crash with recent (>2.2) versions of shp2pgsql when the default is superfluously specified _and_ the shapefile contains a code page.

Connects to #946.


rmarianski:
:+1:

PR #949, opened 2016-08-08
Adds semver statement

### LINKED ISSUE #948: Add semver document
As we approach v1.0.0 of Tilezen, what promises are we making? When should we bump the version on future changes? 

Developers should be able to tell from the Tilezen version increment how much effort it will take them to integrate the new tiles with their map. We use semantic versioning to communicate this.

Add documentation for how Tilezen proposes to implement SemVer for vector tiles.

nvkelso:
Fixes #948. 

As we approach v1.0.0 of Tilezen, what promises are we making? When should we bump the version on future changes?

This PR converts the Google Doc to Markdown in the main repo, with a few modifications.


SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,137 @@
...
+3. **PATCH** version **Z** when fixing backwards-compatible bugs
+4. **pre-release** version **-POSTFIX** when releasing developer previews
+
+**NOTE:** The above applies _after_ software has reached as version `1.0.0`, no promises are made for earlier versions.
+
+#### Developer level of effort:
+
+- Major version X: **high** – significant integration challenges, read the changelog closely
+- Minor version Y: **low** – some integration challenges, read the changelog
+- Bug fixes Z: **none** – simply use the new tiles, ignore the changelog
```
rmarianski: Can we say something like "can skim or ignore the changelog" instead? ;)

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,137 @@
...
+- Pre-release POSTFIX: **low to high** – some to significant integration challenges, read the changelog
+
+## Versioning the Tilezen API
+
+Proposed that upon our `1.0.0` release Tilezen makes the following promises.
+
```
rmarianski: How about adding a quick blurb to define what common, common-optional, optional, and kind are here?

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,137 @@
...
+1. **Remove** `common-optional` feature **property** at zoom 14 or more
+1. **Change** `common-optional` feature **property name**
+1. **Remove** `optional` feature **property** at zoom 14 or more
+1. **Change** `kind` **value name**
+1. **Remove** `kind` **value** completely from all zooms
+1. **Move** `kind` from one layer to another
+1. **Additional merging** across `kind` **values** in zooms 14, 15, or 16 (or max zoom) by removing `common`, `common-optional`, and/or `optional` **properties** or other method
+1. **Merging** within `kind` **values** at zooms 16 (or max zoom) by removing `common`, `common-optional`, and `optional` **properties** or other method
+1. **Change** of <= -3 (earlier) to default `min_zoom` or `max_zoom` **values** to determine when `kind` is included
+1. **Change** of >= +2 (later) to default `min_zoom` or `max_zoom` **values** to determine when `kind` is included
```
rmarianski: Does it make sense to also add in a line for changing the zoom level where we include all data for "overzoom" purposes? Or is that just implied by following the min_zoom values in stylesheets and things would transparently work?

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+### POLITICAL GEOGRAPHY is not versioned
+
+#### Examples:
+
+1. **Major** additions, deletions to country names, borders, disputed territories, and capitals are possible and may be advertized.
+    - Generally the **Natural Earth** source we currently use for low zooms (0-8) doesn’t update frequently.
+    - **OpenStreetMap** (used for zooms 9+) updates minutely
+    - **Who’s On First** (used for zooms 12+) updates minutely.
```
rmarianski: This isn't currently true, and it may be the sort of thing that's better left unspecified. Can we say that it is "constantly updating" or that it's on an "ongoing" or "regular" basis? As it stands this would give me the impression that when a feature is updated in osm, I would expect to see it minutes later.

VERSION
```diff
@@ -0,0 +1 @@
+1.0.0-pre1
```
rmarianski: Will we be updating this file during deployments going forward? And this would reflect the service version?

Because certain changes will be external to this repo, it makes me wonder whether it's better to capture it externally to this repo. Maybe even on an endpoint that is served live and can report the underlying versions and details of a particular deployment.

SEMANTIC-VERSIONING.md
nvkelso: Skim in 36c72231486ff4aabe622dd5b4af5459b7ec51b1.

SEMANTIC-VERSIONING.md
nvkelso: Added in 6ec85ead90225bb710d61122a25068f5f378cce4.

It feels like this needs a little bit more work, but it's a good straw man. Specifically: the descriptions added are pickups of WOF placetype roles, and match fairly well with what's in the Layers documentation now. But it feels like it needs slightly more finesse around which promises are made.

SEMANTIC-VERSIONING.md
nvkelso: For Tilezen API, I think that would be a MINOR change (we're telling people to use the `min_zoom` values, but those aren't present on all features). But for the service it'd be a MAJOR change? bc29eac909b769c2064f1b69c74266b71ad9c157 adds that (we can change it!).

SEMANTIC-VERSIONING.md
nvkelso: Fixed in 47a15f9863ea8c33bed7989c9c404a0e62014a2e.

VERSION
nvkelso: Yes, we should update the **Tilezen API** version in the new [VERSION](VERSION) file. 

Agree we should capture the **service** version external to the vector-datasource repo. Should that go into a new repo and/or one of the chef recipes? 

(I've updated the deployment checklist wiki, but it'll need to be updated again per outcome of above.)

rmarianski: > Agree we should capture the service version external to the vector-datasource repo. Should that go into a new repo and/or one of the chef recipes?

It would probably be easiest to generate a json file in a recipe and push that over to s3. We can read the revisions of the software components from our stack settings. And we can move the version there as well, or we can just read this new version file from the tagged vector-datasource if you think it would be easier for you to manage that way. We have a step that generates the tilejson files already, so we can probably update/rename that step to generate this as well for a particular deploy.

zerebubuth: What would we use this file for? Could we replace it with `git describe HEAD`? (or `git describe --abbrev=0 HEAD` if we don't want the exact commit)

When we're deploying production `vector-datasource`, the Chef recipes point to a specific tag, so the `VERSION` file is superfluous for the purposes of identifying the code version and it could potentially be in conflict with the tag - which would be right?

If we want to keep the `VERSION` file, and the version in `setup.py` as well, should we add a lint / CI check to ensure that the version matches the git tag?

SEMANTIC-VERSIONING.md
```diff
@@ -0,0 +1,169 @@
...
+Additional localized names are available as `common-optional`, but their actual use is the data is not widespread.
+
+
+### POLITICAL GEOGRAPHY is not versioned
+
+#### Examples:
+
+1. **Major** additions, deletions to country names, borders, disputed territories, and capitals are possible and may be advertised.
+    - **Natural Earth** (used for zooms 0-8) updates infrequently (often annually)
+    - **OpenStreetMap** (used for zooms 9+) updates frequently (at least daily)
```
zerebubuth: Is it worth stating an exception here for the `openstreetmapdata` land/water polygons. Hopefully we'll update those regularly, but it's unlikely to be daily.

VERSION
rmarianski: IIUC, this would be to differentiate the version of vector-datasource from the api version. There can be changes that sit outside the software that could affect the api.

We could just track the api version separately in the stack settings. I can see it being confusing that there are 2 separate version specifiers in the same repository.

zerebubuth: I don't understand - when would we tag a version on `vector-datasource` that was different from the semver version that the document is describing? If there's a change in the API, for example something in tilequeue or some upstream rewrite rule, then it's possible we'd still want to update documentation in this repo.

If the version would change and this repo would stay entirely identical (apart from `VERSION`), then does that indicate that the `VERSION` file belongs in a different repo, and it's that other repo which is being versioned?

rmarianski: Yea, I think you're right that the documentation and api version might want to live in a separate repository.

zerebubuth: The documentation for the tile contents seems like it belongs in this repo, along with the code which creates those contents... Is it the URL structure and available formats which is determined externally? In which case, perhaps it's worth splitting the docs about those different things?

rmarianski: Sounds good to me.

nvkelso: Yeah, my intent is to have the VERSION file in this vector-datasource repo establish what the SemVer version is for the Tilezen API (the content), and that should be maintained before a git release is tagged (release wiki has been updated to this effect). The upside is it's easy to read the VERSION file directly in the repo without going thru a bunch of git hoops, and if the files become disassociated from a git checkout the VERSION is still carried along.

+1 for adding a lint / CI check to ensure that the version matches the git tag.

Agree that the Mapzen Vector Tiles service VERSION should be somewhere else, sounds like Chef in json. Should there be a new version endpoint to query this?

rmarianski: We already have the version of vector-datasource specified in setup.py though. Can we just keep it in there?

nvkelso: Suggest adding this line https://github.com/tilezen/vector-datasource/blob/master/setup.py#L4 with:

```
version = open("VERSION").read()
```

rmarianski: Sounds good, I pushed over a safer version in 243fd3b to your branch.

nvkelso: @rmarianski Is it possible to do the same for the tilejson file template, or are we considering that the SERVICE version? Where should the service version go?

SEMANTIC-VERSIONING.md
nvkelso: Yes, added that in 53ea48abe568bc0ffc88f482710ed232a1ea7f62.

Did further rearranging in 7c247eabdc740ada104df99278b7df0560506a90 to move data update frequency to its own sub section above political geography, remove the aspirational advertisement note from the political geography section, but do specify it as a PATCH version trigger for SERVICE.

nvkelso:
Related change in tileserver: https://github.com/tilezen/tileserver/pull/35

nvkelso:
Related change in tilequeue: https://github.com/tilezen/tilequeue/pull/105.

nvkelso:
Related tilejson template change: https://github.com/tilezen/vector-datasource/issues/962

VERSION
nvkelso: Tracking the tilejson bit here: https://github.com/tilezen/vector-datasource/issues/962.

nvkelso:
Woot!

PR #956, opened 2016-08-10
Update road indexes

rmarianski:
The queries only filter on road level currently.


zerebubuth:
Looks like the test fails because [that hotel](http://www.openstreetmap.org/relation/6480430) has recently become a multipolygon, and we're probably just pulling in the way ID.

:+1: on the changes.

Issue #522, opened 2016-02-11
Large hotels should be visible earlier


nvkelso:
This area of SF has many large, famous hotels, but those aren't visible on the map until zoom 17.  We should include `hotel` and `motel` earlier, based on their areas.

The [Fairmount](https://www.openstreetmap.org/way/32947165) and [Ritz-Carlton](https://www.openstreetmap.org/way/32947245) are candidates for zoom 15. The [Standard](https://www.openstreetmap.org/relation/3827943) and [InterContinental Mark Hopkins](https://www.openstreetmap.org/way/288499856) are z16 suitable. No change to the no-area default of z17.

Related: #520 

17/37.79196/-122.40892

<img width="895" alt="screen shot 2016-02-10 at 22 51 53" src="https://cloud.githubusercontent.com/assets/853051/12970892/08cab1a2-d049-11e5-800d-41764de2350e.png">

nvkelso:
This is related to Indy's new year's resolutions post, due out early March.

rmarianski:
z15

before:

![before-hotel-z15](https://cloud.githubusercontent.com/assets/53770/13301029/f4c26114-db10-11e5-9f59-519b0310ef4a.png)

after:

![after-hotel-z15](https://cloud.githubusercontent.com/assets/53770/13301043/0676d20a-db11-11e5-94e8-d0e0768d66e4.png)

z16 is a lot more chaotic locally now because we started metatiling.

The before I took from production so it's got a lot less clutter:

![before-hotel-z16](https://cloud.githubusercontent.com/assets/53770/13301077/2371e6a6-db11-11e5-99bc-b3ced726ceaf.png)

after:

![after-hotel-z16](https://cloud.githubusercontent.com/assets/53770/13301099/35e13684-db11-11e5-88d4-a1ac29d30d21.png)

nvkelso:
LGTM

PR #959, opened 2016-08-10
Data was updated - fix test to match.

zerebubuth:
The test for hotels from #522 included a test for the "Stanford Court Renaissance", which has since [been deleted](https://www.openstreetmap.org/changeset/41355932), and a [new object added](https://www.openstreetmap.org/relation/6480430) with the name "Stanford Court San Francisco Hotel". Perhaps it changed owners or something?


rmarianski:
:+1:

Issue #951, opened 2016-08-08
Venice is under water


nvkelso:
* **What did you see?** Venice is under water, so is the island west of Athens.
* **What did you expect to see?** Roads should be on land.

We think this is caused because "land" kind sorts below water, but should instead sort above water.

zerebubuth:
In the case of Venice, it's because [the Adriatic sea](http://www.openstreetmap.org/relation/4590486#map=13/45.4320/12.3353), which gets `kind:sea, sort_key:200`, covers Venice island, which is `kind:earth, sort_key:10`.

In the case of Salamina, near Athens, [the Aegean Sea](http://www.openstreetmap.org/relation/4594226#map=16/37.9543/23.5497), also `kind:sea, sort_key:200`, covers Salamina island, also `kind:earth, sort_key:10`.

If we alter the sort key of seas to be below earth, then we may get the opposite problem happening elsewhere in the world; bays which are part of seas "draining" because the land now "floats" above them. Related to #446.

There's another possible fix: both of these sea polygons come from a multipolygon tagged `place=sea` and, although it _is_ a polygon, perhaps it would be better to use it only for labelling as we already do for bays, straits and fjords #400 #515? In both cases above there were other polygons ([Venice Lagoon](http://www.openstreetmap.org/relation/3049430) and Salamina's coastline) which cover the water areas, without needing changes to the sort key.

nvkelso:
When we added islands I made a bad assumption that their landmass would already be present in the earth data we get from openstreetmapdata.com (and they'd be cookie-cutttered with at least ocean-like water), so this logic removes their polygons (after creating label placements for them): https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L478-L484. 

Is another option to not remove island land features (to keep kind IN (`island`,`islet`) in the **earth** layer, and sort that above the water but below other man_made features? Example: https://www.openstreetmap.org/relation/2683672#map=18/45.42440/12.32811.

[...]

PR #964, opened 2016-08-10
Olga/951 remove place sea

okavvada:
Connects to #951.
- [x] Update tests
- [ ] Update data migrations
- [ ] Update docs

Ready for review @rmarianski , @zerebubuth , @nvkelso

### DIFF
--- a/data/migrations/v1.0.0-polygon.sql
+++ b/data/migrations/v1.0.0-polygon.sql
@@ -1,6 +1,0 @@
-UPDATE
-  planet_osm_polygon
-  SET mz_water_min_zoom = mz_calculate_min_zoom_water(planet_osm_polygon.*)
-  WHERE
-    tags -> 'place' = 'sea'
-    AND COALESCE(mz_water_min_zoom, 999) <> COALESCE(mz_calculate_min_zoom_water(planet_osm_polygon.*), 999);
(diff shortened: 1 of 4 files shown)


okavvada:
The issue is that although we are correctly dropping the sea polygon so it does not overlap the islands, we are not generating sea labels. This is true for prod as well.

zerebubuth:
The label will be placed at a point within the polygon which we calculate in Python (although @rmarianski is currently moving some of that logic to the database in #963), and the same point can be found by running:

```sql
select osm_id, st_astext(st_transform(st_pointonsurface(way), 4326))
from planet_osm_polygon where osm_id = -4594226;
```

I've updated the test to use that point, and it passes on my computer. We'll have to wait and see if CircleCI agrees!

The migration shouldn't be needed, as the changes don't alter `mz_water_min_zoom`. When processing gets to the post-process steps from `queries.yaml`, the database has already done all its work.

okavvada:
Thank you @zerebubuth for the help! I also deleted the migration as you were right, it was not needed.

zerebubuth:
:+1: looks great!

rmarianski:
Yea, the migration would only have been needed if the rule was removed in the yaml file.

:+1:

Issue #192, opened 2015-09-10
Shield text version of ref


nvkelso:
By [adding the network tags](https://github.com/mapzen/vector-datasource/issues/121) from the relations onto features we know, for instance, that a way for "US 101" is network "US:US" and should have use US Route style artwork. But we can't shove the ref value into the shield's text straight because it duplicates "US ". 

Two options:

* We modify the ref to strip out values like "US ", but this a destructive change and might impact other users of the data, especially non-Tangram users.

* We create a new shield text property and store the result there.

zerebubuth:
Is it possible to add a third option: filter the prefix out of the ref client-side? It seems like it's something that would be style-dependent anyway: If you wanted to render a plain, non-specialised shield then you'd want to see "US 101", as it's the shield symbol itself which tells you it's a "US-" road, and it's the style which determines what shield will be rendered?

nvkelso:
See also: #896.

bcamper:
While it's possible to process the `ref` client-side with regex/JS string operations and we can prototype it this way (and always preserve that option for people who want flexibility), I think it is worth having a server-side provided "shield text" field to avoid this work client-side (both to avoid the cost of all those regexes client-side and to support renderers that don't have this ability).

nvkelso:
We already calculate the shield type and export that as the `network` property on roads (see "mz_get_rel_network" in **functions.sql** . We need to iterate over all the route relations, not just the first one (export multiple with `;` separator?), and do some regEx to pull out the shield text part of the "name" and export that in a new property.

nvkelso:
See also: http://www.openstreetmap.org/user/Minh%20Nguyen/diary/39123

nvkelso:
💯 This is super cool, and verified :)

nvkelso:
Clarified documentation in 7112bf1c22c4f1a76755cae8b2442ebef1e09590.

nvkelso:
Reopening as I'm seeing many cases in Europe where the `shield_text` is **A ##** or **M ##** or **N ##** or **E ##** (there may be others) with a space between the leading character and the numbers, which impacts artwork creation on the spritesheet side & matching logic in Tangram scene files. 

**There shouldn't be a 2nd char space in the `shield_text` values (but the leading character seems convention to keep).**

Examples:

- E road: 
  - http://tile.dev.mapzen.com/vector/v1/all/9/257/174.topojson
[...]

PR #966, opened 2016-08-12
Add 'shield_text' property on roads

zerebubuth:
This adds a `shield_text` property to go with the `network` on roads. There are a few changes from previous behaviour:

* A "most important" network is now chosen, instead of whichever was first. This will give better consistency between adjacent road segments and when routes merge.
* Only road networks are considered for "most important". This means the `network` property disappears from some existing roads where it duplicates the `bicycle_network` property.

The "most important" calculation is currently very simple, and does not have special cases in for various special roads. For example [it is recommended](http://elrond.aperiodic.net/shields/supported.html) that [Historic Route 66](https://en.wikipedia.org/wiki/U.S._Route_66) is mapped without a `ref` tag, which will mean it doesn't appear. In general, `name` is ignored and `ref` is assumed to be numeric.

This seemed good enough for an iteration, and we can improve it if/when good examples are available. @nvkelso does that sound good?

Connects to #192.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -92,9 +92,9 @@ $$ LANGUAGE plpgsql IMMUTABLE;
 --
 CREATE OR REPLACE FUNCTION mz_get_rel_networks(
   way_id bigint)
-RETURNS text AS $$
+RETURNS text[] AS $$
 SELECT
-  string_agg(unnested, '|')
+  array_agg(unnested)
 FROM (
   SELECT
     unnest(tags) AS unnested
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -83,24 +83,33 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
 
--- mz_get_rel_network returns a network tag, or NULL, for a
--- given way ID.
+-- mz_get_rel_networks returns a list of triples of route type,
+-- network and ref tags, or NULL, for a given way ID.
 --
 -- it does this by joining onto the relations slim table, so it
 -- won't work if you dropped the slim tables, or didn't use slim
 -- mode in osm2pgsql.
 --
-CREATE OR REPLACE FUNCTION mz_get_rel_network(
+CREATE OR REPLACE FUNCTION mz_get_rel_networks(
   way_id bigint)
 RETURNS text AS $$
-BEGIN
-  RETURN mz_first_dedup(ARRAY(
-    SELECT mz_rel_get_tag(tags, 'network')
-    FROM planet_osm_rels
-    WHERE parts && ARRAY[way_id]
-      AND parts[way_off+1:rel_off] && ARRAY[way_id]));
-END;
-$$ LANGUAGE plpgsql STABLE;
+SELECT
+  string_agg(unnested, '|')
+FROM (
+  SELECT
+    unnest(tags) AS unnested
+  FROM (
+    SELECT
+      hstore(tags)->ARRAY['route','network','ref'] AS tags
+    FROM
+      planet_osm_rels
+    WHERE
+      parts && ARRAY[way_id] AND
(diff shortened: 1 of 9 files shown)


data/functions.sql
```diff
@@ -83,24 +83,33 @@ BEGIN
...
+-- network and ref tags, or NULL, for a given way ID.
 --
 -- it does this by joining onto the relations slim table, so it
 -- won't work if you dropped the slim tables, or didn't use slim
 -- mode in osm2pgsql.
 --
-CREATE OR REPLACE FUNCTION mz_get_rel_network(
+CREATE OR REPLACE FUNCTION mz_get_rel_networks(
   way_id bigint)
 RETURNS text AS $$
```
rmarianski: Is it easy to have this return back an array, hstore, or json with the values instead of a delimited string? This is just meant for the python transform to consume and process right?

vectordatasource/transform.py
```diff
@@ -3676,3 +3676,68 @@ def normalize_operator_values(shape, properties, fid, zoom):
...
+    """
+
+    networks = properties.pop('mz_networks', None)
+
+    if networks is not None:
+        networks = networks.split('|')
+
+        # take the list and make triples out of it
+        itr = iter(networks)
+        triples = zip(itr, itr, itr)
```
rmarianski: Nice! :zap:

vectordatasource/transform.py
```diff
@@ -3676,3 +3676,68 @@ def normalize_operator_values(shape, properties, fid, zoom):
...
+
+        def is_road_network(t):
+            return t[0] == 'road'
+
+        triples = filter(is_road_network, triples)
```
rmarianski: I see you were in a functional mood ;) If this function is just used in this one location, list comprehension instead?

rmarianski:
Put in some minor comments, but the implementation lgtm. :+1:

data/functions.sql
zerebubuth: Ah, yes! Good point! I had forgotten that was JSON and thought it had to be stringly to fit in an hstore. Fixed in 861ac6d.

vectordatasource/transform.py
zerebubuth: Yeah, that would be more readable... but what would it look like? Can I do `[t for t in triples where t[0] == 'road']`?

rmarianski: :white_check_mark:

nvkelso:
Nice! 

If we're iterating thru all the networks and their shield_text, can we please also export that entire list of shields (even if it's on two new properties instead of the current singular properties)?

vectordatasource/transform.py
zerebubuth: Thanks, fixed in 99ea802. And :mushroom: level up for me, learned new Python syntax :wink:

rmarianski: Whoops, yea I missed that it's `if` instead of `where` in your comment, but you got it.

zerebubuth:
Yup, sure! In 62375da added support for exporting `all_networks` and `all_shield_texts` as lists. This means that they won't be available in MVT format, as we currently can't encode lists as MVT values.

nvkelso:
For lists in MVT values, do we drop those completely for that format now? Could we add new logic in the mapbox-vector-tile library  to concatenate lists into a single string with `;` separator characters? (We need to support these for Eraser Map, for instance, that uses MVT but also needs multi shields.)

zerebubuth:
Yes, we drop them at the moment. We could add that logic, but it might be better to encode them as JSON lists rather than `;` concatenate them just in case there are escaping issues.

I've added https://github.com/tilezen/mapbox-vector-tile/issues/64 to track this. But since multiple road shields aren't supported in any of the renderers using our tiles, and aren't part of the v1.0 milestone, that suggests it's okay to leave until v2 or later.

Issue #583, opened 2016-03-09
zoom 9 tile has excessive landuse, other layer details


nvkelso:
http://vector.mapzen.com/osm/all/9/245/166.topojson?api_key=vector-tiles-HqUVidw

There are many tiny landuse features included in this tile. 

* The large ones should keep their names and independent existence. 

* The smaller ones should loose their names and be merged (like what we started doing in the roads layer in v0.7).

nvkelso:
Download: [9:245:166.topojson.json.zip](https://github.com/mapzen/vector-datasource/files/164276/9.245.166.topojson.json.zip)

nvkelso:
Separate tile previewed in QGIS:

![screen shot 2016-03-28 at 13 31 07](https://cloud.githubusercontent.com/assets/853051/14089385/7980494c-f4e9-11e5-9e06-ba2640385967.png)

[22.json.zip](https://github.com/mapzen/vector-datasource/files/192431/22.json.zip)

For instance, all the Urban Area polygons from Natural Earth at the same min_zoom and sort_key could be merged.

nvkelso:
I suspect the `roads` and `boundary` layers need additional geom simplifying. 

We could also strip some properties off the boundary layer at low zooms (like `name` and the name variants) at zooms less than 8 as the zoom isn't large enough to provide screen space to do the labeling.

![screen shot 2016-03-28 at 13 35 53](https://cloud.githubusercontent.com/assets/853051/14089500/04a88642-f4ea-11e5-8c8c-ede2258343d3.png)

![screen shot 2016-03-28 at 13 35 24](https://cloud.githubusercontent.com/assets/853051/14089488/fa3fa9d8-f4e9-11e5-989a-7f199ab53d3d.png)

zerebubuth:
Tracking changes for the linework in `roads` and `boundaries` layers in #641 - no need to conflate that task, which is about altering a threshold for a feature we already have, and this one which is about adding an entirely new feature.

nvkelso:
**Establish some scale ranks for landuse**

From queries.yaml, these are already applied, but the CSV is empty now.

The sizes will be similar to how buildings.csv is setup, but you'll use some of the area thresholds from other filters. 

```
  - fn: vectordatasource.transform.csv_match_properties
    resources:
      matcher:
        type: file
        init_fn: vectordatasource.transform.CSVMatcher
        path: spreadsheets/scalerank/landuse.csv
    params:
      source_layer: landuse
[...]

PR #967, opened 2016-08-16
Merge landuse polygons

zerebubuth:
Refactor building merge code to do landuse polygons as well. This is configured to drop the name, operator and sport tags from polygons being merged between zooms 9-12, then merge where the properties are the same.

On the example test tile, `9/245/166`, the feature count dropped from 630 to 80, and the tile size dropped by 27%. (Only 27% because most of the features are still there, just all in a multipolygon.)

Connects to #583.


queries.yaml
```diff
@@ -432,6 +432,23 @@ post_process:
...
+  # drop the names of small landuse polygons
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: landuse
+      start_zoom: 9
+      end_zoom: 12
+      properties: [name, operator, sport]
+      where: >-
+        pixel_area > area
+      geom_types: [Polygon, MultiPolygon]
```
rmarianski: Looks like this will drop these properties, even if for some reason they cannot be merged later on. Is that what we want to do?

zerebubuth: This mostly just leaves `kind`, and a bunch of properties which are derived from `kind`, which means things generally do get merged in the next step, even if it's just into a big MultiPolygon. I thought that's what the issue was asking for... perhaps @nvkelso could clarify?

If we're expecting polygons to get merged along seams then we might have to do some much more complex processing to bridge the gaps between them - on the example tile, most were disjoint.

nvkelso:
Neat! Great savings!

FYI: We rely on operator at low zooms for some kinds in Mapzen house styles. Primarily US National Park Service and US Forest Service. Should we split off those combos to new kind values, or whitelist those two operators and strip the rest?

zerebubuth:
> FYI: We rely on operator at low zooms for some kinds in Mapzen house styles. Should we split off those combos to new kind values?

I think that's the best option, if it's an important difference that wouldn't want merged.

That blacklist is bothering me, actually - I should see if I can swap it around to make it a whitelist, otherwise other random tags might stop it from merging.

Also, should we be merging _before_ generating point placements, or _after_?

nvkelso:
For roads I think we merge differently for different kinds (dropping different properties), that's probably what we could do for 'national_park' and a few other kinds wrt operator?

For label placements, it may be okay to not generate those until zoom 12 or so when merging no longer happens. Olga moved most the important label bits to POIS already.

zerebubuth:
Yup, we could keep the `operator` tag for parks and not other kinds.

> ...and a few other kinds wrt operator?

Which other kinds did you have in mind?

nvkelso:
I think these show up starting zoom 2, others are later: 

battlefield
national_park
protected_area
nature_reserve

And sometimes protected_class is meaningful at low zooms.

docs/layers.md
```diff
@@ -242,6 +242,8 @@ Landuse polygons from OpenStreetMap representing parks, forests, residential, co
 
 Zooms 4 and 5, 6 and 7 includes a mix of Natural Earth `urban_area` (zooms 0-9 only) features and OpenStreetMap data for `national_park`, `protected_area`, and `nature_reserve` only. After that more more feature kinds are included, and they have a richer set of properties including `sport`, `religion`, `surface`, `attraction`, `zoo`, and `natural`. Feature selection is filtered per zoom until zoom 15.
 
+At mid-zooms, between 9-12, some landuse polygons are merged to reduce payload size. To facilitate this, the name of the landuse area may be dropped for small polygons. When polygons are merged, the original `id` properties are dropped, and the `area` is re-calculated for the new size.
```
nvkelso: Generally this is true, but for a few kinds we show them as earlier as zoom 2 (https://github.com/tilezen/vector-datasource/blob/master/yaml/landuse.yaml#L5-L13).

queries.yaml
```diff
@@ -432,6 +432,23 @@ post_process:
         - transit
         - water
 
+  # drop the names of small landuse polygons
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: landuse
+      start_zoom: 9
```
nvkelso: We show a few kinds earlier as zoom 2 (https://github.com/tilezen/vector-datasource/blob/master/yaml/landuse.yaml#L5-L13).

queries.yaml
nvkelso: Big MultiPolygons are working for me (and what I expected). We can tackle something more past the v1 milestone.

docs/layers.md
zerebubuth: I can see that `mz_landuse_min_zoom` might be as low as 2, but the [queries](https://github.com/tilezen/vector-datasource/blob/89bd4677c7b0e7be767d079bb31ff02c6602a1ee/queries/landuse.jinja2#L46) only appear to kick in at zoom 4.

Have I overlooked some way we can get polygons in landuse at z<4?

nvkelso: You're correct, zoom 4.

rmarianski:
:+1:

PR #969, opened 2016-08-16
Unify building parts with their building

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/653


rmarianski:
The [relation in the issue, the sf ferry building](http://www.openstreetmap.org/relation/6062613) doesn't end up getting inserted into the database (it's in rels, but not in planet_osm_polygon). But, the [outline](http://www.openstreetmap.org/way/24460886) does. This just means that currently the `root_id` that gets associated with those buildings is the outline and not the relation id (a feature with the relation id doesn't get returned in the tile).

I also tried to find an example of a tile that had a `root_relation_id` set in the `pois` layer, but haven't thus far. I'll continue to hunt down for one. I wanted to find one to write a test for it to verify that the commit in this pr helps.

docs/layers.md
```diff
@@ -151,7 +151,7 @@ Combination of OpenStreetMap administrative boundaries (zoom >= 8) and Natural E
 
 Polygons from OpenStreetMap representing building footprint, building label_placement points, building_part features, and address points. Starts at zoom 13 by including huge buildings, progressively adding all buildings at zoom 16+. Address points are available at zoom 16+, but marked with `min_zoom: 17` to suggest that they are suitable for display at zoom level 17 and higher.
 
-Individual `building:part` geometries from OSM following the [Simple 3D Buildings](http://wiki.openstreetmap.org/wiki/Simple_3D_Buildings) tags at higher zoom levels are now exported as `building_part` features with specified `kind_detail`.
+Individual `building:part` geometries from OSM following the [Simple 3D Buildings](http://wiki.openstreetmap.org/wiki/Simple_3D_Buildings) tags at higher zoom levels are now exported as `building_part` features with specified `kind_detail`. Building parts will receive a `root_id` that will correspond to the id of the `building` feature that they intersect.
```
zerebubuth: Pedantic, but: This makes it sound like `building:part`s will _always_ receive a `root_id`, but its possible for `building:part`s to not match any `buildings`. Would it be clearer if it were written as follows?

> Building parts may receive a `root_id` corresponding to the `building` feature, if any, with which they intersect.

zerebubuth:
Tile `16/32747/21793` has two station points for Waterloo station with `root_relation_id: 1242762`. Pulling in that relation should, I think, be enough to reproduce that in the tests.

One pedantic thing: There are test cases for the non-overlapping and overlapping case, but none for the case where it picks the most overlapping of several candidates.

Other than that, it looks great! :+1:

docs/layers.md
rmarianski: Good call, updated in 3fa38ad.

rmarianski:
> Tile 16/32747/21793 has two station points for Waterloo station with root_relation_id: 1242762. Pulling in that relation should, I think, be enough to reproduce that in the tests.

Thanks. Updated in 8a6cb49.

rmarianski:
> One pedantic thing: There are test cases for the non-overlapping and overlapping case, but none for the case where it picks the most overlapping of several candidates.

Added unit test for best overlap in aa7d144.

zerebubuth:
:+1: looks awesome! Nice work!

PR #973, opened 2016-08-18
Fixes for interaction with the label placement work

zerebubuth:
Shift merging to after label placement extraction and ignore landuse points in test. Should fix failing CI tests.

Connects to #583.

### DIFF
--- a/integration-test/583-merge-landuse.py
+++ b/integration-test/583-merge-landuse.py
@@ -648,8 +648,13 @@ with features_in_tile_layer(9, 245, 166, 'landuse') as landuse:
     feature_props = set()
 
     for feature in landuse:
-        # ignore ID, area when checking for uniqueness
         p = feature['properties'].copy()
+
+        # ignore label placements
+        if p.get('label_placement'):
+            continue
+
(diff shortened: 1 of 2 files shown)


integration-test/583-merge-landuse.py
```diff
@@ -648,8 +648,13 @@ def _freeze(thing):
     feature_props = set()
 
     for feature in landuse:
-        # ignore ID, area when checking for uniqueness
         p = feature['properties'].copy()
+
+        # ignore label placements
+        if p.get('label_placement'):
```
rmarianski: Pedantic, membership tests can be: `'label_placement' in p`

rmarianski:
Thanks for the fix! :zap: :+1:

integration-test/583-merge-landuse.py
zerebubuth: Is `label_placement` always truthy if present? I know we've used values of `"yes"` and `True` before, but are we excluding using `False`?

rmarianski: Ah, good point! It usually is truthy, perhaps even always, but I like your spelling better in case it's not.

PR #974, opened 2016-08-18
Update config_all_layers behavior

rmarianski:
Ensure that when config_all_layers is set, the `all` layer is requested.


zerebubuth:
LGTM :+1:

Issue #972, opened 2016-08-17
Use 2 char language codes not 3


nvkelso:
Following up on https://github.com/tilezen/vector-datasource/pull/816, it looks like iOS, Android, and web all natively prefer 2 char language codes instead of 3 char code. Instead of forcing everyone to shim between, we should just state the 2 char code instead.

/cc @bcamper

nvkelso:
And consider whitelisting country code (e.g. 2nd part of `en-gb`).

nvkelso:
This is working on dev for me. 

But one minor documentation fix required that `ja` for Japanese instead of `jp` (see https://github.com/tilezen/vector-datasource/pull/975#discussion_r76197406).

rmarianski:
@dianashk we had previously talked about using 3 letter codes, but looks like 2 letter codes are the way to go.

nvkelso:
Fixed https://github.com/tilezen/vector-datasource/issues/972#issuecomment-242310941 in 24430af2591163e669d3c3c6c3ffefdbc5ad51dc.

PR #975, opened 2016-08-18
Use 2 letter language codes

zerebubuth:
Use ISO 639-1 two letter language codes for WOF & OSM rather than ISO 639-3 3 letter codes. This partially reverses changes made for #418.

Connects to #972.

### DIFF
--- a/SEMANTIC-VERSIONING.md
+++ b/SEMANTIC-VERSIONING.md
@@ -132,18 +132,18 @@ In addition to the `common` **name** locals call a place, the following `common`
 
 #### Common languages:
 
-1. `name:ara` Arabic
-1. `name:zho` Chinese, traditional or simplified
-1. `name:eng` English
-1. `name:fra` French
-1. `name:rus` Russian
-1. `name:spa` Spanish
-1. `name:deu` German
-1. `name:gre` Greek
-1. `name:ita` Italian
-1. `name:jpn` Japanese
-1. `name:kor` Korean
-1. `name:vie` Vietnamese
+1. `name:ar` Arabic
+1. `name:zh` Chinese, traditional or simplified
+1. `name:en` English
+1. `name:fr` French
+1. `name:ru` Russian
+1. `name:es` Spanish
+1. `name:de` German
+1. `name:gr` Greek
+1. `name:it` Italian
+1. `name:jp` Japanese
+1. `name:ko` Korean
+1. `name:vi` Vietnamese
 
 Arabic, Chinese, English, French, Russian and Spanish are used by the United National for meetings and official documents. The other languages listed are frequently used in OpenStreetMap and Who's On First.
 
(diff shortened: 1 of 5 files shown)


vectordatasource/transform.py
```diff
@@ -386,38 +389,38 @@ def _normalize_country_code(x):
...
     if lang_code_result is None:
         return None
 
-    country_result = _normalize_country_code(country_candidate)
-    if country_result is None:
-        return None
+    if country_candidate:
+        country_result = _normalize_country_code(country_candidate)
+        if country_result is None:
+            return None
```
rmarianski: This matches the previous behavior, but I'm now wondering if it's better to keep the language code rather than dropping it on an invalid country code.

vectordatasource/transform.py
```diff
@@ -386,38 +389,38 @@ def _normalize_country_code(x):
...
+osm_l10n_lookup = set([
+    'zh-min-nan',
+    'zh-yue'
+])
 
 
 def _convert_osm_l10n_name(x):
-    x = osm_l10n_name_lookup(x)
+    if x in osm_l10n_lookup:
+        return x
```
rmarianski: Do we want to preserve these variations, or normalize them?

zerebubuth: I assumed that if we handled them specially before, we'd want to keep them now? @nvkelso what do you think?

vectordatasource/transform.py
zerebubuth: I'm wondering if we have something like `name:en="Foo", name:en_XY="Bar"` do we want the `name:en_XY` to normalise to `name:en` and potentially clash with the other entry?

rmarianski: Good point. We can set it up so that we'll only set the language code if one doesn't already exist in this scenario, but the added complexity might not be worth it.

zerebubuth: I've changed the behaviour in 4df4863 to keep the most specific language code. At the moment, that just means that any with an omitted country are less specific and won't overwrite a code without a country, e.g: `en` always beats `en_NOEXIST`.

rmarianski: Great idea with priorities! :zap: My first thoughts were also to just return a more complex value object, but just whether the country code was invalid or not. The priority notion is a better generalization of that.

rmarianski:
:+1:

SEMANTIC-VERSIONING.md
```diff
@@ -132,18 +132,18 @@ In addition to the `common` **name** locals call a place, the following `common`
...
+1. `name:ar` Arabic
+1. `name:zh` Chinese, traditional or simplified
+1. `name:en` English
+1. `name:fr` French
+1. `name:ru` Russian
+1. `name:es` Spanish
+1. `name:de` German
+1. `name:gr` Greek
+1. `name:it` Italian
+1. `name:jp` Japanese
```
nvkelso: This should probably be `ja` instead of jp.

Issue #970, opened 2016-08-17
Add kind_detail for remaining roads (not NE, not aeroway)


nvkelso:
For the v1 release we added `kind_detail` values for NE roads and OSM aeroway, we need to add that for all the other features (mostly based on highway tag, sometimes not).

nvkelso:
@zerebubuth Did [#978](https://github.com/tilezen/vector-datasource/pull/978/files#diff-e43c77e9884dfed29c891f4935fdad10) remove all the `highway` and related raw OSM tags via changes to the roads.yaml?

I was expecting the new `kind_detail` to be in addition to those tags, but looks like they went missing? 

Removing any reference to the original OSM tags has made our lives in Tilezen easier, and I think will be great for client styling. Is there utility in keeping the raw OSM tags around for v1 to help folks coming from OpenStreetMap tagging world understand how Tilezen is setup? (But propose dropping them in v2?)

nvkelso:
(Updating Bubble Wrap to use `kind_detail` for all the roads layer filters has restored the missing roads I was seeing yesterday.)

zerebubuth:
Yes, the superfluous tags were removed. That information is in `kind_detail` now, when it's relevant to the feature's `kind`. In many cases, irrelevant and extraneous tags have been dropped, for example `leisure` values on `highway`s or `sport` values on `railway`s.

Would keeping the extra tags in until v2 make it more work for anyone depending on those to upgrade from v1 to v2? If the upgrade is hard, perhaps it's better to acclimatise users to the new schema from the beginning or, if the upgrade is simple, then perhaps it wouldn't be a barrier to understanding?

My feeling is that keeping the OSM tags intact makes sense if we do it consistently and permanently. However, we don't. For example, many of the POI features we expose don't retain [the original tag their `kind` was set from](https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L59-L65). We're including many other [special purpose properties specific to our style families](https://github.com/tilezen/vector-datasource/issues/473), which anyone familiar with OSM tagging would have to learn. Might it be equally helpful for folks with an OSM background to have great docs about how our features acquire their properties? That would be helpful to people of all backgrounds, too.

I think both tile size and schema clarity would suffer if we try to be _both_ a lightly-transformed and a heavily opinionated tileset. We seem to have been moving towards the latter, so let's embrace that :smile:

nvkelso:
💯

PR #978, opened 2016-08-19
Kind detail for all roads

zerebubuth:
Converts the various "subkinds" into `kind_detail` values. For example, for a `kind:highway` road we used to also output a `highway: motorway` property, but this is now `kind_detail: motorway`. Similarly for `kind: railway` and so forth. This vastly reduces the number of columns in the spreadsheet for assigning a sort key, and should allow us to be much more precise about which properties get output.

Connects to #970.


rmarianski:
lgtm :+1:

Issue #983, opened 2016-08-22
Investigate slow query


rmarianski:
```sql
SELECT
	    tags->'name' AS name,
	    'openstreetmap.org' AS source,
	    ST_AsBinary(way) AS __geometry__,
	    osm_id AS __id__,
	    min_zoom,
	    way_area AS area,
	    tags->'protect_class' AS protect_class,
	    tags->'operator' AS operator,
	    tags->'cuisine' AS cuisine,
	    tags->'religion' AS religion,
	    tags->'sport' AS sport,
	    (transit_routes).score AS mz_transit_score,
	    (transit_routes).root_relation_id AS mz_transit_root_relation_id,
	    (transit_routes).train_routes AS train_routes,
	    (transit_routes).subway_routes AS subway_routes,
	    (transit_routes).light_rail_routes AS light_rail_routes,
	    (transit_routes).tram_routes AS tram_routes,
	    tags->'ref' AS ref,
	    tags->'attraction' AS attraction,
	    tags->'zoo' AS zoo,
	    tags->'exit_to' AS exit_to,
	    %#tags AS tags,
	    mz_properties
	
	FROM (
	
	  SELECT
	    way, 0::real AS way_area, mz_poi_min_zoom AS min_zoom,
	    osm_id,
	    -- note: the mz_calculate_transit_routes_and_score function is pretty
	    -- expensive, so we only want to calculate it when we actually need the
	    -- result.
	    CASE
	      WHEN tags ? 'railway' AND tags->'railway'='station' AND osm_id > 0
	        THEN mz_calculate_transit_routes_and_score(osm_id, NULL)
	      ELSE NULL
	    END AS transit_routes,
	    mz_calculate_json_pois(planet_osm_point.*) AS mz_properties,
	    tags
	
	  FROM
	    planet_osm_point
	
	  WHERE
	    way && ST_SetSrid(ST_MakeBox2D(ST_MakePoint(8609997.899908404797, 6105309.357059651986), ST_MakePoint(9548994.035744480789, 7044305.492895727977)), 3857)
	
	    AND mz_poi_min_zoom <= 6
	
	
	UNION ALL
	
	  SELECT
	    way, way_area, mz_poi_min_zoom AS min_zoom,
	    osm_id,
	    -- note: the mz_calculate_transit_routes_and_score function is pretty
	    -- expensive, so we only want to calculate it when we actually need the
	    -- result.
	    CASE
[...]

PR #985, opened 2016-08-23
Add pois indexes for polygon and point tables

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/983

### DIFF
--- a/data/migrations/v1.0.0-point.sql
+++ b/data/migrations/v1.0.0-point.sql
@@ -43,6 +43,7 @@ ALTER INDEX new_planet_osm_point_min_zoom_pois_15_index RENAME TO planet_osm_poi
 COMMIT;
 
 DROP INDEX IF EXISTS planet_osm_point_min_zoom_way_index;
+DROP INDEX IF EXISTS planet_osm_point_min_zoom_wa9_index;
 DROP INDEX IF EXISTS planet_osm_point_min_zoom_way_9_index;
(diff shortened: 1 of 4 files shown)


data/migrations/v1.0.0-point.sql
```diff
@@ -10,3 +10,39 @@ UPDATE
...
+DROP INDEX IF EXISTS planet_osm_point_min_zoom_pois_12_index;
+ALTER INDEX new_planet_osm_point_min_zoom_pois_12_index RENAME TO planet_osm_point_min_zoom_pois_12_index;
+COMMIT;
+
+BEGIN;
+DROP INDEX IF EXISTS planet_osm_point_min_zoom_pois_15_index;
+ALTER INDEX new_planet_osm_point_min_zoom_pois_15_index RENAME TO planet_osm_point_min_zoom_pois_15_index;
+COMMIT;
+
+DROP INDEX IF EXISTS planet_osm_point_min_zoom_way_index;
```
zerebubuth: I think the existing index is called `planet_osm_point_min_zoom_wa9_index` for the `IS NOT NULL` constraint.

rmarianski: Good catch.

rmarianski: Updated in 58c85dc.

zerebubuth:
:+1: looks good

Issue #605, opened 2016-03-13
Add highway=corridor and crossing=*, sidewalk=* tags for roads layer


nvkelso:
**New `roads` layer features:**

* `highway=corridor` - starting at zoom 16, to support indoor mapping and layering on top of buildings. [Example in SFO](https://www.openstreetmap.org/way/300867020), read more on [OSM wiki]()

**New attributes on `roads` layer features:**

* `crossing=*` - starting at zoom 17, to support pedestrian mapping. [Example in NYC](http://www.openstreetmap.org/way/382535027), read more on [OSM wiki](http://wiki.openstreetmap.org/wiki/Key:crossing). Drop this attribute at earlier zooms.
* `sidewalk=*` - starting at zoom 15, to support pedestrian mapping. [Example in SF](http://www.openstreetmap.org/way/23766104), read more on [OSM wiki](http://wiki.openstreetmap.org/wiki/Sidewalks). Drop this attribute at earlier zooms.

rmarianski:
It looks like elements can have [crossing](http://taginfo.openstreetmap.org/keys/crossing#values) and [sidewalk](http://taginfo.openstreetmap.org/keys/sidewalk#values) keys, as well as having [footway values](http://taginfo.openstreetmap.org/keys/footway#values). Do we want to normalize these somehow?

I originally read the issue as passing the truthy `crossing` and `sidewalk` tags through to the tile properties, but the [nyc example](http://www.openstreetmap.org/way/382535027) only has `footway=crossing` and no `crossing` tag.

nvkelso:
We added `footway=*` in the v0.10 release so that's already done.

nvkelso:
We added some `footway=*` (mostly sidewalk and crossing) in v0.10 release. Let's take this opportunity to normalize (whitelist) values and share that whitelist with `crossing=*` and `sidewalk*=*` values.

nvkelso:
**Corridors** are coming thru, but we might want to set their default sort order above buildings?

I've adjusted them above in the screenshot below:
18/37.61770/-122.38177

<img width="843" alt="screen shot 2016-08-09 at 22 09 04" src="https://cloud.githubusercontent.com/assets/853051/17542484/04ceda06-5e7e-11e6-9620-7fe03fe275f2.png">

nvkelso:
[...]

PR #986, opened 2016-08-23
Include sidewalk_(left|right) properties

rmarianski:
Connects to #605


rmarianski:
Do we want to also include sidewalk values of `no`? The [test added](http://www.openstreetmap.org/way/438362919) is a case.

nvkelso:
LGTM

Issue #976, opened 2016-08-18
Tiles should include POIs, places that have fractional zoom


nvkelso:
Olga noticed that some tiles didn't include features with a fractional zoom until the next rounded up full integer zoom. For instance: if the POI is min_zoom 13.2 it would be expected to be in the zoom 13 tile, but it is only appearing in the zoom 14 tile.

zerebubuth:
Yup. There's a lot of places where we check `min_zoom <= zoom`, but if `zoom` only takes integer values, then that's wrong, it should be `min_zoom < zoom + 1`. There are a couple of solutions to this:

1. Fix that code in all the places it exists.
2. Don't support fractional `min_zoom`.

I'm in favour of the latter, as `min_zoom` is really only a hint anyway.

nvkelso:
I'd like to take door 1, fix the code. 

We curate some neighbourhood `min_zoom` values upstream in Who's On First assuming they will show up at the fractional zoom, but since the feature is only available at the next integer zoom it doesn't work. We also spend a effort getting fractional zooms on POIs. When we don't include these earlier it creates sparse looking vector maps as the fraction gets closer to the next integer (e.g. zoom 13.8 looks sparse compared to zoom 13 or zoom 14).

rmarianski:
What does fixing the code mean? Would we make the min zoom values floats throughout? And then have the queries ask for features < zoom + 1?

nvkelso:
Yes, I think so.

The min_zoom values with 13.86747474 are not super useful client side, so
maybe pair this with a final transform that clips them to 13.86 or 13.867
(after tile kind budgets might sort and trim features)? Data curation (e.g.
in Who’s On First) would almost only be to 13.8 tenths, but collision
sorting client side is probably useful to hundredths or thousandths.

zerebubuth:
We currently don't have any way of getting fractional zooms from `places` - all the NE data pulls in `scalerank`, which is an integer, and the values in `places.yaml` are all integers. Do we want to make the change anyway so that we can do it in the future?

nvkelso:
I think the `places` layer could contain fractional zooms when sourced from Who's On First: 

- https://github.com/tilezen/vector-datasource/blob/master/queries/places.jinja2#L77

But because our WOF table schema incorrectly converts Who's On First floats to ints we only get ints out?

[...]

PR #989, opened 2016-08-24
Support fractional zoom for POIs and places

zerebubuth:
This changes the database so that places can have fractional `min_zoom` - this was already supported on POIs. The `min_zoom` is rounded in the filters to 2 decimal places. A test was added to ensure that a POI with `min_zoom: z+f` with `0 <= f < 1` is included in zoom `z`.

Connects to #976.


rmarianski:
@zerebubuth fyi I rebased all changes off of latest master and force pushed. Sorry if this causes any trouble for you.

zerebubuth:
Changes in 4ce389c look good to me :+1:

Issue #952, opened 2016-08-09
Drop id property on merged features


nvkelso:
Roads at low and mid zooms are merged, but we're exporting a bogus (randomly picked ID). We should just drop the id instead.

nvkelso:
Zoom 14 and zoom 10 tiles no longer include merged feature IDs. Zoom 14 seems like there is a minor configuration issue, let's track that with #1002.

PR #991, opened 2016-08-24
Remove id property from merge functions

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/952

I took the liberty of consolidating all the merge functions as part of this pr. It was already pretty well factored with pre/post hooks, so I grew support for merging either lines or polygons.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -2816,89 +2816,6 @@ def _thaw(thing):
     return thing
 
 
-def merge_features(ctx):
-    """
-    Merge (linear) features with the same properties together, attempting to
-    make the resulting geometry as large as possible. Note that this will
-    remove the IDs from any merged features.
-
-    At the moment, only merging for linear features is implemented, although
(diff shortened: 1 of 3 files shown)


vectordatasource/transform.py
```diff
@@ -2952,20 +2895,13 @@ def _merge_features_by_property(features, drop_props_fn=None,
             new_features.append((shapes[0], orig_props, fid))
```
zerebubuth: Should we restore the `id` here if the feature is unique and cannot be merged?

rmarianski: I expect that should be the behavior, because the `orig_props` are used here, ie a copy before the `props` are modified. I should probably add a test to verify.

zerebubuth: Ooops! Missed that it was `orig_props`...

zerebubuth:
:+1: Nice refactor, too!

rmarianski:
Added more tests in 205a649.

PR #994, opened 2016-08-25
Start querying for pois at z4

rmarianski:
I meant to include this as part of the index updates, but it slipped through.


zerebubuth:
:+1:

Issue #979, opened 2016-08-19
Update shapefile import process


rmarianski:
Reprojecting some of the shapefiles to mercator can result in some invalid geometries.

The current process imports the shapefiles via make files using ogr2ogr. Previously, we had resolved this just by calling st_makevalid in the database after import, and then re-exported the fixed data into the shapefiles that were checked into the repo.

What we might want to do instead is move the import process out of make files, and instead have a python script that drives the process. This will allow us to "fix" the geometries directly, and we can probably just go from source shape file to database directly. We've needed to grow a python script to tile one of the ne shapefiles, so we're being guiding down this path anyway.

nvkelso:
+1

nvkelso:
We've made progress on this in the last few releases, closing.

Issue #931, opened 2016-08-02
Additional locality changes for places layer


nvkelso:
Picking up where https://github.com/tilezen/vector-datasource/issues/840 left off, there are a couple more changes for the "locality" and "region"-alization of the places layer.

1. When we say `state_capital: true`, we should really be saying `region_capital: true`.
2. The Natural Earth YAML lookup for country and region capitals needs to specify `kind: locality` (not `kind: city`).
3. The Naturl Earth YAML lookup for generic `Populated place` should specify `kind: locality` (not `kind: city`).
4. We should also specify the `kind_detail` for Scientific Stations coming from Natural Earth as `scientific_station`.
5. (NEW) When we say `capital: true`, we should really be saying `country_capital: true`.

nvkelso:
We should also specify the `kind_detail` for Scientific Stations coming from Natural Earth as `scientific_station`.

nvkelso:
**tl;dr**

Problem that `region_capital: false,` is showing up in tiles. It should only ever export when it's `true`.

Test case Washington DC (osm ID: 158368533) in http://tile.dev.mapzen.com/vector/v1/all/8/73/97.topojson.

```
properties: {
...
scalerank: 13,
capital: "yes",
admin_level: "2",
population: 646449,
kind: "locality",
name: "Washington",
region_capital: false,
id: 158368533,
...
}
```

===

**Non-consequential:**

Huh, this is producing odd outputs in the debug tool for house styles:

![image](https://cloud.githubusercontent.com/assets/853051/17960142/287e6980-6a59-11e6-9a2c-3fceb33e3233.png)

But the underlying data in [8/41/98.topojson](https://tile.dev.mapzen.com/vector/v1/all/8/41/98.topojson) looks as we expected it to:

```
{
type: "Point",
properties: {
kind: "locality",
scalerank: 13,
name: "Sacramento",
name:lt: "Sakramentas",
kind_detail: "city",
region_capital: true,
source: "openstreetmap.org",
name:ja: "サクラメント",
name:uk: "Сакраменто",
capital: "4",
name:ru: "Сакраменто",
name:ta: "சேக்ரமெண்டோ",
id: 150959789,
population: 453781
},
coordinates: [
618,
800
]
},
```

nvkelso:
WFM

PR #1003, opened 2016-08-29
Exclude region_capital=false

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/931

I added a little shim in the apply non planet sql to try and adjust for the invalid geometries. This should be taken out once a better fix, like https://github.com/tilezen/vector-datasource/issues/979 lands.


data/apply-updates-non-planet-tables.sql
```diff
@@ -1,3 +1,23 @@
...
+                          THEN ST_MakeValid(the_geom)
+                     ELSE the_geom END)
+  WHERE NOT ST_IsValid(the_geom);
+DELETE FROM ne_10m_admin_0_boundary_lines_land WHERE NOT ST_IsValid(the_geom);
+
+UPDATE ne_10m_admin_1_states_provinces_lines
+  SET
+    the_geom = (CASE WHEN GeometryType(ST_MakeValid(the_geom)) = 'MULTILINESTRING'
+                          THEN ST_MakeValid(the_geom)
+                     ELSE the_geom END)
```
zerebubuth: What do we want to do about geometries which are invalid, but `st_makevalid` results in something other than a multilinestring? Are there any?

data/apply-updates-non-planet-tables.sql
```diff
@@ -83,9 +102,28 @@ UPDATE ne_10m_admin_1_states_provinces_lines
...
-DO $$
-BEGIN
+UPDATE water_polygons SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE land_polygons SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE ne_110m_ocean SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE ne_50m_ocean SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE ne_10m_ocean SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE ne_110m_coastline SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE ne_50m_coastline SET mz_label_placement = ST_PointOnSurface(the_geom);
+UPDATE ne_10m_coastline SET mz_label_placement = ST_PointOnSurface(the_geom);
```
zerebubuth: Coastlines are multilinestrings, and don't have names - do we need a label placement on them?

zerebubuth:
:+1: for the `region_capital` changes. Ignore the other comments - I should have read your comment that they're temporary before reviewing them, sorry!

data/apply-updates-non-planet-tables.sql
rmarianski: In fact, I think most of them fall into this category, and that's why I didn't leave it as a simple update. I drop them on the next line. Is there a better alternative?

data/apply-updates-non-planet-tables.sql
rmarianski: Good point. Should we just leave these all as null then? Or should we have not even added a column for these in the first place to ensure we never query for labels against these tables?

rmarianski:
Yea, they are temporary in the sense that I'm hoping we'll provide a more robust solution soon. But, it's looking likely that this is what will get deployed for 1.0.

data/apply-updates-non-planet-tables.sql
zerebubuth: Ooops, yeah, I missed that next line. Sorry!

data/apply-updates-non-planet-tables.sql
zerebubuth: We should probably never query these tables for labels. We can just return NULL in the query in `water.jinja2`. It's an optimisation anyway, since the query returns `name` as NULL, so it will never generate a label placement point.

rmarianski: Rebased the pr onto master, and updated to remove `mz_label_placement` from ne coastline tables in 5211aa2.

rmarianski:
Also mirror `region_capital` logic to `capital` in query.

rmarianski:
And name it `country_capital` instead of `capital`.

queries/places.jinja2
```diff
@@ -38,7 +38,7 @@ SELECT
     NULL AS labelrank,
     tags->'population' AS population,
     tags->'capital' AS capital,
```
nvkelso: Can you set this as `country_capital`, and looks like Ottowa is getting `capital:yes` exported now, instead of `true`.

nvkelso: `country_capital` would require another change in the YAML file for Natural Earth.

rmarianski:
And for ne data as well.

rmarianski:
Updates in f9330a9.

rmarianski:
Should be good to go now.

nvkelso:
LGTM

nvkelso:
I believe this is ready for merging (I've given an R, Matt gave an R in https://github.com/tilezen/vector-datasource/pull/1003#issuecomment-243418201.)

Issue #896, opened 2016-07-14
Convert Natural Earth's road levels into shield enums, cleanup names


nvkelso:
Picking up a thread in https://github.com/tilezen/vector-datasource/issues/890#issuecomment-232793990...

We're exporting Natural Earth's road `level` values in tiles now, but they're only really useful for generating road shields. But they're not really setup that way. Let's use this as a stub for what highway shield enums should look like per country (country codes in the `sov_a3` and `ref` values are in the `name` column (I've verified there isn't super long text in there).

- `Interstate` – surprisingly these are in all three `USA`, `CAN`, `MEX`, with the vast majority being in `USA` only.
- `E` - E roads in Europe, but Europeans don't always seem to take kindly to these types of shields (preferring their own national shields). 
- `Federal` – with sov_a3 values of `USA`, `CAN`, `MEX`, and ` ` which looks to be New Zealand (and can be limited using the `continent` value of **Oceania**.
- `Ferry` – there are a few ferries with "shield text" ref values in the name column. They look to all be weird E road continuations.
- `State` - No indication of which state is given.
- `Other` - There are a few name values, but not many. Random small roads around North America.
- `` (empty), same as `U/C` below, there are a lot of these.
- `U/C` – there are only 5 of these

zerebubuth:
A European's perspective on "E" roads: They're a post-hoc addition, often not visible on signage. For example, the [E-30](https://en.wikipedia.org/wiki/European_route_E30) follows the M4 motorway east from south Wales towards London and (according to Wikipedia) turns north onto the M25 London orbital motorway. [Mapillary has a good photo of the junction signage](http://www.mapillary.com/map/im/7aUCr0DXqx2rgvegIHfwwA/photo), which shows M25 to Watford, but no E-30 to Moscow. So if E-30 was shown on the map instead of M25, then I'd be pretty confused.

It would be nice to have the information in tiles and perhaps show E roads as extra, secondary shields if there's space, but showing them instead of national shields would probably make it more difficult for people to navigate.

nvkelso:
Merging #900 here as it's all the same work:

* **What did you see?** Refs are actually in the names, there are more potential refs in the alt names.
* **What did you expect to see?** Less junk!

* add `ref: { col: name }` to the `&ne_properties block` when it's not one of (Sunshine Coast Highway, Seward Hwy., and Portage Glacier Hwy.).
* If it is one of those 3 values (or just a long shield text?), then put it into `name`.
[...]

PR #1004, opened 2016-08-30
Add shield text & networks for NE roads.

zerebubuth:
Select networks from NE roads and encode compatible network names to OSM.

Connects to #896.

@rmarianski, @nvkelso could you review, please?


rmarianski:
:+1:

Issue #719, opened 2016-04-14
Add kind_detail values for pois & other feature types


nvkelso:
**tl;dr**

Add new `kind_detail` for:

-  `restaurants` should additionally map the `cuisine` values into `kind_detail`
- `pitches` should additionally map the `sport` values into `kind_detail`

===

We currently provide more granular details needed for displaying roads in various `highway`, `aerialway`, & etc tags. Similarly `restaurants` can be paired with a `cuisine` value for greater granularity. For pitches, the `sport` value. Let's formalize that and store in a generic `kind_detail` property that can be referenced. `kind_detail` values should be unique (within a parent `kind`, and across parent `kind`s).

This can be done for the `pois` layer now, `roads` layer soon. Other layers and victory!

_(An earlier version of this referred to `subkind` instead of `kind_detail`.)_

nvkelso:
Related work was done in https://github.com/tilezen/vector-datasource/issues/970 for roads.

iandees:
Should the new key be `subkind` or `kind_detail`?

nvkelso:
@iandees We've standardized on `kind_detail`.

iandees:
👍  I updated the description to reference `kind_detail` rather than `subkind` so I don't confuse my future self.

nvkelso:
Verified :)

Issue #1008, opened 2016-08-30
Move optional pois layer JINJA tags to YAML


nvkelso:
Following up with https://github.com/tilezen/vector-datasource/issues/719, in moving `cuisine` and `sport` values to `kind_detail` we should remove those properties (in keeping with the changes in the roads layer in the v1 release).

While dropping `cuisine` looks straight forward, there are a couple other features that still are interesting with `sport`, I think, besides pitches (in landuse and pois)? 

These often map to several (but not many) kinds:

```
    tags->'cuisine' AS cuisine,
    tags->'religion' AS religion,
    tags->'sport' AS sport,
    (transit_routes).score AS mz_transit_score,
    (transit_routes).root_relation_id AS mz_transit_root_relation_id,
    (transit_routes).train_routes AS train_routes,
    (transit_routes).subway_routes AS subway_routes,
    (transit_routes).light_rail_routes AS light_rail_routes,
    (transit_routes).tram_routes AS tram_routes,
    tags->'ref' AS ref,
    tags->'attraction' AS attraction,
    tags->'zoo' AS zoo,
    tags->'exit_to' AS exit_to,
```

I think  `tags->'operator' AS operator,` is interesting to have on all features, so propose leaving that in jinja file.

rmarianski:
For everything we drop from the queries, it's worth checking the queries.yaml file as well as the transforms to see if it gets used downstream. For example, there's label placement logic related to the [sport](https://github.com/tilezen/vector-datasource/blob/9526e79b73e01da8cc8b444fcf76a39879b6a89a/queries.yaml#L472) tag, which gets dropped in a subsequent [transform](https://github.com/tilezen/vector-datasource/blob/9526e79b73e01da8cc8b444fcf76a39879b6a89a/queries.yaml#L482-L491) at lower zooms.

nvkelso:
This was completed in Fall 2017.

PR #1014, opened 2016-08-30
Add kind_detail for leisure=pitch and amenity=restaurant

iandees:
- [x] Update tests
- [x] Update data migrations
- [ ] Update docs

Connects to #719.

Also removes the `cuisine` and `sport` attributes from features.


queries/pois.jinja2
```diff
@@ -7,9 +7,7 @@ SELECT
     way_area AS area,
     tags->'protect_class' AS protect_class,
     tags->'operator' AS operator,
-    tags->'cuisine' AS cuisine,
     tags->'religion' AS religion,
-    tags->'sport' AS sport,
```
nvkelso: Please don't remove the general sports tag now (but it's okay to remove the general cuisine tag).

I think there are a few more kinds that have useful sports tags. Let's not take sport out until we've sorted out which is which, and that work is bucketed into https://github.com/tilezen/vector-datasource/issues/1008 in the next release.

nvkelso:
Besides https://github.com/tilezen/vector-datasource/pull/1014#discussion-diff-76898795, LGTM 👍

queries/pois.jinja2
nvkelso: Fixed in 1548a6977e70173ef6974ee768d54c7609c4793c.

nvkelso:
Ready for merge.

Issue #679, opened 2016-04-05
only calculate label placements for landuse layer at later zooms


nvkelso:
**tl;dr** Olga already moved most important landuse features to the pois layer. What remains there should only receive calculated label placements at later zooms. I propose zoom 15+.

===

Looks like we need ensure all landuse kinds in the lookup table are represented **both** in the [landuse.csv](https://github.com/mapzen/vector-datasource/blob/master/spreadsheets/kind/landuse.csv) table and in the [pois.csv](https://github.com/mapzen/vector-datasource/blob/master/spreadsheets/kind/pois.csv) table. Right now the several `park` lookups are only in the landuse table (not the pois table), so attraction wins when it does get to the `pois` layer, or they are orphaned in the `landuse` layer if there is no equivalent rule in the `pois` layer. 

The proposed rule of thumb is: The landuse, buildings, and water layers should generally only have label placements points for things that don't get icons (are shown with text only labels). 

* **What did you see?** (1) In tile [16/19299/24630](http://vector.mapzen.com/osm/all/16/19299/24630.topojson?api_key=vector-tiles-HqUVidw) (download [zip](
https://github.com/mapzen/vector-datasource/files/204439/24630.json.zip)) I expected to see this park label feature (`id: 218327131`) to show up `pois` layer. (2) Similarly I expected to see this park label feature (`id:22727025`) to show up as `kind:park` in the `pois` layer.
* **What did you expect to see?** Instead I see (1) `218327131`'s label position in the `landuse` layer, and (2) `22727025` as a `kind:attraction` in the `pois` layer.
* **What map location are you having problems with?** New York City, near http://tangrams.github.io/bubble-wrap/#18/40.75397/-73.98267. 
* **Screenshot?** 

<img width="965" alt="screen shot 2016-04-05 at 08 33 16" src="https://cloud.githubusercontent.com/assets/853051/14281794/1a6ba6f6-fb09-11e5-892c-9cffd17fec0a.png">

**1) for park in landuse > pois:**

```
{
type: "Polygon",
properties: {
kind: "park",
name: "Special Permit Plaza",
area: 2045,
sort_key: 24,
source: "openstreetmap.org",
min_zoom: 11.757,
id: 218327131
},
arcs: [
[
502
]
]
},
{
type: "Point",
properties: {
label_placement: "yes",
kind: "park",
name: "Special Permit Plaza",
area: 2045,
sort_key: 24,
source: "openstreetmap.org",
[...]

PR #1015, opened 2016-08-31
remove label placements early for landuse and buildings

nvkelso:
Connects with #679 to remove label placements early for landuse and buildings.

- [x] **Update tests** - new test added.
- [ ] **Update data migrations** - none required, these are all post processing steps.
- [ ] **Update docs** - we don't make promises about zoom ranges of label positions.

### DIFF
--- a/integration-test/679-less-landuse-building-label-placements.py
+++ b/integration-test/679-less-landuse-building-label-placements.py
@@ -2,7 +2,7 @@
 # http://www.openstreetmap.org/way/82206919
 assert_no_matching_feature(
     14, 2620, 6330, 'landuse',
-    {'id': 82206919, 'kind': 'pier', 'label_placement': type(None)})
+    {'id': 82206919, 'kind': 'pier', 'label_placement': True})
 
 assert_has_feature(
     15, 5240, 12661, 'landuse',
@@ -13,7 +13,7 @@ assert_has_feature(
 # http://www.openstreetmap.org/way/260520160
 assert_no_matching_feature(
     15, 5242, 12663, 'buildings',
-    {'id': 260520160, 'kind': 'building', 'label_placement': type(None)})
+    {'id': 260520160, 'kind': 'building', 'label_placement': True})
 
 assert_has_feature(
     16, 10484, 25326, 'buildings',
--- a/integration-test/679-less-landuse-building-label-placements.py
+++ b/integration-test/679-less-landuse-building-label-placements.py
@@ -2,7 +2,7 @@
 # http://www.openstreetmap.org/way/82206919
 assert_no_matching_feature(
     14, 2620, 6330, 'landuse',
-    {'id': 82206919, 'kind': 'pier', 'label_placement': None})
+    {'id': 82206919, 'kind': 'pier', 'label_placement': type(None)})
 
 assert_has_feature(
     15, 5240, 12661, 'landuse',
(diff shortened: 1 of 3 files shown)


integration-test/679-less-landuse-building-label-placements.py
```diff
@@ -0,0 +1,21 @@
+# The landuse for a pier
+# http://www.openstreetmap.org/way/82206919
+assert_no_matching_feature(
+    14, 2620, 6330, 'landuse',
+    {'id': 82206919, 'kind': 'pier', 'label_placement': None})
```
rmarianski: To check that a property doesn't exist, you should check for `type(None)`. Otherwise I believe it will try to match that it exists with a value of null. For example:

`{'id': 82206919, 'kind': 'pier', 'label_placement': type(None)}`

integration-test/679-less-landuse-building-label-placements.py
```diff
@@ -0,0 +1,21 @@
...
+assert_has_feature(
+    15, 5240, 12661, 'landuse',
+    {'id': 82206919, 'kind': 'pier', 'label_placement': True})
+
+
+# The building known as 650 California Street
+# http://www.openstreetmap.org/way/260520160
+assert_no_matching_feature(
+    15, 5242, 12663, 'buildings',
+    {'id': 260520160, 'kind': 'building', 'label_placement': None})
```
rmarianski: `type(None)`

nvkelso: Thanks! Fixed in 13640d1b1af9a0e772b52e0e6f8dd211cc6d0bc8.

integration-test/679-less-landuse-building-label-placements.py
nvkelso: Thanks! Fixed in 13640d1b1af9a0e772b52e0e6f8dd211cc6d0bc8.

nvkelso:
Not sure why this is failing, appears to be for unrelated 583-merge-landuse.py test:

```
[  67/141] FAIL: 'integration-test/583-merge-landuse.py'
  File "./scripts/../integration-test.py", line 465, in run_test
    'layers_in_tile': layers_in_tile,
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 240, in run_path
    return _run_module_code(code, init_globals, run_name, path_name)
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 82, in _run_module_code
    mod_name, mod_fname, mod_loader, pkg_name)
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "integration-test/583-merge-landuse.py", line 666, in <module>
    feature_props.add(props)
  File "/opt/circleci/python/2.7.10/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "./scripts/../integration-test.py", line 184, in features_in_tile_layer
    yield features
  File "integration-test/583-merge-landuse.py", line 665, in <module>
    % feature['properties'])
Exception: Tile 'http://localhost:59464/landuse/9/245/166.json': Duplicate properties {u'kind': u'forest', u'area': 259280, u'sort_key': 31, u'source': u'openstreetmap.org', u'min_zoom': 9.0, u'id': 264004272} in landuse layer, but properties should be unique (except ID, area).
```

nvkelso:
@zerebubuth any advice on test failure?

zerebubuth:
Fixed in 7920b2b - the issue was that the zoom cut-off which prevented the `handle_label_placement` step from generating label placements also stopped it from removing the `mz_label_placement` column. That column, being effectively unique, meant that no merging between polygons could happen. I added a couple more filters to remove that column at other lower zooms.

nvkelso:
@zerebubuth thanks!

Issue #1016, opened 2016-08-31
Missing localized names on region boundary


nvkelso:
* **What did you see?** I only see default names (including left and right names, see below).
* **What did you expect to see?** The [source data from OSM](http://www.openstreetmap.org/relation/165473) has names in a couple dozen languages (see below).
* **What map location are you having problems with?** 
  - http://tile.dev.mapzen.com/vector/v1/all/8/42/96.topojson?api_key=vector-tiles-HqUVidw

```
type: "LineString",
properties: {
kind: "region",
name: "Nevada - California",
sort_key: 250,
kind_detail: "4",
id:right: -165475,
id:left: -165473,
name:right: "California",
id: -165473,
osm_relation: true,
name:left: "Nevada"
},
arcs: [
0
]
```

From OSM:

```
name:ar	نيفادا
name:arc	ܢܝܒܐܕܐ
name:bn	নেভাডা
name:el	Νεβάδα
name:en	Nevada
name:eo	Nevado
name:fa	نوادا
name:hak	Nui-fà-tha̍t
name:he	נבדה
name:hi	नेवाडा
name:hy	Նևադա
name:ja	ネバダ州
name:ka	ნევადა
name:ko	네바다
name:ml	നെവാഡ
name:mr	नेव्हाडा
name:nv	Ná'aazhoozh Hahoodzo
name:os	Невадæ
name:pl	Nevada
name:qu	Nevada suyu
name:ru	Невада
name:sr	Невада
name:szl	Newada
name:ta	நெவாடா
name:th	มลรัฐเนวาดา
name:tl	Nebada
name:uk	Невада
name:ur	نیواڈا
name:yi	נעוואדא
name:zh	内华达州
```

nvkelso:
(Found this in completing https://github.com/tilezen/vector-datasource/issues/1012.)

iandees:
So far my exploration has shown that the New York/New Jersey border [on prod](https://vector.mapzen.com/osm/boundaries/15/9643/12327.topojson) has New Jersey on the right while it is currently on the left [on dev](https://vector.dev.mapzen.com/osm/all/15/9643/12327.topojson). Also, the prod tile has a geometry for "New York - New Jersey" and a separate geometry for "New Jersey" while the dev tile has a geometry for "New Jersey - New York" and a separate geometry for "New York".

[...]

PR #1022, opened 2016-09-01
Fix missing localized names on boundaries

iandees:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

Connected to #1016.

### DIFF
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -1,5 +1,6 @@
 {% macro ne_boundary_cols() %}
     gid AS __id__,
+    'naturalearthdata.com' AS source,
     {% filter geometry %}the_geom{% endfilter %} AS __geometry__,
     mz_calculate_json_boundaries(t.*) AS mz_properties
 {% endmacro %}
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -60,6 +60,7 @@ WHERE
 
 SELECT
   tags->'name' AS name,
+  'openstreetmap.org' AS source,
   NULL AS maritime_boundary,
   mz_calculate_json_boundaries(planet_osm_polygon.*) AS mz_properties,
   -- note that we force the RHR, which makes outers clockwise,
@@ -84,6 +85,7 @@ UNION ALL
 
 SELECT
   NULL AS name,
+  'openstreetmap.org' AS source,
   true AS maritime_boundary,
   '{"kind": "maritime"}'::json as mz_properties,
   {% filter geometry %}{{ bounds['polygon']|bbox_intersection('the_geom',3857) }}{% endfilter %} AS __geometry__,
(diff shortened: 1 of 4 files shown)


queries/boundaries.jinja2
```diff
@@ -60,6 +60,7 @@ WHERE
 
 SELECT
   tags->'name' AS name,
+  'openstreetmap.org' AS source,
```
nvkelso: Nice catch!

Please also specify Natural Earth as a source at the top of the file in the macro:

```
{% macro ne_boundary_cols() %}
     gid AS __id__,
     {% filter geometry %}the_geom{% endfilter %} AS __geometry__,
     mz_calculate_json_boundaries(t.*) AS mz_properties
 {% endmacro %}
```

iandees:
The tests appear to be failing because the right- and left-handedness of the New York/New Jersey boundary is flipped again. Is that arbitrary?

zerebubuth:
The code is [intended to work from a stable orientation](https://github.com/tilezen/vector-datasource/blob/f151b30ef8bd3ffe143a9313e204e2ceb005567a/queries/boundaries.jinja2#L65-L69) but it looks like the post-processor is [sensitive to feature order](https://github.com/tilezen/vector-datasource/blob/f151b30ef8bd3ffe143a9313e204e2ceb005567a/vectordatasource/transform.py#L1774). I suspect that the left/right are getting flipped because the linestring direction is being flipped - so I think it would still display correctly, which might explain why we haven't noticed this problem until now.

Should we enforce a given ordering, perhaps by sorting on `fid` or `tags['name']` before [that loop in the transform](https://github.com/tilezen/vector-datasource/blob/f151b30ef8bd3ffe143a9313e204e2ceb005567a/vectordatasource/transform.py#L1774)?

iandees:
What the heck? Now the tags aren't there at all?

zerebubuth:
It looks like there was an error around the boundaries test, and it logged this:

```
./scripts/../integration-test.py:148: UnicodeWarning: Unicode unequal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  elif v != exp_v:
```

So perhaps they are there, and the unicode unequal comparison means it's picking up the maritime boundary as the "best match", which wouldn't have the same left/right.

iandees:
That should have been fixed in 39797ff (part of this branch) – I wonder what broke. I'll keep looking locally.

iandees:
This still behaves differently on my local instance, but appears to pass here now.

iandees:
(By "differently" I mean that the New York and New Jersey left/right are flipped.)

zerebubuth:
Are you running the CI process locally (i.e: getting data from Overpass), or testing against a metro (or other) extract? Would it be possible to grab the GeoJSON for the tile on your local machine and compare it to dev and see if it's just the tags which are flipped or the direction of the linestring too?

iandees:
After discussing in person, I'm going to add a sort as described in https://github.com/tilezen/vector-datasource/pull/1022#issuecomment-245256220 so that we get consistent results.

iandees:
Whew, finally got it working.

selfie-0

nvkelso:
@zerebubuth Please review :)

zerebubuth:
Looks good to me :+1:

PR #1031, opened 2016-09-08
Add min_zoom to templated SQL queries

iandees:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

Connects to #976.


queries/boundaries.jinja2
```diff
@@ -84,6 +86,7 @@ UNION ALL
 
 SELECT
   NULL AS name,
+  TODO_FIX_ME as min_zoom,
```
iandees: Not sure what to use here for `buffered_land`.

queries/roads.jinja2
```diff
@@ -3,6 +3,7 @@ SELECT
     gid AS __id__,
     {% filter geometry %}the_geom{% endfilter %} AS __geometry__,
     'naturalearthdata.com' AS source,
+    TODO_FIX_ME as min_zoom,
```
iandees: Not sure what to use here for `ne_10m_roads`.

queries/boundaries.jinja2
zerebubuth: Because we include this at all zoom levels, does `0 AS min_zoom` make sense?

queries/roads.jinja2
zerebubuth: We seem to include roads where `scalerank <= zoom`, so I think that effectively means `scalerank = min_zoom`.

queries/boundaries.jinja2
```diff
@@ -84,6 +86,7 @@ UNION ALL
 
 SELECT
   NULL AS name,
+  0 as min_zoom,
```
nvkelso: Should this kick in at zoom 8, when we switch over from Natural Earth to OpenStreetMap?

nvkelso:
Generally looks good to me. @zerebubuth for a once over.

nvkelso:
@iandees Can you please add tests to verify `min_zoom` values are exporting in those 4 layers (there should be 8 tests: 4 low zoom Natural Earth tests and 4 OpenStreetMap mid-zoom or high zoom tests).

iandees:
Building the fixtures for the tests here is taking a pretty long time. Still going…

iandees:
@rmarianski, can you take a quick peek at this? The various pieces of this have been looked at, but a quick gander over the whole thing would be helpful.

rmarianski:
:+1:

iandees:
Thanks!

Issue #987, opened 2016-08-24
George Washington National Forest should not be a national park


zerebubuth:
[George Washington National Forest](http://www.openstreetmap.org/way/34416231) currently comes through with [`kind: national_park`](https://github.com/tilezen/vector-datasource/blob/57aeb9cd139c699840940ea8aa96f4b92db92788/integration-test/473-landuse-scalerank.py#L25), but [shouldn't](https://github.com/tilezen/vector-datasource/pull/980#issuecomment-242004509).

Relevant tags on the object:

* `boundary=national_park`
* `landuse=forest`
* `leisure=nature_reserve`
* `operator=US Forest Service`

Do we want some combination of these to override the [choice of `kind: national_park`](https://github.com/tilezen/vector-datasource/blob/57aeb9cd139c699840940ea8aa96f4b92db92788/yaml/landuse.yaml#L5-L7)?

nvkelso:
Good catch! Browsing around the USA this seems fairly common and probably the root cause of a lot of our cludgy filter shims. 

If we see the operator is `United States Forest Service` (we normalize a list of values into that in post-processing) it should be `kind:forest` instead of `kind:national_park`. 

I've had good success in the scene file also paring with **protect_class` (to insist that a national_park can't be operated by USFS and it should generally be in only 2, 3, and 5 protect_class :

```
            national_park:
                filter:
                    all:
                        - kind: [national_park, battlefield, protected_area]
                    any:
                        - not: { operator: [ "United States Forest Service" ] }
                        - protect_class: ['2','3','5']
```

- Example [Stanislaus National Forest](http://www.openstreetmap.org/relation/972008) in California near Yosemite. 
- Stanislaus happened to have protect_class = 6, which is how we end up drawing national forests in the scene file. But this other [Humboldt](http://www.openstreetmap.org/relation/2389847) forest in Nevada has protect_class 5 so that's not guaranteed. 

I suspect these might address edge cases, while the dominant case is the example above.

-  `landuse=forest` I think this isn't worth pursing at this time as I've only found [one case](http://www.openstreetmap.org/relation/5579230) which is already addressed by the other logic, and I suspect this may do some harm unless further researched and it's an edge case so skip.

[...]

PR #1034, opened 2016-09-09
Fix filters for national forests and parks

zerebubuth:
Makes filters much stricter to be a national park - now `boundary=national_park` is no longer sufficient, and some combination of `operator`, `protect_class`, `protection_title` and `designation` are needed to be a national park.

Connects to #987.

@nvkelso could you take a look and make sure I captured all the examples, and that the resulting `kind`s and `tier`s are what you expected, please?

### DIFF
--- a/integration-test/987-national-forests.py
+++ b/integration-test/987-national-forests.py
@@ -57,10 +57,13 @@ assert_has_feature(
 # a national park. Alternatively boundary:type=protected_area could give us
 # kind: protected_area instead.
 #
+# This feature also has protection_title=National Forest, and the name is
+# "Arapaho National Forest", so it should probably be classed as a forest.
+#
 # http://www.openstreetmap.org/relation/396026
 assert_has_feature(
     13, 1683, 3103, 'landuse',
-    { 'kind': 'park', 'id': -396026 })
+    { 'kind': 'forest', 'id': -396026 })
 
 # leisure=nature_reserve in North Farallon Islands State Marine Reserve has
 # boundary=protected_area with protect_class=4 in tile 11/323/791. It's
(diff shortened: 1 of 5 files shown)


integration-test/987-national-forests.py
```diff
@@ -0,0 +1,205 @@
...
+
+# Point Reyes is also leisure=nature_reserve but without an operator but with
+# a protect_class of 2 (it should be kind:national_park, but might need some
+# data work?), until data is fixed I think it'd come thru as
+# kind:nature_reserve?
+#
+# http://www.openstreetmap.org/relation/1250137
+assert_has_feature(
+    15, 5192, 12627, 'landuse',
+    { 'kind': 'national_park', 'id': -1250137 })
```
nvkelso: I see, `kind:nature_reserve` looses out because `protect_class`=`2`, sure.

integration-test/987-national-forests.py
```diff
@@ -0,0 +1,205 @@
...
+    { 'kind': 'national_park', 'id': -215231 })
+
+# operator=United States National Park Service and protect_class=2 in
+# Yellowstone National Park with boundary=national_park and
+# leisure=nature_reserve should end up with kind:national_park.
+#
+# https://www.openstreetmap.org/relation/1453306
+assert_has_feature(
+    13, 1591, 2972, 'landuse',
+    { 'kind': 'national_park', 'id': -1453306 })
```
nvkelso: To make Yellowstone work, Point Reyes looses out (until data is cleaned up). Works for me.

nvkelso:
👍 LGTM, let's get it on dev :)

The one caveat is while the `tier` classes are useful for grouping, I'm not sure we should export them in tiles.

PR #1035, opened 2016-09-09
Update NE road properties

zerebubuth:
Connects to #896.

### DIFF
--- a/integration-test/896-ne-shield-enums-2.py
+++ b/integration-test/896-ne-shield-enums-2.py
@@ -1,0 +1,21 @@
+# I think we should remove for v1 (and never should be been included at zoom
+# less than 7 anyhow):
+#
+# Remove properties:
+#  * level
+#  * namealt
+#  * namealtt
+#
+# E30 (M4) in UK
+assert_no_matching_feature(
+    7, 63, 42, 'roads',
+    { 'source': 'naturalearthdata.com', 'kind': 'highway',
+      'level': None })
+assert_no_matching_feature(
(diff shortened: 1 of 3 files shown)


integration-test/896-ne-shield-enums-2.py
```diff
@@ -0,0 +1,35 @@
...
+    { 'source': 'naturalearthdata.com', 'kind': 'highway',
+      'level': None })
+assert_no_matching_feature(
+    7, 63, 42, 'roads',
+    { 'source': 'naturalearthdata.com', 'kind': 'highway',
+      'namealt': None })
+assert_no_matching_feature(
+    7, 63, 42, 'roads',
+    { 'source': 'naturalearthdata.com', 'kind': 'highway',
+      'namealtt': None })
```
rmarianski: Should these tests be for `None` or `type(None)`?

nvkelso:
👍 (modulo Rob's comment above)

integration-test/896-ne-shield-enums-2.py
zerebubuth: These are negative matches (`assert_no_matching_feature`), so the tests could be translated as "assert that there are no NE highways with any `level`, ans so forth. The aim is to make sure that all of the `level`, `namealt` and `namealtt` properties have been removed.

rmarianski: Ah ok, sorry I didn't catch that when reading through it initially. :+1:

PR #1036, opened 2016-09-12
restore buildings to zoom 13

nvkelso:
- [x] Update tile selections &/or transforms
- [x] Update tests
- [x] Data migrations required
- [ ] No doc updates required (fix regression over prod)

### DIFF
--- a/data/migrations/v1.0.0-polygon.sql
+++ b/data/migrations/v1.0.0-polygon.sql
@@ -27,6 +27,12 @@ UPDATE
     AND COALESCE(mz_landuse_min_zoom, 999) <> COALESCE(mz_calculate_min_zoom_landuse(planet_osm_polygon.*), 999);
 
 UPDATE planet_osm_polygon
+  SET mz_building_min_zoom = mz_calculate_min_zoom_buildings(planet_osm_polygon.*)
+  WHERE mz_calculate_min_zoom_buildings(planet_osm_polygon.*) IS NOT NULL;
+    (tags -> 'building' IS NOT NULL AND (way_area >= 1600 OR volume >= 300000 ))
+    AND COALESCE(mz_building_min_zoom, 999) <> COALESCE(mz_building_min_zoom(planet_osm_polygon.*), 999);
+
+UPDATE planet_osm_polygon
   SET mz_label_placement = ST_PointOnSurface(way)
   WHERE mz_label_placement IS NULL;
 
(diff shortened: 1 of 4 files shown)


nvkelso:
@zerebubuth I added a tight data migration just for z13 features (ignoring edge cases for underground and no – those should still be calculated but its a 1% expense).

integration-test/845-buildings-z13.py
```diff
@@ -0,0 +1,6 @@
+# Earlier work in 845 dropped buildings from zoom 13
+
+# http://www.openstreetmap.org/way/23654700
+# http://www.openstreetmap.org/way/60500069
+assert_has_feature(
+    13, 1310, 3170, 'buildings', 'kind': 'building'})
```
zerebubuth: Syntax error here - you need a `{` to begin the tags `dict` literal.

zerebubuth:
This seems to have broken the following tests:

* integration-test/653-unify-building-part.py
* integration-test/834-park-building.py
* integration-test/842-normalize-building-kind.py

Could you check to make sure that the new behaviour is what you want, and that these aren't new regressions, please?

integration-test/845-buildings-z13.py
nvkelso: Thanks! Fixed in 8d54e2955c4dc4f1a95dbe0001b162f70f1d2da0.

data/migrations/v1.0.0-polygon.sql
```diff
@@ -27,6 +27,12 @@ UPDATE
     AND COALESCE(mz_landuse_min_zoom, 999) <> COALESCE(mz_calculate_min_zoom_landuse(planet_osm_polygon.*), 999);
 
 UPDATE planet_osm_polygon
+  SET mz_building_min_zoom = mz_calculate_min_zoom_buildings(planet_osm_polygon.*)
+  WHERE mz_calculate_min_zoom_buildings(planet_osm_polygon.*) IS NOT NULL;
```
rmarianski: trailing semicolon, probably want an `AND` here instead?

nvkelso: Fixed in 990e58ee9bc19e33cc758c40556ca7a71cfc14ec (by removing most of that WHERE line).

rmarianski:
LGTM :+1:

nvkelso:
I think I've fixed the z13 building problem, but am still working on the 3 other integration test failures @zerebubuth called out in https://github.com/tilezen/vector-datasource/pull/1036#issuecomment-246328811.

nvkelso:
Rockin' it, @rmarianski!

rmarianski:
:guitar:

Issue #992, opened 2016-08-25
Dropping raw NE scalerank and labelrank values


nvkelso:
**tl;dr** is https://github.com/tilezen/vector-datasource/issues/992#issuecomment-243209051.

===

We pass this thru from Natural Earth on features like populated places and roads, but now that we have `min_zoom` it's probably not necessary and is confusing to have both.

nvkelso:
But where we invent new scalerank values (like in https://github.com/tilezen/vector-datasource/pull/867), should we call that `scale_rank` instead to follow Tilezen property name conventions?

nvkelso:
We'd keep the `min_zoom: scalerank` logic in the YAML files for these layers, but generally we'd remove reference to `scalerank` and `labelrank` from the JINJA files and documentation. 

**scalerank**

- [ ] For all of these, make sure `min_zoom` property exports in tiles (with the value of scalerank or other logic). Then remove raw &/or processed `scalerank` on the following 3 layers:
- [ ] boundaries
- [ ] places - must be paired with https://github.com/tilezen/vector-datasource/issues/981 otherwise house styles will break.
- [ ] places transforms (for adding scalerank to OSM values) & related transform tests
- [ ] roads

**labelrank**

_In the future we might look at these to remove more name and shield values from features. But for now let's remove reference to them in exported tile content._

-  [ ] boundaries - is primarily concerned with disputed boundaries, doesn't look at tiny country boundaries (which would really help us).
- [ ] places - confirmed the label rank values are gibberish. 
- [ ] roads - looks like this was only populated in Natural Earth for Europe, and would need more work to make sense of. 

Would also need updates to (manually maintained) tilejson.json as we've removed properties.

- [ ] tilejson.json

nvkelso:
Looks like Bubble Wrap relied on `scalerank` for region (state) boundary lines filtering and there is an explosion of lines all over the map, especially in Europe.

I think those values should be in the `min_zoom` property now (which is also missing, see https://github.com/tilezen/vector-datasource/issues/976#issuecomment-245500564) so this should "work" once that bug has been fixed. 

Except... We should guard against bad `scalerank`=0 values from Natural Earth and instead force those to be `scalerank`=7 (or `min_zoom` of 7).

nvkelso:
In the boundaries YAML use a new CASE statement for the min_zoom assignment.

nvkelso:
WFM

PR #1037, opened 2016-09-12
fix scalerank 0 region boundaries, drop low-zoom names, add map_unit

nvkelso:
connects to #992 to fix scalerank 0 region boundaries
also adds map_unit boundaries and
drops name at low zooms 0,1,2,3,4,5,6

- [x] Update tests
- [x] Update data download and config
- [x] Update docs


nvkelso:
@zerebubuth Looks like my adding `map_unit` as a new Natural Earth data source is making all the tests fail. Can you take a look, please?

zerebubuth:
The first error seems to be:

```
ERROR:  column "mapcolor9" of relation "ne_10m_admin_0_boundary_lines_map_units" does not exist
invalid command \N
invalid command \.
ERROR:  syntax error at or near "5.00000000000"
LINE 1: 5.00000000000 Map unit boundary 
        ^
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

Which is preventing the shapefiles for the map units being loaded. It looks like the [schema](https://github.com/tilezen/vector-datasource/blob/fae00e841d1497cd7dd72396a8728de134b30cf2/data/shapefile_schema/ne_10m_admin_0_boundary_lines_map_units.sql) doesn't contain a column called !mapcolor9", but the data does. We don't currently support loading a subset of the columns - they must all be specified in the schema.

The second error seems to be:

```
psql:apply-updates-non-planet-tables.sql:107: ERROR:  column "scalerank" does not exist
LINE 15: ...recla" = 'Admin-1 region boundary' THEN CASE WHEN scalerank ...
```

Which probably means that `extra_columns: scalerank` needs to be specified when using the global [`ne_region_boundaries_min_zoom`](https://github.com/tilezen/vector-datasource/pull/1037/files#diff-dca2f4be5bc54303506aa7848a476050R10).

nvkelso:
Thanks! Fixed in b39f38bb911dd40b3081af9dc291fd22ecb51117 and 26e1c1562c59d22fc050abb41c96d6f751409c8f, waiting for CirlceCI to confirm.

yaml/boundaries.yaml
```diff
@@ -75,3 +80,21 @@ filters:
...
+    table: ne
+  - filter:
+      featurecla:
+        - Admin-1 boundary
+        - Admin-1 statistical boundary
+        - Admin-1 statistical meta bounds
+    min_zoom: *ne_region_boundaries_min_zoom
+    extra_columns: [scalerank]
+    output: {kind: region, kind_detail: '4'}
+    table: ne
```
rmarianski: no trailing newline

rmarianski:
LGTM

yaml/boundaries.yaml
nvkelso: Fixed in ad617c94a7e25f1154aabcbeaef6e787e9140e08.

rmarianski:
I checked in some fixes. We'll see if the tests pass on circle now :fingers-crossed:

zerebubuth:
> We'll see if the tests pass on circle now

Do they pass locally? :stuck_out_tongue_winking_eye:

rmarianski:
> Do they pass locally? :stuck_out_tongue_winking_eye:

I'm going to wait until the circle build finishes before answering that :stuck_out_tongue_closed_eyes:

nvkelso:
Our current thinking is the data files here need to be in web Mercator (but are probably in geographic now).

nvkelso:
Last test failures are related to OSM zoom 8 content, what gives?

zerebubuth:
Boundaries looked like they just needed the names swapped, possibly because the direction was reversed due to the introduction of stable ordering in #1022.

The second failure looks like it was also a copy-paste issue with the tile coordinates.

Works for me now, let's see if CircleCI agrees :wink:

nvkelso:
Woot! Merging.

PR #1040, opened 2016-09-13
New fixture shapefiles

zerebubuth:
NOTE: This is a PR into a branch, not master.

I've added new shapefile fixtures. One problem was that the existing ones were in latlon, but the test harness expects them to be already projected to EPSG:3857. I've added a section in the TESTS doc about  this to hopefully clarify this unstated assumption.

The third test is still failing. This seems to be because the road chosen as a fixture has a `scalerank=8`, but we use `scalerank=min_zoom` for roads, which means that it would only appear at zoom 8. However, we switch from NE to OSM at zoom 8 - so this bit of NE data will never be shown.

@iandees could you review, please? Let me know if anything is unclear or if the docs could be further improved.


rmarianski:
:+1:

iandees:
This looks good, thanks.

PR #1041, opened 2016-09-13
Add information to make the test pass.

zerebubuth:
@iandees - hope this helps. I'll annotate the changes individually. For #1031.


integration-test/976-fractional-pois.py
```diff
@@ -12,15 +12,15 @@
     'name': 'New Jersey - Pennsylvania' })
 
 assert_has_feature(
-    5, 9, 12, 'roads',
-    { 'min_zoom': 3.0 , 'id': int, 'sort_key': 383,
+    7, 37, 48, 'roads',
+    { 'min_zoom': 3 , 'id': int, 'shield_text': '95',
     'source': 'naturalearthdata.com' })
```
zerebubuth: I changed the tile here to zoom 7, as we strip some properties from roads when `zoom < 7`, including the `shield_text`, `name` and `ref`. This makes the test quite non-specific, especially as `id` is not reliably constant when coming from shapefile fixtures. At zoom 7, it's still using NaturalEarth data, but the `shield_text` is present so we've more confidence it's matching the feature it was intended to.

integration-test/976-fractional-pois.py
```diff
@@ -12,15 +12,15 @@
...
 
 # There is no transit data from Natural Earth
 
 assert_has_feature(
-    5, 9, 12, 'water',
-    { 'min_zoom': 0 , 'id': 1144,
```
zerebubuth: Zoomed "in" again, so that the data comes from `ne_10m_lakes` rather than one of the other, coarser NE tables.

integration-test/976-fractional-pois.py
```diff
@@ -32,13 +32,15 @@
     'source': 'openstreetmap.org',
     'name': 'New Jersey - New York' })
 
+# http://www.openstreetmap.org/relation/568499
 assert_has_feature(
```
zerebubuth: Added relevant links to OSM data that needs to be pulled in.

integration-test/976-fractional-pois.py
```diff
@@ -47,7 +49,7 @@
 
 assert_has_feature(
     9, 150, 192, 'water',
-    { 'min_zoom': 0, 'id':  10613,
+    { 'min_zoom': 0, 'id': int,
     'source': 'openstreetmapdata.com',
```
zerebubuth: IDs from the shapefiles aren't necessarily constant - they get automatically assigned either by `shp2pgsql` or Postgres (I'm not sure which). So it's safer to not match on `id` and instead just say it's present and an `int`.

Arguably this is a bug, as it means that prod/dev are different from the test instance. However, when it was written, I couldn't think of an easy way to make sure the IDs are the same.

iandees:
Thanks @zerebubuth, this helps. I was trying to use the same tile for the zoomed out stuff to make maintenance easier, but this works.

Issue #1043, opened 2016-09-14
Update not equals filter rule


rmarianski:
The not equals filter rule should be updated to first check that the tag exists before checking the inequality, similar to how the equals rule works.

zerebubuth:
I had to do this for `SetRule` before as well. Should we just extend it to cover all the rules?

rmarianski:
> I had to do this for SetRule before as well. Should we just extend it to cover all the rules?

Is there a way to extend it generically to all rules? Seems like the equals, not equals, set, and existence rules are the ones that need it, and the others are either combinators or it's not relevant.

zerebubuth:
Do we need it for the `GreaterOrEqualsRule` too?

One possible way to do it would be in the combinators, as that's where the problem actually arises. For example, if we have `tags->'foo' >= 0 OR tags->'bar' in ('a', 'b')` then the problem is when `'foo'` doesn't exist and `'bar'` might be `'a'`, then the whole expression returns NULL as the first term is NULL. We could swap the OR check for `case when $first then true when $second then true else false end`? Then we wouldn't need the exists checks... I think?

rmarianski:
> Do we need it for the GreaterOrEqualsRule too?

In theory yes, but not in practice because all the tags are strings and the only places that we use these rules are when we know we have numeric values.

> One possible way to do it would be in the combinators, as that's where the problem actually arises.

I see what you mean. I suppose we just need a function that converts the result to a boolean where null is false, and we can call that on the results in the combinators. I feel like ensuring that we always return a boolean in the leaf rules is more intuitive though.

zerebubuth:
I agree that it's more intuitive to do it in the leaves. And I guess `COALESCE($test, false)` does what we want, and would be fewer tests than doing the check first? Or, perhaps `COALESCE($test, true)` where it's a negative test?

nvkelso:
Related to building filters in https://github.com/tilezen/vector-datasource/pull/1036.

PR #1044, opened 2016-09-14
Update not equals rule to check for tag existence

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1043 and https://github.com/tilezen/vector-datasource/pull/1036.

### DIFF
--- a/data/migrations/create-sql-functions.py
+++ b/data/migrations/create-sql-functions.py
@@ -113,6 +113,15 @@ class Table(object):
         return Column(not not_tag, name)
 
 
+def ensure_sql_boolean(column, check):
+    # tags->'foo' is NULL when the hstore doesn't contain that key,
+    # which makes all the following comparisons NULL, so use coalesce
+    # to ensure we never evaluate to NULL
+    if column._is_tag:
+        check = 'COALESCE(%s, FALSE)' % check
+    return check
+
+
 class EqualsRule(object):
 
     def __init__(self, column, value):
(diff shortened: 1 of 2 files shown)


zerebubuth:
:+1: looks good to me

zerebubuth:
:+1:

rmarianski:
@zerebubuth I liked your `coalesce` idea, and updated the code accordingly in f9fd38b.

data/migrations/create-sql-functions.py
```diff
@@ -113,15 +113,14 @@ def create_column(self, name):
...
-        return '(%s) AND (%s)' % (exists_check, check)
-    else:
-        return check
+def ensure_sql_boolean(column, check, null_value):
+    # tags->'foo' is NULL when the hstore doesn't contain that key,
+    # which makes all the following comparisons NULL, so use coalesce
+    # to ensure we never evaluate to NULL
+    if column._is_tag:
+        sql_on_null = 'TRUE' if null_value else 'FALSE'
+        check = 'COALESCE(%s, %s)' % (check, sql_on_null)
```
zerebubuth: Does this need an 'else' branch in case the column isn't a tag?

zerebubuth: Oh, wait, no I see how this works - it just returns the `check` unmodified.

rmarianski:
Whoops, we actually want the sense on the not equals check reversed, ie evaluate to false if the tag doesn't exist. Updated in 82f202c.

zerebubuth:
Hmm.. So we don't have the meaning that `(A ≠ B) ≡ ¬(A = B)`? Our "not equals" really means "(A is not null and not equal to B) or (A is null)"?

rmarianski:
> Hmm.. So we don't have the meaning that (A ≠ B) ≡ ¬(A = B)? Our "not equals" really means "(A is not null and not equal to B) or (A is null)"?

Yea, currently. I did it that way to match the intent in the yaml. But it's used very rarely, so it might have been better to update the usages in the yaml file instead and revert back the not equals meaning. What do you think?

rmarianski:
Or we can also just get rid of not equals and replace the usages with the more generic `not` operator.

zerebubuth:
The only use I see now (that the underground use has gone) is:

```yaml
  - filter: {disused: -no}
```

Which does really mean "disused is present but not 'no'". Should we change that to:

```yaml
  - filter:
    - disused: True
    - not: { disused: no }
```

?

I think I agree with you that it's preferable, as it's clearer about the case where `disused` is NULL.

rmarianski:
Agreed, updated in acc79b6.

zerebubuth:
:+1: looks good to me!

Issue #1048, opened 2016-09-14
Rename sort_key to sort_rank


nvkelso:
As https://github.com/tilezen/vector-datasource/issues/988 for label priorities didn't make the cut for v1 tiles, there was a backwards incompatible change proposed in there to rename `sort_key` to `sort_rank`. Let's see if we can get that in now.

nvkelso:
Everything was looking pretty crazy for a second there, but updating Bubble Wrap to `sort_rank` fixed everything :)

PR #1049, opened 2016-09-14
Rename sort_key to sort_rank

iandees:
- [x] Update tests
- [ ] No data migrations needed
- [x] Update docs

Connected to #1048.


iandees:
This looks OK to me. @rmarianski did I miss anything on this change?

nvkelso:
Looks good to me.

zerebubuth:
Wow, that's a lot of changes! Nice work :smile:

Should we also change:

```
queries.yaml:324:      cutting_attrs: { sort_key: 'sort_key', reverse: True }
queries.yaml:607:      cutting_attrs: { sort_key: 'sort_key', reverse: True }
```

to

```
queries.yaml:324:      cutting_attrs: { sort_key: 'sort_rank', reverse: True }
queries.yaml:607:      cutting_attrs: { sort_key: 'sort_rank', reverse: True }
```

?

This bit of code/config tries to ensure that when we "overlap" buildings with landuse (the first line) or "cut" roads against landuse (the second line) then we do it in order of decreasing `sort_rank` so that landuse polygons "on top" are the first to be considered. The effect is to assign a `landuse_kind` to the road or building which matches the background landuse polygon, for example so that roads through a national park can have a slightly green casing.

rmarianski:
Only thing I noticed is what @zerebubuth mentioned. Otherwise LGTM.

zerebubuth:
Great! Looks good to me :+1:

PR #1052, opened 2016-09-16
Use a more focused buildings migration

rmarianski:
In https://github.com/tilezen/vector-datasource/pull/1036 the buildings migration was broadened to be safer, but we can narrow its scope to cover any affected features to speed it up.


zerebubuth:
:+1: looks good to me

PR #1053, opened 2016-09-16
Fix migration to use variable name

rmarianski:


zerebubuth:
Yup! Looks great :+1:

rmarianski:
thanks for catching this!

PR #1056, opened 2016-09-19
Configure tilejson api version dynamically

rmarianski:

### DIFF
--- a/tilejson/tilejson.json.erb
+++ b/tilejson/tilejson.json.erb
@@ -17,9 +17,9 @@
     "name" : "<%= @params[:name] %>",
     "private" : true,
     "scheme" : "xyz",
-    "tilejson" : "2.0.0",
+    "tilejson" : "<%= node[:tilejson][:tilejson_version] %>",
     "tiles" : [
-        "http://<%= node[:tilejson][:host] %>/mapzen/vector/<%= node[:tilejson][:vector_tiles_version] %>/all/{z}/{x}/{y}.<%= @params[:extension] %>?api_key={api_key}"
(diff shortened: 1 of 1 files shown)


rmarianski:
Connects to https://github.com/tilezen/vector-datasource/pull/1056

tilejson/tilejson.json.erb
```diff
@@ -17,7 +17,7 @@
     "name" : "<%= @params[:name] %>",
     "private" : true,
     "scheme" : "xyz",
-    "tilejson" : "2.0.0",
+    "tilejson" : "<%= @params[:tilejson_version] %>",
     "tiles" : [
         "http://<%= node[:tilejson][:host] %>/mapzen/vector/<%= node[:tilejson][:vector_tiles_version] %>/all/{z}/{x}/{y}.<%= @params[:extension] %>?api_key={api_key}"
```
zerebubuth: If this is related to mapzen/chef-mapzen_tilequeue#27 then does `:vector_tiles_version` need to change to `:url_path_version`?

rmarianski: Nice catch, updated in b18f1b9.

I also noticed that I was referencing the api version incorrectly too. Should be fixed in 82aa063.

zerebubuth: Does that maens the `:tilejson_version` on L20 also needs to come from `node[:tilejson]`?

rmarianski: Yes ;) Thanks for catching that one too. Fixed in 9481e50.

PR #1059, opened 2016-09-19
Work around differences in mvt decode structure

rmarianski:
Got a test failure on dev related to this for integration-test/1030-invalid-wkb-polygons.py
This probably isn't the best way to do it, but it gets the test to pass :)

Also, on a related note it would be good if the mvt features function would honor the config to use the all layer. That way we wouldn't have to go through the tileserver instance to generate the response. If you agree we should just make an issue for it.


zerebubuth:
:+1:

Follow up issue in #1060.

PR #1061, opened 2016-09-20
Fix four failing dev tests.

zerebubuth:
This fixes:

* **473-landuse-tier** so that a point is tested in isolation at a higher zoom. At the lower zoom, it has the same name as a nearby feature and so is merged with it.
* **742-predictable-layers-pois** so that Arlington National Cemetery is tested against one of the two IDs it can have. There are two features in the same place from OSM, but the POIs get merged.
* **841-normalize-boundaries-kind** to use the containing relation for the test. The outer boundary way is also tagged, but is discarded by `osm2pgsql`.
* **993-remove-props-road-merge** switch tests to avoid merging problems. Instead of testing that a specific object with an ID does not have the property, the tests now assert that the general object exists and that no object exists with the dropped property.
* **987-national-forests** was failing, but now is passing. Magic!

Further failures:

* **845-buildings-z13** appears to be failing because the min zooms are wrong - the migration for this this probably needs to be run. If it was already run then there might be some error which is preventing it from updating properly.
* **976-fractional-pois** is failing because the New York boundary isn't present in the output. However, it is present in the database, and appears to have the right properties and min zoom. It isn't clear why the boundary code is failing to cut this with the New Jersey boundary.
* **992-boundaries-min_zoom-and-name** is failing because there are no `Admin-1 boundary` items in the `ne_50m_admin_1_states_provinces_lines` table. There are none in the shapefile, so perhaps this indicates a shapefile processing error?
* **524-peak-kind-tile-rank** is failing because dev pulls in an extra "padded" bounds to make MVTB tiles. This means it pulls in extra nearby peaks, ranks them and discards those ranked >5. It then discards those ranked 2-5 when it clips the padded bounds to the tile bounds to make JSON format. This might need a code change to only rank features _within_ the tile.

@nvkelso what should we do about the `kind_tile_rank` failure? Should we be only ranking within tile boundaries? This will cause weirdness at MVTB boundaries, as peaks in the "padded" area can't get assigned their usual `kind_tile_rank` without pulling in a whole tile's worth of padding.

### DIFF
--- a/integration-test/742-predictable-layers-pois.py
+++ b/integration-test/742-predictable-layers-pois.py
@@ -21,11 +21,14 @@ assert_no_matching_feature(
     15, 5240, 12666, 'landuse',
     {'id': 79457493, 'kind': 'grave_yard', 'label_placement': True})
 
-# Way:41654965 Cemetery in POIs
+# Arlington National Cemetery - there are two objects representing this in
+# the same place. The code will de-duplicate one or the other of them.
+#
 # http://www.openstreetmap.org/way/41654965
+# http://www.openstreetmap.org/relation/2475077
 assert_has_feature(
     12, 1171, 1567, 'pois',
-    {'id': 41654965, 'kind': 'cemetery', 'min_zoom': 12})
+    {'id': set([41654965, -2475077]), 'kind': 'cemetery', 'min_zoom': 12})
 
 # Label placement Cemetery in landuse
 # http://www.openstreetmap.org/way/44580948
--- a/integration-test/473-landuse-tier.py
+++ b/integration-test/473-landuse-tier.py
@@ -32,7 +32,10 @@ assert_has_feature(
       'tier': 2, 'min_zoom': 8 })
 # this one is clipped by the polygon min_zoom, and so appears at the same
 # level.
+#
+# note that the feature _is_ present in the zoom 8 tile, but gets merged
+# with a nearby feature of the same name, so instead we test this at z9.
 assert_has_feature(
-    8, 71, 98, 'pois',
+    9, 142, 196, 'pois',
     { 'kind': 'forest', 'id': 34416231,
       'tier': 2, 'min_zoom': 8 })
--- a/integration-test/605-crosswalk-sidewalk.py
+++ b/integration-test/605-crosswalk-sidewalk.py
@@ -1,7 +1,7 @@
-# http://www.openstreetmap.org/way/367770916
+# http://www.openstreetmap.org/way/367477828
 assert_has_feature(
-    16, 10478, 25323, 'roads',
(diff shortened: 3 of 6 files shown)


nvkelso:
* I'll have a look at **992-boundaries-min_zoom-and-name**, I probably still have the raw files on my machine.
* For **524-peak-kind-tile-rank**, I think for now we should do as you propose and only calculate for the tighter (not buffered) tile boundary and buffered MVT will just be a known issue (and won't be on prod so isn't a service issue for us, yet). When we look at larger meta-tiles this could be a good time to re-example this logic?

integration-test/742-predictable-layers-pois.py
```diff
@@ -21,11 +21,14 @@
...
-# Way:41654965 Cemetery in POIs
+# Arlington National Cemetery - there are two objects representing this in
+# the same place. The code will de-duplicate one or the other of them.
+#
 # http://www.openstreetmap.org/way/41654965
+# http://www.openstreetmap.org/relation/2475077
 assert_has_feature(
     12, 1171, 1567, 'pois',
-    {'id': 41654965, 'kind': 'cemetery', 'min_zoom': 12})
+    {'id': set([41654965, -2475077]), 'kind': 'cemetery', 'min_zoom': 12})
```
nvkelso: Neat, nice syntax option!

nvkelso:
I don't see anything obvious for **992-boundaries-min_zoom-and-name**. @rob, @iandees, and I have all confirmed the CA-NV border is present in that integration-test/figure file, with the correct PRJ. 

And oddly enough https://circleci.com/gh/tilezen/vector-datasource/1155 says `[ 151/152] PASS: 'integration-test/992-boundaries-min_zoom-and-name.py'`?

My database is still in a mixed state so I think this is as far as I can take this one right now.

zerebubuth:
> I don't see anything obvious for 992-boundaries-min_zoom-and-name. @rob, @iandees, and I have all confirmed the CA-NV border is present in that integration-test/figure file, with the correct PRJ.

Ah, sorry - I didn't explain that failure properly. These tests all pass locally and on CircleCI (apart from one unrelated one I'll look at shortly). They fail only on dev - which means we have a difference in behaviour between a full database on the one hand, and the fixtures and Overpass-derived partial database on the other.

The CA-NV border is in the fixtures, and so the test passes locally. The CA-NV border (and all `Admin-1 boundary` features) is missing from the `ne_50m_admin_1_states_provinces_lines` table on dev, and from the shapefile bundle that I downloaded from S3. This perhaps suggests an issue with the NE data conversion and re-projection process.

zerebubuth:
Follow-ups:

* **976-fractional-pois** in #1063
* **992-boundaries-min_zoom-and-name** in #1064
* **524-peak-kind-tile-rank** in #1065

Issue #1065, opened 2016-09-21
Only rank features within the bounds of the tile.


zerebubuth:
The **524-peak-kind-tile-rank** test is failing on dev because it pulls in an extra "padded" bounds to make MVTB tiles. This means it pulls in extra nearby peaks, ranks them and discards those ranked >5. It then discards those ranked 2-5 when it clips the padded bounds to the tile bounds to make JSON format. This might need a code change to only rank features `within` the tile.

@nvkelso says:

>  I think for now we should do as you propose and only calculate for the tighter (not buffered) tile boundary and buffered MVT will just be a known issue (and won't be on prod so isn't a service issue for us, yet).

This does not fail CI as the CI tests do not generate MVTB.

nvkelso:
@rmarianski Please verify this issue.

rmarianski:
Looks good. The peak that wasn't showing up before in http://tile.dev.mapzen.com/vector/v1/all/11/420/779.json is now, and all the tile ranks are in succession.

PR #1066, opened 2016-09-21
Only rank features within the unpadded bounds

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1065


rmarianski:
Whoops, I only saw now that you had self-assigned this issue. If you had already implemented it, maybe we can compare implementations :)

Issue #1063, opened 2016-09-21
NY boundary issue on dev


zerebubuth:
The **976-fractional-pois** test is failing on dev because the New York boundary isn't present in the output. However, it is present in the dev database, and appears to have the right properties and min zoom. It isn't clear why the boundary code is failing to cut this with the New Jersey boundary.

The test passes fine locally and on CI, so is possibly something to do with having a full database of data rather than the cherry-picked data for the tests.

See #1061 for the original issue.

nvkelso:
I still see region boundaries from OSM, and I still see Hoopa reservation boundary.

PR #1067, opened 2016-09-21
Fix admin boundary test failure

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1063

I was able to replicate the failure locally. I stepped through the transform, and took a guess at what was going on based on what I saw. It seemed like there were multiple features with the same id, and they were self intersecting and becoming empty and eliminating themselves. But this was just an assumption that got the test to pass for me to pass locally, and I'm not sure if this was the wrong approach or if it has other consequences.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -1781,6 +1781,7 @@ def admin_boundaries(ctx):
 
         for i, feature in enumerate(features):
             boundary, props, fid = feature
+            prop_id = props['id']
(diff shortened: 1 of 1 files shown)


zerebubuth:
I saw a different problem in #1068, but it's possible that there is more than one bug here. Certainly, it doesn't make much sense for a single boundary to intersect itself. I think the fix is worth having, in addition to any other fixes going around.

:+1:

PR #1068, opened 2016-09-21
Only delete features within bounds. Boundaries fixes.

zerebubuth:
Note that this is a PR into @rmarianski's #1067 PR - which seemed to make sense since we'd mostly solved the problem in the same way.

In addition to only ranking features within the bounds, it's necessary to only delete features within the bounds when they don't satisfy the ranking.

This patch adds configuration to the testing tile server to make it capable of generating MVTB tiles. This exposed another issue, which is that the admin boundary code was getting confused with the mix of oriented polygonal boundaries and non-oriented linear boundaries. The linear boundaries used in the tests could be replaced by polygonal alternatives, so the linear boundaries were dropped. This makes it more likely that the polygonal admin boundaries will be correctly processed.

Connects to #1065. Relates to #1063.

@rmarianski, @iandees could you review, please?

### DIFF
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -112,27 +112,4 @@ FROM buffered_land
 WHERE
   {{ bounds['polygon']|bbox_filter('the_geom',3857) }}
 
-UNION ALL
-
-SELECT
-  tags->'name' AS name,
-  'openstreetmap.org' AS source,
-  mz_boundary_min_zoom as min_zoom,
-  NULL AS maritime_boundary,
-  mz_calculate_json_boundaries(planet_osm_line.*) AS mz_properties,
-  {% filter geometry %}way{% endfilter %} AS __geometry__,
-  osm_id AS __id__,
-  %#tags AS tags
-
-FROM
-  planet_osm_line
-
-WHERE
-  {{ bounds['line']|bbox_filter('way',3857) }} AND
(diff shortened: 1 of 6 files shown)


nvkelso:
For the relation **Hoopa Valley Tribe**, will the feature show up in **both** `boundaries` and `landuse` layers after this change?

queries/boundaries.jinja2
```diff
@@ -112,27 +112,4 @@ FROM buffered_land
 WHERE
   {{ bounds['polygon']|bbox_filter('the_geom',3857) }}
 
-UNION ALL
```
nvkelso: Does removing this UNION prevent export of any boundary line data that is not polygons?

zerebubuth: Yes, that is the intention. The boundary line data was breaking the admin boundaries post-process code where it attempts to assign left/right properties. Lines have no well-defined topological left/right, and that was breaking the code.

The cases where we had tests on linear data turned out to be replaceable with polygonal data. If that isn't the case generally, then we need to find more tests.

zerebubuth:
> For the relation Hoopa Valley Tribe, will the feature show up in both boundaries and landuse layers after this change?

Yes. That's the existing behaviour - do you want to change the behaviour to exclude `protect_class=24` items from `landuse`?

iandees:
This looks good to me 👍

Issue #1064, opened 2016-09-21
NE Admin-1 boundaries missing from processed shapefiles


zerebubuth:
The **992-boundaries-min_zoom-and-name** test is failing because there are no `Admin-1 boundary items` in the `ne_50m_admin_1_states_provinces_lines` table in the dev database. There are none in the shapefile downloaded from S3 [as part of the installation process](https://github.com/tilezen/vector-datasource/wiki/Mapzen-Vector-Tile-Service#load-additional-data-and-update-database), so perhaps this indicates a shapefile processing error?

The shapefile fixture is okay, and the tests pass locally and on CI.

rmarianski:
I checked the [original upstream](http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/50m/cultural/ne_50m_admin_1_states_provinces_lines.zip) shape file, and it too doesn't contain any `Admin-1 boundary` items, all are `1st Order Admin Lines`.

nvkelso:
This is my mistake in
https://github.com/tilezen/vector-datasource/pull/1037/files#diff-dca2f4be5bc54303506aa7848a476050L32

We should add that one selector back to the remaining region filter.

_n

nvkelso:
WFM

PR #1069, opened 2016-09-21
Restore ne 1st Order Admin Lines rule

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1064

@nvkelso is this what you had in mind in https://github.com/tilezen/vector-datasource/issues/1064#issuecomment-248665808 ?


yaml/boundaries.yaml
```diff
@@ -52,6 +52,10 @@ filters:
     table: osm
 
   # ne
+  - filter: {featurecla: 1st Order Admin Lines}
```
nvkelso: Combine with line 98 or so where general `region` filter selection is.

zerebubuth:
We should recreate the `integration-test/fixtures/ne_50m_admin_1_states_provinces_lines/992-ne_50m_admin_1_states_provinces_lines-nv-ca.shp` file, as it currently has `featurecla=Admin-1 boundary` whereas the newer shapefiles from S3 have `featurecla=1st Order Admin Lines`.

yaml/boundaries.yaml
rmarianski: Updated in d724e05.

rmarianski:
Updated fixture in 0c4b529.

Issue #981, opened 2016-08-22
Remove unstyled NE places data (preferring existing OSM data)


meetar:
* **What did you see?**

The Boston `Admin-1 capital` poi shows up in the Places layer at high zooms, where it is seen to be on a home improvement store in Roxbury, which is not near the center of town at that zoom level. Per @nvkelso's suggestion, perhaps NE places data could be removed from higher zooms, maybe > 9?

* **What did you expect to see?**

Figured it would be near the capitol building or some other Schelling point

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

http://tangrams.github.io/explorer/#13.7/42.3294/-71.0834/kind/Admin-1%20capital

* **Screenshot?** _Props for animated gifs._

<img width="451" alt="screen shot 2016-08-22 at 11 02 47 am" src="https://cloud.githubusercontent.com/assets/459970/17862277/8b743e50-6862-11e6-8b1e-eacd292810e5.png">

nvkelso:
Here's the scene file logic pulled out of Bubble Wrap, for all zooms (mostly `scalerank` based, sometimes also `population` range):

```
populated-places-natural-earth-z2:
    filter: { name: true, source: naturalearthdata.com, $zoom: [2], scalerank: 0 }
    # scalerank: [0]
    # EXCLUDES scalerank: [1,2,3,4,5,6,7,8,9,10]

populated-places-natural-earth-z3:
    filter: { name: true, source: naturalearthdata.com, $zoom: [3] }
    # scalerank: [0,1]
    # EXCLUDES scalerank: [2,3,4,5,6,7,8,9,10]

populated-places-natural-earth-z4:
    filter: { name: true, source: naturalearthdata.com, $zoom: [4] }
    # scalerank: [0,1,2]
    # EXCLUDES scalerank: [3,4,5,6,7,8,9,10]

populated-places-natural-earth-z5:
    filter: { name: true, source: naturalearthdata.com, $zoom: [5] }
    # scalerank: [0,1,2,3,4]
    # EXCLUDES scalerank: [5,6,7,8,9,10]

populated-places-natural-earth-z6:
    filter: { name: true, source: naturalearthdata.com, $zoom: [6] }
    # scalerank: [0,1,2,3,4,5,6]
    # EXCLUDES scalerank: [7,8,9,10]

populated-places-natural-earth-z7:
    filter: { name: true, source: naturalearthdata.com, $zoom: [7] }
[...]

PR #1076, opened 2016-09-22
Update migration to include min zoom for places.

zerebubuth:
D'oh! I thought this was a faulty migration, but in fact it was simply that we weren't including `min_zoom` in the query. The silver lining is that it means no migration is needed to fix it :smile: 

Updated the test so it checks for `min_zoom` too.

Connects to #981.

@rmarianski & @iandees could you review, please?


nvkelso:
+1, ha!

Issue #1091, opened 2016-09-30
REGRESSION: Missing name:short on region features


nvkelso:
* **What did you see?** We're dropping `name:short` from v1-pre3 vector tiles (prod v0.10 has them). I suspect we need to whitelist `name:short` in the new language code logic (or just pass thru values that we don't remap)?
* **What did you expect to see?** They should be present.
* **What map location are you having problems with?** Over the USA
  * dev: `http://tile.dev.mapzen.com/mapzen/vector/v1/all/5/7/12.topojson`
  * prod: `http://vector.mapzen.com/osm/all/5/7/12.topojson`
* **Screenshot?** _Props for animated gifs._

<img width="908" alt="screen shot 2016-09-30 at 19 08 54" src="https://cloud.githubusercontent.com/assets/853051/19009392/79155520-8741-11e6-96bb-a9cd58d679ea.png">

**prod v0.10:**
_note also `name:abbreviation` in addition to `name:short`_

```
properties: {
kind: "state",
name:short: "MO",
scalerank: 4,
name:eo: "Misurio",
source: "openstreetmap.org",
name:uk: "Міссурі",
name:ru: "Миссури",
name:be: "Місуры",
population: 6063589,
name:abbreviation: "Mo.",
id: 473849775,
name:es: "Misuri",
name: "Missouri"
```

**dev v1.0-pre3:**

```
properties: {
kind: "region",
name: "Missouri",
kind_detail: "state",
name:eo: "Misurio",
source: "openstreetmap.org",
min_zoom: 4,
name:uk: "Міссурі",
name:be: "Місуры",
name:ru: "Миссури",
id: 473849775,
name:es: "Misuri",
population: 6063589
},
```

nvkelso:
Verified in dev!

PR #1093, opened 2016-10-03
Include name:short as a tag name alternate

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1091

@zerebubuth, @iandees could you review please?


rmarianski:
Added an integration test for Missouri in a32d2bf.

PR #1105, opened 2016-10-13
Translation of name:zh changed in OSM

zerebubuth:
Fixes CI failures. Seems better to test for the presence of the other language names rather than their exact value.

@rmarianski what do you think?


rmarianski:
:+1:

PR #1110, opened 2016-10-19
Tweaked format wording

migurski:
Based on input from a guest at office hours, we'd like to offer a few additional format hints to make MVT findable. Also removed the header to match other docs.

/cc @rmglennon.


nvkelso:
+1

zerebubuth:
Awesome, thanks!

migurski:
Thanks for the merge! I guess it would now be necessary to tag this with a new release so it’s available?

PR #1114, opened 2016-11-01
Kkowalsky/docs update

kkowalsky:
This is just to update the 'Display Tiles' page with all six house styles and a new picture. I edited the code banks so they don't break my markdown in my editor.


kkowalsky:
looking into why the builds failed...

kkowalsky:
looks like 'integration-test/291-483-suppress-historical-closed.py', 'integration-test/857-move_barriers_to_landuse.py' both had exception errors in 'setup_and_run_tests.sh'

nvkelso:
The PNG was a couple MB but is displayed small on the page, I compressed it to a GIF using PS. See https://github.com/tilezen/vector-datasource/pull/1114/commits/f07515d4e38a044873b573270bcb45fa5aada81b.

Now that we have product pages for each map style, we should direct people to mapzen.com instead of Github.io for these. See https://github.com/tilezen/vector-datasource/pull/1114/commits/f07515d4e38a044873b573270bcb45fa5aada81b.

kkowalsky:
thanks for catching that @nvkelso!

nvkelso:
@kkowalsky Merge against master and the test failures will be history.

kkowalsky:
thanks for the fix @zerebubuth!

nvkelso:
LGTM, thanks!

Issue #1111, opened 2016-10-27
[docs] Add info to docs about tile coordinates


kkowalsky:
*\* Copied from https://github.com/tilezen/joerd/issues/109 because relevance **

A support ticket came in where the user was trying to get a tile by passing decimal degrees, instead of tile coordinates.

We should clarify that the the {x}/{y} is a tile coordinate. Currently, it looks pretty much the same as the standard decimal degrees format. Need to tell people that tile coordinates are used and provide instructions for teaching people how to convert

@kkowalsky knows of some resources that we can add to docs.

kkowalsky:
add a statement about best performance using "all" instead of specific layers and filtering data on the client side

nvkelso:
@zerebubuth Is Katie's suggestion in https://github.com/tilezen/vector-datasource/issues/1111#issuecomment-257418470 about suggestion `all` valid anymore with Xonacatl in production?

zerebubuth:
The advice is still good, but the situation is more complicated now. Xonacatl adds an extra "hop" in the request, so for people on low-latency, high-bandwidth connections they might see a reduction in performance using custom layers. If they are on a high-latency, low-bandwidth connection (e.g: mobile) then custom layers are probably better.

To explain what's going on, here's the path of a request through the system:

1. Client asks Fastly for the tile.
  1. If Fastly already has it, return it to client and stop here.
  2. If it's an `all` request, skip to 4.
2. Fastly asks Xonacatl for the custom layers tile.
3. Xonacatl asks Fastly for the corresponding `all` tile.
4. Fastly asks S3 for the `all` tile. If it exists, return it to the client and stop here.
5. Fastly asks tileserver (in US-east) for the tile.

Steps 2 and 3 can be skipped for `all` tiles, at the cost that the response might be considerably larger, depending on which layers were selected. The largest delay in returning a tile to the client is whether it has to perform step 5 or not - but the client has little control over that.

The best thing that they can do to avoid hitting step 5 is handle "over-zooming", which means not requesting tiles at zoom > 16. Instead, they should request the z16 tile which contains the tile they originally wanted and "cut out" the part that they need. This is handled automatically by Tangram. We do handle requests for z > 16, but they will always hit step 5, and be comparatively slow. (This is possibly another service we could run alongside Xonacatl and the Metatile server - but there are no plans to do it at the moment).

[...]

PR #1115, opened 2016-11-02
Add "migration to 1.0" documentation page

rmglennon:
This adds a new file with migration information and best practices for users of the pre-1.0 service.

- Tells people that they need to take action to use 1.0, as it is not automatic
- Gives guidance on what happens if you continue using the deprecated service
- Adds some info on performance of custom layers vs all from https://github.com/tilezen/vector-datasource/issues/1111 
- Adds the detailed migration guide into the docs from @nvkelso's personal account gist (with a few formatting edits and changes to terms like "funky") into the documentation set...we will need to publish this to see what it looks like on the website.

If we can get this installed, maybe @burritojustice and @migurski can reference the link to the docs in the newsletter this week.

This is similar to the Pelias to Mapzen Search v1 document (it was in the docs on mapzen.com for about a year): https://github.com/pelias/pelias-doc/blob/master/transition-from-beta.md 

Please review and add comments and commits to clarify or correct. Thanks!

cc @kkowalsky


kkowalsky:
there was an update to master to fix the build fail, you should merge against master to make it work again

also there needs to be a PR for the docs generator to add this additional page

this is good, looks like you took some of @zerebubuth's recommendation's in the "Some or All" section :+1:

rmglennon:
OK, I pulled in from this master to my fork and added commits to the fork, but tests are still failing. Someone with commit access and knowledge of the tests in this repo, please review and fix! 

I will put in a PR in docs-gen to add the topic when the content is approved. :)

nvkelso:
I'll have a look...

docs/migrate-to-vector-tiles-1.0.md
```diff
@@ -0,0 +1,246 @@
...
+
+## If you continue using vector.mapzen.com
+
+Although Mapzen announced that the deprecated service will be maintained for up to a year after the 1.0 release, you should update sooner for the best experience.
+
+During this transition period, only tiles that were previously archived on Mapzen's servers will be available. This means that you may encounter locations, such as rural areas or certain zoom levels, where no data is returned. This can result in a 404 HTTP status error code.
+
+At the time of the 1.0 release, only the `all` and `buildings` layers were archived and cached from the deprecated service. If you attempted to query for other layers, you may have been unable to access them and received an error. This was happening because tiles for custom layers were generated only on demand, and no fresh tiles were being served from the previous URLs. However, shortly after the 1.0 release, Mapzen added more custom layers to the server cache, which gives you the ability to query for any layer on the previous URLs.
```
nvkelso: This sentence is problematic:

`However, shortly after the 1.0 release, Mapzen added more custom layers to the server cache, which gives you the ability to query for any layer on the previous URLs.`

Probably normal English versus technical language is fighting here...

What about:

`However, shortly after the 1.0 release, Mapzen added special handing for custom layers served out the the archive which gives you the ability to query for any layer on the previous URLs.`

docs/migrate-to-vector-tiles-1.0.md
```diff
@@ -0,0 +1,246 @@
...
+
+Even if tiles are loading and your `vector.mapzen.com` map appears to be working in your area of interest, keep in mind that the content of the tiles reflect the state of OpenStreetMap as of early October 2016. No further changes from OSM will be imported into the archived tiles. For example, if you modify a building in OSM, you will not see that change in the deprecated service.
```
nvkelso: Flagging mixed use of `OpenStreetMap` and `OSM`. Should we say `OpenStreetMap (OSM)` on the first use and then `OSM` on subsequent uses?

docs/migrate-to-vector-tiles-1.0.md
```diff
@@ -0,0 +1,246 @@
...
+
+If you update to 1.0, all these functions to work properly; tiles in previously uncached areas are generated on demand and the OSM data is fresh. OSM data in the 1.0 tiles are updated frequently, often within hours or a day of the edits being made in OSM.
```
nvkelso: Similar comment about `OpenStreetMap` and `OSM`, but in this paragraph we only use `OSM`. I think I prefer to always say `OpenStreetMap` fully spelled out in documentation.

docs/migrate-to-vector-tiles-1.0.md
```diff
@@ -0,0 +1,246 @@
...
+  * some values are now `natural_forest`
+  * but many values are still forest
+
+* **landuse** layer: `park`
+  * some values are now `natural_park`
+  * but many values are still park
+
+* **landuse** layer: `wood`
+  * some values are now `natural_wood`
+  * but many values are still work
```
nvkelso: Should be:

`* but many values are still wood` 

(work > wood)

nvkelso:
Once this PR and Katie's are merged, I'll cut a new `*-doc*` release for mapzen-docs-generator to reference.

docs/migrate-to-vector-tiles-1.0.md
kkowalsky: yes that's correct.

docs/migrate-to-vector-tiles-1.0.md
rmglennon: changed to new sentence!

docs/migrate-to-vector-tiles-1.0.md
rmglennon: changed to wood

docs/migrate-to-vector-tiles-1.0.md
```diff
@@ -0,0 +1,246 @@
...
+
+With the 1.0 release, you may find benefits to requesting `all` layers instead of custom ones. You should run benchmark tests for both custom and `all` layers on your target devices and network connections. If you find that one method is clearly faster than the other, then choose it. If the results are about the same, then use `all` because it has a simpler service architecture.
+
+If you choose `all`, you can use client-side cartography to limit the display of the layers. For example, if you are using Tangram to draw your map, you can use data [filters](https://mapzen.com/documentation/tangram/Filters-Overview/).
+
+Both custom and `all` layers are part of Mapzen's global tile server cache. This is beneficial because tiles from the cache do not count toward your rate limits and can be served faster, yet the map data in the tiles can still be updated regularly.
+
+## If you continue using vector.mapzen.com
+
+Although Mapzen announced that the deprecated service will be maintained for up to a year after the 1.0 release, you should update sooner for the best experience.
```
nvkelso: We should state a date here. I propose rounding up to Oct. 31st, 2017 as the last guaranteed date of availability for archived v0.10 tiles.

docs/migrate-to-vector-tiles-1.0.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Choose whether to use some or all layers
+
+A vector tiles request can be for one or more individual layers, such as `roads` or `landuse`, or for `all` layers in the service.
+
+With the 1.0 release, you may find benefits to requesting `all` layers instead of custom ones. You should run benchmark tests for both custom and `all` layers on your target devices and network connections. If you find that one method is clearly faster than the other, then choose it. If the results are about the same, then use `all` because it has a simpler service architecture.
+
+If you choose `all`, you can use client-side cartography to limit the display of the layers. For example, if you are using Tangram to draw your map, you can use data [filters](https://mapzen.com/documentation/tangram/Filters-Overview/).
+
+Both custom and `all` layers are part of Mapzen's global tile server cache. This is beneficial because tiles from the cache do not count toward your rate limits and can be served faster, yet the map data in the tiles can still be updated regularly.
```
rmglennon: I see. Xonacotl doesn't necessarily mean cache. I can change the last sentence to say just that `all` layers are in the cache...or remove it altogether.

docs/migrate-to-vector-tiles-1.0.md
rmglennon: updated with that date!

docs/migrate-to-vector-tiles-1.0.md
rmglennon: Sounds good. I've added this and removed the rate limits info.

rmglennon:
Thanks for the feedback, @nvkelso and @kkowalsky. Aaanndd, the tests are passing now. 

I think I got all the review points covered. I want a view with the original text, the review part, and the new changes together!

docs/migrate-to-vector-tiles-1.0.md
rmarianski: > @rmarianski to confirm (and make sure we're not double counting against API limits when the custom request is on-demand).

:white_check_mark: That's right.
Specifically, the xonacatl request itself is not rate limited, and it's just the subrequest to the `all` layer that may require a new tile to be generated. This may change in the future if we find that xonacatl itself needs to be guarded (and here we would need to avoid double counting). Anyway, :+1: that the rate limits info has been removed, because it might just add confusion, which I probably just did.

rmglennon:
@nvkelso  Is it possible to preview or precog this before doing a release (like tag to the commit in the makefile)? Maybe you could help with that over in docs-gen, as you're more familiar with how this part of the help is built.

I am not certain how the gist part will look. For example, our markdown parser prefers a blank line between style changes, such as before bullet lists.

burritojustice:
@nvkelso @rmglennon looks like checks have failed -- let me know when this is ready, waiting to send the newsletter

zerebubuth:
@burritojustice please ignore the test failure, it appears to be because Circle is suppressing our custom environment variable because this PR is coming from a fork. This is causing it to try and use the public Overpass instance instead of our internal one. Unfortunately, the public instance is very heavily loaded and rate-limits us before the build is finished.

kkowalsky:
to avoid the forking issues, I made a new PR in #1118 with the latest version of the file

PR #1120, opened 2016-11-04
update attribution requirements

### LINKED ISSUE #1113: [docs] Clarify Mapzen attribution when using service
Reading the vector tile docs for [attribution](https://mapzen.com/documentation/vector-tiles/attribution/) it seems like I would only needs to attribute OSM and WOF, but when using the hosted service (which most anyone is), a developer would also be required to credit Mapzen (see [terms of use](https://mapzen.com/documentation/overview/#terms-of-use)).

We should update the vector tile attribution page to make this clearer.

kkowalsky:
fixes #1113, updates the attribution requirements for Vector Tiles

- also removed 'footers' because the docs pages have that information in the generator.

### DIFF
--- a/docs/attribution.md
+++ b/docs/attribution.md
@@ -1,6 +1,6 @@
 # Attribution
 
-Attribution is required for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) page.
+Attribution is required for using Mapzen's hosted vector tile service as well as for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) and the [Mapzen rights](https://mapzen.com/rights) pages.
 
 **Required attribution:**
 
@@ -12,7 +12,7 @@ Attribution is required for many of our data providers. Example language is prov
 
 ### Where to attribute
 
-Attribution needs to "appear in a place that is reasonable to the medium or means you are utilising." [Specific examples](http://wiki.osmfoundation.org/wiki/License#Where_to_put_it.3F) are given by the OSM Foundation and are generally best practices for giving credit to any source.
+Attribution needs to "appear in a place that is reasonable to the medium or means you are utilising." [Specific examples](http://wiki.osmfoundation.org/wiki/License#Where_to_put_it.3F) are given by the OSM Foundation and are generally best practices for giving credit to any source. More information on attribution is on our [rights](https://mapzen.com/rights) page.
 
 ## The fine print
 
--- a/docs/attribution.md
+++ b/docs/attribution.md
@@ -1,15 +1,14 @@
 # Attribution
 
-
 Attribution is required for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) page.
 
 **Required attribution:**
 
->  © OpenStreetMap contributors, Who's On First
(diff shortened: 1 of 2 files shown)


nvkelso:
For the intro on the attribution, should we say credit to Mapzen for using the hosted vector tile service? We added Mapzento the credit line, but it's otherwise not clear why since the attributions are otherwise for data sources.

migurski:
When changing this, let’s make sure it’s compatible with the advice at https://mapzen.com/rights:

> If you are using a combination of Mapzen services, you may use the following attribution statement:
> 
> _Powered by [Mapzen](https://mapzen.com) - [Attribution](https://www.mapzen.com/rights). Data ©[OSM](https://openstreetmap.org/copyright) contributors._

kkowalsky:
just updated this, let me know if I should explicitly copy the rights page @migurski (I tried to link only to avoid future update problems)

kkowalsky:
I'm also doing this concurrently with [terrain tiles](https://github.com/tilezen/joerd/pull/115), please have a look :eyes: at them as well.

docs/attribution.md
```diff
@@ -1,19 +1,18 @@
 # Attribution
 
-
-Attribution is required for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) page.
+Attribution is required for using Mapzen's hosted vector tile service as well as for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) and the [Mapzen rights](https://mapzen.com/rights) pages.
 
 **Required attribution:**
 
->  © OpenStreetMap contributors, Who's On First
+>  © Mapzen, OpenStreetMap contributors, Who's On First
```
nvkelso: The `©` mark goes with OpenStreetMap, not Mapzen. 

(I'm actually not sure why it needs to be `© OpenStreetMap`, seems overkill if it's in a source credit context.)

A common usage in the wild:

`Leaflet | Tangram | © OSM contributors | Mapzen`

docs/attribution.md
```diff
@@ -1,19 +1,18 @@
...
 
 **Full attribution for all projects:**
 
->  © OpenStreetMap contributors, Who's On First, Natural Earth, and openstreetmapdata.com
+>  © Mapzen, OpenStreetMap contributors, Who's On First, Natural Earth, and openstreetmapdata.com
```
nvkelso: Ditto re `©` mark

docs/attribution.md
nvkelso: And same comment from the terrain docs review where this is now for combination of the hosted service (e.g. Mapzen credit) and the data (existing credit).

docs/attribution.md
```diff
@@ -1,19 +1,14 @@
 # Attribution
 
-
-Attribution is required for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) page.
+Attribution is required for using Mapzen's hosted vector tile service as well as for many of our data providers. Example language is provided below, but you are responsible for researching each project to follow their license terms. More details are available on the [Data Sources](data-sources.md) and the [Mapzen rights](https://mapzen.com/rights) pages.
 
 **Required attribution:**
```
kkowalsky: ahhh, I misunderstood. Double checking the wording on Terrain and adding a commit.

also yay Natural Earth!

nvkelso:
LGTM

Issue #1106, opened 2016-10-14
Merge water and earth polygons


zerebubuth:
- **What did you see?**

![image](https://cloud.githubusercontent.com/assets/271360/19397173/991e934c-923e-11e6-9c3e-bc55093e08e9.png)
- **What did you expect to see?**

Similar, but without the red horizontal and vertical lines.
- **What map location are you having problems with?**

The example above is New York City, but the problem should be evident globally.

We should probably merge the water and earth polygons to get rid of (most of) these seams. (The ones near property changes would remain, e.g: `sea`/`ocean`) It would also make the tiles smaller.

Related #735, /cc @meetar.

nvkelso:
Related to #1107

rmarianski:
The background here is that we used to union the general earth and water queries, and grouped by their properties, but we had to switch to using json instead of hstore, and the version of postgresql we were using at the time didn't support grouping by json.

Since we're using postgresql 9.5 now, we could go back to grouping by it in sql, but that would mean that we require 9.5 to run the tile stack.

Instead, we can do the merging and grouping in python. Transforms should exist to already did this, if not out of the box, close to it. We merge polygons for the buildings and landuse layers.

nvkelso:
I’d like to do this at the python Transform post processing layer so we
don’t need to bump the Postgres requirement.

iandees:
How can I test this? @zerebubuth, what style are you looking at in that screenshot?

nvkelso:
@iandees Try Peter's tag explorer:

- http://tangrams.github.io/explorer/#10.0/40.6233/-73.4422/kind/ocean
- http://tangrams.github.io/explorer/#10.0/40.6233/-73.4422/kind/earth

pnorman:
> The background here is that we used to union the general earth and water queries, and grouped by their properties, but we had to switch to using json instead of hstore, and the version of postgresql we were using at the time didn't support grouping by json.

You should be able to pick a field out of the json and group by that, but I think it's better to do this in post-processing.

nvkelso:
Verified fix on dev using custom Bubble Wrap and in Ikeda, one of the Sandbox styles where this issue was originally reported.

- Ikeda tested with this variant (load into Tangram Play): https://api.github.com/gists/3bba222b99804ccbf81a1188d9f01d45

[...]

PR #1124, opened 2016-11-10
Merge earth and ocean polygons

iandees:
- [x] Update tests
- [ ] ~~Update data migrations~~
- [ ] ~~Update docs~~

Connects to #1106.

### DIFF
--- a/integration-test.py
+++ b/integration-test.py
@@ -311,6 +311,25 @@ def assert_at_least_n_features(z, x, y, layer, properties, n):
                 "%r, found only %d" % (n, properties, num_matching)
 
 
+def assert_less_than_n_features(z, x, y, layer, properties, n):
+    """
+    Downloads a tile and checks that it contains less than `n` features which
+    match the given `properties`.
+    """
+    with features_in_tile_layer(z, x, y, layer) as features:
+        num_features, num_matching = count_matching(
+            features, properties)
+
+        if num_features >= n:
+            raise Exception, "Found %d or more features including " \
+                "properties %r (because layer %r had %d features)" % \
+                (n, properties, layer, num_features)
+
+        if num_matching >= n:
+            raise Exception, "Did not find %d features including properties " \
+                "%r, found only %d" % (n, properties, num_matching)
+
+
 def assert_no_matching_feature(z, x, y, layer, properties):
     with features_in_tile_layer(z, x, y, layer) as features:
         num_features, num_matching = count_matching(
--- a/integration-test.py
+++ b/integration-test.py
@@ -373,8 +373,8 @@ def print_coords(f, log, idx, num_tests):
         runpy.run_path(f, init_globals={
             'assert_has_feature': print_coord,
             'assert_no_matching_feature': print_coord,
-            'assert_at_least_n_features': assert_at_least_n_features,
-            'assert_less_than_n_features': assert_less_than_n_features,
+            'assert_at_least_n_features': print_coord,
+            'assert_less_than_n_features': print_coord,
             'features_in_tile_layer': print_coord_with_context,
             'assert_feature_geom_type': print_coord,
             'layers_in_tile': print_coord_with_context,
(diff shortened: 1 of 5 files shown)


integration-test/1106-merge-ocean-earth.py
```diff
@@ -0,0 +1,4 @@
+# There should be a single (merged) ocean feature in this tile
+assert_less_than_n_features(9, 167, 186, 'ocean', properties, 2)
+# There should be a single (merged) earth feature in this tile
+assert_less_than_n_features(9, 170, 186, 'earth', properties, 2)
```
nvkelso: We should test once for OSM and once again for Natural Earh, so these tests again but for a zoom 5 coordinate.

iandees: Yep, I was starting here while I started figuring out how to replicate the tiled Natural Earth we use.

integration-test.py
```diff
@@ -311,6 +311,25 @@ def assert_at_least_n_features(z, x, y, layer, properties, n):
...
+def assert_less_than_n_features(z, x, y, layer, properties, n):
+    """
+    Downloads a tile and checks that it contains less than `n` features which
+    match the given `properties`.
+    """
+    with features_in_tile_layer(z, x, y, layer) as features:
+        num_features, num_matching = count_matching(
+            features, properties)
+
+        if num_features >= n:
```
zerebubuth: I'm not sure that this fits with the description of the function.

For example, if a tile has 10 features, but only one matches the `properties`, and we test `assert_less_than_n_features(z, x, y, layer, properties, 2)`, then I would expect the test to pass. But the test on the total number of features here means that it won't.

integration-test.py
```diff
@@ -359,6 +378,8 @@ def print_coords(f, log, idx, num_tests):
         runpy.run_path(f, init_globals={
             'assert_has_feature': print_coord,
             'assert_no_matching_feature': print_coord,
+            'assert_at_least_n_features': assert_at_least_n_features,
+            'assert_less_than_n_features': assert_less_than_n_features,
```
zerebubuth: These should be `print_coord`, or similar function with matching signature. This part of the code is used to print the coordinates of all tiles being tested, which can be useful for pre-seeding the dev TOI.

integration-test.py
iandees: I'm just going to remove the `num_features` check altogether and rely on the `num_matching` check. Does that make sense?

zerebubuth: Yup, that's good.

queries.yaml
```diff
@@ -425,6 +425,15 @@ post_process:
...
         'label_placement' not in properties and kind in
         ('archipelago', 'island', 'islet', 'continent', 'valley', 'ridge')
+
+  # merge ocean and earth polygons, see https://github.com/tilezen/vector-datasource/issues/1106
+  - fn: vectordatasource.transform.merge_polygon_features
+    params:
+      source_layer: earth
+  - fn: vectordatasource.transform.merge_polygon_features
+    params:
+      source_layer: ocean
```
rmarianski: Should this be `water` instead?

iandees: Dangit, yes. I thought I had fixed that earlier.

rmarianski:
@iandees btw, you get the :trophy: for best branch name of the day. In my mind, I say "heart" instead of leelu in reference to captain planet. And "multi-pass" when I hear leelu.

iandees:
I'm not seeing any tiled/unjoined `kind:ocean` data at the < z9 range (the Natural Earth data), so a test there wouldn't really be testing anything new.

iandees:
@nvkelso lowzoom tests added. You're thinking of screenshots from something like [the tag explorer map](http://tangrams.github.io/explorer/#10.2/43.8042/-69.0568/kind/ocean) before and after? How do I go about switching the tag explorer to use my local tileserver endpoint?

nvkelso:
@iandees clone that repo, start up local http server and load the index file, point the YAML source at your dev ip & port.

nvkelso:
You could also write quick Tangram Play style...

nvkelso:
I think this is ready to merge, we can visually verify it once it's on dev.

Issue #294, opened 2015-10-05
Refine water / land mask for martime_boundary


nvkelso:
This data file needs updates:
- http://s3.amazonaws.com/mapzen-tiles-assets/curated/buffered_land.zip

nvkelso:
Channel Islands off California need to be added to the maritime SHP (we shouldn't see boundary here):

<img width="836" alt="screen shot 2015-10-03 at 16 57 59" src="https://cloud.githubusercontent.com/assets/853051/10265684/0c4373c4-69f0-11e5-9480-865460931e20.png">

nvkelso:
Near Virginia and Washington DC:
#9/37.8427/-75.6437

<img width="749" alt="screen shot 2015-10-03 at 17 05 57" src="https://cloud.githubusercontent.com/assets/853051/10265711/10867d36-69f1-11e5-9650-f1b747ed6b6d.png">

nvkelso:
South Carolina:
#9/33.2973/-79.3591

<img width="686" alt="screen shot 2015-10-03 at 17 06 44" src="https://cloud.githubusercontent.com/assets/853051/10265717/2dcde5aa-69f1-11e5-994d-c0966b7670ac.png">

nvkelso:
Florida:
#8/27.496/-79.579

<img width="333" alt="screen shot 2015-10-03 at 17 07 32" src="https://cloud.githubusercontent.com/assets/853051/10265724/44a3a134-69f1-11e5-9dbf-b4ff8cb7a4c5.png">

zerebubuth:
TODO: export "maritime" boundaries from OSM to shape for @nvkelso.

nvkelso:
Paging @stepps00!

nvkelso:
(There are a lot more funky areas in Europe.)

stepps00:
@zerebubuth - Can you please dump a shapefile of the country (admin level 2) and region (admin level 4) ways for me, please?

stepps00:
Dates:

**October 24th** - Alpha data revisions in to Nathaniel for review
**October 28th** - Beta data revisions in to Nathaniel for final push (for 1.1 build)

nvkelso:
Related: https://github.com/tilezen/vector-datasource/issues/735

stepps00:
[...]

PR #1128, opened 2016-11-17
Version bump -> 1.0.2

rmarianski:


nvkelso:
Also needs to be updated in [docs] / layers:

- https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#changelog

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+v1.0.2
+------
+* **Release date**: 2016-11-17.
+* **Requires:** [tileserver v1.0.0](https://github.com/mapzen/tileserver/releases/tag/v1.0.0) and [tilequeue v1.0.1](https://github.com/mapzen/tilequeue/releases/tag/v1.0.1) and [mapbox-vector-tile v1.0.0](https://pypi.python.org/pypi/mapbox-vector-tile/v1.0.0).
+* Merge water and earth polygons. See [#1106](https://github.com/tilezen/vector-datasource/issues/1106).
+* Update buffered_land shapefile. See [#294](https://github.com/tilezen/vector-datasource/issues/294).
```
nvkelso: Let's instead say:

* `Improve maritime_boundary tagging in Europe and globally by updating buffered_land shapefile.`

Additionally, in [docs] / layers the following text needs to be removed: `See planned data improvement in #294.`

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+v1.0.2
+------
+* **Release date**: 2016-11-17.
+* **Requires:** [tileserver v1.0.0](https://github.com/mapzen/tileserver/releases/tag/v1.0.0) and [tilequeue v1.0.1](https://github.com/mapzen/tilequeue/releases/tag/v1.0.1) and [mapbox-vector-tile v1.0.0](https://pypi.python.org/pypi/mapbox-vector-tile/v1.0.0).
```
nvkelso: Why do we link to `pypi` here but in [requirements](https://github.com/tilezen/vector-datasource/blob/master/requirements.txt#L22) we link to the Git repo?

Similarly, we link to require specific versions of tileserver and tilequeue. 

Should we have dev and prod install targets instead of defaulting to dev?

nvkelso:
Otherwise looks good to me.

CHANGELOG.md
rmarianski: > Why do we link to pypi here but in requirements we link to the Git repo?

No reason. Which do you prefer?

> Similarly, we link to require specific versions of tileserver and tilequeue.

> Should we have dev and prod install targets instead of defaulting to dev?

I would rather not even have these requirements files checked in these repos in the first place. AFAIK only the getting started instructions use them. Maybe the tests?

CHANGELOG.md
rmarianski: Updated in 72093da.

rmarianski:
> Also needs to be updated in [docs] / layers:
> 
> https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#changelog

Updated in 344bf7b. But do we need to list the version number here as well?

CHANGELOG.md
nvkelso: Final `requirements` question is more existential than for this PR, we can talk about it later.

nvkelso:
Yes, the docs are the only generally publicly documented place we state the version number.

PR #1137, opened 2016-11-22
clarify docs license to CC-BY

nvkelso:
Needs to be paired with docs generator update to `exclude` the new `docs/LICENSE.md` file.


migurski:
I don't think it's necessary to specifically exclude `docs/LICENSE.md` — it’s [not mentioned in `config/vector-tiles.yml`](https://github.com/mapzen/mapzen-docs-generator/blob/master/config/vector-tiles.yml) so it doesn’t appear anyplace in the output.

nvkelso:
@migurski Sounds good, thanks!

PR #1157, opened 2017-01-24
Update test, original node deleted

rmarianski:
@zerebubuth, @iandees, @nvkelso could you review please?

### DIFF
--- a/integration-test/647-cycle-route.py
+++ b/integration-test/647-cycle-route.py
@@ -3,7 +3,7 @@
 # http://www.openstreetmap.org/relation/32386
 assert_has_feature(
     16, 10487, 25327, 'roads',
-    { 'kind': 'major_road', 'cycleway': 'lane', 'bicycle_network': 'lcn' })
+    { 'id': 24335490, 'kind': 'major_road', 'cycleway_right': 'lane', 'bicycle_network': 'lcn' })
 
 # Way: King Street (8920394) http://www.openstreetmap.org/way/8920394
 assert_has_feature(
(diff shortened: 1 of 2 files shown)


zerebubuth:
Looks like `integration-test/647-cycle-route.py` is also failing because

```
Exception: Tile 'http://localhost:57946/roads/16/10487/25327.json': Did not find feature including properties {'kind': 'major_road', 'bicycle_network': 'lcn', 'cycleway': 'lane'}. The closest match was {u'cycleway_right': u'lane', u'name': u'The Embarcadero', u'sort_rank': 380, u'kind_detail': u'primary', u'is_bicycle_related': True, u'kind': u'major_road', u'bicycle_network': u'lcn', u'source': u'openstreetmap.org', u'min_zoom': 8.0, u'oneway': u'yes', u'sidewalk': u'separate', u'id': 88466052}: missed {'cycleway': "None != 'lane'"}.
```

It looks like it's picking up [this way](http://www.openstreetmap.org/way/88466052) instead of [the way it's supposed to](http://www.openstreetmap.org/way/24335490) because the latter was recently edited to have `cycleway:right=lane` rather than `cycleway=lane`. Unless there was a specific reason why we were testing one rather than the other, it's probably good enough just to change that to `cycleway_right`. Is it worth adding an `id` as well, so that it only tries to match the intended element?

zerebubuth:
Got some more fixes in e371f9d, which passes locally for me. Fingers crossed that Circle feels the same way!

rmarianski:
And it does! Ok to merge?

nvkelso:
Woot! Merging.

PR #1169, opened 2017-03-03
Update >z16 test failure coords to use z16

rmarianski:
This updates all the coordinates used in the tests to use the equivalent at z16. I left the water boundaries and highline building high zoom coordinates because they seemed to be more sensitive to the zoom used.


nvkelso:
These generally look good to me. Irwin them on phone, but Circle is unhappy about one or more.

rmarianski:
@zerebubuth thanks for the help getting the errors fixed! :zap:

PR #1180, opened 2017-03-21
Drop small inners

zerebubuth:
At mid and low zooms, polygons with lots of small inners cause problems for simplification, validation, intercut and MVT generation. Dropping these small holes means that some operations are slightly faster, and many are more robust.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -3663,7 +3663,6 @@ def _drop_small_inners(poly, area_tolerance):
     inners = []
     for inner in poly.interiors:
         area = Polygon(inner).area
-        import sys; print>>sys.stderr, "area: %f, area_tolerance: %f" % (area, area_tolerance)
         if area >= area_tolerance:
             inners.append(inner)
 
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -3657,6 +3657,77 @@ def drop_features_mz_min_pixels(ctx):
             feature_layer['features'] = features_to_keep
 
 
+def _drop_small_inners(poly, area_tolerance):
+    ext = poly.exterior
+
+    inners = []
+    for inner in poly.interiors:
+        area = Polygon(inner).area
+        import sys; print>>sys.stderr, "area: %f, area_tolerance: %f" % (area, area_tolerance)
+        if area >= area_tolerance:
+            inners.append(inner)
+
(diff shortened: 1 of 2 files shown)


vectordatasource/transform.py
```diff
@@ -3657,6 +3657,76 @@ def drop_features_mz_min_pixels(ctx):
...
+        "You must provide a pixel_area parameter to drop_small_inners"
+
+    if zoom < start_zoom:
+        return None
+
+    if end_zoom and zoom > end_zoom:
+        return None
+
+    meters_per_pixel_area = calc_meters_per_pixel_area(zoom)
+    area_tolerance = meters_per_pixel_area**2 * pixel_area
```
rmarianski: Do we need to square `meters_per_pixel_area` here?

vectordatasource/transform.py
```diff
@@ -3657,6 +3657,76 @@ def drop_features_mz_min_pixels(ctx):
...
+            if geom_type == 'Polygon':
+                new_shape = _drop_small_inners(shape, area_tolerance)
+                if new_shape:
+                    new_features.append((new_shape, props, fid))
+
+            elif geom_type == 'MultiPolygon':
+                polys = []
+                for g in shape.geoms:
+                    new_g = _drop_small_inners(g, area_tolerance)
+                    if new_g:
```
rmarianski: Is this ever falsey?

vectordatasource/transform.py
zerebubuth: Well spotted!

Yes, I seem to have been confused about whether that was a linear or area ratio. D'oh!

Fixed in 71488a5.

vectordatasource/transform.py
zerebubuth: Ah, yes, as an object it's probably always truthy. I've swapped that for what I really meant; does the remaining polygon have any area? I think we can omit validity checks, as deleting inners seems like it should always leave a valid polygon still valid.

Fixed in 71488a5.

rmarianski: > Ah, yes, as an object it's probably always truthy. I've swapped that for what I really meant; does the remaining polygon have any area?

Perhaps that's what shapely already does? I didn't check, but even so an area check is more explicit.

 >I think we can omit validity checks, as deleting inners seems like it should always leave a valid polygon still valid.

Yea, I had the same reasoning. :+1:

PR #1181, opened 2017-03-22
Specify id in test to fix failure

rmarianski:
There was no id in the `assert_no_matching_feature` call, and it started picking up a different osm way.


zerebubuth:
I should note that it's pretty conventional to _not_ put an ID in the "assert not exists" type tests, as they tend to test that a whole category of features aren't present. In this case, it looks okay, and it's probably not doing a more invasive fix (e.g: testing the feature _is_ present at `z+1` and then testing it's not at `z`). In general, we know we need to change how the test fixtures are done, which will fix this kind of error.

PR #1182, opened 2017-03-22
Fix test failures related to 512px tiles

rmarianski:
WIP - some tests are returning a 404. I'll iterate on enqueueing those and addressing further failures.


integration-test/473-landuse-tier.py
```diff
@@ -13,8 +13,7 @@
 # area 30089300
 assert_has_feature(
     8, 56, 115, 'landuse',
-    { 'kind': 'national_park', 'id': -921675, 'tier': 1,
-      'min_zoom': 8 })
+    { 'kind': 'national_park', 'tier': 1, 'min_zoom': 8 })
 assert_no_matching_feature(
     7, 28, 57, 'landuse',
     { 'kind': 'national_park', 'id': -921675 })
```
zerebubuth: Should we also take the ID off this one? It seems to be trying to test that the particular national park starts to be shown at `z8` on the basis of its area, so it's possible that there's another, bigger one which would cause it to fail without the ID. But if we're not including the ID at `z7`...?

rmarianski: It passed without it, so I didn't mess with it, but happy to change. Would it be better to just bump the zoom level up one instead? Not quite exactly the same test then (testing the min zoom rather than when the tile actually shows up), but probably equivalent enough?

rmarianski: I think I see what you mean here now. Unfortunately, when I take it out the test fails, because it picks up a different `national_park`.

```
Found feature matching properties {'kind': 'national_park'} in layer 'landuse', but was supposed to find none. For example: {u'kind': u'national_park', u'area': 262412992, u'sort_rank': 18, u'source': u'openstreetmap.org', u'min_zoom': 5.0, u'tier': 1, u'id': -930900, u'osm_relation': True}
```

zerebubuth: Should we take the whole `assert_no_matching_feature` test out, then? Or leave it in, but it doesn't seem like it's very useful as a test any more. :wink:

rmarianski: Sounds reasonable to me. Updated in f6c0cee.

integration-test/400-bay-water.py
```diff
@@ -13,5 +13,5 @@
 # Horsens Fjord
 # https://www.openstreetmap.org/relation/1451065
 assert_has_feature(
-    14, 8646, 5112, 'water',
+    14, 8645, 5114, 'water',
```
rmarianski: Works again against prod? It starts failing again for me locally though.

Issue #1185, opened 2017-03-28
1000s of gardens in Edinburgh zoom 12 tile


nvkelso:
- Fix min_zoom calculation for `garden` features
- Followup about requiring (or not requiring) `name` is in https://github.com/tilezen/vector-datasource/issues/1186.

While investigating https://github.com/mapsforge/vtm/issues/341 it looks 1000s of features in the `pois` layer of `kind:garden` that don't have names, creating 18,000+ features in some tiles like:

- http://tile.mapzen.com/mapzen/vector/v1/all/12/2011/1276.topojson

Browsing around OpenStreetMap data here most backyards have been tagged as unnamed, private gardens, example:

- http://www.openstreetmap.org/way/273274870

There appears to be an error in the vector tile filter code where gardens are all getting min_zoom 12 instead of throttling that by area (and really we should require some value of `name` – should we in general  no, but for what kinds of features, feels like separate issue):

- https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L441-L444

Specifically:

```
 min_zoom: { min: [ 12, { max: [ { lit: zoom }, { lit: *tier6_min_zoom } ] }, 16 ] }
```

Should probably drop the `12,` for:

```
 min_zoom: { min: [ { max: [ { lit: zoom }, { lit: *tier6_min_zoom } ] }, 16 ] }
```

Where tier6_min_zoom already is:

```
  - &tier6_min_zoom |
      CASE WHEN way_area >= 500000 THEN 12
           WHEN way_area >= 100000 THEN 13
           WHEN way_area >=  50000 THEN 14
           WHEN way_area >=   5000 THEN 15
           ELSE                         16
      END
```

I don't see any other filters at first glance that have this error, but let's take another pass thru as part of this issue.

nvkelso:
@zerebubuth I'm still seeing lots of small gardens in the example zoom 12 tile, like this one I expected at zoom 15:

```
{"type":"Point",
"properties":
{"source":"openstreetmap.org",
"kind":"garden",
"name":"Moray Place",
"area":44071,"tier":6,"min_zoom":12.0,"id":23187947},
"coordinates":[2020,2508]
}
```

Maybe a bad min_zoom database migration?

nvkelso:
This is working :)

Issue #1085, opened 2016-09-23
Refine poi min_zoom for place_of_worship and other features


nvkelso:
Following up from #1084 which noted a funky min_zoom for Grace Cathedral in SF... 

I suspect:
- significant `place_of_worship` features don't always show up early enough, like Grace Cathedral in SF. I suspect we need to measure areas from landuse polygons differently than areas from buildings to get it more back to the earlier behavior. The current zooms feel good for landuse based areas (cap at zoom 12), but the buildings probably need to reset to be closer to the earlier v0.10 behavior, but with a cap at zoom 14.

<img width="560" alt="screen shot 2016-09-22 at 23 00 12" src="https://cloud.githubusercontent.com/assets/853051/18775695/65e85f62-8118-11e6-8cf7-1958085012b7.png">

<img width="395" alt="screen shot 2016-09-22 at 22 59 49" src="https://cloud.githubusercontent.com/assets/853051/18775696/66639bf0-8118-11e6-8c55-5b2a36ad3cec.png">

<img width="691" alt="screen shot 2016-09-22 at 22 59 38" src="https://cloud.githubusercontent.com/assets/853051/18775698/66690fae-8118-11e6-8081-c5d016c8a69e.png">

<img width="425" alt="screen shot 2016-09-22 at 22 58 34" src="https://cloud.githubusercontent.com/assets/853051/18775697/66666308-8118-11e6-8f89-8d37b9686658.png">
- `garden` zooms just seem way off with far too many at mid-zooms, I suspect because of the 12 in the filter:
  
  `min_zoom: { min: [ 12, { max: [ { lit: zoom }, { lit: *tier6_min_zoom } ] }, 16 ] }`
  
  versus something in the more dominant pattern:
  
  `min_zoom: { min: [ { max: [ 12, { lit: zoom }, { lit: *tier6_min_zoom } ] }, 16 ] }`
  
  In the Bubble Wrap stylesheet I'm using this to throttle `garden` features back:
  
  ```
  garden-area-early:
                   filter:
                       all:
                           - kind: [garden, allotments]
                       any:
                           # limit show smaller landuse areas to higher zooms
                           - { $zoom: [12], area: { max: 500000 } }
                           - { $zoom: [13], area: { max: 100000 } }
                           - { $zoom: [14], area: { max: 10000 } }
                           - { $zoom: [15], area: { max: 5000 } }
                   draw:
                       icons:
                           visible: false
  ```

[...]

PR #1188, opened 2017-03-30
Fix min zoom for garden POIs.

zerebubuth:
Added a test that small gardens mapped in Edinburgh are only visible at z16, not z12. Used the zoom spec from #1085, which seems to fit the general pattern better than just dropping the `12`.

Fixed up a test for which the garden was too small (according to `tier6_min_zoom`) to be visible at z12, bumped down to z13.

Note that tests will fail until #1187 is merged.

Connects to #1185.

### DIFF
--- a/integration-test/1185-garden-min-zoom.py
+++ b/integration-test/1185-garden-min-zoom.py
@@ -1,0 +1,10 @@
+# this garden previously had a min_zoom of 12, but based on its size should be
+# z16 instead.
+#
+# http://www.openstreetmap.org/way/273274870
+assert_has_feature(
+    16, 32182, 20422, 'landuse',
+    { 'kind': 'garden', 'id': 273274870, 'min_zoom': 16, 'tier': 6 })
+assert_has_feature(
+    16, 32182, 20422, 'pois',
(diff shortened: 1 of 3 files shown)


integration-test/1185-garden-min-zoom.py
```diff
@@ -0,0 +1,10 @@
+# this garden previously had a min_zoom of 12, but based on its size should be
+# z16 instead.
+#
+# http://www.openstreetmap.org/way/273274870
+assert_has_feature(
+    16, 32182, 20422, 'landuse',
+    { 'kind': 'garden', 'id': 273274870, 'min_zoom': 16, 'tier': 6 })
+assert_has_feature(
```
nvkelso: This garden is marked `access=private` and doesn't have a name so I was only expecting to see a polygon in the `landuse` layer and no feature in the `pois` layer.

nvkelso: @zerebubuth for above.

nvkelso: (My motivation here is to filter out PONIs without names server side so I can rip that out of the client stylesheet code.)

Issue #775, opened 2016-04-30
Project walking route ref onto roads and paths


nvkelso:
- **What did you see?** Many hiking paths are grouped into relations which have `ref` values, but no `ref` value on the way. The trail's `ref` is the primary method of way finding in this mountain area.
- **What did you expect to see?** I expect to see the relation's `ref` value on the way.
- **What map location are you having problems with?** 
  - https://www.openstreetmap.org/way/127908481
  - https://www.openstreetmap.org/relation/103942

nvkelso:
Followup to: https://github.com/mapzen/vector-datasource/issues/596.

nvkelso:
<img width="1146" alt="screen shot 2016-04-30 at 08 33 39" src="https://cloud.githubusercontent.com/assets/853051/14934362/467c5e58-0eae-11e6-96ac-601a7d72b26e.png">

zerebubuth:
At the moment, we get `mz_networks` from the relations for some road types, which can end up populating the `network` property of the feature. Where we use `mz_networks` we ignore the `network` tag, if present, and set it from the "most important" network from the relations. However, we keep the `ref` tag, even though it might have referred to a network that we ignored.

There are `walking_network` properties, which are fetched from the relations, but only the "most important" single network. The `network` property on lower grade roads is set from the tag, which may or may not be the same as the `walking_network` network. The `ref` will be left as-is.

Does it make sense to rationalise all of these by including the `network` and `ref` from the way along with all the other `network` and `ref` values from the relations, and always setting `network` and `ref` from the "most important" network?

That approach might break a bunch of things, though. :worried:

nvkelso:
Can we isolate the car, walking, cycling refs into their own shield names spaces?

So a way's `walking_network` relation has a ref, than it'd show up as `walking_shield_text` on the way.

And a way's `bicycle_network` relation has a ref, than it'd show up as `bicycle_shield_text` on the way.

nvkelso:
This way has 3 different walking networks:

- https://www.openstreetmap.org/way/373532611
- `JMT`, `PCT`, and ` ` some weird section thing without a ref.

This way has cycle, bus, and driving networks!

- https://www.openstreetmap.org/way/417097119
- `23` (bus), `50` (bicycle), and `35` (driving)

zerebubuth:
Yup, good point.

[...]

Issue #1189, opened 2017-04-04
Calculate line offset for transit route relations


nvkelso:
In San Francisco, New York, London, Paris, Tokyo, and other important transit markets multiple transit "lines" travel along the same "tracks". 

Right now in the **transit** layer we export duplicative features per relation, and in the `roads` layer we have just the track geometry. To offset the lines client-side in Tangram we'd need a single geometry per track part (like in the roads layer), with an ordered list of relations that hang off that geometry.

To allow the Tangram team to prototype I propose we hang an ordered list of transit "lines" (relations) off the `roads` layer geometry with just the line name, color and optionally operator, line shield text (ref), and transit mode (e.g. subway, light rail, rail).

bcamper:
We can prototype with existing data, so no rush there (we can hardcode a few layers with different offsets). But for the "real" feature, it's going to be much easier to implement on the client side with separate geometries for each line (even if that means they're duplicated). With separate lines, we can apply an `offset` to each, using a JS style function that translates a line index number into an offset in pixels. But making the client "multiply" a single geometry into several others based on a feature property would be a significantly greater change to the processing pipeline. While you can currently draw the same feature multiple times, that's based on static layers. There's no facility for dynamically replicating a feature based on arbitrary logic.

nvkelso:
@bcamper Thanks, that's even easier :) I've updated the main issue description.

zerebubuth:
Does the `offset` have to be unique within the tile, or unique "globally"? I had thought that we would assign offsets at junctions, but I'm not sure how we'd encode that into the tile.

Some detailed examples of the output that we want would be very helpful.

bcamper:
We can make a diagram of what I was hoping to have for Tangram.

cc @meetar

zerebubuth:
A diagram would be very helpful, thanks!

nvkelso:
On the Tangram side @meetar has proposed a PR to add support to JS in https://github.com/tangrams/tangram/pull/547. The assumption there is there are different line geoms (like vector tiles have today) for the different lines, but if an `offset` is available it'll push the line away. There's some followup to determine if shader versus geom preprocessing is the way to go (more like Mapbox does it).

When Brett was in the SF office last week we made the following sketches, with make a few assumptions:

- Reuse existing **transit** layer in a backwards compatible way
- Replace existing relation geometry with **ways** (more like we do in the **roads** layer)
- For each way:
[...]

Issue #1175, opened 2017-03-14
Project cycling route ref onto roads and paths


nvkelso:
After we published the new Mapzen Bike Map ([blog post](https://mapzen.com/blog/bike-map/)) we received feedback it would be nice to see special ref shields on the colored bike features.

- **What did you see?** Many cycling roads are grouped into relations which have `ref` values, but no `ref` value on the way. The cycleway's `ref` is useful for way finding in both in cities and for longer distance touring between cities.
- **What did you expect to see?** I expect to see the cycle relation's `ref` value on the way as a new property (similar to what we do for the road shields, there should be a set of singular and plural cycleway network and cycleway shield text properties added)
- **What map location are you having problems with?** 
  - https://www.openstreetmap.org/way/417389551 and it's relation:
    - https://www.openstreetmap.org/relation/32310
    - 16/37.7526/-122.4192

This issue is similar to https://github.com/tilezen/vector-datasource/issues/775 and they can probably be done at the same time.

nvkelso:
Hmm, I'm only rarely seeing this on roads in San Francisco that _do_ have cycle relations, including https://www.openstreetmap.org/way/417389551 which is in the test. Bad dev build?

Oddly I **do** see the right tags on https://www.openstreetmap.org/way/397097600 just a few blocks away.

zerebubuth:
Hmmm... very strange. I just took a look and got a feature with properties:

```json
{
  "kind": "major_road",
  "name": "Valencia St.",
  "sort_rank": 379,
  "kind_detail": "secondary",
  "oneway": "no",
  "is_bicycle_related": true,
  "bicycle_shield_text": "45",
  "cycleway": "lane",
  "source": "openstreetmap.org",
  "min_zoom": 10,
  "landuse_kind": "retail",
  "all_bicycle_shield_texts": [
    "45"
  ],
  "all_bicycle_networks": [
    "lcn"
  ],
  "id": 417389551,
  "bicycle_network": "lcn"
}
```

Which does include the information. But the tile last modified says `Thu, 13 Apr 2017 18:42:03 GMT`. So that's weird. I can't really explain that unless you got a different version of the tile from me, but I'm not sure how that would be the case...

We'll need to rebuild dev anyway because of changes in #1213. Let's check whether this is still an issue after that's all done.

nvkelso:
Verified, looks great!

[...]

PR #1193, opened 2017-04-06
Add walking route refs

zerebubuth:
Networks and refs are taken both from the tags on the way (if they look like `*wn` networks) and from relations that the way might be part of. The logic for road shields is re-used and extended to provide all the networks and `shield_text`s, and select the most important to be `walking_network` and `walking_shield_text`.

This now means that route information is requested for almost all road types, which might make the queries slower. They haven't been noticeably so on my local system, but we'll have to wait and see what they look like on dev.

Connects to #775.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1063,12 +1063,9 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `all_bicycle_networks` and `all_bicycle_shield_texts`: All of the bicycle networks of which this road is a part, and each corresponding shield text. See `bicycle_network` and `bicycle_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
 * `bicycle_network`: Present if the feature is part of a cycling network. If so, the value will be one of `icn` for International Cycling Network, `ncn` for National Cycling Network, `rcn` for Regional Cycling Network, `lcn` for Local Cycling Network.
 * `bicycle_shield_text`: Contains text intended to be displayed on a shield related to the bicycle network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
-* `all_bus_networks` and `all_bus_shield_texts`: All of the bus routes of which this road is a part, and each corresponding shield text. See `bus_network` and `bus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
+* `all_bus_networks` and `all_bus_shield_texts`: All of the bus and trolley-bus routes of which this road is a part, and each corresponding shield text. See `bus_network` and `bus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
 * `bus_network`: Note that this is often not present for bus routes / networks. This may be replaced with `operator` in the future, see [issue 1194](https://github.com/tilezen/vector-datasource/issues/1194).
-* `bus_shield_text`: Contains text intended to be displayed on a shield related to the bus network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
-* `all_trolleybus_networks` and `all_trolleybus_shield_texts`: All of the trolleybus routes of which this road is a part, and each corresponding shield text. See `trolleybus_network` and `trolleybus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
-* `trolleybus_network`: Note that this is often not present for trolley-bus networks / routes.
-* `trolleybus_shield_text`: Contains text intended to be displayed on a shield related to the trolleybus route. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
+* `bus_shield_text`: Contains text intended to be displayed on a shield related to the bus or trolley-bus network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
 * `surface`: Common values include `asphalt`, `unpaved`, `paved`, `ground`, `gravel`, `dirt`, `concrete`, `grass`, `paving_stones`, `compacted`, `sand`, and `cobblestone`.
 
 #### Road properties (optional):
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1035,13 +1035,9 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `all_networks` and `all_shield_texts`: All the networks of which this road is a part, and all of the shield texts. See `network` and `shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
 * `network`: eg: `US:I` for the United States Interstate network, useful for shields and road selections. This only contains _road_ network types. Please see `bicycle_network` and `walking_network` for bicycle and walking networks, respectively.
 * `shield_text`: Contains text to display on a shield. For example, I 90 would have a `network` of `US:I` and a `shield_text` of `90`. The `ref`, `I 90`, is less useful for shield display without further processing. _See planned bug fix in [#1062](https://github.com/tilezen/vector-datasource/issues/1062)._
-* `all_walking_networks` and `all_walking_shield_texts`: All of the walking networks of which this road is a part, and each corresponding shield text. See `walking_network` and `walking_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
-* `walking_network`: e.g: `nwn` for a "National Walking Network". Other common values include `iwn` for international, `rwn` for regional and `lwn` for local walking networks.
-* `walking_shield_text`: Contains text intended to be displayed on a shield related to the walking network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
 
 #### Road properties (common optional):
 
-* `bicycle_network`: Present if the feature is part of a cycling network. If so, the value will be one of `icn` for International Cycling Network, `ncn` for National Cycling Network, `rcn` for Regional Cycling Network, `lcn` for Local Cycling Network.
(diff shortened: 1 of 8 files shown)


integration-test/775-walking-route-refs.py
```diff
@@ -0,0 +1,46 @@
...
+#   type=route, route=road, network=US:CA, ref=35
+assert_has_feature(
+    16, 10469, 25340, 'roads',
+    { 'id': 417097119,
+      'network': 'US:CA',
+      'shield_text': '35',
+      'bicycle_network': 'lcn',
+      'bicycle_shield_text': '50',
+      'bus_network': type(None),
+      'bus_shield_text': '23' })
```
nvkelso: I think this is a new property, and one that may be re-used in the future for other transit shields? 

Propose property names:

- `bus_network` > `transit_network`
- `bus_shield_text` > `transit_shield_text`

vectordatasource/transform.py
```diff
@@ -3862,7 +3862,46 @@ def normalize_operator_values(shape, properties, fid, zoom):
...
+    Take the network and ref tags from the feature and, if they both exist, add
+    them to the mz_networks list. This is to make handling of networks and refs
+    more consistent across elements.
+    """
+
+    network = props.get('network')
+    ref = props.get('ref')
+    mz_networks = props.get('mz_networks', [])
+
+    if network and ref:
```
nvkelso: neat :)

integration-test/775-walking-route-refs.py
```diff
@@ -0,0 +1,46 @@
...
+# https://www.openstreetmap.org/relation/1225378
+#   type=route, route=hiking, network=nwn, ref=PCT
+# https://www.openstreetmap.org/relation/1244828
+#   type=route, route=foot, network=rwn, ref=JMT
+# https://www.openstreetmap.org/relation/1244686
+#   type=route, route=foot, network=rwn, ref="", name="PCT - California Section H"
+assert_has_feature(
+    16, 11044, 25309, 'roads',
+    { 'id': 373532611,
+      'all_walking_networks': ['nwn', 'nwn', 'rwn', 'rwn'],
```
nvkelso: Can we also test for:

- `walking_network`: `nwn`, `walking_shield_texts`: `PCT`

Which might introduce length of shield text sort considerations?

yaml/roads.yaml
```diff
@@ -45,10 +44,9 @@ global:
       sidewalk: {col: sidewalk}
       sidewalk_left: {col: 'sidewalk:left'}
       sidewalk_right: {col: 'sidewalk:right'}
-  - &osm_network_from_relation
+  - &osm_network
```
nvkelso: 👍

integration-test/775-walking-route-refs.py
```diff
@@ -0,0 +1,46 @@
...
+# https://www.openstreetmap.org/relation/1976278
+#   type=route, route=road, network=US:CA, ref=35
+assert_has_feature(
+    16, 10469, 25340, 'roads',
+    { 'id': 417097119,
+      'network': 'US:CA',
+      'shield_text': '35',
+      'bicycle_network': 'lcn',
+      'bicycle_shield_text': '50',
+      'bus_network': type(None),
```
nvkelso: (Or all the other transit already has contextual shield information in the `transit` layer and we're good here? I'm open...)

nvkelso:
I think this also connects to https://github.com/tilezen/vector-datasource/issues/1175 for bike shields and (don't see an issue for bus shields, though https://github.com/tilezen/vector-datasource/issues/1189 mentions it obliquely).

integration-test/775-walking-route-refs.py
zerebubuth: I would imagine this is only bus routes, since it's the roads layer. I suppose it's possible that some tram or trolley-bus routes use the features which end up in the roads layer rather than railways which end up in the transit layer.

zerebubuth:
> I think this also connects to #1175 for bike shields

Yeah, I stopped when I realised I was encroaching on #1175's territory. I'd like to get this merged w.r.t the walking routes and then address the other issues around biking and bus routes in other issues. I find it easier to keep everything organised if issues are pretty focused :smile:

integration-test/775-walking-route-refs.py
zerebubuth: Cool. Added in 44d67e7.

integration-test/775-walking-route-refs.py
zerebubuth: I should add that this won't pick up tram or trolley-bus routes because those have different `route` tags on the relation which haven't been included in this. They could be, in a future relation, if we want. We could mix them all up in a `transit_network` list, or keep them separate in `bus_network`, `tram_network`, `trolleybus_network` lists, whichever makes more sense.

nvkelso: Interesting! Let's leave it as `bus_*` then. But...

Since trolley buses are already in the `is_bus_route` calculation, so I think they should be added to this PR for completeness.

See also:

- https://github.com/tilezen/vector-datasource/blob/master/data/functions.sql#L556
- https://github.com/tilezen/vector-datasource/pull/633

(Trams are already in the transit layer [here](https://github.com/tilezen/vector-datasource/blob/81d415e273a517848b1f338f882f4a88312511ee/yaml/transit.yaml#L15), so trams should be ignored is this PR.)

nvkelso: I'm also fine not doing the `bus_network` && **operator** in this PR – it's more interesting on non-bus transit.

integration-test/775-walking-route-refs.py
nvkelso: Ignore this, these would only ever be bus (or trolley bus which we already consider bus) https://github.com/tilezen/vector-datasource/pull/1193#discussion_r110254398.

integration-test/775-walking-route-refs.py
nvkelso: Operator is tracked in https://github.com/tilezen/vector-datasource/issues/1194.

queries.yaml
```diff
@@ -93,6 +93,7 @@ layers:
       - vectordatasource.transform.normalize_aerialways
       - vectordatasource.transform.normalize_cycleway
       - vectordatasource.transform.add_is_bicycle_related
+      - vectordatasource.transform.merge_networks_from_tags
```
nvkelso: In queries.yaml we'll also want to (selectively) drop shields for new walking, bicycle, and bus at earlier zooms:

- https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L506-L515

Right now `path` isn't included there – and some international and national important bicycle and walking shields will be present on both paths and `minor_road` so this may get tricky.

zerebubuth:
> (I prefer to keep the PR with bike and bus rather than splitting it apart.)

Oops, sorry. I missed this comment. Added that stuff back in, plus trolley-bus stuff, in 0aa1caf and 38bf986.

queries.yaml
zerebubuth: I had a go at this in 5cdf0ec.

docs/layers.md
```diff
@@ -1058,8 +1057,19 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
...
+* `bicycle_network`: Present if the feature is part of a cycling network. If so, the value will be one of `icn` for International Cycling Network, `ncn` for National Cycling Network, `rcn` for Regional Cycling Network, `lcn` for Local Cycling Network.
+* `bicycle_shield_text`: Contains text intended to be displayed on a shield related to the bicycle network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
+* `all_bus_networks` and `all_bus_shield_texts`: All of the bus routes of which this road is a part, and each corresponding shield text. See `bus_network` and `bus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
+* `bus_network`: Note that this is often not present for bus routes / networks. This may be replaced with `operator` in the future, see [issue 1194](https://github.com/tilezen/vector-datasource/issues/1194).
```
nvkelso: 👍

docs/layers.md
```diff
@@ -1058,8 +1057,19 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
...
+* `bus_shield_text`: Contains text intended to be displayed on a shield related to the bus network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
+* `all_trolleybus_networks` and `all_trolleybus_shield_texts`: All of the trolleybus routes of which this road is a part, and each corresponding shield text. See `trolleybus_network` and `trolleybus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
```
nvkelso: Remove the separate `trolleybus` sections here, but note in the `bus` section that it includes trolleybus.

integration-test/1194-bus-route-refs.py
```diff
@@ -0,0 +1,43 @@
...
+#   outbound
+# http://www.openstreetmap.org/relation/2980504
+#   inbound
+assert_has_feature(
+    16, 10477, 25327, 'roads',
+    { 'id': 225516711,
+      'bus_network': type(None),
+      'bus_shield_text': '3',
+      'all_bus_networks': [None, None],
+      'all_bus_shield_texts': ['3', '3'] })
```
nvkelso: Inbound and outbound duplicates are annoying but are workable for this release. But can you file an issue to de-dup these for future followup, please?

queries.yaml
nvkelso: 👍

integration-test/1194-bus-route-refs.py
zerebubuth: Follow-up in #1204.

docs/layers.md
zerebubuth: Ooops! Well spotted. I forgot about the docs. Fixed in 7806fca.

Issue #1171, opened 2017-03-09
Roads with bicycle=designated are missing that property


nvkelso:
From the Bike Map launch for Walkabout ([blog post](https://mapzen.com/blog/bike-map/)) a user [noticed](https://twitter.com/ustroetz/status/839903089872224256) that roads with bicycle=designated tags aren't showing up as blue on the map. It turns out we have config error in our roads filter logic:

Currently we only export `bicycle=*` on footways, but that should be for any road layer feature (so ~~move~~ **copy** it up a couple lines up to the highways section):

- **Needs change:** https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L79

The post-processing operation correctly anticipates the bicycle tag when setting the `is_bicycle_related` boolean, but the data isn't there on most roads because of the bug in the roads.yaml (above).

- **No change:** https://github.com/tilezen/vector-datasource/blob/master/vectordatasource/transform.py#L3577

nvkelso:
This should also expose `bicycle=no` properties as a no-op for motorways and the like that don't allow bikes.

bhousel:
Awesome Bike Map!  

Someone I work with just noticed that `bicycle=dismount` is also being ignored, so there are a bunch of segments styled white that should be off limits.

For example, these pedestrian ways around UPenn:  https://www.openstreetmap.org/way/335523313

nvkelso:
@bhousel Thanks for the report. The `bicycle=dismount` data already existed on paths, not showing it in the map was a style problem so I moved your comment over to https://github.com/tangrams/walkabout-style/issues/152, which is now fixed (but needs a prod deploy to go live).

nvkelso:
(The `is_bicycle_related` logic mentioned above may also need to change for consistency, but for the styling we're good to go.)

bhousel:
Thanks for the quick fix @nvkelso! 👍

nvkelso:
Hmm, I see this property added on the test way in Eureka, California...

But the way in Germany mentioned via Twitter still lacks the `bicycle:designated` tag?

http://www.openstreetmap.org/way/4613300#map=17/52.52742/13.41253

nvkelso:
Verified :) I see that way in Germany now, and a ton more in places like Portland and Berkeley!

[...]

PR #1195, opened 2017-04-06
add bicyle tag to roads

nvkelso:
Connects with #1171 to add bicyle tag to roads

- [x] Update tests
- [x] Updated docs

No change to [transform logic](https://github.com/tilezen/vector-datasource/blob/150b660946fed6f1328450fd9a20eeb88a7d3518/vectordatasource/transform.py#L3590) fir `is_bicycle_related` as it's already working for paths but the required bicycle tag just wasn't present on roads (fixed in this PR).

I don't think this needs a data migration (it's an optional extra tag at select time).


nvkelso:
I thought about adding this to `piste` features, too – but those are a subset duplicate of the paths, and the paths already have the bicycling properties explored so no-op.

Issue #1147, opened 2016-12-27
Add ramp:bicycle tag


nvkelso:
* **What did you see?** Near the main train station in Copenhagen one of the stairs has a ramp for bikes. But which one!?
* **What did you expect to see?** That stairs should be exported with a `ramp:bicycle` boolean.
* **What map location are you having problems with?** 19/55.67089/12.56553
* **Screenshot?** _Props for animated gifs._

- https://www.openstreetmap.org/way/91275149#map=19/55.67089/12.56553

<img width="1029" alt="screen shot 2016-12-27 at 14 31 16" src="https://cloud.githubusercontent.com/assets/853051/21509783/731c7dd4-cc41-11e6-9b3c-cd567653a890.png">

nvkelso:
From TagInfo these values are possible:

- `yes` (2 430) most common
- `no` (959)
- `right` (85)
- `left` (77)
- `both` (6)
- `permissive` (2)
- `dismount` (1)

Since we don't test and exclude `no` on other properties like **bicycle**, I think a simple pass thru is fine here.

nvkelso:
Related tags which have just as much usage:

- `ramp` (most usage)
- `ramp:wheelchair`
- `ramp:bicycle`
- `ramp:stroller`
- `ramp:luggage`

nvkelso:
I don't see the new `ramp_bicycle` tag on the example feature.

nvkelso:
Verified! The Walkabout styling will need to be [updated](https://github.com/tangrams/walkabout-style/issues/125), though as the current bike map styling looks bad there.

<img width="694" alt="screen shot 2017-04-19 at 23 35 39" src="https://cloud.githubusercontent.com/assets/853051/25216777/0a83e444-2559-11e7-8bc5-3695aed8ff86.png">

PR #1196, opened 2017-04-06
add generic ramp and bicycle ramp tags to paths

nvkelso:
Connects to #1147.

- [x] Update tests
- [x] Update docs

I don't think this needs a data migration (it's an optional extra tag at select time).


nvkelso:
Woot!

Issue #1172, opened 2017-03-10
Refine min_zoom for bike related roads


nvkelso:
Now that we launched a bike map style ([blog post](https://mapzen.com/blog/bike-map/)) we're getting feedback that the cycleways would be nice to see earlier generally, and more specifically because the current approach shows large gaps where there is in fact thru connections (see screenshots below).

An [old issue](https://github.com/tilezen/vector-datasource/issues/391) suggested earlier ranges, too. But we ended up being more conservative when the tile feature launched in 2016 trying to make sure that overall tile size remained small, and unexpected minor roads didn't surprise people. 

NOW:

- `icn` THEN **11**
- `ncn` THEN **11**
- `rcn` THEN **12**
- `lcn` THEN **13**

MIGHT TURN INTO:

- `icn` THEN **8** (can't be smaller than 8, as that's when NE roads transition to OSM roads)
- `ncn` THEN **8**
- `rcn` THEN **10** (density in Netherlands is off the charts, looks like rest of EU is filling in like this)
- `lcn` THEN **11** (would fix a few gaps in San Francisco, and since most minor roads appear at zoom 12 anyhow, the earlier zoom 13 setting was probably not that useful)

This issue tracks revisiting the min zoom, reviewing change in tile size and tile generation speeds, and style (Tangram scene file for Mapzen Basemap) implications.

The road `min_zoom` is controlled in the **functions.sql** (which is called on each kind of roads in the road.yaml, and also controls walking path min zooms):

- https://github.com/tilezen/vector-datasource/blob/master/data/functions.sql#L629-L653

```
-- returns TRUE if the given way ID (osm_id) is part of a path route relation,
-- or NULL otherwise.
-- This function is meant to be called for something that we already know is a path.
-- Please ensure input is already a path, or output will be undefined.
CREATE OR REPLACE FUNCTION mz_calculate_path_major_route(osm_id BIGINT)
RETURNS SMALLINT AS $$
BEGIN
  RETURN (
    SELECT
        MIN(
            CASE WHEN hstore(tags)->'network' IN ('iwn', 'nwn', 'icn','ncn') THEN 11
                 WHEN hstore(tags)->'network' IN ('rwn', 'rcn') THEN 12
                 WHEN hstore(tags)->'network' IN ('lwn', 'lcn') THEN 13
            ELSE NULL
            END
        )
        AS p
     FROM planet_osm_rels
    WHERE
      parts && ARRAY[osm_id] AND
[...]

PR #1198, opened 2017-04-07
earlier bike roads (and walking routes)

nvkelso:
Connects to #1172 to show bike and walking related ways a lot earlier, and #1214 for dropping early all_* network and shield texts. 

- [x] Update tests
- [x] Update data migrations (as it's a `min_zoom` change)
- [x] Update docs


nvkelso:
I added new data migration in https://github.com/tilezen/vector-datasource/pull/1198/commits/27a5926fc8663686c9d0a5ed085af1c0897743d9 and removed older v1.0 migrations in https://github.com/tilezen/vector-datasource/pull/1198/commits/826a2393debf3455da19d381252f948f30c24e1c, but had to add the older v1.0 migrations back in https://github.com/tilezen/vector-datasource/pull/1198/commits/d380e9a1926c50c6092bf77c6c229e96a5a030bf as CircleCI barfed when they were missing.

See also: 

- https://github.com/tilezen/vector-datasource/blob/master/CONTRIBUTING.md#7-update-data-migrations

nvkelso:
I've fixed the **migrations** in https://github.com/tilezen/vector-datasource/pull/1198/commits/e6817dfc4ea5af207f748b0483651edc8d8ec51f based on conversation with @rmarianski – it's just the version'd `*.sql` files that need to be removed (the others are required).

For tests, note that @zerebubuth's changes in [vector-datasource/pull/1193](https://github.com/tilezen/vector-datasource/pull/1193/files) for when the walking route and bike route **network** property is dropped impacts which props I can test for in his https://github.com/tilezen/vector-datasource/pull/1193/files#diff-fc49e5c256610bc0b32a7de5a0520835.

nvkelso:
@zerebubuth This is _finally_ ready for review :)

data/migrations/v1.3.0-line.sql
```diff
@@ -0,0 +1,34 @@
...
+  SET mz_road_level = mz_calculate_min_zoom_roads(planet_osm_line.*)
+  WHERE
+      (tags -> 'highway' IN ( 'trunk_link', 'primary_link', 'secondary_link', 'tertiary_link',
+                   'residential', 'unclassified', 'road', 'living_street', 'pedestrian',
+                   'path', 'track', 'cycleway', 'bridleway', 'footway', 'steps',
+                   'service' ) OR
+       tags -> 'whitewater' = 'portage_way')
+      AND mz_calculate_min_zoom_roads(planet_osm_line.*) IS NOT NULL;
+
+CREATE INDEX new_planet_osm_line_roads_geom_index ON planet_osm_line USING gist(way) WHERE mz_road_level IS NOT NULL;
```
zerebubuth: It shouldn't be necessary to re-create the indexes, they should be updated along with the rows in the `UPDATE` statement above.

queries.yaml
```diff
@@ -516,32 +516,24 @@ post_process:
...
-  # previous tests assume we keep walking network information until zoom 11
   - fn: vectordatasource.transform.drop_properties
     params:
       source_layer: roads
       start_zoom: 0
-      end_zoom: 10
+      end_zoom: 13
       properties:
-        - walking_network
-        - walking_shield_text
```
zerebubuth: This means that walking networks will be kept on the roads until zoom 7, but the earliest `min_zoom` for a walking route is zoom 9. This could block opportunities for merging roads at low zooms. Would we display walking data at z7-8? If not, we should remove this data.

nvkelso: Good point, I'll remove the data at those zooms.

data/migrations/v1.3.0-line.sql
nvkelso: Fixed in 392c88e4e1b2015f2b20ca0e4ddf5d698b1777f8.

queries.yaml
nvkelso: Fixed in 7cd5ffcbfc7eab939a55d22f35596285ee1bee54. 

I also removed the path name at early zooms to enable more line merging – that was necessary for v1.0 when we didn't have the walking network information but is now getting in the way.

nvkelso:
Tests pass (again again), comments address. Merging :)

Issue #1191, opened 2017-04-04
Improve road/line merging


bcamper:
We currently merge certain features, such as roads, by shared properties. This can create large MultiLineString geometries, which reduces property duplication and client-side processing time for feature-filtering. However, there are two ways these merged features can be substantially improved:

1. Many (in some cases a large %) of the individual LineStrings within those MultiLineString are often **contiguous** (e.g. the last coordinate on one LineString is the same as the first coordinate of the next LineString), and could be further coalesced into a smaller number of LineStrings, eliminating data duplication (repeating coordinates) and reducing client-side time for line processing (avoid per-LineString setup/overhead).

2. Within these LineStrings, there are often unnecessary extra vertices that are **collinear** (even after the secondary merge step above) -- such as individual nodes at intersections on a long road. These can be removed to further reduce the vertex count and improve client-side geometry build and render time.

I ran some client-side post-processing tests to gauge the impact, but I think it makes most sense to do this once server-side, as an extension of the existing merge transforms. In this example, two additional transforms were applied to the data:

- For each MultiLineString, the constituent LineStrings were iterated in order, and fused where the last/first coordinates matched.
- Then, the coordinates of each resulting LineString were iterated, and intermediate collinear points (according to a configurable tolerance) were removed (so this is a more narrow form of simplification).

In these images, red dots show each road vertex (only minor roads are shown). For the initial (un-modified) tile data, it takes **144,378** polygons to render these roads. After the additional merging steps, the # of polygons was reduced to **80,414** -- a **44% reduction**. (In this example there are a couple places where the geometry is slightly altered, but with proper tolerance tweaking it should be possible to do these transforms with no visual changes.)

Here is the before/after:
![dissolve-roads](https://cloud.githubusercontent.com/assets/16733/24681602/f629a3c2-1963-11e7-9450-8ddc4c227442.gif)

Additionally, this merging can improve **label placement** by creating longer individual LineStrings upon which more labels can fit.

nvkelso:
Great GIF!

zerebubuth:
This is already done server side in the [`merge_line_features`](https://github.com/tilezen/vector-datasource/blob/master/vectordatasource/transform.py#L3055) post-processor. However,

1. The line merging code might not be doing what we expect, so we should add extra tests around that.
[...]

PR #1201, opened 2017-04-07
Re-simplify roads after merging.

zerebubuth:
We have already simplified the road network and clipped it to the tile by the time we merge (multi) linestrings for the same road together. The problem is that the line merging algorithm will make a single linestring from a multi-linestring, but won't drop the vertices at the merge.

This change introduces a new post-processing filter to run simplification on a single layer and runs it on the road layer post-merge. This seems to bring the vertex count down significantly.

Connects to #1191.


nvkelso:
Check back in 30 minutes for CircleCI to finish ;)

iandees:
It might be good to add a test for bridges or tunnels or something where we need to keep the linestrings separate for ordering purposes.

Issue #1103, opened 2016-10-05
Remove natural_forest and natural_wood from pois layer


nvkelso:
These should only be label placements in the landuse layer but are now appearing in the pois layer.

```
  # natural forests - note already clamped to z12
  - filter: {natural: forest}
    min_zoom: { min: [ { max: [ 12, { lit: zoom + 2 }, { lit: *tier2_min_zoom } ] }, 14 ] }
    output: { kind: natural_forest, tier: 2 }

  # natural woods - note already clampted to 12
  - filter: {natural: wood}
    min_zoom: { min: [ { max: [ 12, { lit: zoom + 2 }, { lit: *tier2_min_zoom } ] }, 14 ] }
    output: { kind: natural_wood, tier: 2 }
```

Example:
- http://tile.dev.mapzen.com/mapzen/vector/v1/all/14/8184/5448.topojson

```
{
type: "Point",
properties: {
kind: "natural_wood",
area: 2426,
source: "openstreetmap.org",
min_zoom: 14,
tier: 2,
id: -1756198,
osm_relation: true
},
coordinates: [
388,
933
]
},
```

<img width="394" alt="screen shot 2016-10-04 at 22 08 40" src="https://cloud.githubusercontent.com/assets/853051/19101687/286fcd9e-8a7f-11e6-8135-000d69863955.png">

Curious that we're even showing those without any names!

nvkelso:
Possibly also `village_green`?

<img width="597" alt="screen shot 2016-10-04 at 22 41 54" src="https://cloud.githubusercontent.com/assets/853051/19102278/c9cb4f70-8a83-11e6-88a4-70ed4ec95e43.png">

nvkelso:
I can find the 2nd example for **Mt. Cydonia Ponds Natural Area** label position but not the 3rd example in this test for **Liebesinsel**. Both were fresh generated tiles.

- https://github.com/tilezen/vector-datasource/pull/1205/files#diff-35adbc8f1a5e80a9c3e72c1ed1397caa

zerebubuth:
I think this might have been because tileserver and tilequeue were using different date prefixes - I forgot to update the tileserver one after the dev build finished. It's looking okay for me now. Could you check that it's looking good for you too, please?

nvkelso:
Verified :)

PR #1205, opened 2017-04-12
Remove `natural_wood` and `natural_forest` POIs.

zerebubuth:
These are no longer present in the POIs layer, but will be present in the landuse layer as label placements.

Connects to #1103.

@nvkelso: One side-effect of this is that `natural=wood` or `natural=forest` nodes will disappear, as we do not select point features for the landuse layer. We only create points from polygons as label placements. Is that the desired behaviour, or should we also include landuse nodes as label placement points?

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -870,8 +870,6 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `museum`
 * `music`
 * `national_park`
-* `natural_forest` - _See planned bug fixes in [#1103](https://github.com/tilezen/vector-datasource/issues/1103)._
-* `natural_wood` - _See planned bug fixes in [#1103](https://github.com/tilezen/vector-datasource/issues/1103)._
 * `nature_reserve`
 * `newspaper`
 * `ngo`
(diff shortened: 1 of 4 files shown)


nvkelso:
I think it's fine to drop those, but do you have a counter-example?

zerebubuth:
> I think it's fine to drop those, but do you have a counter-example?

I don't understand, what do you mean by a counter-example?

We've got tests for generated points in `landuse`, e.g: [for `natural_forest`](https://github.com/tilezen/vector-datasource/blob/a29bd5d0da47731e1c3c887ef65a11454e4f9ad0/integration-test/742-predictable-layers-pois.py#L80-L83) and [here too](https://github.com/tilezen/vector-datasource/blob/a29bd5d0da47731e1c3c887ef65a11454e4f9ad0/integration-test/1103-no-natural-pois.py#L26-L28). And also that [no points are generated in POIs](https://github.com/tilezen/vector-datasource/blob/a29bd5d0da47731e1c3c887ef65a11454e4f9ad0/integration-test/1103-no-natural-pois.py#L29-L31)). And also that [nodes don't create POIs either](https://github.com/tilezen/vector-datasource/blob/a29bd5d0da47731e1c3c887ef65a11454e4f9ad0/integration-test/742-predictable-layers-pois.py#L89-L91). Are there other cases you'd want covered too?

Issue #1208, opened 2017-04-17
Bug/Typo in pois.yaml


paulmach:
* **What did you see?**
lines [1135-6](https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L1135) in `yaml/pois.yaml` are:
```
output:
  kind: {col: tourism, tier: 6}
```

* **What did you expect to see?**
I expected to see `output: { kind: { col: tourism}, tier: 6}`  which is what makes sense and found other places like [line 211](https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L211)

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

NA

* **Screenshot?** _Props for animated gifs._

NA

nvkelso:
More legible code is always good, thanks for the fix :)

nvkelso:
Verified in dev.

PR #1209, opened 2017-04-17
Fix bug/type is tourism attraction pois.yaml

paulmach:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

☝️ ??

I think this is a type since `kind: {col: tourism, tier: 6}` doesn't make sense.  Also given the filter the tourism column/tag will always be 'attraction'

Fixes: https://github.com/tilezen/vector-datasource/issues/1208

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -440,7 +440,8 @@ filters:
     # tier5_min_zoom
     min_zoom: 17
     output:
-      kind: {col: attraction, tier: 5}
+      kind: {col: attraction}
+      tier: 5
   # resort
   - filter: {tourism: resort}
     min_zoom: { min: [ { max: [ 14, { lit: zoom + 5.32 }, { lit: *tier5_min_zoom } ] }, 17 ] }
@@ -464,7 +465,8 @@ filters:
     # already clamped below polygon area min
     min_zoom: 17
     output:
-      kind: {col: tourism, tier: 6}
+      kind: {col: tourism}
+      tier: 6
   # common - no POI
   # garden
   - filter:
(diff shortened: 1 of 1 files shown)


yaml/pois.yaml
```diff
@@ -1131,5 +1131,4 @@ filters:
       tourism: attraction
     # already clamped below polygon area min
     min_zoom: 17
-    output:
-      kind: {col: tourism, tier: 6}
+    output: { kind: attraction, tier: 6}
```
nvkelso: While this is correct, I suggest separating onto individual lines like:

```
output:
  kind: attraction
  tier: 6
```

For example:

https://github.com/paulmach/vector-datasource/blob/628a0aa905f3146694644f75cd65f2c247f71fd5/yaml/pois.yaml#L159-L165

nvkelso:
@paulmach Thanks for the contribution! It's more legible, and fixes a potential bug with the tier assignment :)

paulmach:
Thanks @nvkelso  fixedup

nvkelso:
Just to be safe I'm going to hold off merging until CicleCI is happy – it's failing now because of an unrelated Overpass issue.

paulmach:
The first one I caught while reading. But my script caught 3 more cases.

nvkelso:
Oh! Can you share, please?

paulmach:
I'm writing code that, for a given osm element, evaluates the filters and if there is a match, generates the outputs. It got confused by these lines.  I will share once done. Needs a few more things to get it to work on all the YAML files.

nvkelso:
@paulmach cool, please file new issues for those.

PR #1213, opened 2017-04-19
Fix migrations and tests for dev

zerebubuth:
The migrations didn't include changes for things with no names. Those have been added to the migration now.

Other changes:
* Lots of tests were using z15 coordinates and IDs to find the relevant features. In a full database, these are often merged, and so lose their IDs. These were replaced with z16 coordinates instead.
* Because more roads are available in the full database, some tests needed to be adjusted to have higher tolerances.
* Because more relations are available in the full database, some choices of which relation was most important and used as `shield_text` changed. It wasn't important to the test exactly which shield was used, so I made the test insensitive to that.

Once this is merged, I'll deploy again, run the migration and re-seed to a new bucket.

### DIFF
--- a/integration-test.py
+++ b/integration-test.py
@@ -160,8 +160,7 @@ def match_distance(actual, expected):
 
 
 class GatewayTimeout(Exception):
-    def __init__(self, message):
-        self.message = message
+    pass
 
 
 def tile_url(z, x, y, layer):
--- a/integration-test.py
+++ b/integration-test.py
@@ -159,21 +159,57 @@ def match_distance(actual, expected):
     return (distance, misses)
 
 
-@contextmanager
-def features_in_tile_layer(z, x, y, layer):
+class GatewayTimeout(Exception):
+    def __init__(self, message):
+        self.message = message
+
+
+def tile_url(z, x, y, layer):
     assert config_url, "Tile URL is not configured, is your config file set up?"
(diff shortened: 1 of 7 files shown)


integration-test.py
```diff
@@ -159,21 +159,57 @@ def match_distance(actual, expected):
     return (distance, misses)
 
 
-@contextmanager
-def features_in_tile_layer(z, x, y, layer):
+class GatewayTimeout(Exception):
+    def __init__(self, message):
+        self.message = message
```
rmarianski: The constructor should be provided for you, and should work this way. In other words, should be similar to:

```python
class GatewayTimeout(Exception):
    pass
```

But if you want to be more explicit it can be left in. In that case, it might not matter but feels safer to add in a call to super:
```python
    super(GatewayTimeout, self).__init__(message)
```

This is test code anyway, so perfectly fine to just leave as is.

zerebubuth: Thanks! I didn't know about the `pass` thing. I guess that just means it'll use `Exception`'s `__init__` method? Fixed in e03f14a.

rmarianski: Yea, exactly. `pass` is just like a "nothing more to see here" placeholder. You can alternatively add something like a docstring and it'll work the same way. For example:

```python
class GatewayTimeout(Exception):
    """Raised when requests for tiles time out"""
```

Issue #1190, opened 2017-04-04
Calculate colour_name property for transit lines


nvkelso:
Right now we pass thru the raw OpenStreetMap line colors on **transit** layer line features. But many of those are "dirty" colors that don't pair well with basemap colors (and we have at least 6 Mapzen basemap styles now), and are a mix of HEX (`#ff0000`) and named colors (`red`).

We'd like to have an additional set of "Crayola 64" normalized line colors (keep the `color` we have now, but add a curated set). Maybe this is "Crayola" or maybe it's a custom set of transit colors based on what we see in the data (e.g.: `blue`, `blue_light`, `blue_dark`, etc). But each stylesheet could then map those normalized colors to a version that pairs with the rest of the colors in the map, without having to curate by line.

Turns out @thisisaaronland did similar work for Cooper Hewitt a couple years ago that might be useful here:

- https://github.com/cooperhewitt/py-cooperhewitt-swatchbook
- https://github.com/cooperhewitt/py-cooperhewitt-roboteyes-colors
- https://github.com/cooperhewitt/plumbing-palette-server

/cc @sensescape @irees

nvkelso:
Let's export a big list of all the transit route relations, their colors, system operators, and line names for @sensescape to review. I'll ask Ian Rees for the same from Transit.land.

zerebubuth:
Here's a list of all the colours I can find from the bus, train, railway, tram, trolleybus, light_rail and subway route relations; [route_colours.txt](https://github.com/tilezen/vector-datasource/files/930495/route_colours.txt). Note that it's actually a CSV file, but Github doesn't allow attaching those.

Although the selection of blue, green, red and black are most popular, and account for almost 15% of all entries, the majority of entries in the list are hex colours, with a few `GoldenRod`s thrown in for good measure.

Some interesting highlights:

* Apparently something has the colour `*`, which must mean it sparkles at night? Thankfully, I think we have a shader for that.
* One thing has a colour of `fixme`, which I'm unfamiliar with. I must look that up in the Pantone catalogue.
* It's not clear to me whether there's a route with colour `ffffff`, or the author was just frustrated.
* Something is apparently `ex:brown`, although I don't know what the new colour is.

:wink:

zerebubuth:
[...]

PR #1222, opened 2017-04-24
Add transit line palette colours

zerebubuth:
Attempts to parse the `colour` tag into an RGB colour and match it against a fixed palette. The name of the colour is added as an alternative property to allow the style designer to remap it to something which is more appropriate for the design.

Connects to #1190.


docs/layers.md
```diff
@@ -1207,6 +1208,61 @@ Depending on OpenStreetMap tagging, the following properties may be present for
 * `train`
 * `tram`
 
+#### Transit `palette` values:
```
nvkelso: `palette` to `colour_name`

docs/layers.md
```diff
@@ -1184,6 +1184,7 @@ A smaller set is also available for non-`platform` features:
 
 * `colour`: either a `#rrggbb` hex value, or a CSS colour name (like `red`)
 * `layer`
+* `palette`: A colour name from a fixed palette, see description below.
```
nvkelso: `palette` to `colour_name`

integration-test/1190-mz-colours.py
```diff
@@ -0,0 +1,5 @@
+# http://www.openstreetmap.org/relation/366773
+assert_has_feature(
+    16, 19310, 24645, 'transit',
+    { 'kind': 'subway', 'ref': 'Z',
+      'colour': '#996633', 'palette': 'peru' })
```
nvkelso: `palette` to `colour_name`

queries.yaml
```diff
@@ -702,3 +702,58 @@ post_process:
     params:
       source_layer: buildings
       start_zoom: 15
+
+  - fn: vectordatasource.transform.palettize_colours
+    params:
+      layer: transit
+      attribute: palette
```
nvkelso: `palette` to `colour_name`

scripts/csv_colours.py
```diff
@@ -0,0 +1,59 @@
...
+import yaml
+
+
+def to_hex(colour):
+    return "#%02x%02x%02x" % tuple(colour)
+
+
+yaml_config_file = None
+output_file = sys.stdout
+mz_colour_key = 'MZ colour'
```
nvkelso: We should use tilezen instead of mapzen.

scripts/csv_colours.py
```diff
@@ -0,0 +1,59 @@
...
+(options, args) = parser.parse_args()
+
+if options.output:
+    output_file = open(options.output, 'wb')
+
+with open(options.config, 'rb') as yaml_fh:
+    config = yaml.load(yaml_fh)
+
+if options.colour_key:
+    mz_colour_key = options.colour_key
```
nvkelso: `mz` > `tilezen`

scripts/csv_colours.py
```diff
@@ -0,0 +1,59 @@
...
+
+palette = Palette(config['colours'])
+
+for file_name in args:
+    with open(file_name, 'rb') as csvfile:
+        reader = csv.DictReader(csvfile)
+        writer = None
+        for row in reader:
+            if writer is None:
+                keys = list(row.keys()) + [mz_colour_key]
```
nvkelso: mz > tilezen

vectordatasource/transform.py
```diff
@@ -4267,3 +4268,76 @@ def truncate_min_zoom_to_2dp(shape, properties, fid, zoom):
         properties['min_zoom'] = round(min_zoom, 2)
 
     return shape, properties, fid
+
+
+class Palette(object):
```
nvkelso: Please add an comment like "Inspired by https://github.com/cooperhewitt/py-cooperhewitt-swatchbook" here (or elsewhere if there's a good spot).

docs/layers.md
zerebubuth: Fixed in ed2f38c.

integration-test/1190-mz-colours.py
zerebubuth: Fixed in ed2f38c.

queries.yaml
zerebubuth: Fixed in ed2f38c.

scripts/csv_colours.py
zerebubuth: As a general point, we use the `mz_` prefix all over the place, so it doesn't make much sense to use tilezen here rather than `mz_`. I've changed it to `output_colour_key` in ed2f38c. And if we want to do a more general `mz -> tz` change, then we should do that as a follow-up issue.

scripts/csv_colours.py
zerebubuth: Fixed in ed2f38c.

scripts/csv_colours.py
zerebubuth: Fixed in ed2f38c.

vectordatasource/transform.py
zerebubuth: Sure, it's nice to give a hat-tip. Although the idea of palettizing to a small, fixed set of colours goes way back before that. Fixed in 9657644.

docs/layers.md
```diff
@@ -1184,6 +1184,7 @@ A smaller set is also available for non-`platform` features:
 
 * `colour`: either a `#rrggbb` hex value, or a CSS colour name (like `red`)
 * `layer`
+* `colour_name`: A colour name from a fixed palette, see description below.
```
nvkelso: Sort alphabetically

zerebubuth: Thanks for spotting that. Fixed in 5aa1d6e.

Issue #1218, opened 2017-04-21
Allow additional PONIs


nvkelso:
Following up from https://github.com/tilezen/vector-datasource/issues/1186 where we required POIs have names, except for whitelists...

**We should also white-list:**

- `railway=level_crossing` (but make min_zoom 18), example: https://www.openstreetmap.org/node/4608620475
- `highway: [bus_stop, traffic_signals]`
- `amenity: [atm, bus_stop]`
- `tourism: [viewpoint]`
- `tourism: wilderness_hut`
- `amenity: [fuel]`
- `waterway: waterfall`
- `natural: waterfall`
- `man_made: telescope`
- `amenity: [post_box, telephone]`
- `leisure: slipway`
- `amenity: bicycle_rental`
- `railway: subway_entrance`
- `amenity: [bench, waste_basket]`
- `man_made: [mineshaft]`
- `man_made: adit`
- `leisure: dog_park`
- `landuse: quarry`
- `highway: mini_roundabout`
- `lock: 'yes'`
- `kind: lock`
- `man_made: power_wind`
- `natural: cave_entrance`
- `highway: ford`
- `man_made: offshore_platform`
- `railway: [halt, stop, tram_stop]`
- `railway: platform`
- `highway: platform`
- `public_transport: platform`
- `public_transport: stop_area`
- `tags->site: stop_area`
- `tourism: alpine_hut`
- `aeroway: helipad`
- `amenity: [car_sharing, recycling, shelter]`
- `tags->whitewater: [put_in, egress]`
- `tags->whitewater: [hazard, rapid]`

**Remove from list, though:**

- `waterway: dam` - the water body carries this information already (and there is no good dam icon)
- `highway: rest_area` - too many unnamed dinky rest areas marked in the SF Bay Area (and in Paris, Berlin...)

nvkelso:
I'm not seeing this on **dev** :\

I suspect it's because we need to update migrations for _all_ whitelisted PONIs:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L89-L122

But probably we need to run migration for ANY feature that is a POI per general name / no-name filter:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L133-L141

nvkelso:
I've filed https://github.com/tilezen/vector-datasource/pull/1233 which should clean up dev.

[...]

PR #1223, opened 2017-04-26
add additional PONIs to whitelist

nvkelso:
Connects with https://github.com/tilezen/vector-datasource/issues/1218.

- [x] Update tests
- [x] Update data migrations
- [ ] ~~Update docs~~

### DIFF
--- a/data/migrations/v1.3.0-point.sql
+++ b/data/migrations/v1.3.0-point.sql
@@ -5,5 +5,6 @@ UPDATE
     (tags -> 'leisure' = 'garden' OR
      tags -> 'landuse' = 'village_green' OR
      tags -> 'natural' IN ('wood', 'forest') OR
+     tags -> 'tourism' IN ('viewpoint') OR
      (NOT tags ? 'name' AND mz_poi_min_zoom IS NOT NULL))
     AND COALESCE(mz_poi_min_zoom, 999) <> COALESCE(mz_calculate_min_zoom_pois(p.*), 999);
--- a/data/migrations/v1.3.0-point.sql
+++ b/data/migrations/v1.3.0-point.sql
@@ -5,6 +5,7 @@ UPDATE
     (tags -> 'leisure' = 'garden' OR
      tags -> 'landuse' = 'village_green' OR
      tags -> 'natural' IN ('wood', 'forest') OR
+     tags -> 'railway' = 'level_crossing' OR
      tags -> 'tourism' IN ('viewpoint') OR
      (NOT tags ? 'name' AND mz_poi_min_zoom IS NOT NULL))
     AND COALESCE(mz_poi_min_zoom, 999) <> COALESCE(mz_calculate_min_zoom_pois(p.*), 999);
--- a/integration-test/1218-poni-whitelist.py
+++ b/integration-test/1218-poni-whitelist.py
@@ -435,3 +435,10 @@ assert_has_feature(
 assert_has_feature(
     16, 15995, 25090, 'pois',
     {'id': 4216584100})
+
+# originally from 657-natural-man_made.py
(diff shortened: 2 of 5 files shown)


integration-test/1218-poni-whitelist.py
```diff
@@ -0,0 +1,309 @@
+# https://www.openstreetmap.org/node/2207370738
+# 15/5230/12657
+# aeroway"="helipad"
+assert_matching_feature(
```
zerebubuth: Rather confusingly, this is called `assert_has_feature` even though its opposite is `assert_no_matching_feature`.

yaml/pois.yaml
```diff
@@ -87,19 +87,28 @@ global:
...
+            rest_area, traffic_signals, trailhead ]
         - historic: landmark
-        - leisure: [ firepit, fishing, pitch, slipway, swimming_area ]
-        - man_made: [ communications_tower, mast, observatory, petroleum_well,
-            water_tower, water_well, windmill ]
-        - natural: [peak, volcano, geyser, hot_spring, saddle, spring, tree ]
+        - landuse: quarry
+        - leisure: [ dog_park, firepit, fishing, pitch, slipway, swimming_area ]
+        - lock: yes
+        - man_made: [ adit, communications_tower, mast, mineshaft, observatory,
```
zerebubuth: The addition of `mineshaft` here seems to have made a [previous test](https://github.com/tilezen/vector-datasource/blob/bc8c87e7a361ed7424af1cff39390462514b75c0/integration-test/1186-the-pois-with-no-name.py#L29) fail. There's probably a whole load more of these given all the extra PONIs that are being allowed. Might it be shorter and more readable to switch this meaning of the list around and blacklist those for which no name is not allowed?

integration-test/1218-poni-whitelist.py
nvkelso: Doh! Fixed in 0a190c4a12ea0825f526378140c6309e96b9cb06.

yaml/pois.yaml
nvkelso: Thanks for the `mineshaft` tip, I've fixed that and a few other obvious ones in 24d51a1f7b60555f8d05bbecbf790426804ac7a2. Will followup if CircleCI complains. It's kinda 6 of one, 1/2 dozen of the other at this point and these test (yours and mine) – can do that for new tests.

zerebubuth: Yup, it's definitely less useful when the tests fail because we knowingly changed the behaviour. But worth it for the times they help us catch unintended regressions!

nvkelso: _Steadily increasing test coverage since 2016..._

nvkelso:
@zerebubuth I believe all these tests pass now (because the last failure I fixed was the 2nd to last test feature), but CircleCI is having trouble starting the testing server mid way thru. Can you look at this in the AM, please?

zerebubuth:
33f4ea1 seems to have fixed the tests - CircleCI is giving a nice green tick!

data/migrations/v1.3.0-point.sql
```diff
@@ -5,5 +5,6 @@ UPDATE
     (tags -> 'leisure' = 'garden' OR
      tags -> 'landuse' = 'village_green' OR
      tags -> 'natural' IN ('wood', 'forest') OR
+     tags -> 'tourism' IN ('viewpoint') OR
```
zerebubuth: Do we need a migration for `tags->'railway' = 'level_crossing'` as well?

nvkelso: Fixed in adb22653793c816c85057434c6d6ead1099eb2b8.

PR #1228, opened 2017-04-27
Adjust min zoom for piers

zerebubuth:
Piers are part of the recognisable shoreline, so even small piers should be visible earlier.

Connects to #1198.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1148,7 +1148,7 @@ Railway `service` values are:
 
 ![image](images/mapzen-vector-tile-docs-roads-piers.png)
 
-**Piers** start showing up at zoom 11+ with `kind_detail` values of `pier`.
+**Piers** start showing up at zoom 13+ with `kind_detail` values of `pier`.
 
 ## Transit
 
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1148,7 +1148,7 @@ Railway `service` values are:
 
 ![image](images/mapzen-vector-tile-docs-roads-piers.png)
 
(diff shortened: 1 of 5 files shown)


docs/layers.md
```diff
@@ -1148,7 +1148,7 @@ Railway `service` values are:
 
 ![image](images/mapzen-vector-tile-docs-roads-piers.png)
 
-**Piers** start showing up at zoom 13+ with `kind_detail` values of `pier`.
+**Piers** start showing up at zoom 11+ with `kind_detail` values of `pier`.
```
nvkelso: This is the road layer docs, please reset.

We generally don't indicate which zooms for which landuse so I'd leave this change undocumented.

yaml/landuse.yaml
```diff
@@ -390,7 +390,7 @@ filters:
   - filter:
       man_made: pier
       way_area: {expr: "way_area > 0"}
-    min_zoom: *tier4_min_zoom
+    min_zoom: GREATEST(LEAST(zoom + 1.81, 16), 11)
     output: {kind: pier, tier: 4}
```
nvkelso: We should probably remove `, tier: 4` here?

nvkelso: Fixed in 7e8b862789b2e01d553fdcd8dd7a859374162d56.

docs/layers.md
nvkelso: Fixed in 59661b1e7afdd7d4fdc7a8d92412d70b6122802e.

PR #1233, opened 2017-04-30
v1.3 add all whitelisted PONIs to points migration

nvkelso:
Followup to https://github.com/tilezen/vector-datasource/pull/1223 and https://github.com/tilezen/vector-datasource/pull/1192 per comments in https://github.com/tilezen/vector-datasource/issues/1218#issuecomment-298213905.


nvkelso:
I don't think it would be harmful to merge it and take to prod, but I'm also not sure it's necessary since the basic migration already in master would effectively only _demote_ pois, while this one will make sure they are _promoted_ back into legitimacy.

rmarianski:
I went ahead and just merged to master.

PR #1238, opened 2017-05-03
Apply pois migration to polygon

rmarianski:


data/migrations/v1.3.0-polygon.sql
```diff
@@ -3,11 +3,48 @@ UPDATE
...
+      OR tags -> 'tags->nwn_ref' = 'true'
+      OR tags -> 'tags->rcn_ref' = 'true'
+      OR tags -> 'tags->rwn_ref' = 'true'
+      OR tags -> 'tags->whitewater' IN ('egress', 'hazard', 'put_in', 'put_in;egress',
+        'rapid')
+      OR tags -> 'tourism' IN ('alpine_hut', 'information', 'picnic_site', 'viewpoint',
+        'wilderness_hut')
+      OR tags -> 'waterway' IN ('dam', 'lock', 'waterfall')
+      OR (NOT tags ? 'name' AND mz_poi_min_zoom IS NOT NULL)
+      OR tags -> 'man_made' = 'pier'
```
nvkelso: I think this technically works though it could get merged with general IN `man_made` above. This is probably more legible as `pier` is special.

rmarianski: Good call, I've updated it.

PR #1239, opened 2017-05-03
Remove extraneous 'tags->' in migration

rmarianski:


data/migrations/v1.3.0-point.sql
```diff
@@ -29,15 +29,15 @@ WHERE
       OR tags -> 'public_transport' IN ('platform', 'stop_area')
       OR tags -> 'railway' IN ('halt', 'level_crossing', 'platform', 'stop',
         'subway_entrance', 'tram_stop')
-      OR tags -> 'tags->icn_ref' = 'true'
```
nvkelso: `= 'true'` should be more like `NOT NULL` or `tags ? icn_ref`

data/migrations/v1.3.0-polygon.sql
```diff
@@ -29,15 +29,15 @@ UPDATE
       OR tags -> 'public_transport' IN ('platform', 'stop_area')
       OR tags -> 'railway' IN ('halt', 'level_crossing', 'platform', 'stop',
         'subway_entrance', 'tram_stop')
-      OR tags -> 'tags->icn_ref' = 'true'
```
nvkelso: same as above...
= 'true' should be more like NOT NULL or tags ? icn_ref

nvkelso:
Approved pending those changes.

data/migrations/v1.3.0-polygon.sql
rmarianski: Updated in b0e4283.

PR #1241, opened 2017-05-04
Return quickly if network is null

iandees:
Fixes problem we're seeing in production where the network comes through as `None`.


iandees:
This was fixed in https://github.com/tilezen/vector-datasource/commit/d44dbf86bec1c5098578d4ca194159bff710f7ed.

PR #1244, opened 2017-05-05
WIP Add option to save test coordinates in a text file

rmarianski:
This adds a command line option to save the list of coordinates used in the tests in a text file. The idea is that we would instruct the CI test run to emit this file, and then update the list of test coordinates in an s3 location on success. We'd only do this when building the master branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ vector_datasource.egg-info/
 data/Makefile-import-data
 data/Makefile-prepare-data
 data/import-shapefiles.sh
+
+test-coords.txt
--- a/integration-test.py
+++ b/integration-test.py
@@ -170,8 +170,14 @@ def tile_url(z, x, y, layer):
     url = config_url % {'layer': request_layer, 'z': z, 'x': x, 'y': y}
     return url
 
+global_test_coords_file_handle = None
 
-def fetch_tile_http(url):
+
+def fetch_tile_http(url, z, x, y):
+    global global_test_coords_file_handle
+    if global_test_coords_file_handle:
+        global_test_coords_file_handle.write('%d/%d/%d\n' % (z, x, y))
+
     r = requests.get(url)
 
     if r.status_code == 504:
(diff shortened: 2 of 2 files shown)


rmarianski:
The whole way I've implemented this is a bit of a hack, but it seems ok to me for what we're trying to accomplish here.

I didn't update the circleci parts yet, and wanted to first get a sanity check that I'm not going down the wrong path here.

zerebubuth:
Good idea about saving off the list of files required for tests. We added the `-printcoords` argument for this a while back, would it be possible to re-use that?

rmarianski:
It unfortunately doesn't capture everything. But now that I do have the whole list, I diff'ed them, and there were fewer discrepancies than I expected. Here are the tiles that `-printcoords` didn't catch.

524-peak-kind-tile-rank
-----------------------
* 12/662/1443
* 16/12372/26269

507-routes-via-stop-positions
-----------------------------
* 13/2352/3122
* 13/2385/3102

502-water-boundaries-slow.py
----------------------------
* 16/23775/33616


Instead of the approach in this pr, should we just run `-printcoords` and manually tack on these tiles?

I took a quick look at the code, and these tests all seem to be calling `features_in_tile_layer`. But, it looks like that function is being stubbed for printed, so I'm not exactly sure why it's not working. My guess is that the test fails because we don't return back any features, and so the custom assertion fails before trying all the other coordinates in the file.
Assuming this is the case, maybe what's worth doing is just instead of manually raising an exception or using assert directly, we have a `fail` function that gets passed in that the tests use. Then this can be stubbed out the same way as the others. There's just a few of these cases, so that wouldn't be much work. What do you think?

zerebubuth:
> Assuming this is the case, maybe what's worth doing is just instead of manually raising an exception or using assert directly, we have a fail function that gets passed in that the tests use. Then this can be stubbed out the same way as the others. There's just a few of these cases, so that wouldn't be much work. What do you think?

Sounds good to me. I guess this is why unit test frameworks tend to have an object (sometimes which they inherit from) on which they call `.assertWhatever()`. Seems like a good idea to have the same.

rmarianski:
Closing in favor of https://github.com/tilezen/vector-datasource/pull/1245

PR #1245, opened 2017-05-09
Capture all test coordinates with -printcoords

rmarianski:
The biggest change here is to namespace all the test assertion functions using `test`. This is to emphasize that all assertions should be done this way.

### DIFF
--- a/integration-test.py
+++ b/integration-test.py
@@ -395,11 +395,22 @@ def print_coord_with_context(z, x, y, *ignored):
     yield []
 
 
+class TestAPI(object):
+
+    def __init__(self, function_table):
+        self.function_table = function_table
+
+    def __getattribute__(self, name):
+        assert name != 'function_table'
+        function_table = object.__getattribute__(self, 'function_table')
+        return function_table[name]
+
+
 class TestRunner(object):
 
     def __init__(self, mode):
         if mode == 'print':
-            self.test_api = {
+            function_table = {
                 'assert_has_feature': print_coord,
                 'assert_no_matching_feature': print_coord,
                 'assert_at_least_n_features': print_coord,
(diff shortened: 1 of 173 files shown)


iandees:
Any reason we don't use the existing `unittest` framework?

rmarianski:
> Any reason we don't use the existing unittest framework?

IIRC the idea was to create higher level tests that were easier to write than traditional unit tests. @zerebubuth can correct me if I'm wrong here.

zerebubuth:
> IIRC the idea was to create higher level tests that were easier to write than traditional unit tests. 

The [original tests](https://github.com/tilezen/vector-datasource/commit/a431ccec498c58d30d88eee1e18ee69b9a7fdd1e) had to be run against a server, and didn't state their dependencies, which meant they needed a lot of setup and were unlike unit tests. I think, when I wrote them, I didn't use `unittest` because I was thinking of them as "release tests" or "acceptance tests". I guess that's why @rob moved them to be `integration-tests` when he wrote some actual unit tests :smile:

At some point, we added the ability for the code to read its own source to set up the database to run against. It's still not very "unit-y", though, as data from independent tests can interfere with each other because we don't set up the database anew for each test with only the dependencies it needs. Also, changes to the upstream data can make the tests fail.

However, we've been moving closer and closer to having self-contained unit tests. And we might finally get there with `@` annotations for the database contents that the test depends on. Those can be loaded up in a per-test setup and reset in a teardown so that the tests can't interfere with each other and the test result is only a function of the code and fixture, not the wider world.

At that point, it might be worth seeing if using `unittest` has any advantages over the way it is at the moment.

iandees:
Sounds good. We might consider switching to use the `unittest` library at some point simply because it provides a useful framework with a lot of pre-built assertions. Tests don't necessarily have to be "capital-U" unit tests to use the `unittest` Python library.

PR #1246, opened 2017-05-11
Update test coordinates on master builds

rmarianski:
This should update the list of test coordinates in s3 on successful master builds.

### DIFF
--- a/scripts/update-integration-test-coordinates.sh
+++ b/scripts/update-integration-test-coordinates.sh
@@ -1,2 +1,5 @@
 #!/bin/bash
-python integration-test.py -printcoords | python scripts/update-integration-test-coordinates.py
+
+basedir="$(dirname ${BASH_SOURCE[0]})/.."
+
+python ${basedir}/integration-test.py -printcoords | python ${basedir}/scripts/update-integration-test-coordinates.py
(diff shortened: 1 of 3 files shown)


scripts/update-integration-test-coordinates.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+python integration-test.py -printcoords | python scripts/update-integration-test-coordinates.py
```
zerebubuth: Is `integration-test.py` in the current working directory? Should we use the script location to make sure that it finds the right file?

rmarianski: It will be based on how it's run from circleci. The root of the repo is the assumed cwd.

What do you suggest instead? In the setup and run script I see this:

    basedir="$(dirname ${BASH_SOURCE[0]})/.."

Should we do the same thing here?

zerebubuth: I think so. That way it doesn't come as a surprise if someone tries to run it out of the `scripts/` dir, or we end up in a `$PWD` that we didn't expect.

zerebubuth: Btw, I don't mean to suggest that the way it is at the moment is wrong. When I read it, I thought "is `integration-test.py` in the `scripts/` directory?" and it took me a moment to figure out what was going on. I think it's better to avoid confusion (perhaps it's only my confusion :confounded: ) by being explicit about the locations of files where we can.

rmarianski: Updated in 0fffea0.

PR #1261, opened 2017-05-31
WIP Move kind calculation to python

rmarianski:
This is a work in progress pr to track the work to move the kind and output property calculation over to python from sql.

- [ ] Finish initial pass to generate python functions for output calculation
- [ ] Create alternatives for sql expressions in yaml
- [ ] Add tests to exercise cases
- [ ] Normalize sql queries by moving custom logic outside
      (it might be possible that we don't need the jinja files at all any more, but perhaps this is only attainable in a future step)
- [ ] Update code to slot python generation as part of the pipeline (either offline or on startup)
- [ ] Get tests to pass with new setup


rmarianski:
Moved to #1272 for new branch.

Issue #1262, opened 2017-05-31
Fix v1.4.0 series test failures (upstream data changes)


nvkelso:
Circle reports back the following errors (there may be more):

```
[ 172/172] PASS: 'integration-test/993-remove-props-road-merge.py'
FAILED 4 TESTS. For more information, see 'test.log'
[  29/172] FAIL: 'integration-test/291-483-suppress-historical-closed.py'
Traceback (most recent call last):
  File "./scripts/../integration-test.py", line 447, in __call__
    runpy.run_path(f, init_globals=dict(test=self.test_api))
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 240, in run_path
    return _run_module_code(code, init_globals, run_name, path_name)
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 82, in _run_module_code
    mod_name, mod_fname, mod_loader, pkg_name)
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "integration-test/291-483-suppress-historical-closed.py", line 12, in <module>
    {'id': 241643507})
  File "./scripts/../integration-test.py", line 354, in assert_no_matching_feature
    "%r" % (properties, layer, feature['properties'])
  File "/opt/circleci/python/2.7.10/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "./scripts/../integration-test.py", line 225, in features_in_tile_layer
    yield features
  File "./scripts/../integration-test.py", line 354, in assert_no_matching_feature
    "%r" % (properties, layer, feature['properties'])
Exception: Tile u'http://localhost:58563/pois/15/5232/12654.json': Found feature matching properties {'id': 241643507} in layer 'pois', but was supposed to find none. For example: {u'kind': u'electronics', u'name': u'Tarrget', u'area': 8984, u'source': u'openstreetmap.org', u'min_zoom': 13.99, u'id': 241643507}


[  62/172] FAIL: 'integration-test/479-barrier-toll_booth.py'
Traceback (most recent call last):
  File "./scripts/../integration-test.py", line 447, in __call__
    runpy.run_path(f, init_globals=dict(test=self.test_api))
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 240, in run_path
    return _run_module_code(code, init_globals, run_name, path_name)
[...]

PR #1265, opened 2017-06-05
remove duplicate symbol output

paulmach:
Symbol was listed twice under the `osm_piste_properties` outputs in yaml/roads.yaml

I left the one at the end since it was more alphabetical like.


nvkelso:
CircleCI is unhappy, but for un-related test issues already covered in https://github.com/tilezen/vector-datasource/issues/1262. This is ready to merge.

PR #1272, opened 2017-06-14
Move kind calculation to python

rmarianski:
This is a work in progress pr to track the work to move the kind and output property calculation over to python from sql.

- [x] Finish initial pass to generate python functions for output calculation
- [x] Create alternatives for sql expressions in yaml
- [x] Add tests to exercise cases
- [x] Normalize sql queries by moving custom logic outside
      (it might be possible that we don't need the jinja files at all any more, but perhaps this is only attainable in a future step)
- [x] Update code to slot python generation as part of the pipeline (either offline or on startup)
- [x] Get tests to pass with new setup


nvkelso:
Please hold merging this PR until https://github.com/tilezen/vector-datasource/pull/1288 lands. It'll make release management easier :)

nvkelso:
#1288 has been merged and new release tagged. Free to merge this PR.

PR #1278, opened 2017-06-22
Translate new min zoom YAML to SQL

zerebubuth:
This updates the `sql.py` which was previously being used to generate the SQL functions to handle the new structures in YAML that we added to support the Python output. This means that the repo should be self-contained again, and be able to run the tests without having to import the server-side function definitions from `master` branch.

Not to be confused with `sql2.py`, which is the new Python `ast`-based SQL output. That's a separate thing which hopefully will be coming soon.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -933,3 +933,10 @@ BEGIN
   RETURN mz_building_height(tags->'height', tags->'building:levels') * way_area;
 END
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION trim_nz_sh(label TEXT)
+RETURNS TEXT AS $$
+BEGIN
+  RETURN trim(leading 'SH' from label);
+END
+$$ LANGUAGE plpgsql IMMUTABLE;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -927,6 +927,8 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql STABLE;
 
+-- calculates the building height given the footprint area and the tags, which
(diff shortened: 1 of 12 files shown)


zerebubuth:
> Your tests passed on CircleCI!

Yay, finally! :relieved:

data/functions.sql
```diff
@@ -933,3 +933,10 @@ BEGIN
   RETURN mz_building_height(tags->'height', tags->'building:levels') * way_area;
 END
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION trim_nz_sh(label TEXT)
```
nvkelso: Add comment here explaining that this trims the "SH" from New Zealand State Highway road shield refs.

zerebubuth: Yup, done in 8e564c4.

scripts/setup_and_run_tests.sh
```diff
@@ -134,21 +134,28 @@ for tbl in `ls ${basedir}/integration-test/fixtures/`; do
...
-fi
+    # note: don't use postfix ++ as let returns 1 when the arg evaluates to
+    # zero (as (counter=0)++ would), which makes `set -e` terminate the
+    # program.
+    let counter=counter+1
+    if [[ $counter -gt $limit ]]; then
+        echo "Test server didn't start up within ${limit}s."
+        exit 1
+    fi
+done
```
rmarianski: :+1: 

I've had errors locally where it didn't start up in a second.

nvkelso:
@rmarianski This is ready for you're review and merge into your branch :)

Issue #1285, opened 2017-06-26
More test failures: 546-road-sort-keys-roads.py (and friends?)


nvkelso:
Latest CircleCI build failed:

- https://circleci.com/gh/tilezen/vector-datasource/1659

```
[ 172/172] PASS: 'integration-test/993-remove-props-road-merge.py'
FAILED 1 TESTS. For more information, see 'test.log'
[  81/172] FAIL: 'integration-test/546-road-sort-keys-roads.py'
Traceback (most recent call last):
  File "./scripts/../integration-test.py", line 447, in __call__
    runpy.run_path(f, init_globals=dict(test=self.test_api))
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 240, in run_path
    return _run_module_code(code, init_globals, run_name, path_name)
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 82, in _run_module_code
    mod_name, mod_fname, mod_loader, pkg_name)
  File "/opt/circleci/python/2.7.10/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "integration-test/546-road-sort-keys-roads.py", line 36, in <module>
    "name": "Junipero Serra Blvd.", "is_link": True, "sort_rank": 374})
  File "./scripts/../integration-test.py", line 304, in assert_has_feature
    % (properties, closest['properties'], misses)
  File "/opt/circleci/python/2.7.10/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "./scripts/../integration-test.py", line 225, in features_in_tile_layer
    yield features
  File "./scripts/../integration-test.py", line 304, in assert_has_feature
    % (properties, closest['properties'], misses)
[...]

PR #1286, opened 2017-06-27
Update tests to latest data

zerebubuth:
* WOF neighbourhood point for Hollywood moved, making tests fail against live systems (dev, prod). The test was updated, and the fixtures were updated to the most recent version of the data.
* Change the forest used in one test. The previous one had been included in a national park relation and so was "superseded" and no longer generated a forest polygon.
* Update `motorway_link` element to replace previous example, which had changed upstream.


zerebubuth:
Fixes #1285.

PR #1290, opened 2017-06-29
Normalize sql queries

rmarianski:
This doesn't address the problem with handling table sources between the sql and python paths, so the sql calculation is likely to be broken in this pr.

### DIFF
--- a/queries/boundaries.jinja2
+++ b/queries/boundaries.jinja2
@@ -1,13 +1,14 @@
 {% macro ne_boundary_cols(name) %}
-    gid AS __id__,
-    'naturalearthdata.com' AS source,
-    {{ name }} AS name,
-    mz_boundary_min_zoom as min_zoom,
-    NULL AS maritime_boundary,
-    {% filter geometry %}the_geom{% endfilter %} AS __geometry__,
-    mz_calculate_json_boundaries(t.*) AS mz_properties,
-    hstore('source', 'naturalearthdata.com') AS mz_additional_tags,
-    row_to_json(t.*) AS tags
+   gid AS __id__,
+   {% filter geometry %}the_geom{% endfilter %} AS __geometry__,
+   mz_calculate_json_boundaries(t.*) AS mz_properties,
+   jsonb_build_object(
+     'source', 'naturalearthdata.com',
+     'name', {{ name }},
+     'min_zoom', mz_boundary_min_zoom,
+     'featurecla', featurecla,
+     'scalerank', scalerank
+    ) AS __properties__
 {% endmacro %}
 
 {% if zoom < 2 %}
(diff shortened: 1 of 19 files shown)


queries/boundaries.jinja2
```diff
@@ -101,16 +100,14 @@ UNION ALL
...
-  hstore(ARRAY[
-           'source', 'openstreetmap.org',
-           'kind', 'maritime']) AS mz_additional_tags,
-  NULL::json AS tags
+  json_build_object('kind', 'maritime') as mz_properties,
+  jsonb_build_object(
+    'source', 'openstreetmap.org',
+    'min_zoom', 8,
+    'maritime_boundary', true,
+    'kind', 'maritime'
```
zerebubuth: Does `kind` need to be in both `__properties__` and `mz_properties`? Could we move `maritime_boundary` into `mz_properties`?

rmarianski: I only left `mz_properties` there for testing purposes to verify that the output was the same as from the database. I'll remove it from the queries when we're ready to merge.

zerebubuth: Ah, cool. Sounds good :+1:

yaml/pois.yaml
```diff
@@ -3,6 +3,25 @@
...
+    area: {col: area}
+    protect_class: {col: protect_class}
+    operator: {col: operator}
+    religion: {col: religion}
+    sport: {col: sport}
+    mz_transit_score: {col: mz_transit_score}
+    mz_transit_root_relation_id: {col: mz_transit_root_relation_id}
+    train_routes: {col: train_routes}
+    subway_routes: {col: subway_routes}
+    light_rail_routes: {col: light_rail_routes}
```
nvkelso: Missing:

`tram_routes: {col: tram_routes}`

As tye JINJA has:

`'tram_routes', (transit_routes).tram_routes`

queries/places.jinja2
```diff
@@ -65,23 +54,20 @@ WHERE
...
+    {% filter geometry %}label_position{% endfilter %} AS __geometry__,
+    json_build_object(
+      'kind', wof_np.placetype_string
+    ) AS mz_properties,
+    to_jsonb(l10n_name) || jsonb_build_object(
+      'name', name,
+      'source', 'whosonfirst.mapzen.com',
+      'mz_n_photos', n_photos,
+      'area', area::bigint,
+      'max_zoom', max_zoom,
```
nvkelso: Missing:

`'min_zoom', min_zoom,`

yaml/roads.yaml
```diff
@@ -1,5 +1,11 @@
 global:
+  - &output_properties
+    source: {col: source}
```
nvkelso: Surprised `name: { col: name }` isn't here in the global section.

nvkelso: Is that due to the synthetic_columns, is that necessary anymore? 

```
synthetic_columns:
   - name
```

yaml/pois.yaml
rmarianski: Good catch! Fixed in 95d993c.

queries/places.jinja2
rmarianski: Good catch! Fixed in 710e0bd.

yaml/roads.yaml
rmarianski: Do we want pistes to have names? It looks like now they don't. They have `piste_name` and `route_name`.

nvkelso: Ah, right. Maybe add a comment up in the global section noting that? Something like

```
    # not all features (e.g. pistes) include names so names are added below
```

rmarianski: Good call. Added in dd9c991.

PR #1291, opened 2017-06-29
Use https for urls by default, where they are supported

louh:


nvkelso:
WFM, links checkout. Thanks!

This is no longer published to mapzen.com/documentation (instead it links here to master) so merging but not tagging.

PR #1293, opened 2017-06-30
Fix SQL and unit tests

zerebubuth:
* Fixed unit tests to make use of new `meta` parameter.
* Made `perform-sql-updates.sh` stop on error so that issues with setting up the database are easier to debug.
* Added `meta.source` to SQL to match Python implementation.
* Hack to work around vandalism on the OSM feature we use to test theme parks.

### DIFF
--- a/data/perform-sql-updates.sh
+++ b/data/perform-sql-updates.sh
@@ -2,37 +2,51 @@
 
 set -e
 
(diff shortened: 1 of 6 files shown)


rmarianski:
I went ahead and merged this in.

PR #1323, opened 2017-07-20
[docs] CONTRIBUTE file - reduce confusion

matkoniecz:
unify terminology - three lines earlier "database migration" term was used.

"yaml updates don't require restart (but do require a database migration, see below)"

- [x] Update docs


matkoniecz:
I am pretty sure that this change is correct. But there is also chance that docs are more confusing than expected and my change is misleading.

matkoniecz:
Also failed on 429 response from overpass that is unrelated to this change.

PR #1322, opened 2017-07-20
include testing-related debris in .gitignore

matkoniecz:
I was following instructions in https://github.com/tilezen/vector-datasource/blob/master/TESTS.md

During tests git status was displaying some unwanted debris that I think should be in .gitignore.

PR #1324, opened 2017-07-20
introduce waiting and query repeat on 429 response

matkoniecz:
as result overloading overpass will not result in failure of tests
make testing easier when public instance of overpass is used

- [x] Update tests

Should reduce random test failures like one that happened for #1322 PR (.gitignore was changed, tests failed due to 429 and as result PR is misleadingly marked as breaking build).

### DIFF
--- a/integration-test.py
+++ b/integration-test.py
@@ -12,6 +12,7 @@ from contextlib import contextmanager
 import shapely.geometry
 import re
 import lxml.etree as ET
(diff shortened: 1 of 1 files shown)


matkoniecz:
Note that for this PR ci/circleci check passed rather than failing due to 429 error code from overpass - unlike #1323 and #1322

Issue #1327, opened 2017-07-20
[docs] update-the-database-properties - at least create-sql-functions.py was moved


matkoniecz:
https://github.com/tilezen/vector-datasource/blob/master/CONTRIBUTING.md#update-the-database-properties fails with

```
python: can't open file 'create-sql-functions.py': [Errno 2] No such file or directory
psql: FATAL:  database "osm" does not exist
```
For start, create-sql-functions.py was moved to vectordatasource/meta/sql.py in https://github.com/tilezen/vector-datasource/commit/7a1b6d98e16b51386eb0180a164c1a19858d8781#diff-4e2091df1ddaa09aa56175a13674ed7f

I tracked down file movement - but I am not sure whatever something else described in https://github.com/tilezen/vector-datasource/blob/master/CONTRIBUTING.md#update-the-database-properties also changed.

nvkelso:
@rmarianski I think you moved this around in master (while the docs are for the last good release tag). Can you update the docs for master, please?

rmarianski:
Whoops, yea. It looks like it should just be that line though. I've updated it in a [pr](https://github.com/tilezen/vector-datasource/pull/1329).

PR #1328, opened 2017-07-20
add access tag to roads, fixes #1273

### LINKED ISSUE #1273: include access in road layer
* **What did you see?**
https://mapzen.com/documentation/vector-tiles/layers/#roads-transportation is not mentioning access field for roads, for access=private roads inspect in https://mapzen.com/tangram/play/?#17.75000/50.07917/19.89074 is not listing any value that indicates it. Searching in https://mapzen.com/documentation/vector-tiles/layers/#roads-transportation for private, destination and similar reveals nothing.

* **What did you expect to see?**
Access, fundamental values for roads is an exposed attribute.

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

For example at https://www.openstreetmap.org/#map=19/50.07918/19.91407

matkoniecz:
- [x] Update tests
- [x] Update data migrations (not necessary??)
- [x] Update docs

data migrations - at this moment I am not entirely sure what and how needs to be done there (see #1327)

I am not sure what happened in line 128 (Firefox, SublimeText claim that there is no difference, git detects a difference) - I may investigate if this whitespace change is considered unwanted.

### DIFF
--- a/integration-test/1273-roads-access.py
+++ b/integration-test/1273-roads-access.py
@@ -3,7 +3,7 @@
 # restricted access road in military base, Kraków, Poland
 # https://www.openstreetmap.org/way/43322699
 test.assert_has_feature(
-    18, 145574, 88813, 'roads',
+    16, 36393, 22203, 'roads',
     { 'id': 43322699, 'kind': 'minor_road', 'kind_detail': 'service', 'access': 'private'})
 
 test.assert_has_feature(
@@ -20,5 +20,5 @@ test.assert_has_feature(
 # cycleway in Gdańsk, Poland
 # http://www.openstreetmap.org/way/151351130#map=19/54.32715/18.61035
 test.assert_has_feature(
-    19, 289247, 167525, 'roads',
+    16, 36155, 20940, 'roads',
     { 'id': 151351130, 'kind': 'path', 'access': 'yes'})
--- a/integration-test/1273-roads-access.py
+++ b/integration-test/1273-roads-access.py
@@ -1,0 +1,24 @@
+# Add surface properties to roads layer (at max zooms)
+
+# restricted access road in military base, Kraków, Poland
+# https://www.openstreetmap.org/way/43322699
+test.assert_has_feature(
+    18, 145574, 88813, 'roads',
+    { 'id': 43322699, 'kind': 'minor_road', 'kind_detail': 'service', 'access': 'private'})
+
+test.assert_has_feature(
+    14, 9098, 5550, 'roads',
+    { 'id': 43322699, 'kind': 'minor_road', 'kind_detail': 'service', 'access': 'private'})
+
+# motorway bridge in Honk-Kong
+# http://www.openstreetmap.org/way/276506948#map=18/22.48811/113.94748
+# ID may get dropped due to a merge with the other carriageway
+test.assert_has_feature(
+    12, 3344, 1785, 'roads',
+    { 'kind': 'highway', 'kind_detail': 'motorway', 'alt_name:en': 'Shenzhen Bay Bridge', 'access': 'no'})
+
+# cycleway in Gdańsk, Poland
+# http://www.openstreetmap.org/way/151351130#map=19/54.32715/18.61035
+test.assert_has_feature(
+    19, 289247, 167525, 'roads',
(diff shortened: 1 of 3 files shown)


matkoniecz:
Overall, I am not sure but going by analogous https://github.com/tilezen/vector-datasource/commit/2c3edbbaed6aaeaf4d0306f75d1609b239af27e0 I think that this PR is finished

-----------------

> yaml updates don't require restart (but do require a database migration, see below).

so data migrations update is necessary


> Once you make your edits to the YAML file you need to update the database. To recreate SQL functions in Postgres run

this one is not necessary for me as I have no local database. As I understand there is no need to include code for that as it will be done on database update as new version will be released.


> Because some properties (like min_zoom) in the database are pre-computed, we need to update records to recalculate that using the updated functions.

OK, so I need to check whatever road properties are precomputed.


>  5. Verify the new logic by running the test (...) Run the test, hopefully it passes now!

Given that tests are passing - it seems to work without data migration. But maybe because I load database from scratch and database migration is necessary to update existing databases as "migration" indicates?

quotes are from CONTRIBUTING.md

integration-test/1273-roads-access.py
```diff
@@ -0,0 +1,24 @@
+# Add surface properties to roads layer (at max zooms)
+
+# restricted access road in military base, Kraków, Poland
+# https://www.openstreetmap.org/way/43322699
+test.assert_has_feature(
+    18, 145574, 88813, 'roads',
```
nvkelso: Can you set this coordinate to zoom 16 instead, please?

integration-test/1273-roads-access.py
```diff
@@ -0,0 +1,24 @@
...
+# http://www.openstreetmap.org/way/276506948#map=18/22.48811/113.94748
+# ID may get dropped due to a merge with the other carriageway
+test.assert_has_feature(
+    12, 3344, 1785, 'roads',
+    { 'kind': 'highway', 'kind_detail': 'motorway', 'alt_name:en': 'Shenzhen Bay Bridge', 'access': 'no'})
+
+# cycleway in Gdańsk, Poland
+# http://www.openstreetmap.org/way/151351130#map=19/54.32715/18.61035
+test.assert_has_feature(
+    19, 289247, 167525, 'roads',
```
nvkelso: Can you set this coordinate to zoom 16 instead, please? We generally only do the default `max_zoom` for the service.

matkoniecz:
Access tag across all zoom levels was intentional to allow hiding objects - for example it is unwanted to hide road at higher zoom as it is access=no and show it again on zooming out as access data is lost.

It can be done by dropping such ways and making them completely unavailable, but I will expect that on adding such feature tests would be changes.

Lack of transform also was intentional to keep it the same as bicycle tag (and its documentation indicates lack of transformation, from what I have found in code it seems to be true).

integration-test/1273-roads-access.py
matkoniecz: As I understand it will zoom out - is there any reason to expect that data may be dropped on zooming in?

matkoniecz: Done.

integration-test/1273-roads-access.py
matkoniecz: Done.

nvkelso:
> Access tag across all zoom levels was intentional to allow hiding objects - for example it is unwanted to hide road at higher zoom as it is access=no and show it again on zooming out as access data is lost.
>
> It can be done by dropping such ways and making them completely unavailable, but I will expect that on adding such feature tests would be changes.

Looks like our existing `motor_vehicle` property provides similar functionality and we don't drop that property at earlier zooms so not dropping `access` would be consistent. 

I'm still concerned that we shouldn't be exporting `access=yes` tags at lower zooms as it unnecessarily prevents way merging but I'll file a new issue to think track that (since it's already a problem with `motor_vehicle` and maybe other access related tags).

nvkelso:
@matkoniecz Thank you for your contribution to Tilezen! 😀

nvkelso:
Filed #1331 for followup to consider dropping `access=yes` and related tags.

PR #1334, opened 2017-07-25
include surface at mid-zooms

### LINKED ISSUE #1252: surface property missing at mid-zooms
Looks like surface is in the drop properties list twice (oops):

- zoom 11: https://github.com/tilezen/vector-datasource/blob/e324c7d3406895c065aa40bd5de5d66be491d720/queries.yaml#L517
- zoom 14: https://github.com/tilezen/vector-datasource/blob/e324c7d3406895c065aa40bd5de5d66be491d720/queries.yaml#L557

The second instance in zoom 14 is likely an error.

By the time unclassified and some tracks are visible at zoom 11 in https://github.com/tilezen/vector-datasource/issues/1250 and https://github.com/tilezen/vector-datasource/issues/1251 the surface tag needs to be included.

matkoniecz:
enables surface tag on z8 to z14 (as bicycle routes appear on z8)
fixes #1252

- [x] Update tests
- [x] ~~Update data migrations~~ (not needed as doesn't impact `min_zoom`)
- [x] Update docs

WIP, I created PR to run tests independently from my local instance (after reload of testing database test failed with claim that 11/1091/673 has no road layer (and that tile at http://tile.mapzen.com/mapzen/vector/v1/roads/11/1091/673.json has road layer).

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1063,7 +1063,7 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `all_bus_networks` and `all_bus_shield_texts`: All of the bus and trolley-bus routes of which this road is a part, and each corresponding shield text. See `bus_network` and `bus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
 * `bus_network`: Note that this is often not present for bus routes / networks. This may be replaced with `operator` in the future, see [issue 1194](https://github.com/tilezen/vector-datasource/issues/1194).
 * `bus_shield_text`: Contains text intended to be displayed on a shield related to the bus or trolley-bus network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
-* `surface`: Common values include `asphalt`, `unpaved`, `paved`, `ground`, `gravel`, `dirt`, `concrete`, `grass`, `paving_stones`, `compacted`, `sand`, and `cobblestone`.
+* `surface`: Common values include `asphalt`, `unpaved`, `paved`, `ground`, `gravel`, `dirt`, `concrete`, `grass`, `paving_stones`, `compacted`, `sand`, and `cobblestone`. `concrete:plates` and `concrete:lanes` values are transformed to `concrete_plates` and `concrete_lanes` respectively.
 
 #### Road properties (optional):
 
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1016,7 +1016,7 @@ Road names are **abbreviated** so directionals like `North` is replaced with `N`
 
 Mapzen calculates the `landuse_kind` value by intercutting `roads` with the `landuse` layer to determine if a road segment is over a parks, hospitals, universities or other landuse features. Use this property to modify the visual appearance of roads over these features. For instance, light grey minor roads look great in general, but aren't legible over most landuse colors unless they are darkened.
 
-To improve performance, some road segments are merged at low and mid-zooms. To facilitate this, certain properties are dropped at those zooms. Examples include `is_bridge` and `is_tunnel`, `name`, `network`, `oneway` and `ref`. The exact zoom varies per feature class (major roads keep this properties over a wider range, minor roads drop them starting at zoom 14). When roads are merged, the original OSM `id` values are dropped.
+To improve performance, some road segments are merged at low and mid-zooms. To facilitate this, certain properties are dropped at those zooms. Examples include `is_bridge` and `is_tunnel`, `name`, `network`, `oneway`, and `ref`. The exact zoom varies per feature class (major roads keep this properties over a wider range, minor roads drop them starting at zoom 14). When roads are merged, the original OSM `id` values are dropped.
 
 #### Road properties (common):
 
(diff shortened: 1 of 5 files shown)


queries.yaml
```diff
@@ -502,6 +502,13 @@ post_process:
     params:
       source_layer: roads
       start_zoom: 0
+      end_zoom: 7
```
nvkelso: This section isn't needed because we don't source from OpenStreetMap at the low zooms.

nvkelso: (You'd just leave out `surface` from the list of **drop_properties**.)

integration-test/1252-roads-surface.py
```diff
@@ -0,0 +1,52 @@
...
+    13, 4364, 2695, 'roads',
+    { 'id': 58691615, 'kind_detail': 'track', 'surface': 'concrete:lanes'})
+
+test.assert_has_feature(
+    11, 1091, 673, 'roads',
+    { 'id': 58691615, 'kind_detail': 'track', 'surface': 'concrete:lanes'})
+
+test.assert_has_feature(
+    10, 545, 336, 'roads',
+    { 'id': 58691615, 'kind_detail': 'track', 'surface': 'concrete:lanes'})
```
nvkelso: Strangely enough **concrete:lanes** is a valid `surface` value in OSM:

- http://wiki.openstreetmap.org/wiki/Key:surface

But in Tilezen we generally avoid `:` and `-` in property names or values in favor of `_`.

That's controlled in the YAML config. So update where `surface` is added to include a mapping:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L32

More like how we trap aerialway values to prevent `j-bar` from being exported in with the case statement:

https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L633

```
  - filter:
      aerialway:
        - drag_lift
        - platter
        - t-bar
        - goods
        - magic_carpet
        - rope_tow
        - "yes"
        - zip_line
        - j-bar
        - unknown
        - mixed_lift
        - canopy
        - cableway
    min_zoom: 15
    table: osm
    output:
      <<: *osm_highway_properties
      <<: *osm_network
      kind: aerialway
      kind_detail:
        case:
          - { when: { aerialway: 't-bar' }, then: 't_bar' }
          - { when: { aerialway: 'j-bar' }, then: 'j_bar' }
          - { when: { aerialway: 'yes' }, then: null }
          - else: { col: 'aerialway' }
```

nvkelso: We'll want to do this for a few values, TagInfo has the full details:

- https://taginfo.openstreetmap.org/keys/?key=surface#values

matkoniecz: "But in Tilezen we generally avoid : and - in property names or values in favor of _." 

There is also concrete:plates (other are not worth handling).

I will add it to the transform and documentation of surface value.

queries.yaml
matkoniecz: done

integration-test/1252-roads-surface.py
matkoniecz: done

matkoniecz:
I added transformation (on request). Now there are rare/malformed values that have untransformed colon.

Surface values except quite short list are malformed/really rare.

Maybe it would be a good idea to add also filtering and exclude very rare/malformed values?

I would probably reuse surface list from https://github.com/gravitystorm/openstreetmap-carto/blob/2be3a11ac240c82fb81200fbcd8be3aef16a187a/project.mml#L466

rmarianski:
> Maybe it would be a good idea to add also filtering and exclude very rare/malformed values?
> 
> I would probably reuse surface list from https://github.com/gravitystorm/openstreetmap-carto/blob/2be3a11ac240c82fb81200fbcd8be3aef16a187a/project.mml#L466

:+1: to whitelists where feasible

nvkelso:
Filed https://github.com/tilezen/vector-datasource/issues/1335 to follow-up about `is_paved` boolean, which I think should also take care of any more white listing / value transforms.

docs/layers.md
```diff
@@ -1016,7 +1016,7 @@ Road names are **abbreviated** so directionals like `North` is replaced with `N`
...
 
-To improve performance, some road segments are merged at low and mid-zooms. To facilitate this, certain properties are dropped at those zooms. Examples include `is_bridge` and `is_tunnel`, `name`, `network`, `oneway`, `ref`, and `surface`. The exact zoom varies per feature class (major roads keep this properties over a wider range, minor roads drop them starting at zoom 14). When roads are merged, the original OSM `id` values are dropped.
+To improve performance, some road segments are merged at low and mid-zooms. To facilitate this, certain properties are dropped at those zooms. Examples include `is_bridge` and `is_tunnel`, `name`, `network`, `oneway` and `ref`. The exact zoom varies per feature class (major roads keep this properties over a wider range, minor roads drop them starting at zoom 14). When roads are merged, the original OSM `id` values are dropped.
```
nvkelso: Our practice is to use comma before the and in a list:

Examples include `is_bridge` and `is_tunnel`, `name`, `network`, `oneway`, and `ref`.

nvkelso:
Great contribution, thank you!

PR #1336, opened 2017-07-26
add documentation, handle 504

matkoniecz:
- [x] Update tests
- [x] Update data migrations (not relevant)
- [x] Update docs (in comments of code)

documents 429 handling, add handling to 504.

504 also may happen (as overpass is overloaded) and also typically is solvable by retrying.

Today it was happening with some (low) frequency and without that change test relying on public overpass were sometimes failing.

### DIFF
--- a/integration-test.py
+++ b/integration-test.py
@@ -503,9 +503,17 @@ class OsmChange(object):
                          params=dict(data=query))
             if r.status_code == 200:
                 break
-            if r.status_code != 429:
+            if r.status_code != 429 and r.status_code != 504:
+                # "429 Too Many Requests is sent if you pass multiple queries from one IP"
+                # regularly happens with multiple sequential querries
+                # "504 Gateway Timeout is sent if the server has already so much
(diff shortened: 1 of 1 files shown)


integration-test.py
```diff
@@ -503,9 +503,17 @@ def query_result(self, query):
                          params=dict(data=query))
             if r.status_code == 200:
                 break
-            if r.status_code != 429:
+            if r.status_code != 429 and r.status_code != 504:
```
rmarianski: Do you think it's clearer spelling it like:

```python
if r.status_code == 200 or r.status_code not in (429, 504):
    break
```

matkoniecz: I like `r.status_code not in (429, 504)`  - much better than testing 429, 504 separately.

matkoniecz: https://github.com/tilezen/vector-datasource/pull/1336/commits/210227d40bcb6f5088163b9d68c89ebb66dd2246 is inspired by this comment. I kept 200 and 429/504 separately.

Issue #1339, opened 2017-07-26
failed CircleCI tests may be displayed as passing


matkoniecz:
* **What did you see?**

Failing CircleCI tests are listed on Github as passing

* **What did you expect to see?**

Failing CircleCI tests are listed on Github as failing

I looked at [log](https://circleci.com/gh/tilezen/vector-datasource/1788?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) of #1338 and I see

> Exception: Tile u'http://localhost:46680/roads/11/1091/673.json': Did not find feature including properties {'kind_detail': 'track', 'id': 58691615, 'surface': 'concrete_lanes'} (because layer 'roads' was empty)

So, why github marked tests as passing? It seems that integration with CircleCI is not working properly.

matkoniecz:
everything indicates that #1341 fixed this

PR #1341, opened 2017-07-26
Return exit code of test run

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1339

This should make the circleci run fail when the tests fail.


nvkelso:
It is now correctly failing, so CircleCI actually approves of this PR ;)

PR #1343, opened 2017-07-27
normalize filter output to maps

paulmach:
There have been some changes to the yaml file structure introduced in https://github.com/tilezen/vector-datasource/commit/412cb6e0e1d723aa163d1d976c6b1397f49b2629 I think. The inconsistency in the way filter outputs are defined is causing me problems when unmarshaling the yaml.

In some cases (the new updated ones) the "filter output" is a two element array with the first element being a map of the shared `output_properties` and the second being a map of the properties for that filter. For example:
```
output:
  - *output_properties
  - kind: region
    kind_detail: state
```

I'm proposing the filter output should be defined as a map in all cases because that is what the ultimate end results is, a map of properties.  I made the updates so the above is written as:
```
output:
  <<: *output_properties
  kind: region
  kind_detail: state
```

This is how it was before the above mentioned change and there are examples of it being a map (written in the second way) already in the yaml files.

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -192,8 +192,8 @@ filters:
           then: 14
         - else: 15
     output:
-      - *output_properties
-      - {kind: windmill}
+      <<: *output_properties
+      kind: windmill
   - filter: {man_made: lighthouse}
     min_zoom:
       case:
@@ -201,13 +201,13 @@ filters:
           then: 14
         - else: 15
     output:
-      - *output_properties
-      - {kind: lighthouse}
+      <<: *output_properties
+      kind: lighthouse
   - filter: {man_made: observatory}
     min_zoom: 15
     output:
-      - *output_properties
-      - {kind: observatory}
+      <<: *output_properties
+      kind: observatory
 
   ############################################################
   # TIER 2 OVERRIDES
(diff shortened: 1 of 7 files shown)


matkoniecz:
```======================================================================
FAIL: test_ne (test.test_meta.BoundariesTest)
  File "/home/ubuntu/vector-datasource/test/test_meta.py", line 148, in test_ne
    self.assertEquals('macroregion', out_props.get('kind'))
AssertionError: 'macroregion' != 'macroregion,'
```

yaml/pois.yaml
```diff
@@ -926,45 +952,46 @@ filters:
           - [ 15,   1 ]
         default: { clamp: { min: 12, max: 14, value: { sum: [ { col: zoom }, 1.066 ] } } }
     output:
-      - *output_properties
-      - {kind: waterfall, height: {col: height}}
+      <<: *output_properties
+      kind: waterfall,
```
zerebubuth: Similar to @matkoniecz's [comment](https://github.com/tilezen/vector-datasource/pull/1343#issuecomment-318299590), there's a trailing comma here.

paulmach:
Fixed the trailing commas and did a search to make sure they were all gone and now tests pass.

Ultimately I can code around this issue but it would good to be consistent.

nvkelso:
Circle _says_ it's happy, but it's not really per discussion and fix in https://github.com/tilezen/vector-datasource/pull/1341 and related data drift PR that's been merged. 

@paulmach Please merge this branch against master to pick up those changes then I think this PR is ready for final test and approval.

paulmach:
I rebased and tests still look okay

nvkelso:
CircleCI now _really_ says all tests pass. 🎉 

```
[ 176/176] PASS: 'integration-test/993-remove-props-road-merge.py'
PASSED ALL TESTS.
```

zerebubuth:
Thanks!

PR #1352, opened 2017-07-28
WIP Make queries per-table

rmarianski:
Connects to https://github.com/mapzen/tile-tasks/issues/267


integration-test/875-camp-grounds-zoom.py
```diff
@@ -44,6 +44,7 @@
 
 # small campground in point zoom 13
 # http://www.openstreetmap.org/way/417405356
+# min zoom matches in landuse layer, but not in pois for 13
```
nvkelso: Is there anything left for us here? If not, take out the comment to reduce confusion.

queries.yaml
```diff
@@ -104,7 +118,6 @@ layers:
     area-inclusion-threshold: 1
   buildings:
     template: buildings.jinja2
-    start_zoom: 13
```
nvkelso: Where is this handled going forward, or is it only on a per-feature `min_zoom` basis now?

queries.yaml
```diff
@@ -125,7 +138,6 @@ layers:
     area-inclusion-threshold: 1
   pois:
     template: pois.jinja2
-    start_zoom: 4
```
nvkelso: This will probably yield a few more park labels at zooms 2 and 3.

queries.yaml
```diff
@@ -60,7 +76,6 @@ layers:
     area-inclusion-threshold: 1
   landuse:
     template: landuse.jinja2
-    start_zoom: 4
```
nvkelso: This will probably yield a few more park polygons at zooms 2 and 3.

queries.yaml
```diff
@@ -77,7 +92,6 @@ layers:
     area-inclusion-threshold: 1
   roads:
     template: roads.jinja2
-    start_zoom: 5
```
nvkelso: Looks like we source the Natural Earth scale_rank for min_zoom directly, and that starts at 3:

```
    min_zoom: { col: scalerank }
```

So this change would introduce roads at zoom 3 unless paired with a clamp over in the roads layer YAML?

nvkelso: (I'd link but Github is mostly down for me, except these PR comments!)

queries.yaml
```diff
@@ -166,7 +177,6 @@ layers:
     area-inclusion-threshold: 1
   transit:
     template: transit.jinja2
-    start_zoom: 5
```
nvkelso: The smallest feature min_zoom is already 5 so relying on feature min_zoom alone (not guarding here in the layer config) should be fine.

queries.yaml
nvkelso: (Which is fine)

integration-test/875-camp-grounds-zoom.py
rmarianski: Removed in f703a2c.

queries.yaml
rmarianski: In this particular case, per-feature `min_zoom`. The `source`s themselves now have a start zoom for when querying should begin, but there are now cases like this one where we are relying more on the min zoom for filtering.

queries.yaml
rmarianski: If that's the case, I think we should update the yaml to clamp these values. Should we do it in this pr? Or a subsequent one?

queries.yaml
rmarianski: Ok, ditto on the clamp in yaml.

queries.yaml
rmarianski: Ok, I'd like to pair with a clamp over in roads yaml.

queries/planet_osm_line.jinja2
```diff
@@ -0,0 +1,89 @@
...
+  {{ bounds['line']|bbox_filter('way', 3857) }} AND
+
+{% if zoom >= 16 %}
+  (mz_earth_min_zoom IS NOT NULL OR
+   mz_landuse_min_zoom IS NOT NULL OR
+   mz_road_level IS NOT NULL OR
+   mz_transit_level IS NOT NULL OR
+   mz_water_min_zoom IS NOT NULL
+   )
+{% else %}
```
nvkelso: Suspect we need to have the old layer start_zoom at 8 logic near here now?

queries/planet_osm_polygon.jinja2
```diff
@@ -0,0 +1,146 @@
...
+    (mz_boundary_min_zoom IS NOT NULL AND {{ bounds['line']|bbox_overlaps('way', 3857) }}) OR
+    ({{ bounds['polygon']|bbox_filter('way', 3857) }} AND
+      (mz_building_min_zoom IS NOT NULL OR
+       mz_earth_min_zoom IS NOT NULL OR
+       mz_landuse_min_zoom IS NOT NULL OR
+       mz_poi_min_zoom IS NOT NULL OR
+       mz_transit_level IS NOT NULL OR
+       mz_water_min_zoom IS NOT NULL))
+{% else %}
+    (mz_boundary_min_zoom < {{ zoom + 1 }} AND {{ bounds['line']|bbox_overlaps('way', 3857) }}) OR
```
nvkelso: Suspect we need to have the old **earth**, **road**, **water** layer start_zoom at 8 logic near here now? (Road is probably overkill.)

(Landuse and POI was start 4 but it's fine at 0, should fix in layer min_zoom YAML if an issue.)

(Building and transit is already handled i the layer min_zoom YAML.)

queries/planet_osm_point.jinja2
```diff
@@ -0,0 +1,97 @@
...
+    {{ bounds['point']|bbox_filter('way', 3857) }} AND
+
+{% if zoom >= 16 %}
+    (mz_building_min_zoom IS NOT NULL OR
+     mz_earth_min_zoom IS NOT NULL OR
+     mz_places_min_zoom IS NOT NULL OR
+     mz_poi_min_zoom IS NOT NULL OR
+     mz_water_min_zoom IS NOT NULL
+     )
+{% else %}
```
nvkelso: Suspect we need to have the old **earth**, **road**, **water** layer start_zoom at 8 logic near here now?

(Building, poi is already handled in the layer min_zoom YAML.)

(Places is mostly handled in the layer min_zoom YAML, and you've carried over the layer duplicate weeding logic into the new NE table logic so we're good there.)

queries.yaml
nvkelso: It's always been an undocumented bug/feature that they don't show up earlier. If it turns out to be not ideal we can file a followup issue for it.

rmarianski:
Going to merge this in for now. Let's make new issues for the follow ups.

queries.yaml
nvkelso: Let's followup in https://github.com/tilezen/vector-datasource/issues/1353 after this PR is merged.

queries/planet_osm_point.jinja2
nvkelso: Followup via https://github.com/tilezen/vector-datasource/issues/1354.

queries/planet_osm_polygon.jinja2
nvkelso: Followup via https://github.com/tilezen/vector-datasource/issues/1354.

queries/planet_osm_line.jinja2
nvkelso: Followup via https://github.com/tilezen/vector-datasource/issues/1354.

nvkelso:
Follow-up issues in #1353 and #1354.

nvkelso:
Some other followup: https://github.com/tilezen/vector-datasource/issues/1356

Issue #1353, opened 2017-07-31
REGRESSION: clamp NE roads to min_zoom 5


nvkelso:
Following up from https://github.com/tilezen/vector-datasource/pull/1352#discussion_r130407407...

Looks like we source the Natural Earth scale_rank for min_zoom directly, and that starts at 3:

```
    min_zoom: { col: scalerank }
```

So the changes in #1352 would introduce roads at zoom 3. We need to clamp the Natural Earth roads to 5.

nvkelso:
Same as https://github.com/tilezen/vector-datasource/issues/1074.

PR #1355, opened 2017-08-01
Make Natural Earth roads start at zoom 5

zerebubuth:
It looks like the query doesn't (yet?) use the `mz_calculate_*_roads` functions for calculating the properties or min zoom, so both the YAML and queries have been changed to make sure roads only start showing up at zoom 5.

Connects to #1353.

### DIFF
--- a/queries/ne.jinja2
+++ b/queries/ne.jinja2
@@ -225,7 +225,7 @@ WHERE
 {% endif %}
 
 {# conditional roads include #}
-{% if zoom < 8 %}
+{% if zoom < 8 and zoom >= 5 %}
 UNION ALL
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


queries/ne.jinja2
```diff
@@ -268,7 +268,7 @@ FROM
 WHERE
 
   {{ bounds['line']|bbox_filter('the_geom', 3857) }}
-  AND scalerank <= {{ zoom }}
+  AND {{ zoom }} >= 5 AND scalerank <= {{ zoom }}
```
rmarianski: does `between` fit here?

    {{ zoom }} between 5 and scalerank

But, if possible it might be better to just guard the whole query around a conditional to prevent it from executing at all if the zoom is < 5.

zerebubuth: Good point! Fixed in 5208033.

I moved the condition to the `{% if ... %}` static conditional block, as experimentally it seems as if `SELECT * FROM table WHERE FALSE` still executes the scan on `table`, which I had assumed it wouldn't.

Issue #1354, opened 2017-07-31
REGRESSION: duplicate NE and OSM representations at low zooms


nvkelso:
The work done in https://github.com/tilezen/vector-datasource/pull/1352 introduces a few regressions where before we sourced from Natural Earth at low zooms and OpenStreetMap at mid- and high-zooms and now there is both a NE and and OSM representation for the same feature at low zooms. There should only be NE.

**Followup for:**

- queries/planet_osm_line.jinja2 in https://github.com/tilezen/vector-datasource/pull/1352#discussion_r130409196 for **earth**, **road**, **water** layers switch at zoom 8.
- queries/planet_osm_polygon.jinja2 in https://github.com/tilezen/vector-datasource/pull/1352#discussion_r130411636 for **earth**, **road**, **water** layers switch at zoom 8.
- queries/planet_osm_point.jinja2 in https://github.com/tilezen/vector-datasource/pull/1352#discussion_r130412883 for **earth**, **road**, **water** layers switch at zoom 8.

First step, add tests to prove this is failing. Then fix up the logic.

nvkelso:
- **earth** is NE zooms 0 thru 8, and OSM 9+
- **water** is NE zooms 0 thru 8, and OSM 9+
- **roads** is NE zooms 0 thru 7, and OSM 8+

nvkelso:
We've proposed making the OSM/NE transition at zoom 8 for everything. Rather than doing the work twice I propose we act on https://github.com/tilezen/vector-datasource/issues/389 now as part of this fix.

- **earth**, **water**, **roads** to NE zooms 0 thru 7, and OSM 8+

Issue #1356, opened 2017-08-01
Move all jinja2 logic to YAML


nvkelso:
1. There are a few pieces that are better handled in layer YAML config than the jinja2.

https://github.com/tilezen/vector-datasource/blob/master/queries/planet_osm_point.jinja2#L26-L32

```
  CASE WHEN mz_places_min_zoom IS NOT NULL
    THEN jsonb_build_object(
      'country_capital', CASE WHEN tags->'capital' = 'yes' THEN TRUE END,
      'region_capital', CASE WHEN tags->'state_capital' = 'yes' THEN TRUE END,
      'min_zoom', mz_places_min_zoom
    )
  END AS __places_properties__,
```

2. There are many cases when we're setting up query ranges per table in the JINJA (especially Natural Earth) that we can move to the queries.yaml:

https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L12-L28

zerebubuth:
Generalising this to cover all the cases where there are SQL expressions like the example above. We should move across all the expressions where we can comfortably accommodate it in the Jinja2, e.g: all the stuff that doesn't involve joining to other tables, etc...

nvkelso:
Mostly resolved with #1362.

zerebubuth:
There's still a lot of places where we have logic in SQL that, ideally, we'd have in YAML instead. However, I'm not sure that this would be on any kind of near-term roadmap, so I don't think it should be re-opened.

Issue #1306, opened 2017-07-05
Customized min_zoom for transit, bike


nvkelso:
Similar to https://github.com/tilezen/vector-datasource/issues/1248 which suggests custom sort_rank for the bike map mode, there are cases where the `min_zoom` should be different for default (driving) versus bike versus transit.

Primarily we're optimized for driving now, and we sometimes overstuff the tiles other content if it's deemed critical for a different map mode. For example: stations.

Right now we introduce most transit `station` features in the **pois** layer at zoom 10 for our transit overlay, but then in the stylesheet we filter out most of theme at lower zooms based on the `tile_kind_rank` values. So in this case the transit mode `min_zoom` should be 10, but the default `min_zoom` should align more closely with the default Mapzen basemap compilation reveal (a few at zoom 10, a few more at zoom 11, a few more at 12, more at 13, most at 14, all at 15).

This would require changing the `min_zoom` filters at the core of Tilezen to take the minimum across a set of "min_zoom" properties when testing if the feature should be exported for a tile.

```
        station-train-subway:
            filter: { kind: [station, train_station] }
            draw:
                colorized-icons:
                    visible: global.icon_visible_station
                    size: [[13, 12px], [14, 12px], [15, 16px], [17, 18px], [18, 20px]]
                    priority: 42
                    text:
                        visible: global.text_visible_station
            station-early:
                filter:
                    $zoom: { max: 11 }
                draw:
                    colorized-icons:
                        visible: false
            # for a transit style we'd want to see **all**, so don't move this above (it's special to this style)
            low-priority-early:
                filter: { kind_tile_rank: { min: 5 }, $zoom: { min: 0, max: 13 } }
                draw:
                    colorized-icons:
                        visible: false
            low-priority-early-z13:
                filter: { kind_tile_rank: { min: 5 }, $zoom: [13] }
                draw:
                    colorized-icons:
                        visible: false
            low-priority-early-z14:
                filter: { kind_tile_rank: { min: 7 }, $zoom: [14] }
                draw:
                    colorized-icons:
                        text: { visible: false }
```

/cc @sensescape who discovered the hard `min_zoom` of 10 for stations while working on Refill level of detail sprint.

Issue #1279, opened 2017-06-23
Adjust min_zoom of OSM highways and major roads to better match Natural Earth


nvkelso:
@sensescape is adding optional level of detail filters to Refill and is making use of the `min_zoom` to do the filtering. But for the roads layer there's a discontinuity between the Natural Earth data at zooms 0-7 and the OSM data that starts at 8. Many roads that _already_ showed up in earlier zooms only have a `min_zoom` of 8 set, when they should have an earlier zoom.

- https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L185-L204

```
  #############################################################
  #
  # OSM highway
  #
  #############################################################
  - filter:
      highway: motorway
    # min_zoom 8 ≤ any mz_calculate_path_major_route(), so skip
    min_zoom: 8
    table: osm
    output:
      <<: *osm_highway_properties
      <<: *osm_network
      kind: highway
      kind_detail: motorway
  - filter:
      highway: [trunk, primary]
    # min_zoom 8 ≤ any mz_calculate_path_major_route(), so skip
    min_zoom: 8
    table: osm
    output:
      <<: *osm_highway_properties
      <<: *osm_network
      kind: major_road
      kind_detail: {col: highway}
```

nvkelso:
One solution:

- `highway: motorway` at `min_zoom: 5`
- `highway: trunk` at `min_zoom: 6`
- `highway: primary` at `min_zoom: 8`

nvkelso:
Related: https://github.com/tilezen/vector-datasource/issues/1280

nvkelso:
Subtask of https://github.com/tilezen/vector-datasource/issues/1349.

nvkelso:
Verified:

![image](https://user-images.githubusercontent.com/853051/57430370-5095fd80-71e4-11e9-8716-4d1fabb72a96.png)

![image](https://user-images.githubusercontent.com/853051/57430438-86d37d00-71e4-11e9-946f-524e13029564.png)

![image](https://user-images.githubusercontent.com/853051/57430400-6b687200-71e4-11e9-88cb-40d549fe5e2b.png)

PR #1359, opened 2017-08-01
OSM / NE transition

zerebubuth:
Change the zoom level at which the `roads`, `earth`, `water`, and `boundaries` layers switch from Natural Earth to OpenStreetMap as the main data source.

Connects to #1354.


yaml/roads.yaml
```diff
@@ -251,7 +251,7 @@ filters:
       kind_detail: {col: highway}
   - filter:
       highway: secondary
-    min_zoom: { clamp: { min: 0, max: 10, value: { call: { func: util.calculate_path_major_route, args: [ { col: fid } ] } } } }
```
nvkelso: For this and all other changes in this file...

To match up with Natural Earth changes I'd expect this to be a clamp with min of 5 which is when we want to introduce any roads layer content.

Then we need to pair it with a zoom transition in the OpenStreetMap JINJA logic.

```
    min_zoom: { clamp: { min: 5, max: 10, value: { call: { func: util.calculate_path_major_route, args: [ { col: fid } ] } } } }
```

yaml/water.yaml
```diff
@@ -18,8 +18,8 @@ globals:
   - &water_standard_min_zoom
     case:
       - when: { way_area: { min: 102400 } }
-        then: 9
-      - else: { clamp: { min: 9, max: 16, value: { sum: [ { col: zoom }, 1.066 ] } } }
+        then: 8
```
nvkelso: For this and all other changes in this file...

To match up with Natural Earth changes I'd expect this to be a clamp with min of 0 which is when we want to introduce any water layer content.

Then we need to pair it with a zoom transition in the OpenStreetMap JINJA logic for zoom 8.

```
        then: 0
      - else: { clamp: { min: 0, max: 16, value: { sum: [ { col: zoom }, 1.066 ] } } }
```

queries.yaml
```diff
@@ -22,7 +22,7 @@ sources:
     template: ne.jinja2
   osmdata:
     template: osmdata.jinja2
-    start_zoom: 9
+    start_zoom: 8
```
nvkelso: I think this "just works"... but for general OpenStreetMap data we're adding NE-OSM transition logic back to the JINJA because the min_zoom between NE and OSM overlap for the same features.

See related: https://github.com/tilezen/vector-datasource/pull/1359/files#r130750942

nvkelso:
JINJA files are here: 

- https://github.com/tilezen/vector-datasource/blob/master/queries/planet_osm_line.jinja2
- https://github.com/tilezen/vector-datasource/blob/master/queries/planet_osm_point.jinja2
- https://github.com/tilezen/vector-datasource/blob/master/queries/planet_osm_polygon.jinja2

One option is to make the final else:

- https://github.com/tilezen/vector-datasource/blob/master/queries/planet_osm_polygon.jinja2#L136

into another if zoom >= 8 test? 

- https://github.com/tilezen/vector-datasource/blob/master/queries/planet_osm_polygon.jinja2#L127

But the problem with that is some layers like landuse and pois can go all the way to 0 (because their layer content does't conflict with NE).

yaml/roads.yaml
nvkelso: Where the JINJA prevents OSM roads from being included (regardless of their min_zoom) until zoom 8.

yaml/water.yaml
nvkelso: Where the JINJA prevents OSM water from being included (regardless of their min_zoom) until zoom 8.

yaml/roads.yaml
nvkelso: Context here is issues like: https://github.com/tilezen/vector-datasource/issues/1279

zerebubuth:
I'm confused - I thought we were moving towards making the YAML the place for setting `min_zoom`? The YAML is intended to be approximately portable between SQL and Python backends and it's intended to be more readable (if only slightly) than the SQL/JINJA template combination. Putting the zoom transition logic in the JINJA template would mean we'd also need to implement it "upstream" of `vector-datasource` in RAWR tiles to match, meaning it would be much harder to modify on a style-by-style basis later.

zerebubuth:
I've made some changes in ef6d751 and 4848eb5 to move more of the logic to the JINJA templates. I'm not sure that this is what you wanted, and I'm concerned that it's going to cause us more problems in later phases of RAWR tiles. Does it match what you wanted for #1354?

rmarianski:
My understanding is that the min zoom should be set in the yaml, and the guards in the jinja files should be for "max zoom" to suppress pulling ne features in at higher zoom levels.

queries/planet_osm_polygon.jinja2
```diff
@@ -133,7 +133,7 @@ FROM (
        mz_poi_min_zoom IS NOT NULL OR
        mz_transit_level IS NOT NULL OR
        mz_water_min_zoom IS NOT NULL))
-{% else %}
+{% elif zoom >= 8 %}
```
nvkelso: Can you confirm that this limit OSM **boundary** layer related features to zoom >= 8? Are there any other changes needed to support that?

**boundaries** wasn't in the original list of affected layers, but looks like we set the OSM country features to min_zoom 8 in the YAML, but the "same" NE features get min_zoom 0 right now. They should be consistent, and while we shouldn't change that min_zoom in this PR we should allow for them to be equal but "transition" to the OSM feature at zoom 8.

nvkelso:
> I'm confused - I thought we were moving towards making the YAML the place for setting min_zoom? 

The problem is the same feature has two representations, one from OpenStreetMap and one from Natural Earth. They should share the same `min_zoom` value in the output in low-, mid-, and high-zoom tiles regardless of source.

I'm open to using something like `mz_transition_min_zoom` (for OSM?) and `mz_transition_max_zoom` (for NE?) for this as a YAML property that we then use to drop features later in a post-processing step (and drop the property as well). We can't use `max_zoom` as that has other meanings in the schema, namely when any representation of the feature should stop appearing on the map.

rmarianski:
> The problem is the same feature has two representations, one from OpenStreetMap and one from Natural Earth. They should share the same min_zoom value in the output in low-, mid-, and high-zoom tiles regardless of source.

Hhmm, I think I finally understand what you're getting at here. We can discuss further in the upcoming meeting, but my immediate thought to solve this is to create a difference internally between the min zoom that we use for filtering, and the "display" min zoom that we put in the tile.

nvkelso:
We also have a rainy day issue in https://github.com/tilezen/vector-datasource/issues/1306 to generalize min_zoom to support alternatives like `min_zoom_bike` and `min_zoom_transit` and then take the `min( min_zoom, min_zoom_bike, min_zoom_transit )` when considering if a feature is eligible for export in the tile.

nvkelso:
From today's tile meeting we're going to:

1. Now: Extend this PR to handle boundaries as well (adding test, modify jinja) to fix the regression.
2. Later: Work to remove the JINJA completely and update the general queries.yaml with the "layer" / "table" transition logic in https://github.com/tilezen/vector-datasource/issues/1356. _There is still some hand waving around this._

nvkelso:
LGTM, thanks!

PR #1360, opened 2017-08-02
Update db commands to support different user/host

rmarianski:
These changes allow me to fully setup the database on a different host with a different username.

### DIFF
--- a/scripts/setup_and_run_tests.sh
+++ b/scripts/setup_and_run_tests.sh
@@ -63,7 +63,7 @@ function cleanup {
       kill -HUP "${server_pid}"
    fi
    echo "=== Dropping database \"${dbname}\" and cleaning up..."
-   dropdb --if-exists $dbhostargs "${dbname}"
+   dropdb --if-exists $userargs $dbhostargs "${dbname}"
    rm -f empty.osm data.osc test_server.port
 }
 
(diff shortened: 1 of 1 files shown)


scripts/setup_and_run_tests.sh
```diff
@@ -53,7 +63,7 @@ function cleanup {
       kill -HUP "${server_pid}"
    fi
    echo "=== Dropping database \"${dbname}\" and cleaning up..."
-   dropdb --if-exists "${dbname}"
+   dropdb --if-exists $dbhostargs "${dbname}"
```
zerebubuth: Does this need `$userargs` as well?

rmarianski: Good catch! Updated in 5f65ddc.

rmarianski:
> A small thought which might be worth looking at is whether $DBHOST and $USER might be copied, if present, into the $PGHOST and $PGUSER environment variables. That won't affect osm2pgsql, but the PostgreSQL internal tools like createdb, dropdb and psql all look at those variables.

Should we just change the variables that we look at to be the `PGHOST` and `PGUSER` variables?

zerebubuth:
Yup, that would work! Although we'd still need to shoehorn those into the command line arguments that `osm2pgsql` is expecting.

PR #1362, opened 2017-08-03
Move logic from jinja templates -> yaml

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1356

Move the straightforward `case when` jinja changes over to yaml.

### DIFF
--- a/queries/ne.jinja2
+++ b/queries/ne.jinja2
@@ -195,7 +195,6 @@ SELECT
   NULL::jsonb AS __earth_properties__,
 
   jsonb_build_object(
-    'min_zoom', 4,
     'area', way_area,
     'kind', 'urban_area'
   ) AS __landuse_properties__,
@@ -247,7 +246,7 @@ SELECT
   NULL::jsonb AS __places_properties__,
 
   jsonb_build_object(
-      'min_zoom', GREATEST(scalerank, 5),
+      'min_zoom', scalerank,
       'scalerank', scalerank,
       'featurecla', featurecla,
       'type', type,
--- a/queries/ne.jinja2
+++ b/queries/ne.jinja2
@@ -196,7 +196,7 @@ SELECT
 
   jsonb_build_object(
     'min_zoom', 4,
-    'area', way_area::bigint,
+    'area', way_area,
     'kind', 'urban_area'
   ) AS __landuse_properties__,
 
@@ -319,7 +319,7 @@ SELECT
     'boundary', true,
 {% endif %}
 {% if has_area %}
-    'area', way_area::bigint,
+    'area', way_area,
 {% endif %}
     'min_zoom', mz_water_min_zoom,
     'featurecla', featurecla
--- a/queries/wof.jinja2
+++ b/queries/wof.jinja2
@@ -12,7 +12,7 @@ SELECT
       'area', area::bigint,
       'min_zoom', min_zoom,
       'max_zoom', max_zoom,
-      'is_landuse_aoi', CASE WHEN is_landuse_aoi = TRUE THEN TRUE END,
+      'is_landuse_aoi', is_landuse_aoi,
       'kind', wof_np.placetype_string
     ) AS __places_properties__
 
--- a/queries/wof.jinja2
+++ b/queries/wof.jinja2
@@ -9,7 +9,7 @@ SELECT
       'name', name,
       'source', 'whosonfirst.mapzen.com',
       'mz_n_photos', n_photos,
-      'area', area::bigint,
+      'area', area,
       'min_zoom', min_zoom,
       'max_zoom', max_zoom,
       'is_landuse_aoi', is_landuse_aoi,
--- a/queries/wof.jinja2
+++ b/queries/wof.jinja2
@@ -13,7 +13,7 @@ SELECT
       'min_zoom', min_zoom,
       'max_zoom', max_zoom,
       'is_landuse_aoi', is_landuse_aoi,
-      'kind', wof_np.placetype_string
+      'placetype', wof_np.placetype_string
     ) AS __places_properties__
 
 FROM wof_neighbourhood wof_n
--- a/queries/wof.jinja2
+++ b/queries/wof.jinja2
@@ -3,17 +3,20 @@
 SELECT
     wof_id AS __id__,
     {% filter geometry %}label_position{% endfilter %} AS __geometry__,
-    COALESCE(to_jsonb(l10n_name), '{}'::jsonb) AS __properties__,
 
+    COALESCE(to_jsonb(l10n_name), '{}'::jsonb) ||
     jsonb_build_object(
       'name', name,
       'source', 'whosonfirst.mapzen.com',
       'mz_n_photos', n_photos,
       'area', area,
-      'min_zoom', min_zoom,
-      'max_zoom', max_zoom,
       'is_landuse_aoi', is_landuse_aoi,
       'placetype', wof_np.placetype_string
+    ) AS __properties__,
+
+    jsonb_build_object(
+      'min_zoom', min_zoom,
+      'max_zoom', max_zoom
     ) AS __places_properties__
 
 FROM wof_neighbourhood wof_n
--- a/test/test_meta.py
+++ b/test/test_meta.py
@@ -259,6 +259,13 @@ class LanduseTest(unittest.TestCase):
         self.assertTrue(isinstance(area, int))
         self.assertEquals(3, area)
 
+    def test_ne_min_zoom(self):
+        from tilequeue.process import make_metadata
+        meta = make_metadata('ne')
+        props = dict(featurecla='Urban area')
+        out_props = self.landuse.fn(None, props, None, meta)
+        self.assertEquals(4, out_props.get('min_zoom'))
+
 
 class PlacesTest(unittest.TestCase):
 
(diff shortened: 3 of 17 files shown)


test/test_meta.py
```diff
@@ -190,6 +190,44 @@ def test_osm(self):
         out_props = self.landuse.fn(None, props, None, meta)
         self.assertEquals('scree', out_props.get('kind'))
 
+    def test_mz_is_building(self):
```
rmarianski: I added tests for all the changes here to test that the yaml functions were working correctly. It already had the plumbing necessary set up to exercise these paths.

vectordatasource/meta/sql.py
```diff
@@ -167,6 +167,8 @@
     'mz_get_min_zoom_highway_level_gate': 'mz_get_min_zoom_highway_level_gate',
     'util.calculate_path_major_route':    'mz_calculate_path_major_route',
     'mz_calculate_ferry_level':           'mz_calculate_ferry_level',
+    'util.is_building':                   'mz_calculate_is_building_or_part',
+    'util.tag_yes_to_bool':               'notimplemented',
```
rmarianski: @zerebubuth is it important for sql functions to exist for all new functions that we add?

queries/wof.jinja2
```diff
@@ -12,7 +12,7 @@ SELECT
       'area', area::bigint,
       'min_zoom', min_zoom,
       'max_zoom', max_zoom,
-      'is_landuse_aoi', CASE WHEN is_landuse_aoi = TRUE THEN TRUE END,
+      'is_landuse_aoi', is_landuse_aoi,
```
zerebubuth: Since 'min_zoom', 'max_zoom', 'is_landuse_aoi' and 'kind' all come from the WOF record, is it possible to do them all in YAML? Also `area` - all we're doing to it is converting the type, could that be done in a Python side processing function?

vectordatasource/meta/sql.py
zerebubuth: I think it's only necessary if the function is used for calculating the `min_zoom`. Otherwise we can omit it, and it'll only cause problems if we try to use the PostgreSQL-side properties calculation. I don't think we plan on using that again... it's just hanging around for completeness' sake, right?

vectordatasource/util.py
```diff
@@ -75,3 +75,21 @@ def calculate_path_major_route(fid):
 def cycling_network(tags, fid):
     # TODO: implement me! current implementation is a stub.
     return None
+
+
+def tag_yes_to_bool(str_or_none):
+    return True if str_or_none == 'yes' else None
```
zerebubuth: Do we want to also include `true` in that? Sometimes `true` is used as a synonym for `yes` in OSM tags, e.g: `return True if str_or_none in ('yes', 'true') else None`?

zerebubuth: Ooops, I should have read further down below. According to taginfo, there are no `true` values for `capital` or `state_capital`, so we can probably just leave this as is.

rmarianski: Yea, I think it makes sense to make a generic kind of function where we convert typical truthy type values to a boolean. In this case it might not matter, but it could be useful down the road. Updated in 8c2facc.

vectordatasource/meta/sql.py
rmarianski: Ok, sounds good, was just making sure.

queries/wof.jinja2
rmarianski: I'm no following what you mean by moving the min/max zoom to the yaml. Don't we just pass that through here anyway?

I moved the area type conversion to python in ec90ecd.

About the kind for wof records, we just pass that through from the database. But I see where you're coming from, so I hopefully made it clearer by assigning the `kind` name in the yaml off of the field that gets passed through in 1318c55.

vectordatasource/meta/sql.py
```diff
@@ -169,6 +169,7 @@
     'mz_calculate_ferry_level':           'mz_calculate_ferry_level',
     'util.is_building':                   'mz_calculate_is_building_or_part',
     'util.tag_str_to_bool':               'notimplemented',
+    'util.safe_int':                      'notimplemented',
```
nvkelso: What does `'notimplemented'` mean in this context?

queries/ne.jinja2
```diff
@@ -196,7 +196,7 @@ SELECT
 
   jsonb_build_object(
     'min_zoom', 4,
```
nvkelso: `min_zoom` is almost always managed in YAML, should this move there?

vectordatasource/meta/sql.py
rmarianski: See https://github.com/tilezen/vector-datasource/pull/1362#discussion_r131374485

queries/wof.jinja2
zerebubuth: Sorry, I meant that instead of including them as `__places_properties__`, we'd construct a "tags-like" JSON in the general `__properties__` which contains all the rows of the table and then allow the YAML to sort out which ones we pass through. E.g: something like:

```sql
SELECT
  ...
  to_jsonb(w.*) - 'geometry' - 'wof_id' AS __properties__
  ...
FROM wof_neighbourhood w
```

vectordatasource/meta/sql.py
zerebubuth: It's a function which doesn't exist, and will complain about `notimplemented` not existing if anyone tries to calculate the properties in SQL rather than Python. It's on a code path that we don't use any more and might get rid of in the future.

queries/ne.jinja2
rmarianski: Yes, I'm working on moving that over now too.

rmarianski: Updated in d3813c1.

queries/wof.jinja2
rmarianski: We have an hstore column for the l10n names, which is what we stick into `__properties__` at the moment. We could also merge these values from `__places_properties__` into `__properties__`, but  we will still need an empty `__places_properties__` jsonb value to signal that the record should be included in the places layer. That would look something like:

```
COALESCE(to_jsonb(l10n_name), '{}'::jsonb) ||
jsonb_build_object(
  'name', name,
  'source', 'whosonfirst.mapzen.com',
  'mz_n_photos', n_photos,
  'area', area,
  'min_zoom', min_zoom,
  'max_zoom', max_zoom,
  'is_landuse_aoi', is_landuse_aoi,
  'placetype', wof_np.placetype_string
  ) AS __properties__,

'{}::jsonb' AS __places_properties__
```

Is that better?

nvkelso: We'll need to do similar for Natural Earth in the near future.

zerebubuth: Yes, that looks good, and similar to what we've got for other tables. Thanks!

The pattern we seem to be following for other tables is to have the `min_zoom` in the `__$LAYER_properties__`. Would we want to do the same here? Or not, given that WOF min/max zoom are intrinsic and not calculated as they are for other layers?

rmarianski: Sure, updated in 6b150e2.

The way I've been thinking about where properties go for the other sources is more like a base set of properties that are common across all layers, and then the differences that get "mixed in" are what I've been putting in the layer specific namespace. The wof source is `places` layer only, so from this perspective it's somewhat arbitrary where the properties go. Anyway, happy to move them wherever they make sense and is consistent.

rmarianski:
I think I got to all the comments for the immediate work, and the tests are passing now. Ok to merge?

queries/ne.jinja2
```diff
@@ -247,7 +246,7 @@ SELECT
   NULL::jsonb AS __places_properties__,
 
   jsonb_build_object(
-      'min_zoom', GREATEST(scalerank, 5),
+      'min_zoom', scalerank,
       'scalerank', scalerank,
```
nvkelso: Confirmed we still need this for YAML logic like: https://github.com/tilezen/vector-datasource/blob/7b7394482ccd72bb8a46f30137203ea49ce974af/yaml/boundaries.yaml#L19. (It's probably possible to refactor it even more, but I think it's fine as-is for now.)

queries/ne.jinja2
```diff
@@ -195,8 +195,7 @@ SELECT
   NULL::jsonb AS __earth_properties__,
 
   jsonb_build_object(
-    'min_zoom', 4,
-    'area', way_area::bigint,
```
nvkelso: Where do we enforce area to be an int later?

yaml/water.yaml
```diff
@@ -1,7 +1,10 @@
 globals:
   - &output_properties
     name: {col: name}
-    area: {col: area}
+    area:
+      call:
+        func: util.safe_int
+        args: [ { col: area } ]
```
nvkelso: 👍

queries/ne.jinja2
rmarianski: Looks like you saw it in https://github.com/tilezen/vector-datasource/pull/1362#discussion_r131708319 :)

queries/ne.jinja2
rmarianski: This instance in particular was replaced with https://github.com/tilezen/vector-datasource/blob/6b150e2bae1e4d322c6fdf50b978a34aa45a33e1/yaml/roads.yaml#L123

Issue #1028, opened 2016-09-08
Oneway property should be true, false


nvkelso:
We've normalized all the other `yes` related properties in v1 to be True or False, but `oneway` is still `yes`.

Example tile:
- http://tile.dev.mapzen.com/vector/v1/all/15/5242/12664.topojson

`yes` should be True:

```
properties: {
kind: "major_road",
name: "Stockton St.",
sort_key: 377,
kind_detail: "tertiary",
source: "openstreetmap.org",
oneway: "yes",
sidewalk: "both",
id: 417400763
},
```

If it's `no`, should we even export that as False?

```
properties: {
source: "openstreetmap.org",
kind: "minor_road",
service: "alley",
oneway: "no",
sort_key: 356,
kind_detail: "service",
id: 332383249
},
```

nvkelso:
This is a bug fix for tiles in our SemVer doc – and will be an additive change to house styles (which don't show one-way arrows yet). Moving to v1.1

nvkelso:
Related  / duplicate:

- https://github.com/tilezen/vector-datasource/issues/792 Don't export oneway: no on roads
- #782 Normalize cycleway (& left, right, both variant) values & oneway
- https://github.com/tilezen/vector-datasource/issues/453 Normalise oneway property

nvkelso:
We're prepping to show oneway arrows on all Mapzen house styles.

nvkelso:
Well, turns out there are valid values besides yes and no:

- https://wiki.openstreetmap.org/wiki/Tag:oneway%3Dreversible
- https://wiki.openstreetmap.org/wiki/Tag:oneway%3Dalternating

Neither have very many uses in OpenStreetMap today. But thinking about the lanes on I-395 in the Washington, D.C. area that do this, those exceptions are exceptional.

Propose we close this, but still act on some / all of the issues related in https://github.com/tilezen/vector-datasource/issues/1028#issuecomment-251586008.

Issue #453, opened 2015-12-17
Normalise oneway property


zerebubuth:
In #249, the values of `is_link`, `is_tunnel`, and `is_bridge` were normalised so that by default they were not present (indicating "no") and only present when their value was positive. All of these values are comparatively rare, only being set to "yes" on a small proportion of elements. Normalising them makes it slightly easier to extract the meaning from these properties in the stylesheet, and also means that tiles didn't contain extraneous `is_something:no` properties for each road feature, reducing tile size.

The "oneway" tag is similar, except:
1. We don't set it unless it already has a value, unlike `is_link`, `is_tunnel`, and `is_bridge` which were set on every road feature.
2. While `is_link`, `is_tunnel`, and `is_bridge` always default to false, I think there are [some road types for which oneway defaults to true](http://wiki.openstreetmap.org/wiki/Key:oneway#Implied_oneway_restriction), so we would need to include that in the processing if we were to elide `oneway: no`.

Should we normalise the oneway property?

nvkelso:
`is_oneway`, ftw

nvkelso:
We'll probably need to adjust the road merging behavior to take oneway into account, too.

nvkelso:
Outdated, closing.

PR #1364, opened 2017-08-04
Update layers.md

stevage:
`is_link` and `is_tunnel` seem to be `yes`, not `true` in the data.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1046,8 +1046,8 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `is_bicycle_related`: Present and `true` when road features is a dedicated cycleway, part of an OSM bicycle network route relation, or includes cycleway infrastructure like bike lanes, or tagged bicycle=yes or bicycle=designated for shared use.
 * `is_bridge`: `true` if the road is part of a bridge. The property will not be present if the road is not part of a bridge.
 * `is_bus_route`: If present and `true`, then buses or trolley-buses travel down this road. This property is determined based on whether the road is part of an OSM bus route relation, and is only present on roads at zoom 12 and higher.
-* `is_link`: `true` if the road is part of a highway link or ramp. The property will not be present if the road is not part of a highway link or ramp.
-* `is_tunnel`: `true` if the road is part of a tunnel. The property will not be present if the road is not part of a tunnel.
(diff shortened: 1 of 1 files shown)


rmarianski:
That seems like a bug in the data rather than the docs to me. Can you provide a sample tile where you see this?

nvkelso:
Here's another example where `oneway` is **yes** but could be a boolean: https://github.com/tilezen/vector-datasource/issues/1028.

Related: https://github.com/tilezen/vector-datasource/issues/453.

Seems like if we're changing values around from yes to true that's a 2.0 breaking change.

nvkelso:
I looked at two examples in the following tile, they both show `true` (not `yes`). So I think the documentation is correct. 

Closing this for now. But @stevage if you find some counter examples please reopen with those details, please!

- http://tile.mapzen.com/mapzen/vector/v1/all/16/10473/25324.topojson?api_key=YOUR_MAPZEN_API_KEY

```
{"type":"LineString",
"properties":{"kind":"highway",
"sort_rank":374,
"kind_detail":"motorway_link",
"source":"openstreetmap.org",
"min_zoom":11.0,
"is_link":true,
"oneway":"yes",
"landuse_kind":"park",
"id":167952622},
"arcs":[55]},
````

and

```
{"type":"LineString",
"properties":{"bus_network":null,
"bicycle":"no",
"min_zoom":8.0,
"bus_shield_text":"76X",
"all_bus_shield_texts":["76X",
"28",
"91"],
"all_bus_networks":[null,null,null],
"id":251506305,
"network":"US:US",
"kind":"highway",
"name":"Presidio Pkwy.",
"sort_rank":304,
"kind_detail":"motorway",
"is_bus_route":true,
"source":"openstreetmap.org",
"is_tunnel":true,
"all_networks":["US:US"],
"oneway":"yes",
"shield_text":"101",
"landuse_kind":"park",
"ref":"US 101",
"all_shield_texts":["101"]},
"arcs":[24]},
````

stevage:
Thanks - ok, it turns out that I was using vector.tilezen, not tiles.tilezen.

nvkelso:
Ah that explains it! vector.mapzen.com has the earlier pre-v1.0 endpoint
and we standardized property values for v1 on tile.mapzen.com :)

PR #1365, opened 2017-08-07
Split up sources by templates

rmarianski:
This allows specifing start and stop zooms within a particular source, on a template level. The python code will take the templates for a particular source, and merge them together with a `UNION ALL`.

Some of the tests are currently failing. Will still need to look into those.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -11,21 +11,33 @@ all:
 
 sources:
 
-  # NOTE: these are split this way to represent each query that goes to the database
+  # NOTE: each source represents a query to the database
+  # within each source, a list of templates can be used
+  # the templates will be combined with a sql `union all` in between
+  # therefore, it's necessary that the query results of each template
+  # have the same types
+
   planet_osm_polygon:
-    template: planet_osm_polygon.jinja2
+    - template: planet_osm_polygon.jinja2
   planet_osm_line:
-    template: planet_osm_line.jinja2
+    - template: planet_osm_line.jinja2
   planet_osm_point:
-    template: planet_osm_point.jinja2
+    - template: planet_osm_point.jinja2
+
   ne:
-    template: ne.jinja2
+    - template: ne.jinja2
+
   osmdata:
-    template: osmdata.jinja2
-    start_zoom: 8
+
+    - template: osmdata-earth.jinja2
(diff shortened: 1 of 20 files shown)


queries.yaml
```diff
@@ -11,21 +11,126 @@ all:
...
+    - template: ne-water-playas-10m.jinja2
+      zoom_start: 5
+      zoom_stop: 9
+
   osmdata:
-    template: osmdata.jinja2
-    start_zoom: 8
+
+    - template: osmdata-earth.jinja2
+      zoom_start: 9
```
nvkelso: 8?

queries.yaml
```diff
@@ -11,21 +11,126 @@ all:
...
+
+    - template: osmdata-water.jinja2
+      zoom_start: 9
```
nvkelso: 8?

queries/ne-places.jinja2
```diff
@@ -0,0 +1,41 @@
...
+SELECT
+
+  gid AS __id__,
+
+  {% filter geometry %}the_geom{% endfilter %} AS __geometry__,
+  NULL::bytea AS __label__,
+
+  {{ ne_common_properties() }},
+
+  NULL::jsonb AS __boundaries_properties__,
```
zerebubuth: Just checking - we still need these `NULL::jsonb AS __layer_properties__` to make the columns match across the `UNION ALL`s?

queries.yaml
rmarianski: Nice catch! That explains why the tests were failing :) When I rebased off master I didn't catch that this changed upstream.

Fixed in b8ed8e6.

queries.yaml
rmarianski: This too. Also updated in b8ed8e6.

queries/ne-places.jinja2
rmarianski: Yes, I just double checked and it errors out if these are missing.

rmarianski:
Because I missed some of the changes in #1359, do you mind double checking to see if there's anything else in there that I could have overlooked? The tests that @zerebubuth presciently added now pass, but just wanted to verify in case there was some logic in there that might not have been captured by a test.

PR #1366, opened 2017-08-08
Return back ne population as an integer

rmarianski:


nvkelso:
I see a float being returned, where does it get converted to an int?

rmarianski:
https://github.com/tilezen/vector-datasource/blob/1227573a427f879be993cd3fdc43ea4b11004d22/vectordatasource/transform.py#L282-L287

PR #1368, opened 2017-08-09
Rename start/stop to be consistent with rest of system

rmarianski:


rmarianski:
Yea, I think the test failures are just a dep version mismatch.

PR #1372, opened 2017-08-16
Fix failing Marin watershed test

nvkelso:
Upstream data change on OSM.org

- [x] Update tests
- [x] ~~Update data migrations~~
- [x] ~~Update docs~~

### DIFF
--- a/integration-test/987-national-forests.py
+++ b/integration-test/987-national-forests.py
@@ -80,7 +80,13 @@ test.assert_has_feature(
 # Municipal Water District and protect_class=4. It should just be
 # kind:nature_reserve.
 #
+# https://www.openstreetmap.org/relation/7473345
 # https://www.openstreetmap.org/way/297463477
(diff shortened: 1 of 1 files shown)


integration-test/987-national-forests.py
```diff
@@ -80,7 +80,13 @@
 # Municipal Water District and protect_class=4. It should just be
 # kind:nature_reserve.
 #
+# https://www.openstreetmap.org/relation/7473345
 # https://www.openstreetmap.org/way/297463477
+# https://www.openstreetmap.org/way/514928209
```
nvkelso: Not clear if I _need_ the full list of member ways here if I list the relation.

nvkelso:
This should be good to go now, after CircleCI gives it's blessing.

(The earlier Circle run said the feature was back, but I forgot to update the id to new relation id in the test so it failed.)

PR #1378, opened 2017-08-21
Add min zoom filter to boundaries queries

rmarianski:
Connects to https://github.com/tilezen/vector-datasource/issues/1374

The boundaries queries were missing a min zoom filter.


rmarianski:
The [landuse ne queries](https://github.com/tilezen/vector-datasource/pull/1378/files#diff-a837464fc6acd11bcc4402d583962346R101) are also missing a min zoom filter, but those tables don't have a min zoom column set. The yaml looks to always set it 4 though, so it seems like we're fine there.

nvkelso:
> The landuse ne queries are also missing a min zoom filter, but those tables don't have a min zoom column set. The yaml looks to always set it 4 though, so it seems like we're fine there.

Should we set a min_zoom column on them?

rmarianski:
I added the landuse min zoom column to the ne tables. @zerebubuth does 795696a looks right to you?

rmarianski:
We will need a migration, yes. I'll go ahead and add one here for it.

rmarianski:
Migration added in 97ecbe9.

rmarianski:
We'll probably want to do a fresh database build this time anyway, but last time we were in this situation we were adding migrations to make it easier to test in dev, and support anyone else that might want to be on an upgrade path instead of a forcing a fresh build.

Issue #1380, opened 2017-08-24
Slower queries after query by table changes


rmarianski:
The database queries, particularly for lower zoom levels, are now much slower than before. After the query by table changes were made, namely the changes for the queries, the previous database indexes were no longer able to be used. These should be updated to match the new filters.

rmarianski:
There's still a query on the polygon table for z8+ that isn't using a targeted index.

PR #1381, opened 2017-08-24
Database optimizations for query by table changes

rmarianski:
Connects to #1380

* Add start zoom for planet sources
* Stop querying for buildings at lower zoom
* Update indexes for by table queries

### DIFF
--- a/data/apply-planet_osm_line.sql
+++ b/data/apply-planet_osm_line.sql
@@ -32,31 +32,54 @@ UPDATE planet_osm_line
 UPDATE planet_osm_line
   SET mz_label_placement = ST_PointOnSurface(way);
 
-CREATE INDEX planet_osm_line_roads_geom_index ON planet_osm_line USING gist(way) WHERE mz_road_level IS NOT NULL;
-CREATE INDEX planet_osm_line_roads_geom_9_index ON planet_osm_line USING gist(way) WHERE mz_road_level <= 9;
-CREATE INDEX planet_osm_line_roads_geom_12_index ON planet_osm_line USING gist(way) WHERE mz_road_level <= 12;
-CREATE INDEX planet_osm_line_roads_geom_15_index ON planet_osm_line USING gist(way) WHERE mz_road_level <= 15;
+-- only these 2 columns are relevant in lower zoom queries
+CREATE
+  INDEX planet_osm_line_geom_min_zoom_8_index
+  ON planet_osm_line USING gist(way)
+  WHERE
+    mz_landuse_min_zoom < 8 OR
+    mz_transit_level < 8;
 
-CREATE INDEX planet_osm_line_transit_geom_index ON planet_osm_line USING gist(way) WHERE mz_transit_level IS NOT NULL;
-CREATE INDEX planet_osm_line_transit_geom_6_index ON planet_osm_line USING gist(way) WHERE mz_transit_level <= 6;
-CREATE INDEX planet_osm_line_transit_geom_9_index ON planet_osm_line USING gist(way) WHERE mz_transit_level <= 9;
-CREATE INDEX planet_osm_line_transit_geom_12_index ON planet_osm_line USING gist(way) WHERE mz_transit_level <= 12;
-CREATE INDEX planet_osm_line_transit_geom_15_index ON planet_osm_line USING gist(way) WHERE mz_transit_level <= 15;
+-- ladder the higher zoom level indexes
+CREATE INDEX
+  planet_osm_line_geom_min_zoom_9_index
+  ON planet_osm_line USING gist(way)
+  WHERE
+    mz_earth_min_zoom < 9 OR
+    mz_landuse_min_zoom < 9 OR
+    mz_road_level < 9 OR
+    mz_transit_level < 9 OR
+    mz_water_min_zoom < 9;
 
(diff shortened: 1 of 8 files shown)


data/apply-planet_osm_line.sql
```diff
@@ -32,31 +32,54 @@ UPDATE planet_osm_line
...
-CREATE INDEX planet_osm_line_water_geom_index ON planet_osm_line USING gist(way) WHERE mz_water_min_zoom IS NOT NULL;
-CREATE INDEX planet_osm_line_water_geom_9_index ON planet_osm_line USING gist(way) WHERE mz_water_min_zoom <= 9;
-CREATE INDEX planet_osm_line_water_geom_12_index ON planet_osm_line USING gist(way) WHERE mz_water_min_zoom <= 12;
-CREATE INDEX planet_osm_line_water_geom_15_index ON planet_osm_line USING gist(way) WHERE mz_water_min_zoom <= 15;
-
-CREATE INDEX planet_osm_line_boundary_geom_index ON planet_osm_line USING gist(way) WHERE mz_boundary_min_zoom IS NOT NULL;
-CREATE INDEX planet_osm_line_boundary_geom_9_index ON planet_osm_line USING gist(way) WHERE mz_boundary_min_zoom <= 9;
-CREATE INDEX planet_osm_line_boundary_geom_12_index ON planet_osm_line USING gist(way) WHERE mz_boundary_min_zoom <= 12;
-CREATE INDEX planet_osm_line_boundary_geom_15_index ON planet_osm_line USING gist(way) WHERE mz_boundary_min_zoom <= 15;
+-- only these 2 columns are relevant in lower zoom queries
```
nvkelso: The pois layer needs to accompany any landuse layer early start_zoom (so labels can be shown for the polygons).

nvkelso: Ah, this is lines. Never mind!

queries.yaml
```diff
@@ -21,9 +21,14 @@ sources:
 
   planet_osm_polygon:
     - template: planet_osm_polygon.jinja2
+      # landuse and pois start at 4
+      start_zoom: 4
```
nvkelso: Historically we had these at 4, but the data (min_zoom) values I think supports to zoom 2. Perhaps zoom 3 is a middle ground?

nvkelso:
LGTM, let's get a review from @zerebubuth, too.

queries.yaml
rmarianski: Like you mentioned, this just matches what the behavior was previously. The `landuse` and `pois` layers queries started at zoom 4. If we go to 3 and we have some records that have a min zoom < 4, then we will start showing those features at the lower zooms. Is this what we want? If so, perhaps we should do it in a follow up issue/pr just to make that clearer?

Btw, I'm running a query on the dev database just to see what the smallest min zoom values are for pois and landuse in the polygon table. I'll report back when the results come through.

data/apply-planet_osm_line.sql
rmarianski: Yea, I thought about making some more targeted changes, but since we know that the previous indexes aren't going to be relevant any more, I figured it would be cleaner just to remove all the old ones and make new ones that would be relevant. This file is responsible for setting up new databases, and there's a migration that will remove the old indexes and create new ones.

queries.yaml
rmarianski: You were right, there are some min zooms that are set to 3 in the database. Should we update this behavior in another issue?

nvkelso:
Agree changing to start at 3 should be a followup PR.

queries/planet_osm_polygon.jinja2
```diff
@@ -147,7 +147,6 @@ FROM (
        mz_water_min_zoom < {{ zoom + 1 }}))
 {% elif zoom >= 7 %}
     ({{ bounds['polygon']|bbox_filter('way', 3857) }} AND
-      (mz_building_min_zoom < {{ zoom + 1 }} OR
```
zerebubuth: I think that deleting this line would leave the parentheses mismatched, and not sure why we wouldn't need the building min zoom. Perhaps this was a typo?

data/apply-planet_osm_line.sql
zerebubuth: If we just have these two columns as part of the query, is Postgres able to use the index to accelerate it? I'm not sure how "clever" it is at knowing that the index is a superset of these requirements.

rmarianski: I wasn't sure, so I opted to make the index match the query exactly here. To make sure I understand, are you suggesting keeping all the columns in the index where clauses, even if they aren't used, for consistency purposes? Assuming that postgres would continue to use the index, which I can verify.

queries/planet_osm_polygon.jinja2
rmarianski: Yes, it was a typo, good catch! Corrected in 0180c69.

I took buildings out here because this case would only get triggered for zoom level 7, and we know buildings won't be in the results at that zoom.

data/apply-planet_osm_line.sql
zerebubuth: Ah, no, sorry. I had misunderstood. What you've described sounds right, I just wanted to make sure that our indexes are the same "shape" as our queries. (Or that we're verifying Postgres can handle the different "shapes".)

queries.yaml
rmarianski: Follow up issue: https://github.com/tilezen/vector-datasource/issues/1382

PR #1384, opened 2017-08-30
Shorten index names

rmarianski:
Connects to https://github.com/mapzen/tile-tasks/issues/296

Postgresql has a limit on index names. Shorten the naming convention
used to ensure that they are all valid.


rmarianski:
They must have thought, "who would use such long names for indexes"? :)

PR #1394, opened 2017-09-12
Cache fixtures in S3 after a successful test.

zerebubuth:
This should mean that anyone starting from a blank slate should be able to download fixtures from S3 rather than needing to regenerate them from scratch, and this should be much quicker.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -34,6 +34,7 @@ deployment:
       - pip install 'Circle-Beacon == 2.0.0'
       - alert-circle mapzen documentation master $CIRCLE_TOKEN
       - ./scripts/update-integration-test-coordinates.sh
+      - s3cmd --continue-put put ~/.cache/vector-datasource/*.geojson s3://mapzen-tiles-assets/integration-test-fixtures/
   release:
     tag: /.+/
     commands:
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -34,6 +34,7 @@ deployment:
       - pip install 'Circle-Beacon == 2.0.0'
       - alert-circle mapzen documentation master $CIRCLE_TOKEN
       - ./scripts/update-integration-test-coordinates.sh
+      - s3cmd --continue-put put ~/.cache/vector-datasource/*.geojson s3://mapzen-tiles-assets/integration-test-fixtures/
```
nvkelso: Should `s3://mapzen-tiles-assets/integration-test-fixtures/` go into a similar `FIXTURE_CACHE` environmental var like in https://github.com/tilezen/vector-datasource/pull/1394/files#diff-c54941aa99aa0e6832038880055c51aaR58?

zerebubuth: It could. I thought that might make it more prone to misconfiguration, though. I do not expect anyone other than CircleCI to run that code. If someone else has their own fork of `vector-datasource` with different behaviour, then they'd be maintaining their own edited source files anyway, so might as well edit this file rather than set an environment variable in CircleCI.

PR #1396, opened 2017-09-15
Fix root relation IDs for stations

### LINKED ISSUE #1392: Unit test follow-ons
In https://github.com/tilezen/vector-datasource/pull/1386 @zerebubuth noted a few outstanding TODOs:

- [x] `OsmFixtureTest` seemed like a good name at the time. It turned out that having different classes for the different data sources wasn't useful, so this should be `s/^Osm//`.
    - Addressed in https://github.com/tilezen/vector-datasource/pull/1391.
- [ ] Station relation root IDs don't yet work - the list of relations needs to be passed into the fixture data source rather than having each feature already merged to its `__relations__`.
- [x] It's assumed that there's a store (e.g: S3 bucket) of pre-generated fixtures, but this isn't currently implemented. The downloading part is stubbed out, and the uploading part should (perhaps?) be done by a successful `master` branch CircleCI run.
    - https://github.com/tilezen/vector-datasource/pull/1394
- [x] There's a _lot_ of logic in `tilequeue` for this which really feels like it ought to be in `vector-datasource`, but there wasn't an obvious way to extract it cleanly.
    - _Eh, future future work. Ignore for now._
- [x] the branch of `tilequeue` that CircleCI is pulling is set to `zerebubuth/local-fixtures`. This should be changed back after https://github.com/tilezen/tilequeue/pull/241 is merged, but before this PR is merged!
    - At the time this issue was filed https://github.com/tilezen/tilequeue/pull/241 still hadn't been merged.

zerebubuth:
Fetches recursive tree of relations rather than just relations directly including features in the fixture. This means the station relations can be walked to provide access to all the station points, lines and site relations. We use this to add a `root_id` property for grouping items related to the same station.

* Add a test option to always regenerate fixtures - useful for testing.
* Add (previously commented-out) test for root relation ID.
* Added a generative test for station root relations.
* We renamed root_relation_id to root_id some time ago, updating docs to match.

Fixes #1392.

### DIFF
--- a/integration-test/__init__.py
+++ b/integration-test/__init__.py
@@ -957,7 +957,10 @@ def _load_fixture(fh):
         props = feature['properties']
         rel_ids = set(feature.get('relation_ids', []))
         if rel_ids:
-            rels = filter(lambda r: r['id'] in rel_ids, relations)
+            rels = []
+            for r in relations:
+                if r['id'] in rel_ids:
+                    rels.append(r)
             if rels:
                 props['__relations__'] = rels
         geom_lnglat = make_shape(feature['geometry'])
@@ -1239,8 +1242,13 @@ class RunTestInstance(object):
         # or tuples of (fid, shape, properties). the rels should be dicts with
         # keys for id, tags, way and rel offsets and "parts" array of IDs, as
         # if they had come from osm2pgsql's planet_osm_rels table.
-        rows = filter(lambda o: isinstance(o, tuple), objs)
-        rels = filter(lambda o: isinstance(o, dict), objs)
+        rows = []
+        rels = []
+        for o in objs:
+            if isinstance(o, tuple):
+                rows.append(o)
+            elif isinstance(o, dict):
+                rels.append(o)
 
         feature_fetcher = FixtureFeatureFetcher(rows, rels, self.env)
         self.assertions = Assertions(feature_fetcher, self.test)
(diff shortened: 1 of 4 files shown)


integration-test/__init__.py
```diff
@@ -1194,7 +1230,19 @@ def load_fixtures(self, urls, clip, simplify):
...
+        feature_fetcher = FixtureFeatureFetcher(rows, rels, self.env)
+        self.assertions = Assertions(feature_fetcher, self.test)
+
+    def generate_fixtures(self, objs):
+        # features in rows will be dsl.Feature objects (which are namedtuples)
+        # or tuples of (fid, shape, properties). the rels should be dicts with
+        # keys for id, tags, way and rel offsets and "parts" array of IDs, as
+        # if they had come from osm2pgsql's planet_osm_rels table.
+        rows = filter(lambda o: isinstance(o, tuple), objs)
+        rels = filter(lambda o: isinstance(o, dict), objs)
```
rmarianski: I'm surprised the hateful flake8 let this slide :)

zerebubuth: Good point! Do you think it would be better as

```python
rows = []
rels = []
for o in objs:
    if isinstance(o, tuple):
        rows.append(o)
    elif isinstance(o, dict):
        rels.append(o)
```
?

rmarianski: Yea, that looks good to me.

zerebubuth: Cool, thanks. Fixed in 2129454.

PR #1413, opened 2017-10-24
add default value to landuse dam min_zoom

paulmach:
I would like to propose this change so that all min_zoom expression are guaranteed to return a number. In this case, if a point geometry has `waterway: dam` the expression would eval to null.

As far as I know, this is the only place where the min_zoom expression does not always eval to a number.

### DIFF
--- a/yaml/landuse.yaml
+++ b/yaml/landuse.yaml
@@ -935,16 +935,16 @@ filters:
       kind: water_park
   - filter:
       waterway: dam
-    min_zoom:
-      case:
-        - when: { geom_type: line }
-          then: 12
-        - when: { geom_type: polygon }
-          then:
-            clamp:
-              min: 9
-              max: 16
-              value: { col: zoom }
+      geom_type: line
+    min_zoom: 12
+    output:
+      <<: *output_properties
+      kind: dam
+    extra_columns: [way]
+  - filter:
+      waterway: dam
+      geom_type: polygon
+    min_zoom: { clamp: { max: 16, min: 9, value: { col: zoom } } }
     output:
       <<: *output_properties
       kind: dam


yaml/landuse.yaml
```diff
@@ -945,6 +945,7 @@ filters:
               min: 9
               max: 16
               value: { col: zoom }
+        - else: 20
```
nvkelso: Generally our max is 17 and in rare cases 18. I’m fine using 18 instead of 20.

zerebubuth: My reading of [`pois.yaml`](https://github.com/tilezen/vector-datasource/blob/babee5dd98cf063f9c7118c781db4ae954b80714/yaml/pois.yaml#L1526) is that we would generate a POI label point for all dams, regardless of geometry type (although it would be the same point for a point dam). Would be want to have an additional point feature in the `landuse` layer?

Is it useful when we *don't* want to symbolise a POI label *unless* it's a point? In other words, not have an icon for the dam, just draw the original geometry?

nvkelso: > Would be want to have an additional point feature in the landuse layer?

No. The landuse layer should only contain lines and polygons and label points – with the icon bit destined for the **pois** layer as-is (no change requested).

> Is it useful when we don't want to symbolise a POI label unless it's a point? In other words, not have an icon for the dam, just draw the original geometry?

I'm not sure / how this relates to Pauls' PR. 

@paulmach is there an example feature that generates a null zoom now we can discuss, please?

zerebubuth: > I'm not sure / how this relates to Pauls' PR.

I was trying to understand under what circumstances it would be useful to add the dam's point geometry into the `landuse` layer, as this PR does. One possibility, that I was unsure of the usefulness of, was if someone wanted to *not* draw dam POIs, but only draw their original geometries.

If we don't want the POI geometry to appear in `landuse`, which is what you appear to be saying, then we should additionally / instead add `geom_type: [line, polygon]` to the corresponding `filter` section.

paulmach: I don't have a specific feature, this is all theoretical and kind of a implementation detail.

To not show a feature, the `min_zoom` could be [completely null](https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L171) or, in this **one** case, the `min_zoom` evaluates to null.

I'm trying to eliminate this second case as it violates the guarantee that a min_zoom expression is always a number. And since it was only one case, I thought I was a valid change to propose.  Maybe a `else: 99` would better communicate the meaning or "don't display this if it's not a line or polygon"?

If you like, I can share more details about why I care so much about this tiny detail :) but I have code in golang that converts osm data -> mapzen vector tile format/layers. It uses the information in the queries.yaml, yaml/* and spreadsheets/* to do this.  I'm working now on cleaning it up so that I can make it public and this is part of that.

zerebubuth: > I have code in golang that converts osm data -> mapzen vector tile format/layers.

Wow! That sounds awesome! Is that up on Github somewhere?

My problem with `else: 99` would be that it would still give `min_zoom` a value, and we check `min_zoom is not None` to indicate whether the feature should be present or not. We include all features with `min_zoom is not None` at the maximum zoom (currently 16), so this would mean including a feature even though the intention is that it would never be displayed.

One possible way around this would be to rewrite the current single statement into two different statements, e.g:

```yaml
  - filter:
      waterway: dam
      geom_type: line
    min_zoom: 12
    output:
      <<: *output_properties
      kind: dam
      extra_columns: [way]
  - filter:
      waterway: dam
      geom_type: polygon
    min_zoom: { clamp: { min: 9, max: 16, value: { col: zoom } }
    output:
      <<: *output_properties
      kind: dam
      extra_columns: [way]
```

I think this should preserve the exact same behaviour as the current code, but means that no `min_zoom` property is `None`. Although that's a dangerous property to rely on, since we've got no checks to ensure we don't accidentally break it in the future.

How does that look to you?

paulmach:
Thanks @zerebubuth that works for me. I made the changes above.

My code "compiles" all the min_zoom and outputs expressions so if a min_zoom can return a 'non-number' it'll return an error at "load config files" time.

Basically the code compiles all the expressions into objects implementing the `Eval(*Context) interface{}`.  If it can it upgrades this to a `EvalNum(*Context) float64` interface.  If possible it is faster and safer to work in numbers vs. casting the `float64` all the time.

If all the parts of the case statement don't return a number (e.g. else is implicitly null) it can't implement the `EvalNum` interface. Later when I make sure that all min_zooms are EvalNum it'll return the error. outputs don't have to be numbers, they can be numbers, strings or null, so this check isn't required but using `EvalNum` where possible can help.

Anyway, I'm really close to finishing it up, it is my top non-work, non-family priority (whatever that means 😄 ). I'll let you guys know when it's public.

zerebubuth:
@nvkelso, please could you have another look at this PR? Does it still need changes?

(We can ignore the CI fail, that's because of `flake8`'s increased strictness and has already been fixed on `master`.)

nvkelso:
👍 New changes work for me.

paulmach:
Thank you

PR #1420, opened 2017-11-01
Use label placement point for representative point, if available.

zerebubuth:
We generate and store a representative point for the whole geometry in the `__label__` and, later, `mz_label_placement` properties. The geometry is clipped at an earlier point in the processing pipeline, so calculating the representative point again results in some duplicate labels. If the original label placement point is available, we can use that instead.


nvkelso:
@zerebubuth does this fix the missing hospital POIs in the comparison image you posted in Slack?

PR #1435, opened 2017-11-19
Add a configuration file

pnorman:
This follows the standard pattern of seperating the config file of code that uses software (vector-datasource) from the software itself (tileserver), and leaves the detailed documentation to tileserver.

### DIFF
--- a/config.dev.yaml
+++ b/config.dev.yaml
@@ -1,3 +1,8 @@
+# This is a minimal configuration to get vector-datasource running in
+# development. Documentation about the tileserver configuration syntax
+# can be found in it's repository.
+# In production the debug, reloading, and caching settings will probably
+# need to be changed, at a minimum.
 postgresql:
   dbnames: [osm]
 queries:
--- a/config.yaml
+++ b/config.yaml
@@ -1,0 +1,52 @@
+postgresql:
+  dbnames: [osm]
+queries:
+  config: queries.yaml
+  template-path: queries
+  reload-templates: true
+
+formats: [json, topojson, mvt]
+buffer: {}
+
+server:
+  host: localhost
+  port: 8080
+  debug: true
+  reload: true
+  threaded: false
+
+http:
+  # whether to include cors headers on responses
+  cors: true
+  # client cache time in seconds ("Cache-Control: max-age" header)
+  max-age: 300
+
+health:
+  url: /_health
+
+path_tile_size:
+  # NOTE: the keys should be strings
+  "256": 1
+  "512": 2
+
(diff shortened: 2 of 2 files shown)


zerebubuth:
I'd prefer to call this `config.yaml.example` or `config.yaml.tileserver` or something like that. I think this would make it clearer that the file should be copied and modified for local use, rather than any local modifications being committed.

rmarianski:
Are we planning on moving the tileserver configuration over to this repo? I'd suggest removing all the yaml bits from the tileserver repo if so (and code to parse it). It's confusing to have the tileserver yaml configuration in both repos.

pnorman:
> I'd prefer to call this config.yaml.example or config.yaml.tileserver or something like that

Done, and added a comment too the file. I went for `config.dev.yaml` to preserve syntax highlighting, and because most people won't need to modify it for a dev setup.

> Are we planning on moving the tileserver configuration over to this repo? I'd suggest removing all the yaml bits from the tileserver repo if so (and code to parse it). It's confusing to have the tileserver yaml configuration in both repos.

I'm not quite sure what you mean, because there's so many parts of the chain that are yaml. Tileserver obviously needs to be able to parse its own config file, so it's not possible to remove yaml bits from tileserver.

The immediate goal here is to make it so that you need the tileserver **binary** to run vector-datasource, not the entire tileserver repo.

pnorman:
circleci test failure seem unrelated.

rmarianski:
> The immediate goal here is to make it so that you need the tileserver binary to run vector-datasource, not the entire tileserver repo.

What do you mean by running vector-datasource? What I'm confused about is what the tileserver binary provides for you here along with its config file. Isn't the only thing you can do is run tileserver to serve tiles? How would this configuration be different than the default example one in tileserver?

Is the idea that each data source, vector datasource being one, would get its own tileserver config? Maybe we can work towards settings things up so that different data sources don't need a separate tileserver config. Perhaps the default binary can try to read default configuration from a location relative to the package, so that just running the tileserver binary without specifying an optional configuration file would just work in the typical case?

I'm not against this, just trying to get at the underlying motivation here, because now we'll have two locations, across repo boundaries, where we have to maintain configuration changes. Just want to make sure it's worth it and know where we ideally want to end up :)

nvkelso:
(Noting here that we'll return to refactoring various Tilezen pieces in mid Dec.)

pnorman:
The configuration file currently in tileserver contains vector-datasource specific information - the path to the queries, the path to the templates, database name, the yaml path, and possibly other yaml information.

Issue #491, opened 2016-01-15
Backfill tile tests for existing features


nvkelso:
Building off @zerebubuth's work in https://github.com/mapzen/vector-datasource/pull/464, let's flesh out more tests (probably in several rounds).

nvkelso:
We've back filled most tests now, closing.

PR #1439, opened 2017-11-22
Add tests for some shop POIs

pnorman:
- [x] Update tests
- [x] ~~Update data migrations N/a~~
- [x] ~~Update docs N/a~~

I'm doing a PR this early to make sure where I'm adding the test is reasonable. I thought about a new file, but it fits in well here.

### DIFF
--- a/integration-test/491-feature-tests.py
+++ b/integration-test/491-feature-tests.py
@@ -1,0 +1,26 @@
+from . import FixtureTest
+
+
+class FeaturesTest(FixtureTest):
+    def test_shops(self):
+        self._run_test(
+            'http://www.openstreetmap.org/node/2893904480',
+            '16/19299/24631', {'kind': 'bakery'})
+        self._run_test(
+            'http://www.openstreetmap.org/node/886395953',
+            '16/19297/24636', {'kind': 'books'})
+        self._run_test(
+            'http://www.openstreetmap.org/node/2709493928',
+            '16/19297/24629', {'kind': 'butcher'})
+        self._run_test(
+            'http://www.openstreetmap.org/node/2565702300',
+            '16/19295/24638', {'kind': 'car'})
+        self._run_test(
+            'http://www.openstreetmap.org/node/2065155887',
+            '16/19310/24632', {'kind': 'car_repair'})
+
+    def _run_test(self, url, zxy, props):
+        z, x, y = map(int, zxy.split('/'))
+        self.load_fixtures([url])
+        self.assert_has_feature(
+            z, x, y, 'pois', props)


pnorman:
> We generally name test files using the issue number that the tests relate to. But, here we don't have an issue number to use in the filename... so I'm a little unsure what to do.

Ah. I'm unsure then too. The stuff I add is likely to be split up across multiple issues/PRs, so by number doesn't really make sense, since the split will seem arbitrary. I'd like to group all shop tests into the same file, possibly with other things in the file as well.

Could we do something other than issue-numbered tests?

rmarianski:
If the tests were motivated by #526, I think it's fine to name the file using that issue number. If not, we don't actually need to have an issue number for the tests right, technically speaking?

nvkelso:
Hang it off https://github.com/tilezen/vector-datasource/issues/491.

pnorman:
I can do that, but wonder if #491 will then end up being a catch-all for unrelated tests. Perhaps something to be considered long-term.

nvkelso:
If the coat rack gets overloaded then we'll build another one. In the meantime start hanging tests off it ;)

pnorman:
Moved to their own ~~coatrack~~ file.

pnorman:
pinging on this

nvkelso:
Please choose a test case with a name to match the existing logic.

pnorman:
> Please choose a test case with a name to match the existing logic.

I'm not sure what you mean.

nvkelso:
Now I’m not sure either ;) I’ll have a closer look after dinner.

pnorman:
> Now I’m not sure either ;) I’ll have a closer look after dinner.

I did the file rename, perhaps you were thinking of that? Or the name of an object in the file?

integration-test/491-feature-tests.py
```diff
@@ -0,0 +1,26 @@
+from . import FixtureTest
+
+
+class InclusivePois(FixtureTest):
```
nvkelso: @rmarianski @zerebubuth I think this class name is just re-using some other test file's class name (one about International Women's Day and inclusive POIs). Can we rename it to be something more generic here?

nvkelso:
Other than small bit in https://github.com/tilezen/vector-datasource/pull/1439#discussion_r154566896, LGTM.

nvkelso:
@zerebubuth Can you review in Monday morning, please?

integration-test/491-feature-tests.py
zerebubuth: Yup. The class name is not important for the functioning of the test but, like a misleading comment, might contribute to confusion. Since #491 is very generic, it would probably be okay to rename this to `FeaturesTest` or similar. Finer granularity issues could help inspire finer granularity names.

pnorman:
Updates done

Issue #1186, opened 2017-03-28
Require that most POIs have names


nvkelso:
* **What did you see?**

Gardens without names are appearing in Scotland as POIs.

* **What did you expect to see?**

The landuse layer can have unnamed polygons, but POIs should _generally_ only include named features.

* **What map location are you having problems with?** 

- http://tile.mapzen.com/mapzen/vector/v1/all/12/2011/1276.topojson
- http://www.openstreetmap.org/way/273274870
- https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L441-L444

Currently the Mapzen basemap styles have logic to insist on names for most features, except the following kinds:

- https://github.com/tangrams/walkabout-style/blob/gh-pages/walkabout-style.yaml#L4298-L4413
- https://github.com/tangrams/walkabout-style/blob/gh-pages/layers/bike-interlay.yaml#L249-L276

Namely:

Namely it disabled all POIs from displaying if they don't have a name, **unless**:

## No name okay at any zoom:

_The pois.yaml filter already clamps to sane min_zoom value._

- [ ] attraction
- [ ] bbq
- [ ] beacon
- [ ] bicycle_junction
- [ ] bicycle_parking
- [ ] bicycle_repair_station
- [ ] boat_storage
- [ ] communications_tower
- [ ] cross
- [ ] dam
- [ ] drinking_water
- [ ] dune
- [ ] egress
- [ ] fee_station ??? not sure about this one, did we change the kind name before v1?
- [ ] fire_lookout ??? not sure about this one, did we change the kind name before v1?
- [ ] firepit
- [ ] fishing_area
- [ ] gate
- [ ] geyser
- [ ] hazard
- [ ] horse ??? not sure about this one, did we change the kind name before v1?
- [ ] hot_spring
- [ ] information
- [ ] landmark - has a area logic, too
- [ ] life_ring
- [ ] lifeguard_tower
- [ ] observatory
- [ ] parking
- [ ] petroleum_well
- [ ] picnic_site
- [ ] picnic_table
- [ ] pitch
- [ ] playground
- [ ] power_pole
- [ ] power_tower
- [ ] put_in
- [ ] put_in_egress
- [ ] pylon
- [ ] ranger_station
- [ ] rapid
- [ ] rest_area (this is a very early min_zoom 11)
- [ ] rock
- [ ] saddle
- [ ] shower
- [ ] silo ??? not sure about this one, did we change the kind name before v1?
[...]

PR #1442, opened 2017-11-27
Accept shops without name tags

pnorman:
- [ ] Update tests
- [x] Update data migrations
- [x] Update docs

I considered adding a test for this. The problem is that stores without names in the data tend to acquire them as more detailed mapping gets done, which would render an OSM ID based test worthless.

There are no references to no names in the docs.

### DIFF
--- a/data/migrations/v1.5.0-planet_osm_point.sql
+++ b/data/migrations/v1.5.0-planet_osm_point.sql
@@ -1,3 +1,12 @@
+-- Add zoom for POIs without a name
+UPDATE planet_osm_point
+  SET mz_poi_min_zoom = mz_calculate_min_zoom_pois(planet_osm_point)
+  WHERE
+    shop IS NOT NULL
+    AND name IS NULL
+    AND mz_poi_min_zoom IS DISTINCT FROM mz_calculate_min_zoom_pois(planet_osm_point);
+
+-- Index work should be done after UPDATEs
 -- ladder the point indexes
 CREATE INDEX IF NOT EXISTS
   planet_osm_point_geom_min_zoom_6_index
--- a/data/migrations/v1.5.0-planet_osm_polygon.sql
+++ b/data/migrations/v1.5.0-planet_osm_polygon.sql
@@ -1,3 +1,12 @@
+-- Add zoom for POIs without a name
+UPDATE planet_osm_polygon
+  SET mz_poi_min_zoom = mz_calculate_min_zoom_pois(planet_osm_polygon)
+  WHERE
+    shop IS NOT NULL
+    AND name IS NULL
+    AND mz_poi_min_zoom IS DISTINCT FROM mz_calculate_min_zoom_pois(planet_osm_polygon);
+
+-- Index work should be done after UPDATEs
 -- polygon low zoom
 CREATE INDEX IF NOT EXISTS
   planet_osm_polygon_geom_min_zoom_7_index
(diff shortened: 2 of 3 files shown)


data/migrations/v1.5.0-planet_osm_point.sql
```diff
@@ -1,3 +1,12 @@
+-- Add zoom for POIs without a name
+UPDATE planet_osm_point
+  SET mz_poi_min_zoom = mz_calculate_min_zoom_pois(planet_osm_point)
+  WHERE
+    shop IS NOT NULL
+    AND name IS NULL
+    AND mz_poi_min_zoom IS NULL;
```
rmarianski: Is it possible that this could alter the existing min zoom of a poi? If so, can update the condition to something like `coalesce(mz_poi_min_zoom, 999) <> coalesce(mz_calculate_min_zoom_pois(planet_osm_point), 999)`. If not, disregard :)

nvkelso:
We decided in https://github.com/tilezen/vector-datasource/issues/1186 in the last couple releases to drop POIs without a name except in exceptional circumstances gearing the map towards consumer mapping versus survey mapping. 

Please remove `shop: yes` from the no-names whitelist list in https://github.com/tilezen/vector-datasource/pull/1442/files#diff-85e6f75825fd3fd79970e1da0dd6cdaeR164.

nvkelso:
(Not sure this PR is necessary at all?)

data/migrations/v1.5.0-planet_osm_point.sql
pnorman: Isn't the condition you gave just `mz_calculate_min_zoom_pois(planet_osm_point) <> mz_poi_min_zoom` This would be a more restrictive condition, and help cut down on the number of updates.

rmarianski: It's because NULLs are "infectious" as I like to call it, ie if any of them are null then it won't update.
`select NULL <> 42` => `NULL`

pnorman:
> Please remove shop: yes from the no-names whitelist list in https://github.com/tilezen/vector-datasource/pull/1442/files#diff-85e6f75825fd3fd79970e1da0dd6cdaeR164.

That would render the PR moot.

> gearing the map towards consumer mapping versus survey mapping.

This isn't about survey mapping. Displaying a shop icon without a name is a fairly common cartographic choice. Taking the example of a gas station, I generally care more about there being a gas station, than knowing its name. I want to see all gas stations, not just named ones. This wouldn't impact gas stations, them being `amenity` not `shop`, but the same applies to most shops. If I'm after a greengrocer, unless I happen to be local and know what the names mean, I care about presence, not the name.

@zerebubuth wrote in https://github.com/tilezen/vector-datasource/issues/1186#issuecomment-290392298

>I think we can use the name as a proxy for importance, and the access restriction as a measure of public utility.

This is true for gardens, where the difference importance between a named one and an unnamed one is huge, with so many small unnamed gardens. For shops, it's much less the case. They generally all have names on the ground, so the name tag doesn't indicate importance.

This won't make a big impact on tile size, because 80% of shops already have a name.

If we decide not to do this, I'll just switch around my test for the other issue so it uses one with a name, but I think we should include unnamed shops.

data/migrations/v1.5.0-planet_osm_point.sql
pnorman: Oh yes, because it's an inequality, it should be `IS DISTINCT FROM`

rmarianski: Didn't know about `is distinct from`, nice!

nvkelso:
I'm fine with a subset of shops, but not all shops. Update the PR with proposed subset.

pnorman:
> I'm fine with a subset of shops, but not all shops. Update the PR with proposed subset.

My subset would likely include every shop we render.

Closing this PR to work on other stuff that involves shops - I'll take another look later on.

PR #1444, opened 2017-12-01
WIP: add more data, upgrade to natural earth v4

nvkelso:
- [ ] Update tests
- [ ] Update data migrations
- [ ] Update docs

### DIFF
--- a/data/assets.yaml
+++ b/data/assets.yaml
@@ -1,5 +1,5 @@
 bucket: mapzen-tiles-assets
-datestamp: 20161110
+datestamp: 20171130
 
 shapefiles:
 
--- a/data/assets.yaml
+++ b/data/assets.yaml
@@ -9,6 +9,8 @@ shapefiles:
   #   - prj: source projection (defaults to 4326)
   #   - tile: whether to tile the data source before importing
 
+  # water and land layer sources from timestamped openstreetmapdata.com source
+
   - name: water_polygons
     url: http://data.openstreetmapdata.com/water-polygons-split-3857.zip
     prj: 3857
@@ -21,11 +23,16 @@ shapefiles:
     directory: land-polygons-split-3857
     shapefile-name: land-polygons-split-3857/land_polygons.shp
 
+  # used to intercut line features in boundaries layer with ocean marine areas
+  # hand curated by Mapzen for the Tilezen project
+
   - name: buffered_land
     url: http://s3.amazonaws.com/mapzen-tiles-assets/curated/buffered_land_10272016.zip
     prj: 3857
     shapefile-name: buffered_land_10272016.shp
     tile: true
+
+  # Natural Earth
 
   - name: ne_110m_lakes
     url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_lakes.zip
--- a/data/assets.yaml
+++ b/data/assets.yaml
@@ -68,7 +68,7 @@ shapefiles:
     prj: 4326
 
   # this should be scale_rank instead (instead of ne_10m_ocean)
-  - name: ne_10m_ocean
+  - name: ne_10m_ocean_scale_rank
     url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/physical/ne_10m_ocean_scale_rank.zip
     prj: 4326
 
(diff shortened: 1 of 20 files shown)


nvkelso:
@rmarianski I've updated the assets.yaml with the new files, but when I run `python bootstrap.py` in the `data` dir nothing happens. What's my next move?

nvkelso:
@rmarianski if you could walk me thru how to refactor the places layer now that I'm proposing a bunch more features there that'd also be helpful. 

Or perhaps we just wait until the Wikidata IDs in NE v4.1 are added and we just do a data join on all/some of those? Else we loose all the translations which would be a significant regression. But for the disputed area labels we'll definitely need to add those in this PR...

nvkelso:
I'm also pretty confused by:

- signatures in the ne_boundaries queries – they seem opposite for some 10m versus 50m for same theme?
- how to generate label points from ne_10m_geography_regions_polys, and if I just need that one or all the 50m and 110m themes, too? If it should follow the populated places template or the normal template.
- how do I set this to only export label position lines? the scale_rank versions (next lines) are for the geometries that would be stroked on the map:
  `{{ ne_water_query(False, False, False, True, False, 'ne_10m_rivers_lake_centerlines') }}`
- how to add data driven `min_zoom` versus calculated min_zoom in the `ne.jinja2` template.

nvkelso:
(The other outstanding pieces here are to update all the YAML filters to target whatever `featurecla` need to be added. I'll start working thru that Friday and Monday.)

rmarianski:
> @rmarianski I've updated the assets.yaml with the new files, but when I run python bootstrap.py in the data dir nothing happens. What's my next move?

That updates the makefile. The next step is: `make -f Makefile-import-data` which generates a shellscript to import the shapefiles into postgres, `import-shapefiles.sh`. If you're adding a few shapefiles, you can probably extract out just the relevant ones from that file and run those directly. Note that this will download the shapefiles tarball that's on s3 first, but it'll only happen once the first time you run it.

> @rmarianski if you could walk me thru how to refactor the places layer now that I'm proposing a bunch more features there that'd also be helpful.
> 
> Or perhaps we just wait until the Wikidata IDs in NE v4.1 are added and we just do a data join on all/some of those? Else we loose all the translations which would be a significant regression. But for the disputed area labels we'll definitely need to add those in this PR...

Do you mean how to alter the sql queries for NE, or something more? In either case, I think the specifics will dictate the approach. I'm happy to check in and talk more about the details here; I'm just not familiar with it.

> I'm also pretty confused by:
> 
> signatures in the ne_boundaries queries – they seem opposite for some 10m versus 50m for same theme?

For boundaries? I'm taking a quick look and the difference seems to be just whether a `name` is returned back. Am I looking at the same thing you are?

https://github.com/tilezen/vector-datasource/blob/847726832d6433166cf806725eb5c98e009f7b2d/queries/ne-boundaries-50m.jinja2#L2-L6

vs

https://github.com/tilezen/vector-datasource/blob/847726832d6433166cf806725eb5c98e009f7b2d/queries/ne-boundaries-10m.jinja2#L2-L8

where the first arg is just whether the name should be returned.

> how to generate label points from ne_10m_geography_regions_polys, and if I just need that one or all the 50m and 110m themes, too? If it should follow the populated places template or the normal template.

Returning back the point in the `__label__` property should do it. Downstream [python](https://github.com/tilezen/tilequeue/blob/563e6d98a1347b7e4b42c44ee669eb8120d671bf/tilequeue/process.py#L314-L319) code adds that into the properties as `mz_label_placement`.

> how do I set this to only export label position lines? the scale_rank versions (next lines) are for the geometries that would be stroked on the map:
> {{ ne_water_query(False, False, False, True, False, 'ne_10m_rivers_lake_centerlines') }}

Is this what we do for boundaries currently? If so, [this](https://github.com/tilezen/vector-datasource/blob/847726832d6433166cf806725eb5c98e009f7b2d/queries/planet_osm_polygon.jinja2#L21-L28) is what we do there.

> how to add data driven min_zoom versus calculated min_zoom in the ne.jinja2 template.

Do you mean that in some cases you want it just passed through vs calculated? The queries need to provide any of the data needed for calculation, and the yaml generates the actual value based off that. If it's known at query time and you just want to pass that on through to the property, then the yaml can just read it directly via `{col:  min_zoom}`.

nvkelso:
Close this, will remake in the future.

Issue #1226, opened 2017-04-27
landuse_kind not always present on buildings


nvkelso:
* **What did you see?**
Reported via support, the buildings at LAX Airport in Los Angeles _mostly_ have the expected `landuse_kind` properties, but not always, and even the same building can include or not include the property at different zooms.

* **What did you expect to see?**
All buildings on the airport landuse polygon should get `landuse_kind` consistently.

* **What map location are you having problems with?** 
- http://tangrams.github.io/bubble-wrap/#14/33.9412/-118.4190

* **Screenshot?** 

zoom 14 has the main international terminal without, while terminal 2 includes it:

<img width="827" alt="screen shot 2017-04-26 at 20 44 03" src="https://cloud.githubusercontent.com/assets/853051/25467174/781d827c-2ac1-11e7-86dc-28cb62c9de69.png">
<img width="941" alt="screen shot 2017-04-26 at 20 43 40" src="https://cloud.githubusercontent.com/assets/853051/25467175/781ea79c-2ac1-11e7-8dca-0bc1d03b895c.png">

While zoom 14 didn't include the property, now zoom 16 does on nearby United building:
<img width="759" alt="screen shot 2017-04-26 at 20 44 12" src="https://cloud.githubusercontent.com/assets/853051/25467173/780ca2b8-2ac1-11e7-9342-f78324215156.png">

nvkelso:
We think this is because the centroid for the building is calculated on a larger clip than the landuse polygon. They both need to be based on the same clip.

zerebubuth:
Not the centroid, but the landuse polygon must [overlap at least 80% of the building area](https://github.com/tilezen/vector-datasource/blob/b7bd4036967d81c1451a57dbdf5af8a6fbf60293/vectordatasource/transform.py#L1073). The effect is the same, though.

nvkelso:
Also affects `building` features especially in relations, here in New York city (red has it, grey doesn't):

- https://www.openstreetmap.org/way/137564642#map=17/40.75345/-73.97684

<img width="818" alt="screen shot 2017-05-16 at 5 29 33 pm" src="https://cloud.githubusercontent.com/assets/853051/26138194/25b6f2fa-3a7c-11e7-9168-a3b3cb58b7d5.png">

/cc @sensescape

nvkelso:
Related: https://github.com/tilezen/vector-datasource/issues/1142 Should we clip buildings completely to tile boundaries?

[...]

PR #1446, opened 2017-12-01
Clip buildings to tiles.

### LINKED ISSUE #1142: Clip buildings
Buildings are clipped at [3x3 tile boundaries](https://github.com/tilezen/vector-datasource/blob/1ab3ebadcf4c8fa5eab3ef55d2d0e6ed6cb003a2/queries.yaml#L107), and has its own side effects, alpha issues? cc @meetar @bcamper 

We should try clipping buildings like other features on dev to see if that introduces any artifacts.

zerebubuth:
Looks like this was done already as part of #1352, so this PR just updates the test to have the behaviour that we want; that all building geometry is within the tile bounds.

Fixes #1142.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -194,6 +194,8 @@ Label position points may also have `closed` or `historical` kind_detail values 
 
 Values for `kind_detail`  are sourced from OpenStreetMap's `building` tag for building footprints and from `building:part` tag for building parts.
 
+Note that building geometries, like most geometries in Tilezen tiles, are clipped to the bounds of the tile, even if the building extends beyond the tile. This means that it might be necessary to assemble geometry from several neighbouring tiles to recreate the full building. Some buildings are exceptionally large and span many tiles, so this can be tricky.
+
 #### Building properties (common):
 
 * `name`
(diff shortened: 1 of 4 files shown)


nvkelso:
This is not obvious, but if you follow all the issue references, we're hoping the building clip solves the landuse_kind intercut problems in https://github.com/tilezen/vector-datasource/issues/1226. Can you add tests in this PR to prove that's true, please?

zerebubuth:
This won't solve all the `landuse_kind` intercut problems, but it might make it better. I'll add some examples.

Just to be clear: We'll get different kinds of problems instead when buildings cross tile boundaries and one half gets assigned a different `landuse_kind` than the other. E.g:

image

nvkelso:
Well, sorry to be the bearer of bad news but now https://github.com/tilezen/vector-datasource/issues/1226 is assigned to you, too ;) 

Thanks for the illustration. I agree it's a related but different issue than this one.

zerebubuth:
@nvkelso how do those new tests look?

zerebubuth:
@nvkelso all good to merge?

nvkelso:
Yes.

Issue #1425, opened 2017-11-03
Add new kinds for full OSM.org / iD icon library support


nvkelso:
Around 39:

- [x] **art** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dart
    - This is different than tourism=artwork
- [x] **beach** kind_detail – add `surface` as `kind_detail` in the **landuse** and **pois** layers.
	- [x] beach_coarse from http://wiki.openstreetmap.org/wiki/Key:surface???
- [x] **beauty** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dbeauty
- [x] **car_wash** - http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_wash
- [x] **charging_station** - http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station
	- bicycle=yes - electric bikes (also called pedalec) can be charged
	- scooter=yes - electric scooters can be charged
	- car=yes - electric cars can be charged
	- truck=yes - electric trucks can be charged
- [x] **chemist** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dchemist
- [x] **coffee** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dcoffee
- [x] ~~~**communications** ???~~~
- [x] **deli** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Ddeli
- [x] **elevator** - http://wiki.openstreetmap.org/wiki/Tag:highway%3Delevator
- [x] **embankment** - http://wiki.openstreetmap.org/wiki/Key:embankment *** this one gets a little complicated – put into the **landuse** layer.
- [x] **furniture** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dfurniture
- [x] **grave_yard** kind_detail (already loose in religion:*)
	- [ ] grave_yard_christian
	- [ ] grave_yard_generic
	- [ ] grave_yard_jewish
	- [ ] grave_yard_muslim
	- [ ] Also add `denomination` = * to allow for `jehovahs_witness` who don't like a cross symbol, and to allow other uses.
- [x] **hifi** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dhifi
- [x] **hunting_stand** - http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dhunting_stand
[...]

PR #1447, opened 2017-12-03
Add shops from osm.org and iD

pnorman:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

This is a big part of #1425


yaml/pois.yaml
```diff
@@ -1360,8 +1360,11 @@ filters:
       <<: *output_properties
       kind: cycle_barrier
   - filter:
-      shop: [bakery, books, butcher, car, car_repair, clothes, computer, convenience,
-        fashion, florist, gift, greengrocer, hairdresser, jewelry, mobile_phone, optician, pet]
+      shop: [art, bakery, beauty, books, butcher, car, car_repair, clothes,
```
nvkelso: `shop=coffee` and `shop=deli` seem more like `amenity=cafe` and should get that **min_zoom** range.

`shop=variety_store` and `shop=furniture` and `shop=shoes` seems more like `craft: [brewery, carpenter...]` and should get that **min_zoom** range.

I'm fine with the rest being `min_zoom: 17`.

pnorman:
Pushed in a missing file, haven't handled the review comments yet

yaml/pois.yaml
pnorman: > shop=variety_store and shop=furniture and shop=shoes seems more like craft: [brewery, carpenter...] and should get that min_zoom range. Even though this has been around awhile, shop=pet should also be in this set.

I'm making the changes, but `craft` is a bit of a weird thing in OSM, and a better word would probably have been `trade`. I don't see the `craft` values and those `shop` values forming much of a coherent set - some are places a customer goes to buy stuff, others don't interact directly with customers at their locations.

pnorman:
Everything should be good, and PEP style issues are fixed.

PR #1449, opened 2017-12-04
Add a script to get z16 tile for a node

pnorman:
Just a script to get a z16 tile for a node, which I found useful when writing tests.


nvkelso:
LGTM, thanks!

PR #1450, opened 2017-12-04
Add some basic performance recommendations

pnorman:
This starts a file with selected general performance recommendations and Tilezen specific ones.

There's a some changes from the current settings that are significant

- `work_mem` changed from ~2GB to 128MB. At 2GB I have real concerns about queries taking all the RAM, and I haven't seen much in performance gains past 128MB. 256 would also be reasonable.
- `autovacuum_work_mem` set explicitly. With `maintenance_work_mem` set so high for index builds, it's useful to turn it back down for autovacuum.
- `random_page_cost` reflects current recommendations
- The scale factors will cause autovacuum to run much more often. This is necessary, as otherwise it only runs every few months, trying to do a big giant cleanup.

### DIFF
--- a/PERFORMANCE.md
+++ b/PERFORMANCE.md
@@ -1,0 +1,34 @@
+# Improving performance
+
+Performance doesn't matter much for creating tiles for an extract, but when handling production loads for the entire planet, some thought needs to be given to it
+
+## PostgreSQL
+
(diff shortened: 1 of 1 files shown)


nvkelso:
Can you also link to this document in the vector-datasource Wiki, please?

nvkelso:
Added wiki link in https://github.com/tilezen/vector-datasource/wiki/Mapzen-Vector-Tile-Service/8258bda80cb30c9b8401b5786530ec54c20d5fd0.

Issue #1062, opened 2016-09-21
E-road and A-road shield_text cleanup


nvkelso:
This is followup refinement for work originally performed in https://github.com/tilezen/vector-datasource/issues/192.

For E-roads, A-roads, and probably others, the leading character should – it's the 2nd space character that should be dropped.

Here are those examples again as OSM ways & relations:
- shield_text of `A151` from ref of `A 151` with [way](https://www.openstreetmap.org/way/208288552) and [relation](https://www.openstreetmap.org/relation/1159812)
- shield_text of `E402` from ref of `E 402` with [way](https://www.openstreetmap.org/way/121496753) and [relation](https://www.openstreetmap.org/relation/88503)

Looking at TagInfo these aren't commonly tag'd in Europe so this is a v1.1 cleanup task.

nvkelso:
Another example from user report:

Some labels for A roads in UK Derbyshire area are mislabelled

- An example is the A52 Stoke on Trent to Derby City. Its labelled just 52 without the A.
    - @nvkelso comment: This is an example of the focus of this issue, that the A should be retained by the space dropped.
- M1 is labelled - 13
    - @nvkelso comment: This is a data problem in that the E-Road network is present but M1 is only on the road segment attributes and not in a network relation. Once it's in a network relation (and I proposed adding it just for that one road in OSM for testing purposes) then the M1 should sort before the E13 in the shield and network lists.

nvkelso:
Here's an A-road that's loosing it's number in the `shield_text` conversion in Rome, /cc @sensescape.

![image 1](https://user-images.githubusercontent.com/853051/38340455-6795df24-3828-11e8-8166-722840e1b648.png)
![image](https://user-images.githubusercontent.com/853051/38340456-67b03bda-3828-11e8-871b-e836ddbf9e5f.png)

nvkelso:
Related: https://github.com/tilezen/vector-datasource/issues/1483

nvkelso:
This is working in "prod" Nextzen endpoint, closing.

Issue #1389, opened 2017-09-07
Support new US and international shield artwork


nvkelso:
Mapzen basemap styles are adding artwork for all the state-level shields in the USA and top-level shields for several dozen other countries. 

- Report the correct network values and shield_text lengths
- For shield_length > 5, what are the values? We may need to add shield_text cleaning logic &/or ask for wider shield artwork.
- Add assertions (integration tests) to for a sampling of features

nvkelso:
Amazon Athena can be used to report, see https://aws.amazon.com/blogs/big-data/querying-openstreetmap-with-amazon-athena/.

Example query:

```
SELECT tags['network'] as state, length(tags['ref']) AS ref_length, count(distinct tags['ref']) AS count
FROM planet
WHERE type='relation'
        AND tags['network'] LIKE 'US:%'
        AND tags['ref'] IS NOT NULL
GROUP BY tags['network'], length(tags['ref'])
ORDER BY tags['network'], length(tags['ref']), count(*) desc
```

Results in a table like:

| state | ref_length | count
-- | -- | --
1 | US:AAR | 2 | 1
2 | US:ADHS | 1 | 23
3 | US:ADHS | 3 | 6
4 | US:AK | 1 | 9
5 | US:AK | 2 | 3
6 | US:AL | 1 | 2
7 | US:AL | 2 | 72
8 | US:AL | 3 | 153
9 | US:AL | 4 | 1
10 | US:AL | 5 | 2
11 | US:AL | 8 | 1
12 | US:AL | 9 | 1
13 | US:AL | 11 | 1
14 | US:AR | 1 | 6
15 | US:AR | 2 | 83
16 | US:AR | 3 | 326
17 | US:AR | 4 | 32
18 | US:AZ | 1 | 1
19 | US:AZ | 2 | 32
20 | US:AZ | 3 | 39
21 | US:AZ | 4 | 1
22 | US:AZ | 8 | 1
23 | US:AZ:CR | 3 | 1
24 | US:AZ:Spur | 3 | 1
25 | US:AZ:Yavapai | 2 | 2
26 | US:Auto Trail | 2 | 1
27 | US:BIA | 1 | 1
28 | US:CA | 1 | 6
29 | US:CA | 2 | 72
30 | US:CA | 3 | 128
31 | US:CA | 5 | 1
32 | US:CA | 8 | 3
33 | US:CA | 11 | 1
34 | US:CA | 12 | 1
[...]

Issue #135, opened 2015-08-17
Add country codes on roads


nvkelso:
Sometimes we need to localize road colors by country.

nvkelso:
Pulling this into v1.5.0 milestone so we can clean up road shields and networks.

Let's have a look at the country polygons that Valhalla uses.

nvkelso:
We pull this in for intermediate processing and then strip them on final export. Verified in dev.

PR #1452, opened 2017-12-06
Allow some kinds of non-numeric shield_text

zerebubuth:
This normalises the formatting of road `ref`s which consist of a single letter, optional space and then a number. These are often displayed with the letter as part of the signage.

Connects to #1062.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -3932,6 +3932,7 @@ def _guess_type_from_network(network):
         # given, so this should default to roads.
         return 'road'
 
+
 # a mapping of operator tag values to the networks that they are (probably)
 # part of. this would be better specified directly on the data, but sometimes
 # it's just not available.
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 # transformation functions to apply to features
 
 from collections import defaultdict, namedtuple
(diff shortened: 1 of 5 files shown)


vectordatasource/transform.py
```diff
@@ -3943,15 +3996,29 @@ def merge_networks_from_tags(shape, props, fid, zoom):
     ref = props.get('ref')
     mz_networks = props.get('mz_networks', [])
 
+    # if there's no network, but the operator indicates a network, then we can
+    # back-fill an approximate network tag from the operator. this can mean
+    # that extra refs are available for road networks.
+    if network is None:
+        operator = props.get('operator')
+        backfill_network = _NETWORK_OPERATORS.get(operator)
```
nvkelso: This may need to get slightly more sophisticated so account for multiple road shield types per country also based on the A, N, M, & etc shield text prefixes &/or the road class?

nvkelso:
Besides my comment in https://github.com/tilezen/vector-datasource/pull/1452/files#r155352236, I think this is good to go. But I think that's already covered in https://github.com/tilezen/vector-datasource/issues/1389.

So LGTM.

vectordatasource/transform.py
zerebubuth: I think this needs to get a _lot_ more sophisticated!

The addition of the `operator`-based fall-back was only because the M1 section that #1062 mentioned had a `ref` but no `network` on the way, and was only part of the E13 relation (which has both `ref` and `network`). The current logic is built around tuples of `(route_type, network, ref)`, so the simplest approach was to guess a `network` based on the information available on the way.

This is a kludge, but it's an improvement over what went before.

Once we have #135, then we should change the logic so that we're additionally able to match on the country and road class. For example, a road in the UK with `highway=motorway` would be generally expected to have a `ref` starting with `M` (apart from [the ones which don't](https://en.wikipedia.org/wiki/A329(M)_motorway), which have been left scattered around like rakes in long grass by the trickster spirits to frustrate us), and could be assigned a `network` more accurately, where one wasn't otherwise specified.

Issue #1408, opened 2017-10-18
Vector Tiles: "building:material" property


msmollin:
From @nrdegroot: For 3D applications, it would be nice if Mapzen Vector Tiles passed through the "building:material" tag. Currently it only provides the "building:roof_material" tag.

nvkelso:
Seems reasonable. Scheduled for v1.6

nvkelso:
Verified in "prod" Nextzen tiles, closing.

PR #1455, opened 2017-12-13
Add building material tag to output.

zerebubuth:
Connects to #1408.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -210,6 +210,7 @@ Values for `kind_detail`  are sourced from OpenStreetMap's `building` tag for bu
 * `addr_housenumber`: value from OpenStreetMap's `addr:housenumber` tag
 * `addr_street`: value from OpenStreetMap's `addr:street` tag
 * `area`: in square meters (spherical Mercator, no real-world), `polygon` features only. _See planned bug fix in [#1095](https://github.com/tilezen/vector-datasource/issues/1095)._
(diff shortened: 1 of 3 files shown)


docs/layers.md
```diff
@@ -210,6 +210,7 @@ Values for `kind_detail`  are sourced from OpenStreetMap's `building` tag for bu
 * `addr_housenumber`: value from OpenStreetMap's `addr:housenumber` tag
 * `addr_street`: value from OpenStreetMap's `addr:street` tag
 * `area`: in square meters (spherical Mercator, no real-world), `polygon` features only. _See planned bug fix in [#1095](https://github.com/tilezen/vector-datasource/issues/1095)._
+* `building_material`: A description of the material covering the outside of the building or building part, if the information is available. Common values are: `cement_block`, `brick`, `plaster`, `wood`, `concrete`, `metal`, `stone`, `traditional`, `glass`, `mud`, `masonry`, `steel`, `semi-permanent`, `tin`, `permanent`, `timber-framing`, `sandstone`, `clay` and `other`, and there are many other less common values.
```
nvkelso: Generally we alphabetize these sorts of lists.

zerebubuth: This was in order of the number of uses, descending. I thought that would be more useful to anyone who only wants to support the top N values. Would it be better to call that out, or alphabetise the list and let people pick values they think are important?

nvkelso: Yes, please sort a-z.

PR #1458, opened 2017-12-16
Add amenity=* POIs from OpenStreetMap Carto

pnorman:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

Related to #1425.

I tried to pick zoom levels and places within the file similar to similar existing features (e.g. nightclubs like bars)


docs/layers.md
```diff
@@ -678,13 +678,13 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `operator`: Who actually runs the bike share station, eg: "NYC Bike Share".
 * `ref`: The reference of this rental station, if one is available.
 
-#### POI properties (only on `kind:bicycle_parking`):
+#### POI properties (only on `kind:bicycle_parking` and `kind:motorcycle_parking`):
```
nvkelso: Should these properties apply to (car) parking as well?

nvkelso: Yes, per: https://github.com/pnorman/vector-datasource/blob/5cd439892d50b0e8a70205f1d3f3597b4515199b/yaml/pois.yaml#L1349-L1377, please add to parking POIs YAML section:

```
      access: {col: access}
      operator: {col: operator}
      capacity: {col: tags->capacity}
      covered:
        case:
          - when:
              covered: 'yes'
            then: true
          - when:
              covered: 'no'
            then: false
      fee:
        case:
          - when:
              fee: true
              not:
                fee: ['no', 'Free', 'free', '0', 'No', 'none']
            then: true
        # TODO tests expect this to be false and not omitted, is that what we want?
          - else: false
      maxstay: {col: tags->maxstay}
      surveillance:
        case:
          - when:
              surveillance: true
              not:
                surveillance: ['no', 'none']
            then: true
```

(Which is the same as the others except omitting the cyclcehire id lines, so could be made into a global.)

nvkelso: Let's follow-up in https://github.com/tilezen/vector-datasource/issues/1475.

PR #1459, opened 2017-12-18
Remove unused wooded area tags + natural=park and add leaf_type to wooded areas

pnorman:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

~~natural=forest, landuse=wood and natural=park are all mistakes with about 0.02% occurrence in the data.~~

~~This isn't a schema-breaking change because these objects aren't present except as mistakes anyways.~~

I did this change first because I had to modify all the wooded areas for the other change, leaf_type. This adds the leaf_type to kind_detail for wooded areas.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -572,6 +572,9 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 
 The value of the OpenStreetMap `wetland` tag. Common values are `bog`, `fen`, `mangrove`, `marsh`, `reedbed`, `saltmarsh`, `string_bog`, `swamp`, `tidalflat`, and `wet_meadow`.
 
+#### Wood and forest `kind_detail` values
+* The value of the OpenStreetMap `leaf_type` tag. [Common values](https://taginfo.openstreetmap.org/keys/leaf_type#values) are `broadleaved`, `needleleaved`, or `mixed`.
+
 ## Places
 
 ![image](images/mapzen-vector-tile-docs-places.png)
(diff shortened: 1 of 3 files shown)


yaml/landuse.yaml
```diff
@@ -339,12 +339,14 @@ filters:
...
       <<: *output_properties
       kind: forest
+      kind_detail: &woods_detail { col: leaf_type }
       tier: 2
   - filter: { landuse: forest }
     min_zoom: { max: [ 10, *tier2_min_zoom ] }
     output:
       <<: *output_properties
       kind: forest
+      kind_detail: *woods_detail
```
nvkelso: While this may be technically correct YAML, it's doesn't follow our convention.

Instead of declaring the YAML global on line 342 with `kind_detail: &woods_detail { col: leaf_type }` instead add to the globals section around line 134 like:

```
- &leaftype_kind_detail
    kind_detail: { col: leaf_type }
```

and then the usage down below (like 342, 349, etc) would be:

```
      <<: * leaftype_kind_detail
```

docs/layers.md
```diff
@@ -507,8 +507,6 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `military`
 * `national_park`
 * `nature_reserve`
-* `natural_forest` - _See planned bug fix in [#1096](https://github.com/tilezen/vector-datasource/issues/1096)._
```
nvkelso: Please revert these changes for `natural_forest` and `natural_park` throughout this PR, and add the leaf type kind_detail to both.

pnorman: Moved to #1469

Issue #1387, opened 2017-09-03
Need to distinguish business and spur routes from basic networks


nvkelso:
* **What did you see?**

**First Capital Dr** in St. Charles, MO receives an interstate 70 shield when it is in fact a surface street! The motorway south of there also receives an interstate 70 shield. One of these is in error.

* **What did you expect to see?**

Technically First Capital Dr is a **business route** alternate of I-70 which is somewhat common but a little odd.

We should make sure such routes get sometime like `US:I-B` instead of `US:I` network values exported. The OSM `ref` value we draw from does correctly indicate this as a business route. I've seen the same type of thing for `US:US` routes (were we should also support something like `US:US-B`).

This should probably be paired with an alternate shield shown in grayscale and maybe with BUS. added to it on the Mapzen basemap style side that's only labeled zoom 14+? (Looks like these mostly get labeled by Tangram at zoom 15+ now so not really a change.)

* **What map location are you having problems with?** 

St. Charles, MO
http://tangrams.github.io/walkabout-style/#15/38.7781/-90.4949

* **Screenshot?** 

<img width="1078" alt="screen shot 2017-09-02 at 21 13 57" src="https://user-images.githubusercontent.com/853051/30000507-c964dcee-9023-11e7-8870-9305917d19ff.png">

nvkelso:
See also: modifier (thanks @mojodna!):

- https://www.openstreetmap.org/relation/1933234

zerebubuth:
What do you want the behaviour to be if the `network` is already set to something more than just `US:??`?

For example:

* [`modifier=Business`, `network=US:I:Business:Loop`](https://www.openstreetmap.org/relation/1935116) - assume this should stay as `US:I:Business:Loop`?
* [`modifier=Spur`, `network=US:NJ:Essex`](https://www.openstreetmap.org/relation/945855) - ignore the `Spur` and leave as `US:NJ:Essex`?
* [`modifier=Spur`, `network=US:NJ:CR`](https://www.openstreetmap.org/relation/941526) - should be `US:NJ:CR:Spur`? How do we tell the difference between the `CR` and `Essex` cases?

nvkelso:
> * [`modifier=Business`, `network=US:I:Business:Loop`](https://www.openstreetmap.org/relation/1935116) - assume this should stay as `US:I:Business:Loop`?

Yes, do nothing.

[...]

PR #1460, opened 2017-12-18
Add route modifier information to network.

zerebubuth:
Add `modifier` information to `network` property. For example, a route with `modifier=Business` and `network=US:I` would result in an output network property of `US:I:Business`.

Connects to #1387.


docs/layers.md
```diff
@@ -1048,7 +1048,7 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `min_zoom`: a suggestion for which zoom to draw a feature. The value is a float.
 * `ref`: Commonly-used reference for roads, for example "I 90" for Interstate 90. To use with shields, see `network` and `shield_text`. Related, see `symbol` for pistes.
 * `all_networks` and `all_shield_texts`: All the networks of which this road is a part, and all of the shield texts. See `network` and `shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
-* `network`: eg: `US:I` for the United States Interstate network, useful for shields and road selections. This only contains _road_ network types. Please see `bicycle_network` and `walking_network` for bicycle and walking networks, respectively.
+* `network`: eg: `US:I` for the United States Interstate network, useful for shields and road selections. This only contains _road_ network types. Please see `bicycle_network` and `walking_network` for bicycle and walking networks, respectively. Note that networks may include "modifier" information, for example `US:I:Business` for a business route or `US:I:Truck` for a truck route.
```
nvkelso: Since we're whitelisting those in code instead of pass-thru, let's list the values in alphabetical order here (reorder from the SQL below):

`('Business', 'Spur', 'Truck', 'Alternate', 'Bypass', 'Connector', 'Historic', 'Toll', 'Scenic')`

nvkelso: (Noting here there isn't a good OSM Wiki page for modifier to link to, so quoting it alone suffices.)

PR #1461, opened 2017-12-18
Add wetland detail to kind_detail

pnorman:
- [x] Update tests
- [x] Update data migrations
- [x] Update docs

No migration changes needed.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -570,7 +570,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 
 ##### Wetland `kind_detail` values:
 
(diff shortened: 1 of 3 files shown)


docs/layers.md
```diff
@@ -568,6 +568,9 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `works`
 * `zoo`
 
+##### Wetland `kind_detail` values:
+
+The value of the OpenStreetMap `wetland` tag.
```
nvkelso: We should list common values here in a-z alphabetical sorting.

pnorman: ✔️

Issue #1250, opened 2017-05-16
Improve min_zoom of highway=unclassified roads


nvkelso:
* **What did you see?**

When the `min_zoom`s were setup initially there was an assumption that `highway=unclassified` were literally unclassified / worse than highway=residential at `min_zoom: 12`.

- https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L282

But in fact, the [OSM wiki states](http://wiki.openstreetmap.org/wiki/Key:highway):

```
The least most important through roads in a country's system – i.e. minor roads
of a lower classification than tertiary, but which serve a purpose other than access
to properties. Often link villages and hamlets. (The word 'unclassified' is a historical
artefact of the UK road system and does not mean that the classification is unknown; 
you can use highway=road for that.)
```

* **What did you expect to see?**

If that's the case, they should be `min_zoom: 11` like tertiary (but let's keep the `kind: minor_road`): 

- https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L226

* **What map location are you having problems with?** 

- http://tangrams.github.io/walkabout-style/#16/38.2964/-111.5420

The fish hatchery road is just as important arterial as the grey road leading south of Loa. 

* **Screenshot?** 

<img width="382" alt="screen shot 2017-05-15 at 23 15 56" src="https://cloud.githubusercontent.com/assets/853051/26092205/7b39727e-39c4-11e7-96f7-fffa81c68432.png">

<img width="371" alt="screen shot 2017-05-15 at 23 13 18" src="https://cloud.githubusercontent.com/assets/853051/26092159/32278f9e-39c4-11e7-8286-aaf354c3b8c9.png">

<img width="520" alt="screen shot 2017-05-15 at 23 13 36" src="https://cloud.githubusercontent.com/assets/853051/26092172/426120b4-39c4-11e7-9ae2-0d2d5bea2492.png">

nvkelso:
Different example with surface gravel:

- https://www.openstreetmap.org/way/315437880#map=13/37.8959/-111.6677

nvkelso:
- https://tangrams.github.io/explorer/#12.0/38.3663/-111.6309/kind_detail/unclassified

<img width="1440" alt="screen shot 2017-05-15 at 23 27 48" src="https://cloud.githubusercontent.com/assets/853051/26092524/37935fd8-39c6-11e7-8c55-c191d537654c.png">

[...]

PR #1462, opened 2017-12-21
show unclassed roads earlier

nvkelso:
Connects with #1250.

- [x] Update tests
- [x] Update data migrations
- [x] Update docs


nvkelso:
@zerebubuth tests added (and they pass!). This is ready for review.

nvkelso:
> My only concern is that this might increase the size of the tile. We'll have to check on that after the next build.

My hypothesis is the size delta will be small in urban areas and more in rural areas – but tiles are pretty small in rural areas already so there's still "budget" for this.

Issue #1251, opened 2017-05-16
Improve min_zoom of important highway=track roads


nvkelso:
* **What did you see?**

Major tracks are visible too late in forest / mountainous areas in the USA. 

- with **surface** of `paved`, `asphalt`, `concrete` at zoom 11.
- with **[tracktype](http://wiki.openstreetmap.org/wiki/Key:tracktype)** of `grade1` at zoom 11
- with **[tracktype](http://wiki.openstreetmap.org/wiki/Key:tracktype)** of `grade1` but `access=private` at zoom 12
- with **surface** of `gravel` (but not **tracktype** of `grade3`, `grade4`, or `grade5`) at zoom 12.
- with **[tracktype](http://wiki.openstreetmap.org/wiki/Key:tracktype)** of `grade2` at zoom 12

Sister issue to https://github.com/tilezen/vector-datasource/issues/1250.

Right now they don't come in until `min_zoom: 13`. 

- https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L311

* **What did you expect to see?**

Major connector 2 lane routes are often unpaved tracks in the western mountains of USA and are different quality than 1 lane or 1 lane tracks with soft dirt.

These major connectors should be visible earlier at zoom 12 (or even 11). _But if someone wanted them even earlier they should be `highway=unclassified`, see related issue._

* **What map location are you having problems with?**

Southern Utah, examples:

- grade1 / asphalt: http://www.openstreetmap.org/way/329375413
- grade1 / paved: http://www.openstreetmap.org/way/10103739
- grade1 alone: http://www.openstreetmap.org/way/10100375
- grade1 alone (private): http://www.openstreetmap.org/way/10611894
- grade1 / dirt: http://www.openstreetmap.org/way/286309045
- grade1 / unpaved: http://www.openstreetmap.org/way/10317361
- grade2: https://www.openstreetmap.org/way/330951783#map=17/38.28728/-111.58440
- gravel alone: http://www.openstreetmap.org/way/313839575
- gravel alone: http://www.openstreetmap.org/way/14351002

Counter example (should keep their existing min_zoom):

- grade5 / gravel: http://www.openstreetmap.org/way/10103047

* **Screenshot?** 

- https://tangrams.github.io/explorer/#13.0/38.3113/-111.5650/tracktype/grade2

[...]

PR #1463, opened 2017-12-21
show some highway=track earlier

nvkelso:
Connects with #1251.

- [x] Update tests
- [x] Update data migrations
- [x] Update docs


nvkelso:
@zerebubuth Can you take a look at my YAML so far to sanity check it, please?

yaml/roads.yaml
```diff
@@ -353,6 +353,36 @@ filters:
       <<: *osm_footway_properties
       kind: path
       kind_detail: pedestrian
+  - filter:
+      highway: [track]
```
zerebubuth: You don't need the `[]` around a single value.

yaml/roads.yaml
```diff
@@ -353,6 +353,36 @@ filters:
...
+      any:
+        - {surface: [paved, asphalt, concrete]}
```
zerebubuth: You don't need the `{}` around single items in a list context.

nvkelso: Fixed in 11d447c59f029103f1a6fdc40125e4771b05519e.

yaml/roads.yaml
nvkelso: Fixed in 11d447c59f029103f1a6fdc40125e4771b05519e.

nvkelso:
I should add some tests, yes :)

nvkelso:
@zerebubuth Please take another 👀 now that there are tests, data migrations, and doc updates.

integration-test/1251-early-track-roads.py
```diff
@@ -0,0 +1,89 @@
...
+             'kind_detail': 'track'})
+
+    def test_early_track_road_z12_grade1_private(self):
+
+        # private, grade1, track (since private demoted from zoom 11)
+        self.load_fixtures([
+            'https://www.openstreetmap.org/way/10611894',
+        ])
+
+        self.assert_has_feature(
```
zerebubuth: Might it also be worth checking that there is no feature at zoom 11? Just in case we make a change later on which breaks the demotion.

For example, something like:

```python
self.assert_no_matching_feature(
  11, 330, 781, 'roads',
  {'kind': 'path', 'kind_detail': 'track'})
```

(With the fixtures, because we know they're the only features in the database, we don't need to specify `id`, but if you wanted to run this against the tiles then it would be harder, as we'd probably merge away the IDs by zoom 11. If we expose `private`, then that might be worth trying to match on?)

yaml/roads.yaml
```diff
@@ -353,6 +353,36 @@ filters:
       <<: *osm_footway_properties
       kind: path
       kind_detail: pedestrian
+  - filter:
+      highway: track
+      any:
+        - {surface: [paved, asphalt, concrete]}
```
zerebubuth: I think the `{}` are not needed (but also not harmful) around this line.

yaml/roads.yaml
```diff
@@ -353,6 +353,36 @@ filters:
...
+      <<: *osm_highway_properties
+      <<: *osm_network
+      <<: *osm_footway_properties
+      kind: path
+      kind_detail: track
+  - filter:
+      highway: track
+      any:
+        - all:
+          - surface: [gravel]
```
zerebubuth: No need for `[]` around `gravel`.

PR #1465, opened 2017-12-21
add cranes (points, lines) and crane:type

### LINKED ISSUE #1417: cannot seem to find pois like Wind Turbines and Crane;
I am not able to get points of interests of type wind turbine and container gantry crane to show up on the zoom levels that I tested (15-18). Since these are rather large structures I think it would be nice to include them in vector tiles.

Example of vector tile where I would like the cranes to show up: https://tile.mapzen.com/mapzen/vector/v1/all/17/103341/65074.json which is the vector tile for this location: http://www.openstreetmap.org/edit#map=17/1.26857/103.83637

nvkelso:
Fixes #1417 to add cranes and crane:type

- [x] Update tests
- [x] Update data migrations
- [x] Update docs


nvkelso:
Needs to test both **landuse** and **pois** layers.

nvkelso:
All tests pass, @zerebubuth can you have a 👀?

data/migrations/v1.5.0-planet_osm_point.sql
```diff
@@ -1,11 +1,13 @@
 -- Add new shop types
 UPDATE planet_osm_point
   SET mz_poi_min_zoom = mz_calculate_min_zoom_pois(planet_osm_point.*)
-  WHERE mz_poi_min_zoom <> mz_calculate_min_zoom_pois(planet_osm_point.*)
+  WHERE (mz_poi_min_zoom <> mz_calculate_min_zoom_pois(planet_osm_line.*)
+    AND mz_poi_min_zoom <> mz_calculate_min_zoom_pois(planet_osm_polygon.*))
```
zerebubuth: This looks like you're comparing `mz_poi_min_zoom` on points to lines and polygons, which I think is not valid SQL. What did you want to happen here?

Issue #1403, opened 2017-10-05
barrier=wall is missing


matkoniecz:
* **What did you see?**

barrier=wall is not exported

* **What did you expect to see?**

barrier=wall data

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

It seems to be documented as missing - see https://mapzen.com/documentation/vector-tiles/layers/

> Also includes of some barrier and waterway tags: city_wall (zoom 12+), dam (zoom 12+), retaining_wall (zoom 15+), snow_fence (zoom 15+), fence (zoom 16+ only) and gate (zoom 16+ only).

For example https://www.openstreetmap.org/way/330603734#map=19/50.07639/19.93181 is not present in data

https://mapzen.com/products/maps/cinnabar/more-labels/#lat=50.07627&lng=19.93155&z=20

Note that barrier=wall is a standard barrier tag, used over 2M times - https://taginfo.openstreetmap.org/tags/barrier=wall

* **Screenshot?** _Props for animated gifs._

nvkelso:
Good find!

Suggest we modify `fence` so it considers `wall` in the same set also at zoom 16, like:

`- filter: { barrier: [fence, wall] }`

https://github.com/tilezen/vector-datasource/blob/4eb1fbd1cab6194713761470b2fdb57eb43d3c01/yaml/landuse.yaml#L1020-L1024

Thanks for providing a testing feature.

nvkelso:
Verified on dev.

PR #1466, opened 2017-12-21
Add wall lines to landuse layer

nvkelso:
Connects with #1403 to add wall lines to landuse layer

- [x] Update tests
- [x] Update data migrations
- [x] Update docs


yaml/landuse.yaml
```diff
@@ -1030,11 +1030,18 @@ filters:
     output:
       <<: *output_properties
       kind: retaining_wall
+  - filter: {barrier: wall}
+    min_zoom: 16
+    output:
+      <<: *output_properties
+      kind: wall
+      kind_detail: { col: wall }
```
zerebubuth: Is it worth whitelisting the most common values for the `wall` tag value?

nvkelso: From https://taginfo.openstreetmap.org/keys/wall#values:

wall_value | count
---------- | -----
no | 11704961
dry_stone | 92635
noise_barrier | 7787
brick | 7282
yes | 5209
stone | 2311
pise | 1829
castle_wall | 891
null | 575
jersey_barrier | 566
seawall | 492
flood_wall | 420
drystone | 357
concrete | 292
retaining_wall | 238
stone_bank | 211
gabion | 181
D | 154
wind_barrier | 130
kő | 126
alvenaria | 113
wood | 101
kőfal | 89

nvkelso: The general list summary (see table above) is already added to documentation – we're split 50/50 on when to just pass thru and document common values versus white listing. 

I can see black listing a few types like:

wall_value | count
-- | --
no | 11704961
yes | 5209
null | 575

nvkelso: I'm not seeing many combinations of `barrier=wall` and `wall=[no,yes,null]` in the wild so I'm going to merge this as-is. If the problem exists in the wild we can revisit later.

zerebubuth:
> we're split 50/50 on when to just pass thru and document common values versus white listing

As a future task, I'd like us to move towards white-listing by default, with rare, deliberate and documented exceptions for things such as `name`. I think the benefit is to cartographers and map publishers who are able to know in advance from the documentation what they have to cover (e.g: icons, `filter` clauses) to display the features on the map.

This would mean that we should regularly review the white-lists to ensure they stay up to date with the tagging, and perhaps this is something we need tooling around.

Issue #1096, opened 2016-10-03
BUG: New natural_* landuse need custom sort_key


nvkelso:
In https://github.com/tilezen/vector-datasource/pull/980 we added a few more kind values to differentiate from the landuse tags with the same values:
- `natural_wood`
- `natural_forest`
- `natural_park`

We also need to add those new kind values to the documentation, and they're getting default sort_rank values right now (so they z-stack _below_ every other landuse incorrectly).
- [x] Update layer [docs](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md)
- [x] Update sort_rank CSVs

The other "natural" features are here:

```
wetland,*,35
scrub,*,34
meadow,*,33
grass,*,32
```

So we need to add the 3 new ones, and offset the rest up (including ALL the rest, not just these 4):

```
wetland,*,38
scrub,*,37
meadow,*,36
grass,*,35
natural_wood,34
natural_forest,33
natural_park,32
```
- **What map location are you having problems with?** New York city's Prospect Park has some `natural_wood` features:
  - http://tile.dev.mapzen.com/mapzen/vector/v1/all/15/9650/12326.topojson

/cc @bcamper, thanks for finding the bug!

nvkelso:
Verified.

PR #1469, opened 2017-12-23
Remove natural=park, natural=forest, and landuse=wood

pnorman:
All three are tagging errors that occur very infrequently. The first occurs at a 0.022% rate and the latter two at a 0.017% rate in current OSM data.

- [x] Update tests
- [x] Update data migrations
- [x] Update docs


nvkelso:
This relates to https://github.com/tilezen/vector-datasource/issues/1096 but instead of setting sort_rank it removes the features entirely, which would trigger a major version increment per https://github.com/tilezen/vector-datasource/blob/master/SEMANTIC-VERSIONING.md#major-version-increments.

I'm not opposed to that – since it's a breaking change we stash them for later. Short term we should just fix the bug in #1096.

nvkelso:
Closing PR but tracking future breaking change work in https://github.com/tilezen/vector-datasource/issues/1544.

PR #1484, opened 2018-04-05
WIP: Create 1482-maritime_boundary-buffered_land.py

nvkelso:
- [ ] Update tests
- [x] ~~Update data migrations~~
- [x] ~~Update docs~~

### DIFF
--- a/integration-test/1482-maritime_boundary-buffered_land.py
+++ b/integration-test/1482-maritime_boundary-buffered_land.py
@@ -1,0 +1,44 @@
+from . import FixtureTest
+
+
+# Adds tests for OSM features (but not NE features)
+class SortKeyBoundary(FixtureTest):
+
+    def test_usa_country_boundary(self):
+        # country boundary of USA
+        self.load_fixtures([
+            'https://www.openstreetmap.org/relation/148838',
+        ], clip=self.tile_bbox(8, 39, 95, padding=0.1))
+        self.assert_has_feature(
+            8, 39, 95, "boundaries",
+            {"kind": "country", "sort_rank": 262})
+
+    def test_nevada_california_boundary(self):
+        # region boundary between Nevada - California
+        self.load_fixtures([
+            'https://www.openstreetmap.org/relation/165473',
+            'https://www.openstreetmap.org/relation/165475',
+        ], clip=self.tile_bbox(8, 42, 96, padding=0.1))
+        self.assert_has_feature(
+            8, 42, 96, "boundaries",
+            {"kind": "region", "sort_rank": 256})
+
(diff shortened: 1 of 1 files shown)


integration-test/1482-maritime_boundary-buffered_land.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Adds tests for OSM features (but not NE features)
+class MaritimeBoundary(FixtureTest):
+
+    def test_usa_canada_country_boundary(self):
+        # country boundary of USA and Canada
+        self.load_fixtures([
+            'https://www.openstreetmap.org/relation/148838',
+            'https://www.openstreetmap.org/relation/1428125',
+            'file://integration-test/fixtures/buffered_land/'
```
nvkelso: By hosting the shapefile in a `buffered_land` directory FixtureTest will set that as the database table name on import.

zerebubuth:
Following 36ad71b, we now have two test failures (rather than the errors we had before). I think these failures represent the actual regressions we have in the code for dealing with maritime boundaries, so in that sense the tests are "working". The code is the next thing we need to fix.

zerebubuth:
After c0768c9, test failures just say `AssertionError: Unknown source: tilezen.org`, which is fixed in https://github.com/tilezen/tilequeue/pull/330

nvkelso:
Circle is happy now! @zerebubuth, can you give a formal review, please?

circle.yml
```diff
@@ -33,9 +33,10 @@ deployment:
     commands:
       - pip install 'Circle-Beacon == 2.0.0'
       - alert-circle mapzen documentation master $CIRCLE_TOKEN
-      - ./scripts/update-integration-test-coordinates.sh
-      - pip install awscli
-      - aws s3 sync ~/.cache/vector-datasource/ s3://mapzen-tiles-assets/integration-test-fixtures/ --size-only --exclude="*" --include="*.geojson"
+      # TODO: either set up new buckets, or deprecate these things.
+      #- ./scripts/update-integration-test-coordinates.sh
+      #- pip install awscli
+      #- aws s3 sync ~/.cache/vector-datasource/ s3://mapzen-tiles-assets/integration-test-fixtures/ --size-only --exclude="*" --include="*.geojson"
```
nvkelso: This one exists now:

- https://s3.amazonaws.com/nextzen-tile-assets/wof/wof-neighbourhoods.pgdump

In retrospect we should probably call the bucket `tilezen-assets` instead...

rmarianski: I don't think it's a big deal to rename it. It's just a matter of updating the getting started instructions and configuration for the shapefile import/prepare processes right?

PR #1499, opened 2018-05-15
Add country code to roads

zerebubuth:
* Added admin areas curated shapefile, converted from Valhalla's `admins.sqlite`, with data originally from OSM.
* Bumped assets bundle version.
* Added support for admin areas in all the places that we need to list new tables.
* Added admin areas as a new layer, which gets dropped before the tile is written out, so doesn't alter the layers in the output tile.
* Used the admin areas layer joined to roads to add a `country_code` property to output roads.
* Used the `country_code` property to back-fill road network values.
* Added tests for `country_code` and back-filling in GB & AR.

Connects to #135.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -8,6 +8,7 @@ all:
   - pois
   - boundaries
   - transit
+  - admin_areas
 
 sources:
 
@@ -137,6 +138,11 @@ sources:
     - template: wof.jinja2
       start_zoom: 11
 
+  admin_areas:
+    - template: admin_areas.jinja2
+      # starts at the min zoom where we have OSM roads
+      start_zoom: 5
+
 layers:
   water:
     geometry_types: [Point, MultiPoint, Polygon, MultiPolygon, LineString, MultiLineString]
@@ -212,9 +218,6 @@ layers:
       - vectordatasource.transform.normalize_aerialways
       - vectordatasource.transform.normalize_cycleway
       - vectordatasource.transform.add_is_bicycle_related
-      - vectordatasource.transform.merge_networks_from_tags
-      - vectordatasource.transform.extract_network_information
-      - vectordatasource.transform.choose_most_important_network
       - vectordatasource.transform.road_trim_properties
       - vectordatasource.transform.remove_feature_id
       - vectordatasource.transform.tags_remove
@@ -291,6 +294,15 @@ layers:
       - vectordatasource.transform.remove_feature_id
       - vectordatasource.transform.truncate_min_zoom_to_2dp
     sort: vectordatasource.sort.transit
+    area-inclusion-threshold: 1
+  admin_areas:
+    geometry_types: [Polygon, MultiPolygon]
+    simplify_before_intersect: false
+    simplify_start: 9
+    transform:
+      - vectordatasource.transform.tags_create_dict
+      - vectordatasource.transform.add_id_to_properties
+      - vectordatasource.transform.remove_feature_id
     area-inclusion-threshold: 1
 post_process:
   - fn: vectordatasource.transform.numeric_min_filter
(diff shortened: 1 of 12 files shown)


vectordatasource/transform.py
```diff
@@ -4525,3 +4584,54 @@ def backfill_from_other_layer(ctx):
...
+    for idx, feature_layer in enumerate(ctx.feature_layers):
+        layer_datum = feature_layer['layer_datum']
+        layer_name = layer_datum['name']
+
+        if layer_name == layer_to_delete:
+            found_idx = idx
+            break
+
+    if found_idx is not None:
+        del ctx.feature_layers[found_idx]
```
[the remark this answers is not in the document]
zerebubuth: Ah, good point, thanks! I had thought that it wasn't allowed to modify a list while iterating over it, but I think we get away with it here since we exit the loop immediately. Fixed in 7379c6a.

queries.yaml
```diff
@@ -137,6 +138,11 @@ sources:
     - template: wof.jinja2
       start_zoom: 11
 
+  admin_areas:
+    - template: admin_areas.jinja2
+      # starts at the min zoom where we have OSM roads
```
nvkelso: `# starts at the min zoom where we have roads layer content`

Right now it's NE roads at 5+, and switches to OSM later at 9.

queries.yaml
```diff
@@ -292,6 +295,15 @@ layers:
       - vectordatasource.transform.truncate_min_zoom_to_2dp
     sort: vectordatasource.sort.transit
     area-inclusion-threshold: 1
+  admin_areas:
+    geometry_types: [Polygon, MultiPolygon]
+    simplify_before_intersect: false
+    simplify_start: 9
```
nvkelso: If we start including these features at zoom 5 we should probably also simplify starting at zoom 5? Or if we're not simplifying at all before intersect and we're dropping this layer does it matter?

queries.yaml
```diff
@@ -924,3 +951,8 @@ post_process:
         white: [255, 255, 255]
         yellow: [255, 255, 0]
         yellowgreen: [154, 205, 50]
+
+  - fn: vectordatasource.transform.drop_layer
+    params:
+      layer: admin_areas
+      start_zoom: 15
```
nvkelso: Should this be dropped for all zooms?

yaml/admin_areas.yaml
```diff
@@ -0,0 +1,9 @@
+filters:
+  - filter:
+      kind: admin_area
+      iso_code: true
+    min_zoom: 5
```
nvkelso: Do we need to keep re-stating zoom 5?

nvkelso:
@zerebubuth I added a few dangling comments above from today's PR (above).

queries.yaml
zerebubuth: I think it would be better to not simplify at all. I've removed this parameter in #1501, and made sure that it drops the `admin_areas` layer _before_ simplification happens. Just to make doubly-sure, I've set `simplify_start` beyond any zoom we'll ever render!

queries.yaml
zerebubuth: Yup, well spotted! Thanks, fixed in #1501.

yaml/admin_areas.yaml
zerebubuth: Unfortunately we do, because I'm shoe-horning the admin areas in through the same pipeline we use for features. This means the admin area features need a `min_zoom` and all the other trappings of "real" layer features, even though we're going to drop this layer and none of the features will ever appear in the output.

The alternative design would be to set up a parallel set of data structures for non-feature layers, but this seemed more invasive. I thought we might as well try the "pretend they're features" approach and see how that works - if we're happy enough with it, then it'll save us from doing a bunch more deep changes to how `tilequeue` works.

va2ron1:
shapefiles.tar.gz doesn't exist at this (20180514) datestamp in this bucket, in the previous datestamp (20180313) exist.

zerebubuth:
Hi @va2ron1, thanks very much for pointing that out!

It should have been done after this issue was merged, but I forgot. It has been fixed now. Sorry!

Issue #1482, opened 2018-04-04
Boundaries layer features tagged maritime_boundary = 1 in error


nvkelso:
* **What did you see?** Land boundaries are being tagged `maritime_boundary` = **1**, like the USA-Canada `country` boundary, and the Idaho-Washington `region` boundary.
* **What did you expect to see?** Land boundaries should be tagged `maritime_boundary` = **0**.
* **What map location are you having problems with?** https://tangrams.github.io/bubble-wrap/#8/47.874/-119.243. 
   - Bad at zoom 8+ at OSM sourced zooms. 
    - Natural Earth low zooms use different process and are fine (they are **0** and display correctly in Bubble Wrap).
* **Screenshot?** 

<img width="975" alt="screen shot 2018-04-04 at 15 10 32" src="https://user-images.githubusercontent.com/853051/38338373-2a6cb79e-381e-11e8-9aea-3ef78a6d308b.png">
<img width="1559" alt="screen shot 2018-04-04 at 15 11 00" src="https://user-images.githubusercontent.com/853051/38338374-2a90580c-381e-11e8-8c71-65ea97e86e39.png">
<img width="371" alt="screen shot 2018-04-04 at 15 12 43" src="https://user-images.githubusercontent.com/853051/38338375-2aab9374-381e-11e8-91af-387afa77aeec.png">

nvkelso:
Looking at the tile content in QGIS it seems the features are correctly in the tiles, but with the wrong property attributes (`martime_boundary` = **1** instead of **0**):

Zoom 7 in QGIS at 512 px size:
https://tile.nextzen.org/tilezen/vector/v1/512/all/7/22/44.topojson?api_key=xyz

<img width="824" alt="screen shot 2018-04-04 at 15 40 24" src="https://user-images.githubusercontent.com/853051/38338619-1ddbc5a0-381f-11e8-9ced-dcc8b6620526.png">
<img width="312" alt="screen shot 2018-04-04 at 15 40 30" src="https://user-images.githubusercontent.com/853051/38338603-0fd777b0-381f-11e8-8133-08c346922280.png">

Zoom 8 in QGIS at 512 px size:
https://tile.nextzen.org/tilezen/vector/v1/512/all/8/44/88.topojson?api_key=xyz

<img width="822" alt="screen shot 2018-04-04 at 15 40 49" src="https://user-images.githubusercontent.com/853051/38338604-0ff5db74-381f-11e8-9f53-41947a731a2f.png">
[...]

PR #1502, opened 2018-05-16
Maritime boundaries fix

zerebubuth:
This fixes buffered land for SQL queries (`zoom < 10`), taking the query from https://github.com/tilezen/raw_tiles/commit/93f0b7d1bbeacc62d2939a65be5abda520f4604d.

Added extra tests to be sure of the tags we need to pass in.

Added a small check to `generate_fixtures` to validate the type of its arguments, as I've made the mistake of passing in a `list` a few times now.

I've run a test tile through both the SQL and RAWR paths, and it seems to work:

image

Which shows the boundaries on `10/158/380` (tile from OpenStreetMap):

Tile 10/158/380 from OpenStreetMap, data available under the ODbL

Connects to #1482.

### DIFF
--- a/integration-test/1482-maritime_boundary-buffered_land.py
+++ b/integration-test/1482-maritime_boundary-buffered_land.py
@@ -51,3 +51,51 @@ class MaritimeBoundary(FixtureTest):
         self.assert_no_matching_feature(
             8, 44, 88, "boundaries",
             {"kind": "region", "maritime_boundary": 1})
+
+    # this test is to state the properties explicitly, so that we can make sure
+    # they don't get changed unintentionally.
+    def test_generative_non_maritime(self):
+        import dsl
+
+        z, x, y = (8, 44, 88)
+
+        self.generate_fixtures(
+            dsl.way(1, dsl.tile_box(z, x, y), {
+                'source': 'tilezen.org',
+                'maritime_boundary': True,
+                'min_zoom': 0,
+                'kind': 'maritime',
+            }),
+            dsl.way(2, dsl.tile_diagonal(z, x, y), {
+                'source': 'openstreetmap.org',
+                'boundary': 'administrative',
+                'admin_level': '2',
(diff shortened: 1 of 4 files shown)


queries/buffered_land.jinja2
```diff
@@ -0,0 +1,28 @@
...
+
+    '{}'::jsonb AS __boundaries_properties__
+
+FROM (
+  SELECT
+    gid,
+    -- extract only polygons. we might get linestring and point fragments when
+    -- the box and geometry touch but don't overlap. we don't want these, so
+    -- want to throw them away.
+    fiahdlfauisdhluhST_CollectionExtract({% filter geometry %}the_geom{% endfilter %}, 3) AS geom
```
rmarianski: I assume this is a frustration typo? :)

zerebubuth: D'oh! Thanks for spotting that! Removed in 85a54fb.

It wasn't frustration, more that I was testing the RAWR vs. SQL paths and wanted to make absolutely certain that the data was coming from the RAWR tile. I figured an easy way to check that was to have PostgreSQL tell me that function didn't exist. Of course, then I forgot to take it out. :man_facepalming:

integration-test/1482-maritime_boundary-buffered_land.py
```diff
@@ -51,3 +51,51 @@ def test_washington_idaho_region_boundary_not_maritime_boundary(self):
...
+            dsl.way(2, dsl.tile_diagonal(z, x, y), {
+                'source': 'openstreetmap.org',
+                'boundary': 'administrative',
+                'admin_level': '2',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'id': 2,
```
nvkelso: We don't usually test for id, but I think it's fine in this case.

Issue #1491, opened 2018-04-20
Additional shield_text cleanup


sensescape:
Followup to https://github.com/tilezen/vector-datasource/issues/1062... this will be running tally.

- [x] **Argentina**
  - [x] **Argentina** roads are generally prefixed with 2 characters and then a third [A-Z] character can prefix the number part of the ref, and should be included in the generated shield_text (but is currently not):
    - https://www.openstreetmap.org/way/85235685
        - **ref:** `RNA002`
        - **shield_text:** bug: `002`, expected `A002`
  - [x] **Argentina** provincial routes often aren't in relations with proper **network** value (expected to be `AR:provincial`), but they do have a `RP` prefix in he **ref**.
    - https://www.openstreetmap.org/way/132808888
        - **ref:** `RP21`
        - **shield_text:** bug: not exported, expected `21`
        - **network:** bug: not exported, expected `AR:provincial`
- [x] **Australia**
  - _Some roads have network, but almost none of those are prefixed with a country code (they should be)._
  - _Sometimes the network has `addr:country`=`AU` (and even `addr:state`=`QLD`), which could be used like we have `operator` country assignments now._
  - [ ] **Australia** `AU:N-highway` national highway (slowly being replaced with A-road and M-road)
  - [ ] **Australia** `AU:A-road`
  - [ ] **Australia** `AU:M-road`
      - https://www.openstreetmap.org/relation/198331
  - [ ] **Australia** `AU:B-road`
  - [ ] **Australia** `AU:C-road`
  - [ ] **Australia** `AU:N-route` for national route
  - [ ] **Australia** `AU:S-route`
      - https://www.openstreetmap.org/relation/7936070#map=11/-25.5780/152.3691
  - [ ] **Australia** `AU:Metro-road`
  - [ ] **Australia** `AU:T-drive`
      - https://www.openstreetmap.org/relation/7936092#map=12/-27.9501/153.2236
      - https://www.openstreetmap.org/way/17796959#map=17/-26.36878/152.70198
- [x] **Brazil**
  - [ ] Network values of `BR` should sort first for Brazil, then `BR:Trans-Amazonian`, than the various `BR:*` state networks.
  - [ ] **Brazil** `BR:Trans-Amazonian` Trans-Amazonian Highway
    - if( shield_text == 230) && (feature.network == 'BR'), and should be "sorted" before general BR networks (see second relation), and BR:* state networks.
      - https://www.openstreetmap.org/way/258644490
[...]

PR #1503, opened 2018-05-21
Australian road shields

zerebubuth:
Added logic to do per-country road network processing. The first case for this, other than breaking out the logic for AR, GB and US which already existed, was Australia.

This adds the concept of a `CountryNetworkLogic`, which bundles logic for a particular country together (previously it was spread over several functions). The initial version contains up to three functions:

* `backfill`: this is called to expand the `ref`tag into a list of `(network, ref)` pairs . For example; in Australia, where network relations are less commonly used than in the US, a road might have `ref=S1;T2` to indicate it's part of both the S- and T-route networks. This function is where that splitting can happen.

* `fix`: this is called to try and fix whatever problems it can with the `network` and `ref`, returning replacements. For example, in Australia it's common to give the `network=A`, when we'd prefer a "fully-qualified" network name such as `network=AU:A-road`. Also, it's possible for the `network` to be `None` if there was no network tag, but the `ref` gives the network name (e.g: `ref=A1`).

* `sort`: this is called with a `network` and `ref` and should return a numeric value where lower numeric values mean _more_ important networks. This is used to sort the roads into order to pick the "most important" one to put in the `shield_text` and `network` properties on the output element. In Australia, as in the US, it's common for roads to be part of multiple networks and we might only have space to show one shield.

Connects to #1491.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -4071,9 +4071,7 @@ def _guess_network_au(ref):
     for part in ref.split(';'):
         if not part:
             continue
-        network = _AU_NETWORK_EXPANSION.get(part[0])
-        if network:
-            part = part[1:].strip()
+        network, ref = _normalize_au_netref(None, part)
         networks.append((network, part))
     return networks
 
@@ -4148,7 +4146,7 @@ def _splitref(ref):
 
     for i in xrange(0, len(ref)):
         if not ref[i].isalpha():
-            return ref[0:i], ref[i:]
+            return ref[0:i], ref[i:].strip()
 
     # got to the end, must be all "prefix", which probably indicates it's not
     # a ref of the expected prefix-suffix form, and we should just return the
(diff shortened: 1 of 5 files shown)


vectordatasource/transform.py
```diff
@@ -4029,26 +4029,203 @@ def _guess_type_from_network(network):
...
+        sort=_sort_network_au,
+    ),
+    'GB': CountryNetworkLogic(
+        backfill=_guess_network_gb,
+    ),
+    'US': CountryNetworkLogic(
+        backfill=_do_not_backfill,
+        sort=_sort_network_us,
+    ),
+}
```
rmarianski: Great job with the comments and organizing it this way! 👍

zerebubuth: Thanks!

I was wondering whether you thought it might be better organised as a set of classes? For example:

```python
class GB(CountryRoadNetwork):
  def backfill(self, ref):
    ...
```

Or do you think the extra syntax and state-handling of classes (i.e: `self`) get in the way of understanding what's going on?

rmarianski: I prefer the way you have it now. I think it makes it easier to see what's going on, especially because you picked good names for the underlying functions. It'll make it easier to share primitives between them if we ever need to, and because it's more declarative we have more flexibility around how it's organized.

PR #1504, opened 2018-05-22
Add country-specific rules for Canada.

zerebubuth:
Adds shield and network rules for Canada. Because the information about the Trans-Canada Highway isn't stored in a relation (instead seems indicated by `nat_name`), this meant expanding the interface of `backfill` to be able to examine all the tags.

@nvkelso, it seems like in some states the TCH has no number and in other states it takes the [number of the state highway](https://en.wikipedia.org/wiki/British_Columbia_Highway_16). In this version, I've left it unnumbered (`ref=None`). Do we want to try and figure out which state it's in to apply extra logic there?

Connects to #1491.

### DIFF
--- a/integration-test/1491-canada-shields.py
+++ b/integration-test/1491-canada-shields.py
@@ -1,0 +1,120 @@
+from . import FixtureTest
+
+
+class CanadaShieldTest(FixtureTest):
+    def test_tch(self):
+        import dsl
+
+        z, x, y = (16, 18027, 23200)
+
+        self.generate_fixtures(
+            dsl.is_in('CA', z, x, y),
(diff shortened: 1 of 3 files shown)


nvkelso:
> @nvkelso, it seems like in some states the TCH has no number and in other states it takes the number of the state highway. In this version, I've left it unnumbered (shield_text=None). Do we want to try and figure out which state it's in to apply extra logic there?

Yes, the `shield_text` should come thru, please. Generally this looks to be either `1` or `16` (and seems to be a pass thru of the "local" network's ref.

zerebubuth:
OK, thanks. Fixed in d2d26d8.

PR #1506, opened 2018-05-24
Brazilian road shields.

zerebubuth:
Brazil has a bunch of different road networks in each state, plus some extra ones (see `_BR_NETWORK_EXPANSION`). I'm pretty sure this covers all of the different network types which have more than a few tens of instances in OSM.

Connects to #1491.

### DIFF
--- a/integration-test/1491-brazil-shields.py
+++ b/integration-test/1491-brazil-shields.py
@@ -93,8 +93,11 @@ class BrazilShieldTest(FixtureTest):
         )
 
         self.assert_has_feature(
-            z, x, y, 'roads',
-            {'id': 31532787, 'network': 'BR:AL'})
+            z, x, y, 'roads', {
+                'id': 31532787,
+                'shield_text': '105',
+                'network': 'BR:AL',
+            })
 
     def test_ap(self):
         import dsl
@@ -112,8 +115,11 @@ class BrazilShieldTest(FixtureTest):
         )
 
         self.assert_has_feature(
-            z, x, y, 'roads',
-            {'id': 29111026, 'network': 'BR:AP'})
+            z, x, y, 'roads', {
+                'id': 29111026,
+                'shield_text': '010',
+                'network': 'BR:AP',
+            })
 
     def test_am(self):
         import dsl
@@ -130,8 +136,11 @@ class BrazilShieldTest(FixtureTest):
         )
 
         self.assert_has_feature(
-            z, x, y, 'roads',
-            {'id': 28958366, 'network': 'BR:AM'})
+             z, x, y, 'roads', {
+                 'id': 28958366,
+                 'shield_text': '010',
+                 'network': 'BR:AM',
+             })
 
     def test_ba(self):
         import dsl
@@ -150,7 +159,7 @@ class BrazilShieldTest(FixtureTest):
 
         self.assert_has_feature(
             z, x, y, 'roads',
-            {'id': 8099519, 'network': 'BR:BA'})
+            {'id': 8099519, 'shield_text': '263', 'network': 'BR:BA'})
 
     def test_ce(self):
         import dsl
@@ -168,7 +177,7 @@ class BrazilShieldTest(FixtureTest):
 
         self.assert_has_feature(
             z, x, y, 'roads',
-            {'id': 23328809, 'network': 'BR:CE'})
+            {'id': 23328809, 'shield_text': '060', 'network': 'BR:CE'})
 
     def test_es(self):
         import dsl
@@ -193,7 +202,7 @@ class BrazilShieldTest(FixtureTest):
 
         self.assert_has_feature(
             z, x, y, 'roads',
-            {'id': 24602299, 'network': 'BR:ES'})
+            {'id': 24602299, 'shield_text': '480', 'network': 'BR:ES'})
 
     def test_go(self):
         import dsl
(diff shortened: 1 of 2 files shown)


integration-test/1491-brazil-shields.py
```diff
@@ -0,0 +1,827 @@
...
+            dsl.way(29111026, dsl.tile_diagonal(z, x, y), {
+                'source': 'openstreetmap.org', 'ref': 'AP-010',
+                'name': 'Rodovia Juscelino Kubitschek', 'highway': 'primary',
+                'surface': 'paved',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads',
+            {'id': 29111026, 'network': 'BR:AP'})
```
nvkelso: Should shield_text be `010` or `10` here?

integration-test/1491-brazil-shields.py
```diff
@@ -0,0 +1,827 @@
...
+                'name': u'Rua Lopes Trov\xe3o', 'highway': 'secondary',
+            }),
+            dsl.relation(1, {
+                'name': 'Caminho Novo', 'tourism': 'yes', 'route': 'road',
+                'source': 'openstreetmap.org', 'historic': 'yes',
+                'type': 'route',
+            }, ways=[13076835]),
+        )
+
+        # note: we don't pick up the tourist route... should we?
```
nvkelso: Okay to not pick up the tourism route for now...

integration-test/1491-brazil-shields.py
```diff
@@ -0,0 +1,827 @@
...
+            dsl.way(26757190, dsl.tile_diagonal(z, x, y), {
+                'source': 'openstreetmap.org', 'ref': 'VRS-851',
+                'highway': 'secondary', 'oneway': 'yes',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 26757190, 'shield_text': '851',
+                'network': 'BR:RS:secondary',
```
nvkelso: What's your reference for saying `BR:RS:secondary` versus just `BR:RS`? 

I do see it listed as a "secondary highway" on https://en.wikipedia.org/wiki/List_of_state_highways_in_Rio_Grande_do_Sul.

But it's not supported by other map providers &/or photos and I think "secondary" here still means state route – since we're not drilling down on any of the other "subtypes" of RS roads so I think the network should be `BR:RS`?

integration-test/1491-brazil-shields.py
```diff
@@ -0,0 +1,827 @@
...
+            # https://www.openstreetmap.org/way/4964220
+            dsl.way(4964220, dsl.tile_diagonal(z, x, y), {
+                'source': 'openstreetmap.org', 'ref': 'SPA-344/055',
+                'highway': 'primary',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads',
+            {'id': 4964220, 'shield_text': '344', 'network': 'BR:SP:access'})
```
nvkelso: I'd drop the `:access` part.

nvkelso: And I think lower numbers should sort before so `055` for the shield text?

https://en.wikipedia.org/wiki/SP-55

integration-test/1491-brazil-shields.py
```diff
@@ -0,0 +1,827 @@
...
+                'network': 'BR', 'source': 'openstreetmap.org',
+            }, ways=[31611447]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 31611447,
+                'shield_text': '466', 'network': 'BR',
+                'all_shield_texts': ['466', '466'],
+                'all_networks': ['BR', 'BR:PR:connecting']
```
nvkelso: drop `:connecting` from network.

integration-test/1491-brazil-shields.py
```diff
@@ -0,0 +1,827 @@
...
+                'maxspeed': '40', 'name': u'Avenida Jo\xe3o Vieira',
+                'source': 'openstreetmap.org', 'postal_code': '13145-754',
+                'oneway': 'yes', 'ref': 'PLN-346', 'highway': 'secondary',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 116514858,
+                'shield_text': '346', 'network': 'BR:SP:paulinia',
```
nvkelso: Should this be `BR:SP:PLN` in keeping with the data more closely?

nvkelso:
Artwork followup filed in: https://github.com/tangrams/icons/issues/28.

integration-test/1491-brazil-shields.py
zerebubuth: I added it on the basis that `VRS` was giving us more information than just `RS`, and we could always ignore it by matching on `BR:RS.*`.

There's a [list of roads in Rio Grande do Sul](https://pt.wikipedia.org/wiki/Lista_de_rodovias_do_Rio_Grande_do_Sul#Rodovias_vicinais) on the Portuguese Wikipedia, which lists the `VRS` as "Rodovias vicinais", which seems to translate to something like "secondary", "local" or "service".

I've removed it in c68d6eb, but easy to add back again if we want.

integration-test/1491-brazil-shields.py
zerebubuth: I've gone with lower numbers sort first, updated in c68d6eb.

zerebubuth:
Updated to address comments in c68d6eb.

PR #1507, opened 2018-05-24
Chinese road shields

zerebubuth:
Add road shield logic for China. This is for the `G` and `S` networks, `AsianHighway` and also an `X` type of network that I couldn't find any documentation about, but is reasonably well-used in the data.

Added a new `shield_text` function to `CountryNetworkLogic` to apply country-specific logic to what the shield text looks like. Things such as:

* does it have an alphabetic prefix, or do we strip that?
* does it have spaces, dashes or nothing between any prefix and suffix?

There's currently a whole bunch of this in `_road_shield_text`, which was using the `network` and `ref` to try and guess this. It'll be cleaner to move it into country-specific functions.

Connects to #1491.

### DIFF
--- a/integration-test/1491-china-shields.py
+++ b/integration-test/1491-china-shields.py
@@ -1,0 +1,244 @@
+from . import FixtureTest
+
+
+class ChinaShieldTest(FixtureTest):
+    def test_cn_expressways(self):
+        import dsl
+
+        z, x, y = (16, 54413, 26599)
+
+        self.generate_fixtures(
+            dsl.is_in('CN', z, x, y),
+            # https://www.openstreetmap.org/way/127553001
+            dsl.way(127553001, dsl.tile_diagonal(z, x, y), {
+                'name:en': 'Huning Expressway', 'lanes': '4',
+                'name': u'\u6caa\u5b81\u9ad8\u901f', 'surface': 'asphalt',
+                'source': 'openstreetmap.org', 'maxspeed': '120',
+                'int_ref': 'AH5', 'oneway': 'yes', 'ref': 'G42',
+                'highway': 'motorway',
+            }),
+            dsl.relation(1, {
+                'name:en': 'G42 Hurong Expressway',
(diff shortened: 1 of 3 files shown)


integration-test/1491-china-shields.py
```diff
@@ -0,0 +1,244 @@
...
+                'source': 'openstreetmap.org', 'ref': 'X032;X059',
+                'highway': 'secondary',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 137979647,
+                'shield_text': 'X032',
+                'network': 'CN:jx-roads',
```
nvkelso: What's the provenance of the `CN:jx-roads` designation?

vectordatasource/transform.py
```diff
@@ -4224,6 +4254,46 @@ def _normalize_ca_netref(network, ref):
...
+        network = 'CN:jx-roads'
+
+    elif network == 'CN-expressways':
+        network = 'CN:expressways'
+
+    elif network == 'CN-expressways-regional':
+        network = 'CN:expressways:regional'
+
+    elif network == 'JX-roads':
+        network = 'CN:jx-roads'
```
nvkelso: These look to all be "region" roads in the [Jiangxi](https://en.wikipedia.org/wiki/Jiangxi) province and https://overpass-turbo.eu/s/zbf. Should the convention be `CN:JX`?

integration-test/1491-china-shields.py
nvkelso: This comes from OSM tagging, see next comment [down](https://github.com/tilezen/vector-datasource/pull/1507#discussion_r191597625).

nvkelso:
> does it have an alphabetic prefix, or do we strip that?

We keep the G, S & etc.

vectordatasource/transform.py
zerebubuth: Yup, sounds good. Updated in 89a1881.

PR #1508, opened 2018-05-30
French road shields.

zerebubuth:
Added handling for French `A`, `N` and `D` roads, plus the `e-road` network. One interesting thing is that there are also [`RNIL` roads](https://fr.wikipedia.org/wiki/Route_nationale_d'int%C3%A9r%C3%AAt_local), which appear to be old `N` roads which have been delegated to local control (and budget!). I've kept these as designated `N`, because I did some spot checks on StreetView and they appear to be still signed that way.

Connects to #1491.

### DIFF
--- a/integration-test/1491-france-shields.py
+++ b/integration-test/1491-france-shields.py
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 from . import FixtureTest
 
 
@@ -12,7 +13,7 @@ class FrenchShieldTest(FixtureTest):
             # https://www.openstreetmap.org/way/16108247
             dsl.way(16108247, dsl.tile_diagonal(z, x, y), {
                 'horse': 'no', 'hazmat': 'no', 'bicycle': 'no',
-                'name': u'Boulevard P\xe9riph\xe9rique Int\xe9rieur',
+                'name': u'Boulevard Périphérique Intérieur',
                 'toll': 'no', 'surface': 'asphalt', 'lit': 'yes',
                 'source': 'openstreetmap.org', 'maxspeed': '70',
                 'int_ref': 'E 05', 'oneway': 'yes', 'foot': 'no', 'lanes': '3',
@@ -21,12 +22,12 @@ class FrenchShieldTest(FixtureTest):
             }),
             dsl.relation(1, {
                 'type': 'route', 'route': 'road', 'bicycle': 'no',
-                'name': u'Boulevard P\xe9riph\xe9rique de Paris',
+                'name': u'Boulevard Périphérique de Paris',
                 'source': 'openstreetmap.org',
             }, ways=[16108247]),
             dsl.relation(2, {
                 'network': 'e-road', 'ref': 'E 05', 'route': 'road',
-                'wikipedia': u'fr:Route europ\xe9enne 5',
+                'wikipedia': u'fr:Route européenne 5',
                 'source': 'openstreetmap.org', 'wikidata': 'Q693493',
                 'type': 'route', 'section': 'France (north-south)',
             }, ways=[16108247]),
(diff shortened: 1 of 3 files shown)


integration-test/1491-france-shields.py
```diff
@@ -0,0 +1,255 @@
...
+        import dsl
+
+        z, x, y = (16, 33186, 22554)
+
+        self.generate_fixtures(
+            dsl.is_in('FR', z, x, y),
+            # https://www.openstreetmap.org/way/16108247
+            dsl.way(16108247, dsl.tile_diagonal(z, x, y), {
+                'horse': 'no', 'hazmat': 'no', 'bicycle': 'no',
+                'name': u'Boulevard P\xe9riph\xe9rique Int\xe9rieur',
```
rmarianski: I may be wrong, but we normally get these as utf-8 encoded strings right? Might be more representative if these were all utf-8 as input since that's what's going to happen in practice? eg `u'Boul...'.encode('utf-8')`

(As an aside, I think it would be much better if immediately after reading the data from the input source all strings were forced to unicode, but that's a separate bit of cleanup work)

integration-test/1491-france-shields.py
```diff
@@ -0,0 +1,255 @@
...
+            dsl.way(20225575, dsl.tile_diagonal(z, x, y), {
+                'source': 'openstreetmap.org', 'ref': 'D 977bis',
+                'highway': 'secondary',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 20225575,
+                'shield_text': 'D977bis', 'network': 'FR:D-road',
```
nvkelso: Does the `bis` represent "express"? Do we need to somehow differentiate those shields from normal D-roads?

vectordatasource/transform.py
```diff
@@ -4288,6 +4299,27 @@ def _sort_network_cn(network, ref):
...
+    elif network == 'FR:A-road':
+        network_code = 0
+    elif network == 'FR:N-road':
+        network_code = 1
+    elif network == 'FR:D-road':
+        network_code = 2
+    elif network == 'FR':
+        network_code = 3
+    elif network == 'e-road':
+        network_code = 4
```
nvkelso: nit: Sometimes this is set to higher value (and lower sort order), like 99?

integration-test/1491-france-shields.py
zerebubuth: A random site [about driving in France] says that "bis" routes are "secondary" routes. Apparently, they're the opposite of express, and closer in meaning to "scenic"!

What would make sense here:

* `shield_text: D977`, `network: FR:D-road:scenic`? Or,
* `shield_text: D977bis`, `network: FR:D-road:scenic`? Or,
* something else?

integration-test/1491-france-shields.py
zerebubuth: Yup, good point. I've changed the whole file encoding in 4197bc1 - is that closer to what you were thinking?

vectordatasource/transform.py
zerebubuth: Yup, thanks! Better to be consistent. Fixed in 78622b7.

integration-test/1491-france-shields.py
rmarianski: That still keeps the input that the code sees as unicode though, and not utf-8 right? I was coming at it more that in practice the code will always see utf-8 strings and not unicode. ie as a pedant :P 

My understanding is that currently the test case is: `isinstance(name, str) == False and isinstance(name, unicode) == True`. Whereas the input tag that the transforms see when running in production will always be: `isinstance(name, str) == True`.

That make sense?

nvkelso:
Let’s leave them all just D-road network and keep the bis in shield text.

integration-test/1491-france-shields.py
zerebubuth: Ah! Sorry, yes, I think I understand now.

I think it's easier to read to keep the `unicode` in the test file (especially since 4197bc1 changed it to `u'Boulevard Périphérique Intérieur'` without all the `\xe9`s). So in 8e45cf6, I made it so that the `dsl` functions convert `unicode` to UTF-8 encoded `str`. Hopefully this will replicate the real inputs.

What do you think?

rmarianski: You could always just do: `u'Boulevard Périphérique Intérieur'.encode('utf-8')`, but what you did is probably better to ensure it's all utf-8. By the way, [we have a little function](https://github.com/tilezen/tilequeue/blob/4062c1a579bdf965a4ebc711f238d8b37de8b926/tilequeue/utils.py#L57-L74) that converts to utf-8, but it's a little heavier than what's needed here and doesn't fit exactly, so what you have already looks fine to me.

👍

PR #1509, opened 2018-05-30
German road shields

zerebubuth:
Added handling for German road shields.

In addition to the `A`, `B`, `L`, `K` and `e-road` networks, there also appear to be `St` or `S` designated roads, perhaps "Staatstrasse"? Also, Hamburg has local "Ring" designated roads.

Connects to #1491.

### DIFF
--- a/integration-test/1491-german-shields.py
+++ b/integration-test/1491-german-shields.py
@@ -284,3 +284,180 @@ class GermanShieldTest(FixtureTest):
                 'all_shield_texts': ['K40'],
                 'all_networks': ['DE:KS'],
             })
+
+    # https://de.wikipedia.org/wiki/Liste_der_Staatsstra%C3%9Fen_in_Bayern
+    def test_de_staatstrasse_bayern(self):
+        import dsl
+
+        z, x, y = (16, 34765, 22335)
+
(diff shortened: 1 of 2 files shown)


integration-test/1491-german-shields.py
```diff
@@ -0,0 +1,463 @@
...
+            dsl.relation(3, {
+                'name': 'B433', 'type': 'route', 'route': 'road',
+                'source': 'openstreetmap.org', 'ref': 'B 433',
+            }, ways=[4329503]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 4329503,
+                'shield_text': 'B433', 'network': 'DE:BS',
```
nvkelso: The ref order makes the ring road seem more important than the general B road. Please confirm and if so switch the sort order?

zerebubuth: It doesn't seem to be consistent. For example, in the screenshot below it flips between `Ring 2; B5` and `B 5; Ring 2`.

image

We can do it either way around - which would you prefer?

nvkelso:
Ring last (so it’s in all list but not default), please.

PR #1510, opened 2018-05-30
Mexican road shields.

zerebubuth:
Added handling of Mexican road shields.

In addition to national `MEX` roads, there are road networks in most states (although I couldn't find examples in Edomex or Tlaxcala), plus interior and exterior ring roads in Mexico City (`CDMX`). Sometimes these are in relations, but mostly not.

Connects to #1491.

### DIFF
--- a/integration-test/1491-mexican-shields.py
+++ b/integration-test/1491-mexican-shields.py
@@ -1,0 +1,1074 @@
+from . import FixtureTest
+
+
+class MexicanShieldText(FixtureTest):
+    def test_mx_mex_prefix(self):
+        import dsl
+
+        z, x, y = (16, 13978, 28930)
+
+        self.generate_fixtures(
+            dsl.is_in("MX", z, x, y),
+            # https://www.openstreetmap.org/way/285597347
+            dsl.way(
+                285597347,
+                dsl.tile_diagonal(z, x, y),
+                {
+                    "lanes": "2",
+                    "name": "Guadalajara - Zapotlanejo",
+                    "oneway": "yes",
+                    "source": "openstreetmap.org",
+                    "toll": "yes",
+                    "ref": "MEX 15D",
+                    "highway": "motorway",
+                },
+            ),
+        )
+
+        self.assert_has_feature(
+            z, x, y, "roads", {
+                "id": 285597347,
+                "shield_text": "15D",
+                "network": "MX:MEX",
+            },
+        )
+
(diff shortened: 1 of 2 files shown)


integration-test/1491-mexican-shields.py
```diff
@@ -0,0 +1,1074 @@
...
+                    "highway": "motorway",
+                },
+            ),
+        )
+
+        self.assert_has_feature(
+            z, x, y, "roads", {
+                "id": 285597347,
+                "shield_text": "15D",
+                "network": "MX:MEX",
```
nvkelso: Should this be `MX:MX` to follow 2-char convention elsewhere?

nvkelso: Ah, ISO uses 3-char for subdivision codes for Mexican states so using `MEX` here would be more consisent with that. 👍

integration-test/1491-mexican-shields.py
```diff
@@ -0,0 +1,1074 @@
...
+                    "ref": "Q. ROO 5",
+                    "highway": "secondary",
+                },
+            ),
+        )
+
+        self.assert_has_feature(
+            z, x, y, "roads", {
+                "id": 399152055,
+                "shield_text": "5",
```
nvkelso: That's an epic transform!

integration-test/1491-mexican-shields.py
```diff
@@ -0,0 +1,1074 @@
...
+                'alt_name': 'Circuito Bicentenario', 'oneway': 'yes',
+                'foot': 'no', 'ref': 'INT', 'highway': 'motorway',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 123899735,
+                'shield_text': type(None),
+                'network': 'MX:CMX:INT',
```
nvkelso: Are there any of these that are just `MX:CMX` to test for?

zerebubuth: I couldn't find any. Mexico City seems to have a bunch of different `ref`s, e.g: [3 NTE](https://www.openstreetmap.org/way/332816679), [Eje Central](https://www.openstreetmap.org/way/24706165), [Periférico](https://www.openstreetmap.org/way/182927016). Unfortunately, I wasn't able to find any more information about them on the internet, such as whether they're any kind of official designation of road network. For the moment, I've left them out - please follow up with an issue if they ought to be `MX:CMX` networks.

PR #1511, opened 2018-05-30
Japanese road shields.

zerebubuth:
Added handling for Japanese road shields.

The main bit of extra logic here comes from when there's a way with a `ref`, which is in a relation that has `network` but no `ref`. We now detect that case and handle it specially by merging the network information from the relation onto the way. (The convention in most other places seems to be to have `ref` on both the way and relation, which we can also handle.)

Connects to #1491.

### DIFF
--- a/integration-test/1491-japanese-shields.py
+++ b/integration-test/1491-japanese-shields.py
@@ -56,7 +56,7 @@ class JapaneseShieldTest(FixtureTest):
                 'shield_text': '305', 'network': 'JP:prefectural',
             })
 
-    def test_jp_e_expressways(self):
+    def test_jp_e_expressway(self):
         import dsl
 
         z, x, y = (16, 57350, 26039)
@@ -98,9 +98,9 @@ class JapaneseShieldTest(FixtureTest):
         self.assert_has_feature(
             z, x, y, 'roads', {
                 'id': 18576866,
-                'shield_text': 'E28', 'network': 'JP:expressways'})
+                'shield_text': 'E28', 'network': 'JP:expressway'})
 
-    def test_jp_ea_expressways(self):
+    def test_jp_ea_expressway(self):
         import dsl
 
         z, x, y = (16, 58512, 24062)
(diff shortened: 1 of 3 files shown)


integration-test/1491-japanese-shields.py
```diff
@@ -0,0 +1,202 @@
...
+                u'\u81ea\u52d5\u8eca\u9053', 'ref': 'E28', 'route': 'road',
+                'name:ja': u'\u795e\u6238\u6de1\u8def\u9cf4\u9580\u81ea'
+                u'\u52d5\u8eca\u9053',
+            }, ways=[18576866]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 18576866,
+                'shield_text': 'E28', 'network': 'JP:expressways'})
```
nvkelso: Flagging that this network value is plural while others for Japan are singular, but consistent with `CN:expressways` so okay.

zerebubuth: Do you think we should make both `JP:expressways` and `CN:expressways` singular instead?

nvkelso:
Expressways is plural in the data, right? But we’re already changing it from - to : so yeah, more Tilezen consistency is better to singular please.

vectordatasource/transform.py
```diff
@@ -4515,6 +4559,20 @@ def merge_networks_from_tags(shape, props, fid, zoom):
...
+        # then use the network from the relation instead.
+        if network is None:
+            solo_networks_from_relations = []
+            for i in xrange(0, len(mz_networks), 3):
+                t, n, r = mz_networks[i:i+3]
+                if t == 'road' and n and r is None:
+                    solo_networks_from_relations.append((n, i))
+            if len(solo_networks_from_relations) == 1:
+                network, i = solo_networks_from_relations[0]
+                del mz_networks[i:i+3]
```
rmarianski: Is this right? Seems that we would need to remember the index `i`, otherwise we could have skipped past it and just removed the last one? I'm assuming that the intent is to remove the solo network from `mz_networks`, but I could just be misunderstanding.

zerebubuth: The code only does the `del` if there's only one match. We won't be doing a second deletion, so `i` doesn't need to be adjusted to account for prior deletions.

Does that sound right?

rmarianski: Oh I'm sorry, you are remembering the index in the `solo_networks_from_relations` variable. I somehow missed where you saved it, *and* where you pulled it out. Carry on! :)

zerebubuth: Added a better comment in 3a56e2a.

PR #1512, opened 2018-06-01
Greek road shields

zerebubuth:
Added support for Greek road shields.

The provincial network in Greece sometimes has "ΕΠ" `ref`s, but there are [some indications that these aren't supposed to be used as `ref`s but instead `reg_ref` (regional reference)](https://wiki.openstreetmap.org/wiki/WikiProject_Greece/Provincial_Road_Network). I couldn't find anywhere that they were signed as "ΕΠ" roads, so I've dropped those where we see them.

Refactored the test harness a little so that the test rules are applied inside lists. Previously it would have just checked for strict equality (`==`), but I think it's more readable and consistent to follow the same processing on the elements of the list.

Connects to #1491.

### DIFF
--- a/integration-test/1491-greek-road-shields.py
+++ b/integration-test/1491-greek-road-shields.py
@@ -1,0 +1,112 @@
+# -*- encoding: utf-8 -*-
+from . import FixtureTest
+
+
+class GreekShieldTest(FixtureTest):
+    def test_8_grnational(self):
+        import dsl
+
+        z, x, y = (16, 37084, 25282)
+
+        self.generate_fixtures(
+            dsl.is_in('GR', z, x, y),
+            # https://www.openstreetmap.org/way/299496935
+            dsl.way(299496935, dsl.tile_diagonal(z, x, y), {
+                'highway': u'primary',
+                'int_name': u'Athinon',
+                'lit': u'yes',
+                'maxspeed': u'50',
+                'name': u'Αθηνών',
+                'nat_ref': u'ΕΟ8α',
+                'oneway': u'yes',
+                'ref': u'ΕΟ8',
+                'source': u'openstreetmap.org',
+                'source:ref': u'ΦΕΚ Β 319/23.07.1963',
+                'wikipedia': u'en:Athinon Avenue',
+            }),
+            dsl.relation(1, {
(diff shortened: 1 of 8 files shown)


vectordatasource/transform.py
```diff
@@ -4597,6 +4631,24 @@ def _normalize_de_netref(network, ref):
     return network, ref
 
 
+def _normalize_gr_netref(network, ref):
+    ref = _make_unicode_or_none(ref)
+
+    prefix, ref = _splitref(ref)
+    if prefix in (u'ΕΟ', 'EO'):
```
rmarianski: If we know that ref is unicode, then doesn't `prefix` also have to be unicode here?

zerebubuth: Good point! Although it seems to work OK without it, I've fixed them both to be unicode in df0b524, so hopefully it's clearer that `unicode` is being compared with `unicode`.

When are we moving to Python3 again? :wink:

rmarianski: I'm glad you put that comment in about the characters looking the same, because it initially looked like a redundant comparison to me. 👍 

> When are we moving to Python3 again? 😉

When python4 comes out :trollface:

integration-test/1491-greek-road-shields.py
```diff
@@ -0,0 +1,408 @@
...
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[19832532]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 19832532,
+                'network': u'GR:national',
+                'shield_text': u'8α',
```
nvkelso: Just curious: what does the `α` signify?

zerebubuth: As far as I can tell, [the EO8a was built as a grade-separated road paralleling route of the EO8](https://el.wikipedia.org/wiki/%CE%95%CE%B8%CE%BD%CE%B9%CE%BA%CE%AE_%CE%9F%CE%B4%CF%8C%CF%82_8%CE%B1_(%CE%95%CE%BB%CE%BB%CE%AC%CE%B4%CE%B1)), but at some point most of it was renamed to A8 and there are only a few stretches keeping the name EO8a (presumably which didn't make sense as part of the A-network).

PR #1514, opened 2018-06-01
Added United Kingdom road shields.

zerebubuth:
The difference between green and white signage for A-roads is that green signs are [on a "primary route"](https://en.wikipedia.org/wiki/Road_signs_in_the_United_Kingdom#Colours) and white ones are not. "Primary routes" are defined as being between certain sets of named places in the UK, so it would be super-complicated to figure that out. Good for us that the [OSM tagging guidelines for the UK](https://wiki.openstreetmap.org/wiki/United_Kingdom_Tagging_Guidelines#UK_roads) basically encode the colour of the sign in the `highway` tag (which we pass through as `kind_detail`).

Other than that, it was all relatively straightforward!

Connects to #1491.

### DIFF
--- a/integration-test/1491-british-road-shields.py
+++ b/integration-test/1491-british-road-shields.py
@@ -1,0 +1,250 @@
+# -*- coding: utf-8 -*-
+from . import FixtureTest
+
+
+class BritishRoadShields(FixtureTest):
+    def test_m25_gbmroad(self):
+        import dsl
+
+        z, x, y = (16, 32685, 21846)
+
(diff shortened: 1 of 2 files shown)


zerebubuth:
> A late May banking holiday present!

Looks like I got that wrong! Apparently it's called the ["Spring bank holiday"](https://www.gov.uk/bank-holidays). Can't wait to see what's in store for the also imaginatively named "Summer bank holiday". :rofl:

PR #1515, opened 2018-06-04
Various mostly generic network fixes

zerebubuth:
The main things attempted in this are:

1. Upper case the country code where the network is `xx:yyy`, i.e: `XX:yyy`, and
2. Replace dash with colon where the network is `XX-yyy`, i.e: `XX:yyy`.

Additional things that were covered along the way:

* Many networks seem to be of the form `XX-roads` or `XX:Roads`, but generally we prefer the lower case and singular `XX:road`.
* Special handling for roads in Riksvei, Fylkesvei and Ring-roads in Norway
* Handling regional roads in Peru
* Handling expressways and motorways in Poland.

Connects to #1491.

### DIFF
--- a/integration-test/1491-generic-network-fixes.py
+++ b/integration-test/1491-generic-network-fixes.py
@@ -1404,10 +1404,15 @@ class GenericNetworkTest(FixtureTest):
             }, ways=[261067663]),
         )
 
+        # TODO: sort e-road or AH first/last?
         self.assert_has_feature(
             z, x, y, 'roads', {
                 'id': 261067663,
-                'network': u'KZ:national',
+                'shield_text': 'M-32',
+                'network': 'KZ:national',
+                'all_shield_texts': ['M-32', 'E38', 'AH61', 'AH62', 'E123'],
+                'all_networks': ['KZ:national', 'e-road', 'AsianHighway',
+                                 'AsianHighway', 'e-road'],
             })
 
     def test_ltnational(self):
--- a/integration-test/1491-generic-network-fixes.py
+++ b/integration-test/1491-generic-network-fixes.py
@@ -1583,33 +1583,36 @@ class GenericNetworkTest(FixtureTest):
                 'network': u'LV:regional',
             })
 
-    def test_mklocal(self):
+    def test_r29278_mknational(self):
         import dsl
 
-        z, x, y = (16, 36653, 24587)
+        z, x, y = (16, 36578, 24359)
 
         self.generate_fixtures(
             dsl.is_in('MK', z, x, y),
-            # https://www.openstreetmap.org/way/219834131
-            dsl.way(219834131, dsl.tile_diagonal(z, x, y), {
-                'highway': u'tertiary',
-                'name': u'Велушина - Острец',
-                'ref': u'04-01',
+            # https://www.openstreetmap.org/way/193364172
+            dsl.way(193364172, dsl.tile_diagonal(z, x, y), {
+                'highway': u'secondary',
+                'ref': u'R29278',
                 'source': u'openstreetmap.org',
             }),
             dsl.relation(1, {
-                'network': u'mk:local',
-                'ref': u'04-01',
+                'from': u'Negotino',
+                'name': u'R29278',
+                'network': u'mk:national',
+                'ref': u'R29278',
                 'route': u'road',
                 'source': u'openstreetmap.org',
+                'to': u'A2',
                 'type': u'route',
-            }, ways=[219834131]),
+            }, ways=[193364172]),
         )
 
         self.assert_has_feature(
             z, x, y, 'roads', {
-                'id': 219834131,
-                'network': u'MK:local',
+                'id': 193364172,
+                'network': u'MK:national',
+                'shield_text': u'R29278',
             })
 
     def test_mknational(self):
--- a/integration-test/1491-generic-network-fixes.py
+++ b/integration-test/1491-generic-network-fixes.py
@@ -1065,6 +1065,12 @@ class GenericNetworkTest(FixtureTest):
             })
 
     def test_irfreeways(self):
+        # it looks like roads in Iran are signed in both Arabic and Latin
+        # scripts for the numbers, and that the Asian Highway is signed
+        # something like "A1Tr", although the "Tr" is in a little box.
+        #
+        # https://en.wikipedia.org/wiki/Freeway_2_(Iran)
+
         import dsl
 
         z, x, y = (16, 41219, 25277)
@@ -1127,6 +1133,9 @@ class GenericNetworkTest(FixtureTest):
             z, x, y, 'roads', {
                 'id': 549630346,
                 'network': u'IR:freeway',
+                'shield_text': '2',
+                'all_networks': ['IR:freeway', 'AsianHighway'],
+                'all_shield_texts': ['2', 'A1'],
             })
 
     def test_irnational(self):
@@ -1191,7 +1200,10 @@ class GenericNetworkTest(FixtureTest):
         self.assert_has_feature(
             z, x, y, 'roads', {
                 'id': 35693445,
-                'network': u'IR:national',
+                'network': 'IR:national',
+                'shield_text': '44',
+                'all_networks': ['IR:national', 'AsianHighway'],
+                'all_shield_texts': ['44', 'A1'],
             })
 
     def test_japrefectural(self):
--- a/integration-test/1491-generic-network-fixes.py
+++ b/integration-test/1491-generic-network-fixes.py
@@ -3877,7 +3877,7 @@ class GenericNetworkTest(FixtureTest):
         z, x, y = (16, 51383, 28940)
 
         self.generate_fixtures(
-            dsl.is_in('LO', z, x, y),
+            dsl.is_in('LA', z, x, y),
             # https://www.openstreetmap.org/way/477601864
             dsl.way(477601864, dsl.tile_diagonal(z, x, y), {
                 'highway': u'primary',
@@ -3911,7 +3911,7 @@ class GenericNetworkTest(FixtureTest):
         self.assert_has_feature(
             z, x, y, 'roads', {
                 'id': 477601864,
-                'network': u'LO:network',
+                'network': u'LA:national',
             })
 
     def test_mzroads(self):
--- a/integration-test/1491-generic-network-fixes.py
+++ b/integration-test/1491-generic-network-fixes.py
@@ -1,0 +1,3501 @@
+# -*- encoding: utf-8 -*-
+from . import FixtureTest
+
+
+class GenericNetworkTest(FixtureTest):
+    def test_3_myfederal(self):
+        import dsl
+
+        z, x, y = (16, 51656, 32491)
+
+        self.generate_fixtures(
+            dsl.is_in('MY', z, x, y),
+            # https://www.openstreetmap.org/way/201941041
+            dsl.way(201941041, dsl.tile_diagonal(z, x, y), {
+                'alt_name': u'Johor Bahru-Kota Tinggi Highway',
+                'alt_name:ms': u'Lebuhraya Johor Bahru-Kota Tinggi',
+                'highway': u'trunk',
+                'name': u'Tebrau Highway',
+                'name:ms': u'Jalan Tebrau',
+                'oneway': u'yes',
+                'ref': u'3',
+                'source': u'openstreetmap.org',
+            }),
+            dsl.relation(1, {
+                'distance': u'739.06',
+                'network': u'my:federal',
+                'ref': u'3',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+                'wikipedia': u'en:Malaysia Federal Route 3',
+            }, ways=[201941041]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 201941041,
+                'network': u'MY:federal',
+                'shield_text': u'3',
+            })
+
+    def test_baautoceste(self):
+        import dsl
+
+        z, x, y = (16, 35912, 23608)
+
+        self.generate_fixtures(
+            dsl.is_in('BA', z, x, y),
+            # https://www.openstreetmap.org/way/88590170
+            dsl.way(88590170, dsl.tile_diagonal(z, x, y), {
+                'highway': u'motorway',
+                'int_ref': u'E 661',
+                'lanes': u'2',
+                'maxspeed': u'100',
+                'oneway': u'yes',
+                'ref': u'E-661',
+                'source': u'openstreetmap.org',
+            }),
+            dsl.relation(1, {
+                'description:fr': u'E 661 Balatonkeresztúr - Zenica',
+                'name': u'European route E 661',
+                'name:fr': u'Route européenne E 661',
+                'name:ru': u'Европейский маршрут E 661',
+                'network': u'e-road',
+                'ref': u'E 661',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+                'wikidata': u'Q676174',
+                'wikipedia': u'en:European route E661',
+            }, ways=[88590170]),
+            dsl.relation(2, {
+                'name': u'Autoput E-661',
+                'network': u'ba:Autoceste',
+                'ref': u'E-661',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[88590170]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 88590170,
+                'network': u'BA:Autoceste',
+            })
+
+    def test_banational(self):
+        import dsl
+
+        z, x, y = (16, 35920, 23988)
+
+        self.generate_fixtures(
+            dsl.is_in('BA', z, x, y),
+            # https://www.openstreetmap.org/way/114668830
+            dsl.way(114668830, dsl.tile_diagonal(z, x, y), {
+                'description': u'Gorica - Drinovci - Klobuk',
+                'highway': u'tertiary',
+                'ref': u'R 854;R-421',
+                'source': u'openstreetmap.org',
+                'surface': u'asphalt',
+            }),
+            dsl.relation(1, {
+                'network': u'ba:national',
+                'ref': u'R-421',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[114668830]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 114668830,
+                'network': u'BA:national',
+            })
+
+    def test_bgmunicipal(self):
+        import dsl
+
+        z, x, y = (16, 37348, 24090)
+
+        self.generate_fixtures(
+            dsl.is_in('BG', z, x, y),
+            # https://www.openstreetmap.org/way/291607259
+            dsl.way(291607259, dsl.tile_diagonal(z, x, y), {
+                'highway': u'tertiary',
+                'ref': u'GAB1005',
+                'source': u'openstreetmap.org',
+            }),
+            dsl.relation(1, {
+                'network': u'bg:municipal',
+                'note': u'See: http://wiki.openstreetmap.org/wiki/WikiProject_Bulgaria/Road_network',
+                'ref': u'GAB1005',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[291607259]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 291607259,
+                'network': u'BG:municipal',
+            })
+
+    def test_bgnational(self):
+        import dsl
+
+        z, x, y = (16, 36930, 23831)
+
+        self.generate_fixtures(
+            dsl.is_in('BG', z, x, y),
+            # https://www.openstreetmap.org/way/236539372
+            dsl.way(236539372, dsl.tile_diagonal(z, x, y), {
+                'highway': u'secondary',
+                'int_name': u'Shiroka',
+                'name': u'Широка',
+                'name:en': u'Shiroka',
+                'oneway': u'yes',
+                'ref': u'12',
+                'source': u'openstreetmap.org',
+                'surface': u'asphalt',
+            }),
+            dsl.relation(1, {
+                'description': u'О.п. Видин — Иново — Брегово — граница Югославия',
+                'name': u'Второкласен път 12',
+                'network': u'bg:national',
+                'note': u'see: http://wiki.openstreetmap.org/wiki/WikiProject_Bulgaria/Road_network',
+                'ref': u'12',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[236539372]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 236539372,
+                'network': u'BG:national',
+            })
+
+    def test_bgregional(self):
+        import dsl
+
+        z, x, y = (16, 37868, 23962)
+
+        self.generate_fixtures(
+            dsl.is_in('BG', z, x, y),
+            # https://www.openstreetmap.org/way/141204241
+            dsl.way(141204241, dsl.tile_diagonal(z, x, y), {
+                'highway': u'residential',
+                'source': u'openstreetmap.org',
+            }),
+            dsl.relation(1, {
+                'name': u'29608',
+                'network': u'bg:regional',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[141204241]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 141204241,
+                'network': u'BG:regional',
+            })
+
+    def test_bonational(self):
+        import dsl
+
+        z, x, y = (16, 20374, 35802)
+
+        self.generate_fixtures(
+            dsl.is_in('BO', z, x, y),
+            # https://www.openstreetmap.org/way/333920841
+            dsl.way(333920841, dsl.tile_diagonal(z, x, y), {
+                'highway': u'trunk',
+                'lanes': u'2',
+                'maxspeed': u'80',
+                'name': u'RN3: La Paz-Cotapata-Caranavi',
+                'oneway': u'no',
+                'ref': u'F3',
+                'source': u'openstreetmap.org',
+                'surface': u'asphalt',
+            }),
+            dsl.relation(1, {
+                'from': u'Calajahuira',
+                'name': u'Ruta Nacional 3',
+                'network': u'bo:national',
+                'operator': u'ABC',
+                'ref': u'F3',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'to': u'Trinidad',
+                'type': u'route',
+                'wikipedia': u'es:Ruta 3 (Bolivia)',
+            }, ways=[333920841]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 333920841,
+                'network': u'BO:national',
+            })
+
+    def test_bsnational(self):
+        import dsl
+
+        z, x, y = (16, 35968, 24047)
+
+        self.generate_fixtures(
+            dsl.is_in('BS', z, x, y),
+            # https://www.openstreetmap.org/way/279974519
+            dsl.way(279974519, dsl.tile_diagonal(z, x, y), {
+                'description': u'Mostar - Čitluk - Ljubuški - Bijača',
+                'highway': u'primary',
+                'ref': u'M 120;R-423',
+                'source': u'openstreetmap.org',
+                'surface': u'asphalt',
+            }),
+            dsl.relation(1, {
+                'network': u'bs:national',
+                'ref': u'R-423',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[279974519]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 279974519,
+                'network': u'BS:national',
+            })
+
+    def test_bynational(self):
+        import dsl
+
+        z, x, y = (16, 37059, 21611)
+
+        self.generate_fixtures(
+            dsl.is_in('BY', z, x, y),
+            # https://www.openstreetmap.org/way/282927316
+            dsl.way(282927316, dsl.tile_diagonal(z, x, y), {
+                'bicycle': u'no',
+                'bus': u'no',
+                'foot': u'no',
+                'hgv': u'designated',
+                'highway': u'trunk',
+                'horse': u'no',
+                'maxaxleload': u'11.5',
+                'oneway': u'yes',
+                'ref': u'М1',
+                'source': u'openstreetmap.org',
+                'surface': u'asphalt',
+            }),
+            dsl.relation(1, {
+                'distance': u'611 km',
+                'int_name': u'Brest (Kazlovičy) — Minsk — miaža Rasijskaj Federacyi (Redźki)',
+                'loc_name': u'Олимпийка',
+                'name': u'Брест (Козловичи) — Минск — граница Российской Федерации (Редьки)',
+                'name:be': u'Брэст (Казловічы) — Мінск — мяжа Расійскай Федэрацыі (Рэдзькі)',
+                'name:lt': u'Brestas — Minskas — Rusijos Federacijos siena',
+                'name:pl': u'Brześć (Kozłowicze) — Mińsk — granica Federacji Rosyjskiej',
+                'network': u'by:national',
+                'ref': u'М1',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+                'wikidata': u'Q267850',
+                'wikipedia': u'ru:Магистраль М1 (Белоруссия)',
+            }, ways=[282927316]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 282927316,
+                'network': u'BY:national',
+            })
+
+    def test_byregional(self):
+        import dsl
+
+        z, x, y = (16, 37586, 20847)
+
+        self.generate_fixtures(
+            dsl.is_in('BY', z, x, y),
+            # https://www.openstreetmap.org/way/198119075
+            dsl.way(198119075, dsl.tile_diagonal(z, x, y), {
+                'highway': u'tertiary',
+                'maxaxleload': u'10',
+                'official_name': u'Ворняны — Гервяты — Жодишки — Р95',
+                'ref': u'Н6009',
+                'source': u'openstreetmap.org',
+                'surface': u'unpaved',
+            }),
+            dsl.relation(1, {
+                'name': u'Ворняны — Гервяты — Жодишки — Р95',
+                'network': u'by:regional',
+                'ref': u'Н6009',
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[198119075]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 198119075,
+                'network': u'BY:regional',
+            })
+
+    def test_cdrrig(self):
+        # CD:RRIG appears to be the regional route network, with the national
+        # network as CD:RN (route nationale).
+        import dsl
+
+        z, x, y = (16, 38053, 32520)
+
+        self.generate_fixtures(
+            dsl.is_in('CD', z, x, y),
+            # https://www.openstreetmap.org/way/423970668
+            dsl.way(423970668, dsl.tile_diagonal(z, x, y), {
+                'highway': u'secondary',
+                'oneway': u'yes',
+                'ref': u'RP430',
+                'source': u'openstreetmap.org',
+                'surface': u'compacted',
+            }),
+            dsl.relation(1, {
+                'name': u'Régionale prioritaire 430',
+                'network': u'cd:rrig',
+                'ref': u'RP430',
+                'route': u'road',
+                'rrig:id': u'243; 195',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[423970668]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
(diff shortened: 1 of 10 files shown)


vectordatasource/transform.py
```diff
@@ -5062,8 +5342,16 @@ def merge_networks_from_tags(shape, props, fid, zoom):
...
-            # at least we know what country it is in.
-            network = country_code
+            # at least we know what country it is in. but don't add if there's
+            # an entry in mz_networks with the same ref!
+            if ref:
+                found = False
+                for i in xrange(0, len(mz_networks), 3):
+                    t, _, r = mz_networks[i:i+3]
+                    if t == 'road' and r == ref:
+                        found = True
```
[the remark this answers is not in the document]
zerebubuth: Yes, thanks, I'd love one :wink:

Fixed in 28c0600.

At least I won't make so many of these mistakes now that Github's new owners are adding Clippy support. "Hi, it looks like you're trying to write some Python! ..." :paperclip:

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[26686903]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 26686903,
+                'network': u'CH:Nationalstrasse',
```
nvkelso: kinda odd that one is `CH:national` and the other is `CH:Nationalstrasse` but data is king?

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[388323188]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 388323188,
+                'network': u'CO:national:old',
```
nvkelso: okay...

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[25144610]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 25144610,
+                'network': u'GR:provincial:52',
```
nvkelso: We omitted the province code in France in the network value, should we here as well? Else please provide list of province numbers so the client-side code knows what to deal with as a look up instead of a regEx.

nvkelso: Also, is the `52` part of the ref actually used in the wild on road signage or is it just `1` or `01`?

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+
+        z, x, y = (16, 36967, 24661)
+
+        self.generate_fixtures(
+            dsl.is_in('GR', z, x, y),
+            # https://www.openstreetmap.org/way/14429261
+            dsl.way(14429261, dsl.tile_diagonal(z, x, y), {
+                'highway': u'secondary',
+                'name': u'Άγιος Βασίλειος - Χορτιάτης',
+                'reg_ref': u'ΕΠ20',
```
nvkelso: It this supposed to end up in the network list via your other earlier Greek work?

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'type': u'route',
+                'wikidata': u'Q4812011',
+                'wikipedia': u'en:Freeway 2 (Iran)',
+            }, ways=[549630346]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 549630346,
+                'network': u'IR:freeway',
```
nvkelso: Should we also be testing AH transform here?

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'type': u'route',
+                'wikidata': u'Q289474',
+                'wikipedia': u'en:European route E38',
+            }, ways=[261067663]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 261067663,
+                'network': u'KZ:national',
```
nvkelso: a test of the `shield_text` would be useful here since `M-1` ref input should either be `1` or `M1` on export.

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[219834131]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 219834131,
+                'network': u'MK:local',
```
nvkelso: That's a crazy ref of `04-01`, what should the **shield_text** be on export? Does the 04 signify which "local" area it's in? Should it be modified on export?

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'type': u'route',
+                'wikidata': u'Q6367615',
+                'wikipedia': u'no:Fylkesvei 555',
+            }, ways=[106337203]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 106337203,
+                'network': u'NO:Fylkesvei',
```
nvkelso: What's our convention on initial cap on `AA:bb` bb portion? Is it always data driven or do we force it up or down?

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'type': u'route',
+                'wikidata': u'Q6396950',
+                'wikipedia': u'no:Riksvei 15',
+            }, ways=[521803129]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 521803129,
+                'network': u'NO:Riksvei',
```
nvkelso: Does the shield text need a transform and test?

nvkelso: (The next test example seems similar but just has ref value of 7 excluding the Rv prefix. Let's be consistent.)

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[460910438]),
+            dsl.relation(2, {
+                'int_ref': u'AH26',
+                'int_ref:colour': u'blue',
+                'name': u'National Highway',
+                'name:en': u'National Highway',
+
```
nvkelso: Why is this upper `NHN` when source data was lowercase `nhn`? See earlier comments about consistency of `CC:rr` rr values...

nvkelso: Singular, nice!

nvkelso: Please add shield text test – we generally only have 1, 2, 3, 4, and 5-char artwork so this would not render unless the space were taken out of the ref, and possibly drop the `EN`? Needs research.

nvkelso: `:centre`? Okay...

nvkelso: omg, Russian refs can be 8 and 9 characters long!? But is the `-` shown on signs in the wild or should we drop that?

nvkelso: Is this shield_text of `M2` correctly (versus ref of M-2)?

nvkelso: lordy, that's an insane ref value!

nvkelso: which shield_text exports? I think it would be the shorter P382 but what do the locals say? Looking at the way information it seems like that's ref_old so the longer one should export? Eeek!

nvkelso: See earlier comments about `CC:rr:??` and what is ZHT? A county code? Unless these are each signed differently I think we should strip that off. (region level is manageable in client-side lookup, but a 1000 locality codes is not.)

nvkelso: Also noting here that Ukraine shield artwork will need the extra character widths like we added in Russia. /cc @sensescape.

nvkelso: If each territory (admin-1 region?) has a definition we need to provide client-side with a list for lookup, please.

nvkelso: What is the shield_text? `T-10-09`, `T 1009`, or `T1009` or ??? Shorter is generally better.

nvkelso: is this a territorial abbreviation? See earlier `UA:territorial:*` comment.

nvkelso: What is the shield_text? This one begins with 0 so different pattern than T?

nvkelso: Are M prefixed refs generally national `M-road` and this one just happens to fall in Capetown? Are the Capetown motorways signed different than other admin-1 regions in South Africa?

nvkelso: There's a lot going on in this one! Where will the shield_text_all endup for `SADC 231` (assuming that's an all not the default network info)?

nvkelso: Is this the weird `ΕΟ` in `ΕΟ57` that tripped us up last PR? What's the shield_text here?

nvkelso: So this is a admin-1 level road in the Brazilian state of BA?

nvkelso: What should the shield text be, `BA-099` or `99`?

nvkelso: Should the shield_text be `188`?

nvkelso: Please test that the shield_text is `N2` not the raw ref value of `N2 (Cameroon)`.

nvkelso: Should the shield_text be `RN2` or `2`?

nvkelso: Should the network be `IN:SH` here instead of `IN:road? I think a state highway should win over a district road.

nvkelso: why not `IQ:highway`? Or data driven so nothing more to see here?

nvkelso: that `LO:network` is an odd value but data driven so nothing more to see here?

nvkelso: What's the shield_text supposed to be? `O-30` or `O30` or `30`?

nvkelso: Please test shield_text of `D555` (remove space)?

nvkelso: Is this an admin-1 region code? If so please specify all admin-1 region codes so we can add that as a lookup client-side instead of regEx?

integration-test/1491-norwegian-shields.py
```diff
@@ -0,0 +1,135 @@
...
+                'route': u'road',
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[71218142]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 71218142,
+                'network': u'NO:oslo:ring',
```
nvkelso: Do all Norway admin-1 regions get region names like this? If so we need to enumerate them for client-side perf as a comment here.

integration-test/1491-norwegian-shields.py
```diff
@@ -0,0 +1,135 @@
...
+                'type': u'route',
+                'wikidata': u'Q1770921',
+                'wikipedia': u'no:Fylkesvei 51',
+            }, ways=[26119489]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 26119489,
+                'network': u'NO:Fylkesvei',
```
nvkelso: An asside: this https://en.wikipedia.org/wiki/Norwegian_County_Road_55 looks like an epic drive!

nvkelso: `Fylkesvei` here means "county level road". Per comments elsewhere, should this be lowercase?

integration-test/1491-norwegian-shields.py
```diff
@@ -0,0 +1,135 @@
...
+                'type': u'route',
+                'wikidata': u'Q6367270',
+                'wikipedia': u'no:Fylkesvei 270 (Hordaland)',
+            }, ways=[3648469]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 3648469,
+                'network': u'NO:Fylkesvei',
```
nvkelso: Thank you for removing ` (Hordaland)` for just `NO:Fylkesvei`!

integration-test/1491-peruvian-shields.py
```diff
@@ -0,0 +1,135 @@
...
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[414535595]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 414535595,
+                'shield_text': '104',
+                'network': u'PE:AY',
```
nvkelso: Please comment here all Peru admin-1 region codes so those can be moved to client-side lookup versus regEx.

integration-test/1491-peruvian-shields.py
```diff
@@ -0,0 +1,135 @@
...
+                'source': u'openstreetmap.org',
+                'type': u'route',
+            }, ways=[68875825]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 68875825,
+                'shield_text': '30A',
+                'network': u'PE:PE',
```
nvkelso: What is the shield_text supposed to read? `30A`?

integration-test/1491-peruvian-shields.py
```diff
@@ -0,0 +1,135 @@
...
+                'type': u'route',
+            }, ways=[262205335]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 262205335,
+                'shield_text': '103',
+                'network': 'PE:AM',
+                'all_networks': ['PE:AM'],
```
nvkelso: Why isn't `PE:municipal`  in list of all_networks?

vectordatasource/transform.py
```diff
@@ -4556,6 +4617,25 @@ def _sort_network_mx(network, ref):
     return network_code * 10000 + min(ref, 9999)
 
 
+def _sort_network_no(network, ref):
+    if network is None:
+        network_code = 9999
+    elif network == 'NO:oslo:ring':
+        network_code = 0
+    elif network == 'e-road':
```
nvkelso: Nice

vectordatasource/transform.py
```diff
@@ -4577,6 +4657,25 @@ def _sort_network_gb(network, ref):
     return network_code * 10000 + min(ref, 9999)
 
 
+def _sort_network_pl(network, ref):
+    if network is None:
+        network_code = 9999
+    elif network == 'PL:motorway':
+        network_code = 0
+    elif network == 'PL:expressway':
+        network_code = 1
```
nvkelso: motorway sorts after expressway in Poland? I must be getting my English translations scrambled ;)

vectordatasource/transform.py
```diff
@@ -4879,6 +4997,107 @@ def _normalize_kr_netref(network, ref):
...
+        network = 'NO:Riksvei'
+        ref = number
+
+    elif prefix == 'Fv':
+        network = 'NO:Fylkesvei'
+        ref = number
+
+    elif prefix == 'E':
+        network = 'e-road'
+        ref = 'E ' + number.lstrip('0')
```
nvkelso: Do we strip out the space after the E somewhere else? (The space shouldn't be present by the time it exports in shield_text.)

vectordatasource/transform.py
```diff
@@ -4879,6 +4997,107 @@ def _normalize_kr_netref(network, ref):
...
+
+    elif prefix == 'Ring':
+        network = 'NO:oslo:ring'
+        ref = 'Ring ' + number
```
nvkelso: I'm dubious about this one for ref `Ring #` as shield_text...

integration-test/1491-generic-network-fixes.py
nvkelso: Should this be shield_text `H-03`, `H03`, or `H3`?

nvkelso: Should this be shield_text `P-17`, `P 17`, or `P17`? Shorter is better unless it's actually signed that way.

nvkelso: I also saw evidence of `TR:DY` network during research last year – what's the right network value here?

integration-test/1491-generic-network-fixes.py
zerebubuth: If we think they're the same thing, we can normalise them to the same network. I haven't (yet) looked in depth at Switzerland, so I'm not sure what their road network hierarchy looks like.

Do you think we should normalise them?

integration-test/1491-generic-network-fixes.py
zerebubuth: I wasn't sure what to do with this one... Leave it as-is, knowing it won't match anything? Or remove it? Probably shouldn't normalise it to `CO:national`...

integration-test/1491-generic-network-fixes.py
zerebubuth: According to [this OSM wiki page](https://wiki.openstreetmap.org/wiki/WikiProject_Greece/Provincial_Road_Network) there aren't any provincial signs. I did a few spot-checks, and wasn't able to find any, so I can't verify whether "5201" or "ΕΠ01" would be more appropriate to show.

integration-test/1491-generic-network-fixes.py
zerebubuth: It does (via the relation). The "ΕΠ20" by itself doesn't contain enough information, but the relation has the "full `ref`" of "5420". As with the comment above, it's not clear that this is signed anywhere, so might not be useful to put a shield on.

zerebubuth:
With this PR, I didn't dig much into the country details, so a lot of the raw data is showing through. Most of the tests were only that the country code was being transformed correctly, so are lacking tests for `shield_text`. We could fix them here, or follow up in another PR?

Regarding India, Malaysia, Singapore and others; I have follow-up PRs ready for India and Malaysia, currently working on Spain. But they're developed on top of this PR, so it would be easier to get this one merged first!

integration-test/1491-generic-network-fixes.py
zerebubuth: Yup. Added test in 87d1555. Looks like they (might?) show the Asian Highway slightly differently in Iran, so I added a transform for that.

integration-test/1491-generic-network-fixes.py
zerebubuth: Removed the province code in dc9ca6b.

integration-test/1491-generic-network-fixes.py
```diff
@@ -0,0 +1,4231 @@
...
+                'type': u'route',
+                'wikidata': u'Q11589265',
+                'wikipedia': u'ja:神奈川県道23号原宿六ツ浦線',
+            }, ways=[386400216]),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 386400216,
+                'network': u'JP:prefectural:kanagawa',
```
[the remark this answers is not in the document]
zerebubuth: Dropped in b321d53.

integration-test/1491-generic-network-fixes.py
zerebubuth: Added test in 4cb814a.

Went down a bit of a rabbit hole here! Looks like road networks in Kazakhstan are fairly complicated. Firstly, it looks like a ref like `M1` would be the way to go [according to this page](https://en.wikipedia.org/wiki/Roads_in_Kazakhstan), but that's contradicted by [the (original?) page in Russian which has `M-1`](https://ru.wikipedia.org/wiki/%D0%90%D0%B2%D1%82%D0%BE%D0%BC%D0%BE%D0%B1%D0%B8%D0%BB%D1%8C%D0%BD%D1%8B%D0%B5_%D0%B4%D0%BE%D1%80%D0%BE%D0%B3%D0%B8_%D0%9A%D0%B0%D0%B7%D0%B0%D1%85%D1%81%D1%82%D0%B0%D0%BD%D0%B0).

For what it's worth, the `ref`s are pretty consistently `M-1` (and `A-1`, `P-1`), so I think that's how they're intended to be seen. I tried, but wasn't able to find a picture of a road sign in Kazakhstan which showed how it's displayed there.

Finally, there appear to be two network types, `KZ:national` and `KZ:regional`. I was expecting them to map to `M`, `A` and `P`, but instead:

* `KZ:national` maps to  `M`, `A` and `P`,
* `KZ:regional` maps to `KB`, `KC`, `KH`, `KP`, `KR` and `КХHS`.

The [Russian page about Kazakh roads](https://ru.wikipedia.org/wiki/%D0%90%D0%B2%D1%82%D0%BE%D0%BC%D0%BE%D0%B1%D0%B8%D0%BB%D1%8C%D0%BD%D1%8B%D0%B5_%D0%B4%D0%BE%D1%80%D0%BE%D0%B3%D0%B8_%D0%9A%D0%B0%D0%B7%D0%B0%D1%85%D1%81%D1%82%D0%B0%D0%BD%D0%B0) says:

> "K" - local value; Subsequent letters in the index of highways of local significance are determined by the belonging of the road to administrative-territorial units.

So perhaps the network value isn't much use for shields? Would we prefer to replace with `KZ:M-road`, `KZ:A-road`, then `KZ-national` for the P-roads and `KZ:regional` for the K-roads?

integration-test/1491-generic-network-fixes.py
zerebubuth: Yup, that appears to be the only `NN-NN` ref in FYROM... So I'm not sure if it's a mistake or not. I can't find any references to it elsewhere, but since it's only a local road it might not get signed.

I've swapped it out for another example in 150f72a, since all I wanted to test was the transform of `mk:something` to `MK:something`.

integration-test/1491-generic-network-fixes.py
zerebubuth: I've mostly kept it close to what's in the data, unless there seemed to be a reason to change it. We can easily force it to lower case, if that's what you'd prefer?

integration-test/1491-generic-network-fixes.py
zerebubuth: Added tests in 436d1bb. Worth pointing out there's more extensive tests of Norway's roads in [`integration-test/1491-norwegian-shields.py`](https://github.com/tilezen/vector-datasource/pull/1515/files/28c06004411a57c4fabc29241d7315d81e182fc6#diff-2dcb4bdaeb11a50e6c101facf8f9ad66).

integration-test/1491-generic-network-fixes.py
zerebubuth: It's listed [on the OSM wiki](https://wiki.openstreetmap.org/wiki/Relation:route#Road_routes) as being an acronym for "National Highway Network". However, a deeper dive suggests it's not a useful piece of information, since almost all `network` values in the Philippines are `ph:nhn`.

Would it be more useful to categorise as Wikipedia does, with [highways](https://en.wikipedia.org/wiki/Philippine_highway_network) and [expressways](https://en.wikipedia.org/wiki/Philippine_expressway_network), i.e?

* `E#`: Expressways, `PH:expressway`,
* `N#` & `N##`: Primary national roads, `PH:primary`,
* `N###`: Secondary national roads, `PH:secondary`.

It looks like there are national tertiary roads, but the Wikipedia article suggests these aren't generally numbered, so wouldn't have shields.

zerebubuth: Looks like the Asian Highway shield merges with the main shield! That's cool, but looks like it might be a pain to implement :smile: 

Philippine N1 and AH26 merged

zerebubuth: Looks like that doesn't get signposted as `EN 230`, but as `N230`. I added that to the tests in 1603dc5 and did a deeper dive into the Portuguese road network in 46c3a02.

Basically, there are [several different network/shield types](https://en.wikipedia.org/wiki/Roads_in_Portugal#Current_classification_according_PRN_2000):

* `A` autoestradas, similar to motorways, so I've synthesized the network `PT:motorway`,
* `IP` itinerários principais, similar to primary routes, so I've synthesized the network `PT:primary`,
* `IC` itinerários complementares, similar to secondary routes, so I've synthesized the network `PT:secondary`,
* `EN` Estradas nacionais, national roads, for which I've reused the network `PT:national` (note that it's signed with just an `N`),
* `ER` Estradas regionais, regional roads, for which I've reused the network `PT:regional` (note that it's signed with just an `R`),
* `EM` Estradas municipais, municipal roads, so I've synthesized the network `PT:municipal` (note that these don't appear to be signed at all - I could not find an example),
* `VR` Via Rapida, seems to be a motorway class road only on Madeira, so I've synthesized the network `PT:rapid`,
* `VE` Via Expresso, seems to be a primary class road only on Madeira, so I've synthesized the network `PT:express`.

zerebubuth: Yup, good point. Removed that in 46c3a02.

zerebubuth: As far as I can tell, yes. Looks like the shield should be without the dash. Fixed in 4771ca8.

zerebubuth: I've had a look around, and I can't find any actual road sign showing a ref like that. Provisionally, I'd say let's just not output `shield_text` for those kinds of `ref`s. Tested in 2dccb26.

zerebubuth: Added a test in 2dccb26 - looks like the K type roads don't have shields. At least, as far as I can tell.

zerebubuth: I've changed it so that it prefers the `P382` from the relation, and doesn't output shield text for K-type roads, which don't seem to get signed anyway.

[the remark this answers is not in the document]
zerebubuth: [Looks like](https://uk.wikipedia.org/wiki/%D0%90%D0%B2%D1%82%D0%BE%D0%BC%D0%BE%D0%B1%D1%96%D0%BB%D1%8C%D0%BD%D1%96_%D1%88%D0%BB%D1%8F%D1%85%D0%B8_%D0%A3%D0%BA%D1%80%D0%B0%D1%97%D0%BD%D0%B8#/media/File:E40_Road_Sign.JPG) the shields are like `M05`. Updated in 33089ea.

zerebubuth: I've removed the long refs, they are minor roads and don't appear to be signed. I haven't been able to find an example. The longest are the territorial, which are `Txxyy`, so I think you only have to deal with 5 characters.

zerebubuth: As far as I can tell, there's no difference in the shield from territory to territory, so I've dropped that last component in 33089ea.

zerebubuth: Dropped the last component of the network, I mean. The `T` refs appear to concatenate across the `-`. That change also in 33089ea.

zerebubuth: `T1009`, it seems. Fixed in 33089ea.

integration-test/1491-generic-network-fixes.py
nvkelso: I'd leave as-is.

It'll render as a generic shield which I think in this case is fine. This occasionally happens in the USA as well when the "old" route stays signed for a few years while people get used to the "new" route (which should receive the proper artwork).

integration-test/1491-generic-network-fixes.py
nvkelso: We should generally speaking have "motorway" and "other national roads" network values Switzerland. 

Looks like in Switzerland the ways are in networks, but the network value is often just `motorway` for the RED ones ([example](https://www.openstreetmap.org/edit?way=271750328#map=17/46.97784/7.46862)) – while I expect `CH:motorway` to be exported.

- RED Number plate for motorway and highway (national and cantonal)
    - List of examples: https://de.wikipedia.org/wiki/Liste_der_Autobahnen_in_der_Schweiz
- BLUE Number plate Hauptstrasse (national, cantonal and municipal)
    - List of examples: https://de.wikipedia.org/wiki/Liste_der_Hauptstrassen_der_Schweiz

From the [occurrence counts on TagInfo](https://taginfo.openstreetmap.org/keys/network#values) it seems like `CH:Nationalstrasse` should be replaced with `CH:national`.

See also:

- https://de.wikipedia.org/wiki/Liste_der_Nationalstrassen_der_Schweiz
- https://de.wikipedia.org/wiki/Hauptstrasse
- https://de.wikipedia.org/wiki/Strassensystem_in_der_Schweiz_und_in_Liechtenstein

integration-test/1491-generic-network-fixes.py
nvkelso: 🤷‍♂️ moving on...

nvkelso:
Sure. I'm fine merging now... But we do need to fix-up Switzerland and address the Poland comment during this sprint. The rest are nice to haves / we can revisit if they come up again.

Great work!

integration-test/1491-generic-network-fixes.py
zerebubuth: Removed the extra Ukrainian territorial suffixes in 33089ea.

zerebubuth: The `O` (and `C`) networks seem to be local, minor roads. I've taken these out, since I haven't been able to find any evidence that they're displayed as shields physically or on maps.

zerebubuth: Although the network was given as `za:capetown`, the [numbering system of South Africa](https://en.wikipedia.org/wiki/Numbered_routes_in_South_Africa) calls these "metropolitan". Apparently, they're present in a lot of cities. Since we're using this for shields, it doesn't matter what city they're in (shields seem to be the same), so I've mapped it to `ZA:metropolitan`. That, and other, ZA fixes in d7cb3aa.

zerebubuth: I couldn't find any info about what the SADC network is. I wasn't able to find any shields for it on a bunch of road signs that I looked at. I have a feeling it's an international network similar to `e-road` or `AsianHighway`, but I couldn't find any documentation to back that up.

For the moment, I've removed it in d7cb3aa.

zerebubuth: The tests for the Latin EO versus Greek Epsilon-Omicron are in a separate [Greek shields test](https://github.com/tilezen/vector-datasource/blob/c6ace1195c9225d2318239e9496b3bc55581baa6/integration-test/1491-greek-road-shields.py). This one was only about the upper-casing of `gr:national` to `GR:national`. Added shield text test in 0b6dad7.

zerebubuth: Brazilian shields are a little confusing to me. Some of them show without leading zeros ([pic](http://storage8.openstreetcam.org/files/photo/2018/5/9/lth/1178257_75d77_5af26139e468e.jpg) from [OpenStreetCam](http://openstreetcam.org/details/1178257/1972/track-info)) and [some with the leading zero and the state code](https://www.t1noticias.com.br/estado/estado-doa-trecho-da-to-080-ao-governo-federal-para-duplicacao/50826/). Presumably there are other combinations that appear as well.

Since the [state code appears above the number](https://pt.wikipedia.org/wiki/Sistema_rodovi%C3%A1rio_do_estado_de_S%C3%A3o_Paulo#Identifica%C3%A7%C3%A3o), I thought we'd probably want the `shield_text` to be just the number, without the state-dash prefix. What do you think?

For the moment, I've left the leading zero on, but it's easy to take off if we prefer it without?

Added shield text tests in a258b93.

zerebubuth: Added shield text tests in a258b93.

zerebubuth: Verified in 583e6f2.

vectordatasource/transform.py
```diff
@@ -4471,6 +4471,19 @@ def _guess_network_ua(tags):
     return networks
 
 
+def _guess_network_za(tags):
+    ref = tags.get('ref', '')
+    networks = []
+
+    for part in re.split('[;/,]', ref):
```
rmarianski: Think it's worth compiling regular expressions in advance? Or will it just not matter here?

integration-test/1491-generic-network-fixes.py
zerebubuth: Good catch, thanks!

You're right, it looks like the [national road network in Gabon](https://en.wikipedia.org/wiki/Transport_in_Gabon#Road_transport) is usually just `N`. It looks like it's "N" for "National", so I changed the network to `GA:national`. Also, I found an `L` network which I've added as `GA:L-road`.

Fixed in 1a95b15.

zerebubuth: Yup, fixing that was part of the separate Indian shields branch, but we're so far ahead now that I cherry-picked it in as 9b32a4c.

zerebubuth: Data driven. I dug a bit deeper into the Iraqi road network, and it seems that the data is a bit of a mess. I've added #1516 to come back to it on a rainy day.

zerebubuth: Yup, thanks for pointing that out! The ISO 3166-1 country code for Laos is actually `LA`, not `LO`. Fixed in b3ea982.

zerebubuth: Looks like it should be `O30`, although [there's a dot between the `O` and the number in the signage](https://en.wikipedia.org/wiki/Otoyol_1), so `O.30` could be a possibility as well.

Following up in #1517.

zerebubuth: `DY` would seem to be an abbreviation of [Devlet yolu](https://en.wikipedia.org/wiki/Turkish_State_Highway_System), which means "State road". However, I can't find any relations with that `network` value at the moment. The ones I see are:

* `TR/35-road`
* `TR-motorway`
* `TR-road`
* `TR-roads`

However, most roads aren't part of any relation. But all the information we need seems to be in the `ref` anyway. Perhaps the `TR:DY` ones got fixed up in some way since last year?

Follow up in #1517.

zerebubuth: That looks like a provincial road network ([tỉnh lộ](https://en.wikipedia.org/wiki/Transport_in_Vietnam#Road_transport)).

The `network` values on relations are:

* `VN:national`
* `VN-roads`
* `VN-TL`

Although there's very few of them. Looks like we'd have better luck using the `ref` to backfill.

Follow up in #1518.

integration-test/1491-norwegian-shields.py
zerebubuth: No, it's only the [Oslo ring road](https://en.wikipedia.org/wiki/Ring_3_(Oslo)).

The full set of `network` values output in Norway should be:

* `NO:Riksvei`
* `NO:Fylkesvei`
* `NO:oslo:ring`
* `e-road`

integration-test/1491-norwegian-shields.py
zerebubuth: Fixed in ec7b47c.

integration-test/1491-peruvian-shields.py
zerebubuth: All state codes listed here: https://github.com/tilezen/vector-datasource/blob/ec7b47c91b69728631778ffe258651ebdb723df2/vectordatasource/transform.py#L5371-L5396

integration-test/1491-peruvian-shields.py
zerebubuth: [The shield](https://es.wikipedia.org/wiki/Ruta_nacional_PE-30) looks a bit like a US Interstate one, with just the number (and possibly suffix letter). I think `30A` is the right `shield_text`.

integration-test/1491-peruvian-shields.py
zerebubuth: Although the `network=pe:municipal`, the `ref` starts with with `AM-` which indicates it's a departmental route in Amazonas.

vectordatasource/transform.py
zerebubuth: Indeed! I thought that was confusing too, but [motorways are a more important class of road in Poland than expressways](https://en.wikipedia.org/wiki/Highways_in_Poland#Definitions_of_highways_according_to_Polish_law). I think a closer translation into US terms, although not very close, would be that Polish motorways are like US Interstates (or British motorways), and Polish expressways are like US State highways (or British A-roads).

vectordatasource/transform.py
zerebubuth: Yes, when we use `_splitref`, we get back a `ref` that has been stripped of all leading and trailing white space: https://github.com/tilezen/vector-datasource/blob/ec7b47c91b69728631778ffe258651ebdb723df2/vectordatasource/transform.py#L4955

vectordatasource/transform.py
zerebubuth: It appears to be shown on highway signage as ["Ring 3" in black on a white background inside a rounded lozenge](https://upload.wikimedia.org/wikipedia/commons/b/b3/Ring_3_and_E6_Oslo_Quevaal.jpg), like this:

Ring 3 shield

integration-test/1491-generic-network-fixes.py
zerebubuth: `H03` seems to be the way it's displayed in the shield [on Wikipedia](https://uk.wikipedia.org/wiki/%D0%90%D0%B2%D1%82%D0%BE%D1%88%D0%BB%D1%8F%D1%85_%D0%9D_03). Although I haven't yet found an example of one on an actual road sign.

zerebubuth: [On Wikipedia](https://uk.wikipedia.org/wiki/%D0%90%D0%B2%D1%82%D0%BE%D1%88%D0%BB%D1%8F%D1%85_%D0%A0_17) it looks like there might be a space in there, but we generally seem to take spaces out, so I did the same here for consistency. I think most people would understand what's meant, right?

PR #1523, opened 2018-06-07
Malaysian road shields.

zerebubuth:
This was sitting on the shelf while we dealt with #1515. I've rebased it to current master.

Connects to #1491.

### DIFF
--- a/integration-test/1491-malaysian-road-shields.py
+++ b/integration-test/1491-malaysian-road-shields.py
@@ -1,0 +1,509 @@
+# -*- encoding: utf-8 -*-
+from . import FixtureTest
+
+
+class MalaysianShieldTest(FixtureTest):
+    def test_54_myfederal(self):
+        import dsl
+
+        z, x, y = (16, 51239, 32174)
+
+        self.generate_fixtures(
+            dsl.is_in('MY', z, x, y),
+            # https://www.openstreetmap.org/way/258246542
+            dsl.way(258246542, dsl.tile_diagonal(z, x, y), {
+                'bicycle': u'yes',
+                'foot': u'yes',
+                'highway': u'trunk',
+                'lanes': u'2',
+                'maxspeed': u'90',
+                'motor_vehicle': u'yes',
+                'name': u'Jalan Kepong - Kuala Selangor',
+                'oneway': u'yes',
+                'ref': u'FT54',
+                'source': u'openstreetmap.org',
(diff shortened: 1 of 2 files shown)


integration-test/1491-malaysian-road-shields.py
```diff
@@ -0,0 +1,509 @@
...
+        # TODO! should the shield text be '1' or '1-11B1'? not sure from
+        # available sources whether the '-11B1' part is important and/or
+        # usually on signs.
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 68058997,
+                'network': 'MY:federal',
+                'shield_text': '1',
+                'all_networks': ['MY:federal', 'AsianHighway'],
+                'all_shield_texts': ['1', 'AH150'],
```
nvkelso: `AH150` > `150`?

integration-test/1491-malaysian-road-shields.py
```diff
@@ -0,0 +1,509 @@
...
+                'ref': u'MBSA 15',
+                'source': u'openstreetmap.org',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 24701880,
+                'network': u'MY:SGR:municipal',
+                'shield_text': u'MBSA 15',
```
nvkelso: Drop the space in the `shield_text`?

nvkelso: Also, this is more than 5 char total so either needs special handling client side or more server transform. Probably use 15 instead?

nvkelso: From IRL conversation: `BSA15`

integration-test/1491-malaysian-road-shields.py
nvkelso: @nvkelso and @sensescape to research...

nvkelso: From the Wikipedia page looks like we should really do AsianHighway:CountryCode to make the artwork functional per country:

- https://en.wikipedia.org/wiki/Template:AHN-AH

nvkelso: Let's just file a followup issue for this and ignore it for this sprint.

zerebubuth:
Re: `BSA15`, fixed in 2b7c81b.

Issue #1519, opened 2018-06-07
Swiss Road Shields


zerebubuth:
Use `CH:motorway` and `CH:national` (not  `CH:nationalstrasse`), and `CH:regional`.

See https://github.com/tilezen/vector-datasource/pull/1515#pullrequestreview-125733546 and https://github.com/tilezen/vector-datasource/issues/1491#issuecomment-394894677

nvkelso:
There's not a whole lot of CH:national or CH:regional set in the sites but this is working as expected.

PR #1525, opened 2018-06-08
Swiss road shields.

zerebubuth:
It looks like there are perhaps 3 different levels of roads in Switzerland:

* Autobahnen / autostrassen. These seem to generally have a ref of `A#`, and/or be in `ch:motorway` relations.
* National roads / hauptstrassen. These just seem to have a numeric ref, some in relations with `ch:national`.
* Regional roads. These just seem to have a numeric ref, some in relations with `ch:regional`.

I've mapped these to `CH:motorway`, `CH:national` or `CH:regional`.

Connects to #1519. Connects to #1491.

### DIFF
--- a/integration-test/1491-generic-network-fixes.py
+++ b/integration-test/1491-generic-network-fixes.py
@@ -464,7 +464,10 @@ class GenericNetworkTest(FixtureTest):
         self.assert_has_feature(
             z, x, y, 'roads', {
                 'id': 26686903,
-                'network': u'CH:Nationalstrasse',
+                'shield_text': '28',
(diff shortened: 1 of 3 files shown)


zerebubuth:
Heres a [map of motorways in Switzerland](https://en.wikipedia.org/wiki/Motorways_of_Switzerland#/media/File:Image-Swiss-Highway-network-en.png) from Wikipedia, and [road shield pictures](https://en.wikipedia.org/wiki/Road_signs_in_Switzerland_and_Liechtenstein#..._numbering).

Issue #1517, opened 2018-06-07
Turkish Road Shields


zerebubuth:
Normalise Turkish `ref`s and `network`s.

See https://github.com/tilezen/vector-datasource/issues/1491#issuecomment-391825067, https://github.com/tilezen/vector-datasource/pull/1515#pullrequestreview-125754713 and https://github.com/tilezen/vector-datasource/pull/1515#pullrequestreview-125754833.

Subtask of #1491.

nvkelso:
Verified on dev, closing.

PR #1527, opened 2018-06-08
Turkish road shields.

zerebubuth:
Turkish road network seems to be in three classes:

* `O` [Otoyol](https://en.wikipedia.org/wiki/Otoyol), mapped to `TR:motorway`,
* `D` [State highways](https://en.wikipedia.org/wiki/Turkish_State_Highway_System), mapped to `TR:highway`,
* Provincial roads, numbered `00-00`, mapped to `TR:provincial`.

Connects to #1517. Connects to #1491.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -4491,6 +4491,21 @@ def _guess_network_ru(tags):
     return networks
 
 
+def _guess_network_tr(tags):
+    ref = tags.get('ref')
+    networks = []
+
+    for part in _COMMON_SEPARATORS.split(ref):
+        part = part.strip()
+        if not part:
+            continue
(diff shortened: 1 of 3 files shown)


vectordatasource/transform.py
```diff
@@ -4903,6 +4918,39 @@ def _sort_network_ru(network, ref):
...
 
+def _sort_network_tr(network, ref):
+    ref = _make_unicode_or_none(ref)
+
+    if network is None:
+        network_code = 9999
+    elif network == 'TR:motorway':
+        network_code = 0
+    elif network == 'TR:highway':
+        # some highways are "main highways", so it makes sense to show them
```
nvkelso: Nice work!

Issue #1518, opened 2018-06-07
Vietnamese Road Shields


zerebubuth:
Backfill network and shield text for [roads in Vietnam](https://en.wikipedia.org/wiki/Transport_in_Vietnam#Road_transport).

Subtask of #1491.

zerebubuth:
See also https://github.com/tilezen/vector-datasource/pull/1515/files/28c06004411a57c4fabc29241d7315d81e182fc6#diff-e2b4972fb4f09e75e9a85c43ef5d8d41.

nvkelso:
Verified on dev.

PR #1530, opened 2018-06-11
Vietnamese road shields.

zerebubuth:
Vietnam has [several levels of road hierarchy](https://en.wikipedia.org/wiki/Transport_in_Vietnam#Road_transport), and the ones which are commonly mapped are:

* `CT`, đường cao tốc, expressways, mapped to `VN:expressway`,
* `QL`, quốc lộ, national routes, mapped to `VN:national`,
* `ĐT` and `TL`,  đường tỉnh and tỉnh lộ, provincial routes, mapped to `VN:provincial`.

There are many other roads tagged with a `ref`, but no `network` tag and not in a relation. For any where the road class is unknown, I've left them in a generic `VN:road` network.

Connects to #1491. Connects to #1518.


zerebubuth:
The shields generally seem to have a dot in them, such as [`CT.04`](https://en.wikipedia.org/wiki/Hanoi-Haiphong_Expressway), but I removed that to keep the length of the shield text down.

PR #1532, opened 2018-06-12
Add first version of a test generator.

zerebubuth:
The idea is to replace a lot of the external data which we currently have either loaded from OSM or in shapefile fixtures with data that's explicitly in the test file. To make this easier, the `scripts/mktest.py` script can handle loading and transforming data, writing out Python code which can be appended to a test.

For example, to generate a test of the road transformations, one might run `python scripts/mktest.py road --way-id 389741906 --is-in ES --expect '{"shield_text": "A-132", "network": "ES:autonoma"}' >> integration-test/1521-spanish-road-shields.py`. The template is set up in such a way that appending to an existing test results in relatively well-formatted code.

These tests still usually need a bit of cleanup to make sure it's valid Python syntax (e.g: names of test functions), fit with PEP8, and remove unnecessary tags or properties which make the test less readable.

### DIFF
--- a/scripts/mktest.py
+++ b/scripts/mktest.py
@@ -1,0 +1,277 @@
+import requests
+import xml.etree.ElementTree as ET
+import tilequeue.tile as tile
+from ModestMaps.Core import Coordinate
+from jinja2 import Template
+import os
+from contextlib import contextmanager
+
+
+def way_full(way_id, zoom=16):
+    url = 'https://api.openstreetmap.org/api/0.6/way/%d/full' % (way_id,)
+    headers = {'user-agent': 'mktest.py/0.0.1 (https://github.com/tilezen)'}
+    r = requests.get(url, headers=headers)
+    root = ET.fromstring(r.content)
+    assert root.tag == 'osm'
+
+    nodes = {}
+    nds = []
+    tags = {}
+
+    for child in root:
+        if child.tag == 'node':
+            node_id = int(child.attrib['id'])
+            nodes[node_id] = child
+
+        elif child.tag == 'way':
+            assert way_id == int(child.attrib['id'])
+            for wchild in child:
+                if wchild.tag == 'nd':
+                    nds.append(int(wchild.attrib['ref']))
+                elif wchild.tag == 'tag':
+                    k = wchild.attrib['k']
+                    v = wchild.attrib['v']
+                    tags[k] = v
+
+    assert nds
+    first_node = nodes[nds[0]]
+    lat = float(first_node.attrib['lat'])
+    lon = float(first_node.attrib['lon'])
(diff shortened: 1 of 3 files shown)


scripts/mktest.py
```diff
@@ -0,0 +1,277 @@
...
+import xml.etree.ElementTree as ET
+import tilequeue.tile as tile
+from ModestMaps.Core import Coordinate
+from jinja2 import Template
+import os
+from contextlib import contextmanager
+
+
+def way_full(way_id, zoom=16):
+    url = 'https://api.openstreetmap.org/api/0.6/way/%d/full' % (way_id,)
```
rmarianski: nit: just `% way_id` would work here

scripts/mktest.py
```diff
@@ -0,0 +1,277 @@
...
+                elif wchild.tag == 'tag':
+                    k = wchild.attrib['k']
+                    v = wchild.attrib['v']
+                    tags[k] = v
+
+    assert nds
+    first_node = nodes[nds[0]]
+    lat = float(first_node.attrib['lat'])
+    lon = float(first_node.attrib['lon'])
+    x, y = tile.deg2num(lat, lon, zoom)
```
rmarianski: And we only care about the first one because we're interested in tags here, and not the geometry? (Just making sure I understand).

scripts/mktest.py
zerebubuth: If `way_id` somehow becomes a tuple then we get an error saying "not all arguments converted during string formatting", rather than "a number is required, not tuple". Generally, this isn't going to happen... but I've shot myself in the foot that way a number of times! And I've found it terribly confusing. So now I try to default to using `% (thing,)` when I think there should only be one thing.

scripts/mktest.py
zerebubuth: Yup. I was thinking that, because some of the tests are mildly geometry-dependent with the country code tests, I wanted to have some representative geometry but not the whole thing. So I figured it would probably work out okay to find the tile containing any node in the way and use the `dsl.tile_diagonal` function to fake up some geometry in a concise statement.

That works great for the roads, but we might want to do something else as we add extra test generators - sometimes using the real geometry when it's important, sometimes generating it when it isn't.

zerebubuth:
> how close were you to going with the python ast module over the jinja templates

I figured a whole page of https://github.com/tilezen/vector-datasource/blob/45d73715e62d94d5d5777c75e8b82b2913b50262/vectordatasource/meta/python.py#L460-L469 wouldn't be terribly readable :wink: 

Jinja2 ends up being a pain with escaping and whitespace in a way that the AST module wouldn't, though. So I'm definitely not a huge fan of Jinja2 for this. If there's something that would be both more readable and handle escaping and PEP8, then let's use that instead!

Issue #977, opened 2016-08-18
🌍 Low zoom locality names are missing translations + country, region zooms


meetar:
**tl;dr**

The new Natural Earth v4.1 populated places file now includes localized names in a `name_*`, specifically:

1) Expands the name localization added in v4.0 to 21 languages (up from 7) and several
   dozen themes expanding from populated places to include all admin-0, admin-1,
   rivers, lakes, playas, geographic lines, physical labels, parks, airports, ports, and
   more. As part of this work a new unique and stable "ne_id" has been added for any
   feature with a name translation &/or a Wikidata ID concordance. The full list of
   languages is: 

`name_ar`*,
`name_bn`*,
`name_de`,
`name_en`,
`name_es`,
`name_fr`,
`name_el`*,
`name_hi`*,
`name_hu`*,
`name_id`*,
`name_it`*,
`name_ja`*,
`name_ko`*,
`name_nl`*,
`name_pl`*,
`name_pt`,
`name_ru`,
`name_sv`*,
`name_tr`*,
`name_vi`*, and 
`name_zh`

   (Names with * indicate
   new language in v4.1 series.) A 2-character language code decoder ring is here:
   https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes. Props to Wikidata for their
   CC0 license: https://www.wikidata.org/wiki/Wikidata:Introduction.

This is for many themes, not just populated places. So work here should be generic for many theme imports.

===

_Original title:_ **Include OpenStreetMap language data for Natural Earth features**

For example, ocean names are available in many languages at zooms 9 and higher, but only in English at zoom 8 and lower. Example map: https://mapzen.com/tangram/play/?scene=https%3A%2F%2Fapi.github.com%2Fgists%2F9f177672cdaac4640ea1b519cde3420d#9.4233/13.3585/-38.4174

If you zoom out below zoom 9, the label disappears.

nvkelso:
We'd need to match the selection of features shown in earlier zooms when changing the source, but the extra names would sure be nice.

nvkelso:
Related: https://github.com/tilezen/vector-datasource/issues/1140

nvkelso:
OpenMapTiles uses some nifty logic to merge city ranks between OpenStreetMap and Natural Earth, we could probably do similar for the names? (Keeping the Natural Earth features, but decorating them with additional OpenStreetMap names at runtime.)

[...]

PR #1533, opened 2018-06-13
Low zoom translations

zerebubuth:
Here's a first attempt at solving #977. I've included translations in the `places` and `water` layers, where available. For example:

output

It looks like we're not using any data from the physical labels theme. Is that new in v4, or are we getting that data from OSM? Let me know what you think, and let's file follow-up issues for bringing in any new (or missing) data we might want.

Connects to #977.

### DIFF
--- a/data/apply-schema-update.sql
+++ b/data/apply-schema-update.sql
@@ -1,4 +1,17 @@
-CREATE OR REPLACE function tmp_add_column(tbl_name text, cln_name text, typ_name text)
+-- this function creates a new column named `cln_name` on table `tbl_name` with
+-- type `typ_name` - UNLESS that column already exists (the type is not checked)
+-- in which case it does nothing.
+--
+-- the function is idempotent, so can be run multiple times to make the same
+-- change without erroring, unlike the `ALTER TABLE ... ADD COLUMN ...`
+-- statement. this is useful when the import process errors out part way through
+-- and rather than starting from scratch, this section can just be re-run.
+--
+-- NOTE: this function should be deleted at the end of the file, when it's no
+-- longer needed. it won't error if not (see `CREATE OR REPLACE`), but it's not
+-- a function that needs to be left lying around in the database.
+--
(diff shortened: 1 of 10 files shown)


data/schema.sql
```diff
@@ -1,5 +1,10 @@
 -- wof
 
+-- drop table before creating so that we can run this over and over. this
+-- can be useful when the script fails part of the way through and needs to
+-- be re-run.
+DROP TABLE IF EXISTS wof_neighbourhood_placetype CASCADE;
```
nvkelso: 👍

scripts/mktest.py
```diff
@@ -271,6 +272,9 @@ def naturalearth_test(args):
     ne_test_parser.add_argument(
         '--ne-id', type=int, default=1,
         help='Natural Earth ID to use.')
+    ne_test_parser.add_argument(
+        '--layer-name', default='pois',
```
nvkelso: Nit: Not many NE features ending up in `pois` layer, maybe `places` is better?

vectordatasource/meta/sql.py
```diff
@@ -671,7 +671,7 @@ def table_is_polygonal(name):
 
 
 def calculate_meta_source_for_table(name):
-    if table_is_osm(name):
+    if table_is_osm(name) or name == 'admin_areas':
```
nvkelso: good catch

data/apply-schema-update.sql
```diff
@@ -1,94 +1,113 @@
+CREATE OR REPLACE function tmp_add_column(tbl_name text, cln_name text, typ_name text)
```
nvkelso: Can you explain the name of this function more? Seems like the column is persistent so seems like a weird name.

nvkelso:
There are a few followup issues that are already in the milestone (and we can add more), for physical labels and others.

I think the next step is `country` and `region` labels, and the min_zoom, max_zoom in a new PR.

scripts/mktest.py
zerebubuth: Good point. Fixed in f585eff.

PR #1535, opened 2018-06-15
Join countries and regions to Natural Earth for min and max zoom.

zerebubuth:
Join countries and regions in the `places` layer to Natural Earth using common `wikidata` IDs in order to get the `min_label` and `max_label` properties from the Natural Earth features and use them in the output. Where that's not available, fall back to country-specific defaults in the new files under `spreadsheets/min_zoom/*.yaml`. Although they're YAML files, they felt more like spreadsheets in the sense that they're static data tables rather than sets of rules.

Connects to #977.

### DIFF
--- a/integration-test/977-min-zoom-from-ne-join.py
+++ b/integration-test/977-min-zoom-from-ne-join.py
@@ -53,3 +53,67 @@ class MinZoomFromNETest(FixtureTest):
         x, y = deg2num(self.lat, self.lon, zoom)
         self.assert_no_matching_feature(
             zoom, x, y, 'places', {'id': 838090640})
+
+
+class MinZoomFromAdminAreaBasedDefault(FixtureTest):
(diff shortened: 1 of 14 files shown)


integration-test/977-min-zoom-from-ne-join.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.generate_fixtures(
+            # https://www.openstreetmap.org/node/838090640
+            dsl.point(838090640, (self.lon, self.lat), {
+                'name': u'United Kingdom',
+                'place': u'country',
+                'population': u'61792000',
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q145',
+                'wikipedia': u'de:United Kingdom',  # LOL, de:
```
rmarianski: :)

Issue #1534, opened 2018-06-14
Strip off the intermediate country code value on roads


nvkelso:
We strip out the intermediate `admin` layer. We use the country code to calculate road shields. Let's remove the intermediate country code property completely before cutting the release.

NOTE: We can always add it back later for select countries (like the UK and Australia) where road colors are sometimes different.

nvkelso:
People can also use the `network` values now for this if someone wanted to custom color them :)

nvkelso:
Verified on dev.

PR #1538, opened 2018-06-18
Strip country_code from roads.

zerebubuth:
We use this internally to provide country-specific logic for interpreting road `ref` and `network` tags. However, it's wasteful to include it on all road features, especially given that it's included as a prefix of any `network` properties that we output.

Connects to #1534.


nvkelso:
We added this for places, too? Does that need to come out as well?

zerebubuth:
We use [a function called `point_in_country_logic`](https://github.com/tilezen/vector-datasource/blob/b150863bb294fb38f11c4ac02658828eaaf8a794/queries.yaml#L555) to do country-specific processing on places and mini roundabouts, and that function doesn't leave a `country_code` parameter on the feature.

(The only reason the roads one did was because #135 is worded that way.)

Issue #1423, opened 2017-11-02
Add new kinds for full Maki icon library support


nvkelso:
Around 47:

- [ ] !!! **airfield** - http://wiki.openstreetmap.org/wiki/Tag:military%3Dairfield (also landuse, and pois if has name)
- [x] **barrier** - we have a lot of these already, maybe add a few more? https://taginfo.openstreetmap.org/keys/barrier#values http://wiki.openstreetmap.org/wiki/Key:barrier
- [x] **blood_bank** – http://wiki.openstreetmap.org/wiki/Tag:healthcare%3Dblood_donation and http://wiki.openstreetmap.org/wiki/Tag:healthcare%3Dblood_bank
- [x] **danger** - could add http://wiki.openstreetmap.org/wiki/Tag:military%3Ddanger_area (also landuse and pois if has name), and for sea marks (see https://github.com/tilezen/vector-datasource/issues/822). 16/51.8513/0.8902
- [x] **defibrillator** - http://wiki.openstreetmap.org/wiki/Tag:emergency%3Ddefibrillator
- [ ] !!! **entrance** - http://wiki.openstreetmap.org/wiki/Key:entrance – would this go into the buildings layer not pois?
- [x] **gaming** - http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dadult_gaming_centre
- [x] **garden_centre** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dgarden_centre
- [x] **golf** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dgolf (we already have golf_coarse, so I think this is for a golf store
- [x] **grocery** - http://wiki.openstreetmap.org/wiki/Tag:shop%3Dgrocery and related http://wiki.openstreetmap.org/wiki/Tag:shop%3Dgreengrocer
- [x] **harbor** - http://wiki.openstreetmap.org/wiki/Harbour
	- [x] harbour=yes or landuse=harbour 15/51.4963/-2.7199
	- [x] military=**naval_base** 17/52.37419/4.91787
	- landuse=* tag values with their specific usage:
		- [x] landuse=**port_terminal** 18/-27.36735/-55.98510
		- [x] landuse=**ferry_terminal** (we have poi but not landuse?) 19/54.14756/-4.47411
		- [x] landuse=**container_terminal** 16/49.4824/0.1768
		- [x] landuse=**shipyard** 16/45.0595/12.2517
[...]

Issue #822, opened 2016-05-17
feature request: seamark-layer


Dirk--:
as add-on to recreation data-source, I would like to have a seamark vector layer:

Details: http://wiki.openstreetmap.org/wiki/Seamarks

nvkelso:
A few of these might end up in the general `pois` layer (for instance, we already have **lighthouse** there), but many of these are not general purpose and should probably end up in a new `seamarks` layer, especially as they need to be visible earlier than general purpose representation.

nvkelso:
Related: https://github.com/mapzen/vector-datasource/issues/823

nvkelso:
Closing for inactivity.

PR #1539, opened 2018-06-19
Kinds for Maki icon library support

zerebubuth:
Partial for #1423. Thought it was worth PRing this before it became unwieldy!

Adds support for:

* military airfields
* several additional barrier types (e.g: ditches, kerbs, cattle grids, etc...)
* blood banks / donation centres
* danger zones (although I couldn't figure out which seamark corresponded to that, so seamarks are still a TODO).
* defibrillators
* entrances and exits

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -230,6 +230,8 @@ Note that building geometries, like most geometries in Tilezen tiles, are clippe
 * `building`
 * `building_part`
 * `address`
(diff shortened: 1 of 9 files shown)


nvkelso:
This looks good to me so far... and agree sea marks should be excluded here (they're tracked separately in #822 which doesn't have a milestone right now). Thanks!

PR #1541, opened 2018-06-20
More kinds for maki icon library support

zerebubuth:
Added some more kinds related to #1423.

* `gaming`
* `garden_centre`
* `golf`
* `grocery`
* `harbor`
* `naval_base`
* `port_terminal`
* `ferry_terminal` landuse
* `container_terminal`
* `shipyard`
* `wharf`
* `quay`
* `boatyard`
* `boat_lift`
* `ship_chandler`
* `customs`

(NOTE: we already had `man_made=crane` mapped to `kind: crane`, so there wasn't a need to add that.)

Some follow-ups:

* Added a kind for `grocery`. We already had `greengrocer` and, looking at the Wikipedia pages for [grocery](https://en.wikipedia.org/wiki/Grocery_store) compared with [greengrocer](https://en.wikipedia.org/wiki/Greengrocer), I'm not sure that they're the same thing. For the moment, I've left them as separate kinds.
* Many items tagged as `harbour=yes` are also tagged as other things. For example, this [port terminal](https://www.openstreetmap.org/way/256125920) is tagged `harbour=yes, landuse=port_terminal, man_made=pier`; should that be output as a harbour, port terminal or pier - or should we split the geometry to allow multiple features for the harbour landuse and pier, with a POI for the port terminal? Also, this [pier](https://www.openstreetmap.org/way/183049783) is tagged `harbour=yes, man_made=pier` - should that be harbour, pier or split into two features?
* The `customs` kind mentions the [`amenity=customs`](http://wiki.openstreetmap.org/wiki/Tag:amenity=customs) tag, which has only 742 uses. Should we also include the [`barrier=border_control`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dborder_control) tag, which has over 5,000 uses? Should they _both_ be mapped to kind `customs`, or are customs (i.e: taxation) and border control sufficiently distinct things to deserve distinct kinds?

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -891,7 +891,8 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `golf` - Shop selling golf equipment.
 * `golf_course`
 * `government`
-* `greengrocer`
+* `greengrocer` - Shop selling fruits and vegetables.
+* `grocery` - Shop selling non-perishable foods.
 * `group_home`
 * `guest_house`
 * `hairdresser`
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -526,7 +526,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `groyne`
 * `guard_rail`
 * `hanami`
-* `harbor` (a.k.a harbour)
+* `harbour`
 * `hospital`
 * `industrial`
 * `kerb`
@@ -762,6 +762,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `accountant`
 * `adit`
 * `administrative`
+* `adult_gaming_centre`
 * `advertising_agency`
 * `aerodrome`
 * `aeroway_gate`
@@ -890,7 +891,6 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `fuel` - Fuel stations provide liquid gas (or diesel) for automotive use.
 * `furniture`
 * `gallery` - An art gallery.
-* `gaming`
 * `garden` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
 * `gardener`
 * `garden_centre`
@@ -903,7 +903,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `golf_course`
 * `government`
 * `greengrocer` - Shop selling fruits and vegetables.
-* `grocery` - Shop selling non-perishable foods.
+* `grocery` - Shop selling non-perishable food often similar to, but smaller than, a `supermarket`. See also [grocery store on Wikipedia](https://en.wikipedia.org/wiki/Grocery_store).
 * `group_home`
 * `guest_house`
 * `hairdresser`
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -879,6 +879,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `fuel` - Fuel stations provide liquid gas (or diesel) for automotive use.
 * `furniture`
 * `gallery` - An art gallery.
+* `gaming`
 * `garden` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
 * `gardener`
 * `gas_canister` - Shop selling bottled gas for cooking. Some offer gas canister refills.
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -888,6 +888,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `generator`
 * `geyser`
 * `gift`
+* `golf` - Shop selling golf equipment.
 * `golf_course`
 * `government`
 * `greengrocer`
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -882,6 +882,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `gaming`
 * `garden` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
 * `gardener`
+* `garden_centre`
 * `gas_canister` - Shop selling bottled gas for cooking. Some offer gas canister refills.
 * `gate` - with `kind_detail` one of `chain`, `gate`, `kissing_gate`, `lift_gate`, `stile`, `swing_gate`.
 * `generator`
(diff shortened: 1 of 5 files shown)


yaml/landuse.yaml
```diff
@@ -1116,3 +1123,23 @@ filters:
...
       kind: crane
       kind_detail: {col: "crane:type"}
+  - filter:
+      any:
+        harbour: 'yes'
+        landuse: harbour
+    min_zoom: 16
+    output:
+      <<: *output_properties
+      kind: harbor
```
nvkelso: Our practice on the kinds is generally to follow British English, especially if that's what the original OSM tag was. 

Please set this to `harbour`.

yaml/pois.yaml
```diff
@@ -1614,6 +1614,13 @@ filters:
     output:
       <<: *output_properties
       kind: petroleum_well
+  # a boat lift is a more specific type of object than a crane, although it may
```
nvkelso: 👍 Useful note explaining the order of filters.

yaml/pois.yaml
```diff
@@ -1697,6 +1704,21 @@ filters:
...
       kind: quarry
+  - filter:
+      leisure:
+        - adult_gaming_centre
+        # fairly common misspelling
+        - adult_gaming_center
+    min_zoom: 17
+    output:
+      <<: *output_properties
+      kind: gaming
```
nvkelso: I think the kind should be `adult_gaming_centre`

yaml/landuse.yaml
nvkelso: See PR comments – this should move to `landuse: harbour` only.

yaml/landuse.yaml
```diff
@@ -458,6 +458,13 @@ filters:
       kind: airfield
       kind_detail: military
       tier: 3
+  - filter: {military: naval_base}
+    min_zoom: *tier3_min_zoom
+    output:
+      <<: *output_properties
+      kind: naval_base
+      kind_detail: military
```
nvkelso: Why is the `kind_detail` set here?  Are there naval bases that aren't military related?

integration-test/1423-kinds-for-maki-icon-library-support.py
```diff
@@ -1283,3 +1283,646 @@ def test_fire_exit_entrance_node(self):
...
+                'seamark:harbour:category': u'tanker;bulk',
+                'seamark:type': u'harbour',
+                'source': u'openstreetmap.org',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'landuse', {
+                'id': 183049783,
+                # should be 'kind': u'harbor',?
```
nvkelso: Resolve this before merging.

yaml/landuse.yaml
zerebubuth: Ah, sorry, I think that was copy-pasta from the above `airfield`, where it makes sense to say it's a military thing. Civilian airfields exist, but civilian naval bases, as you rightly point out, not so much. Fixed in 96110df.

yaml/landuse.yaml
zerebubuth: `harbor` -> `harbour` in 96110df.

yaml/pois.yaml
zerebubuth: Changed in 96110df.

integration-test/1423-kinds-for-maki-icon-library-support.py
zerebubuth: Resolved in 96110df.

docs/layers.md
```diff
@@ -524,6 +526,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `groyne`
 * `guard_rail`
 * `hanami`
+* `harbor` (a.k.a harbour)
```
nvkelso: Also fix the documentation spelling (to `harbour` without the parenthetical note)

docs/layers.md
```diff
@@ -879,16 +890,20 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `fuel` - Fuel stations provide liquid gas (or diesel) for automotive use.
 * `furniture`
 * `gallery` - An art gallery.
+* `gaming`
```
nvkelso: This name also changes to adult something something?

docs/layers.md
```diff
@@ -879,16 +890,20 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
...
 * `gate` - with `kind_detail` one of `chain`, `gate`, `kissing_gate`, `lift_gate`, `stile`, `swing_gate`.
 * `generator`
 * `geyser`
 * `gift`
+* `golf` - Shop selling golf equipment.
 * `golf_course`
 * `government`
-* `greengrocer`
+* `greengrocer` - Shop selling fruits and vegetables.
+* `grocery` - Shop selling non-perishable foods.
```
nvkelso: `Shop selling non-perishable foods, also known as a "supermarket".`?

yaml/landuse.yaml
```diff
@@ -1116,3 +1132,14 @@ filters:
       <<: *output_properties
       kind: crane
       kind_detail: {col: "crane:type"}
+  - filter:
+      landuse:
+        - harbour
```
nvkelso: I think `harbour` should go into the general port group above, with the min zoom clamp of 11 to 16 ... Possibly also `port`.

yaml/landuse.yaml
```diff
@@ -1116,3 +1132,14 @@ filters:
...
+        - port
+        - shipyard
+        - wharf
+        - quay
+    min_zoom: 16
```
nvkelso: 15 for port and shipyard, 16 for wharf and quay?

docs/layers.md
zerebubuth: I changed it to:

> Shop selling non-perishable food often similar to, but smaller than, a `supermarket`. See also [grocery store on Wikipedia](https://en.wikipedia.org/wiki/Grocery_store).

I get the impression from that Wikipedia article that a grocery store doesn't generally sell fresh fruits and vegetables, nor non-food items. I can't remember ever having been in a store which had such a laser focus - all the ones which come to mind either additionally sold stationery, cards, household products and similar non-food items (and might be called a newsagents or convenience store), or had a fridge/freezer section for selling perishable food.

Should we merge this into `supermarket` or `convenience`?

yaml/landuse.yaml
zerebubuth: OK. Moved `harbour` and `port` in 745f2e5.

yaml/landuse.yaml
zerebubuth: Updated, and added tests in 745f2e5.

docs/layers.md
nvkelso: In the USA `grocery` is probably more akin to "corner store" or "bodega" versus a supermarket which is a usually "big box" style store. Fine as-is from OSM.

PR #1542, opened 2018-06-22
More kinds for Maki icon library support

zerebubuth:
Added new kinds:

* `harbourmaster`
* `mooring`
* `sanitary_dump_station`
* `path` with `kind_detail: quay` in `roads` layer, similar to `pier`.

Additional properties:

* `mooring` on `pier` and `quay`
* `sanitary_dump_station` on `marina`, `camp_site` and `caravan_site`.

Connects to #1423.

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -165,6 +165,7 @@ global:
         - public_transport: [ platform, stop_area ]
         - railway: [ halt, level_crossing, platform, stop, subway_entrance, tram_stop ]
         - "seamark:building:function": harbour_master
+        - "seamark:small_craft_facility:category": fuel_station
         - tags->icn_ref: true
         - tags->iwn_ref: true
         - tags->lcn_ref: true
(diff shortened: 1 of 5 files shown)


yaml/pois.yaml
```diff
@@ -769,6 +773,16 @@ filters:
     output:
       <<: *output_properties
       kind: level_crossing
+  # filter the waterway / boat fuel before generic amenity=fuel, as they appear
```
nvkelso: 👍

yaml/pois.yaml
```diff
@@ -769,6 +773,16 @@ filters:
...
+  # to be often tagged with both.
+  - filter:
+      any:
+        - waterway: fuel
+        - "seamark:small_craft_facility:category": fuel_station
+    min_zoom: { clamp: { min: 0, max: 16, value: { sum: [ { col: zoom }, 2.7 ] } } }
```
nvkelso: should this be min zoom of say 14 instead of 0? Seems less important than general fuel which I don't think is clamped?

zerebubuth: OK. Updated in fd0f922.

PR #1543, opened 2018-06-27
New kinds for OSM and iD presets

zerebubuth:
New kinds:

* `chemist`
* `elevator`
* `embankment` (also an `embankment` property on roads and railways, and `cutting`)
* `miniature_golf`
* `mud`
* `orchard`
* `plant_nursery`
* `plaque` (merging both `memorial=plaque` and `historic=memorial_plaque`)
* `reef`
* `cosmetics`
* `fishmonger`
* `bunker`
* `wayside_cross`
* `obelisk`

Added `kind_detail` whitelists on:

* `beach`
* `forest` and `wood`
* `wetland`

Verified that all the kinds mentioned in #1425 are either implemented here or have been previously implemented. Checked docs and tests, too.

Connects to #1425.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -493,6 +493,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `caravan_site`
 * `carousel`
 * `cemetery`
+* `chemist` - A shop selling household chemicals, often including soaps, toothpaste and cosmetics.
 * `cinema`
 * `city_wall`
 * `college`
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1018,6 +1018,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `pitch`
 * `place_of_worship` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
 * `plant` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
+* `plaque` - A memorial plaque.
 * `playground` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
 * `plumber`
 * `police` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -567,7 +567,6 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `railway`
 * `recreation_ground`
 * `recreation_track`
-* `reef` - A solid feature just under the surface of the ocean, usually made from rock, sand or coral. If known, the `kind_detail` will be given as one of `coral`, `rock`, `sand`.
 * `residential`
 * `resort`
 * `rest_area`
@@ -1446,6 +1445,7 @@ Mapzen calculates the composite exterior edge for overlapping water polygons and
 * `lake` - polygon
 * `ocean` - polygon, point is intended for label placement only
 * `playa` - polygon
+* `reef` - polygon. A solid feature just under the surface of the ocean, usually made from rock, sand or coral. If known, the `kind_detail` will be given as one of `coral`, `rock`, `sand`.
 * `river` - line
 * `riverbank` - polygon
 * `sea` - point, intended for label placement only
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1129,6 +1129,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `waterfall`
 * `watering_place`
 * `watermill` - A structure for using water power to do work. Note that this is different from a modern structure to generate electric power from water, which would be a `generator`.
+* `wayside_cross`
 * `wilderness_hut`
 * `wildlife_park`
 * `windmill`
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -627,7 +627,7 @@ The value of the OpenStreetMap `wetland` tag. Common values include `bog`, `fen`
 
 #### Wood and forest `kind_detail` values
 
-* The value of the OpenStreetMap `leaf_type` tag. [Common values](https://taginfo.openstreetmap.org/keys/leaf_type#values) include `broadleaved`, `needleleaved`, or `mixed`.
+* The value of the OpenStreetMap `leaf_type` tag, whitelisted to `broadleaved`, `needleleaved`, `mixed` or `leafless`.
 
 ## Places
 
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -502,6 +502,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `container_terminal`
 * `crane`
 * `cutline`
+* `cutting` - A lowered area of land, usually to carry a road or railway.
 * `danger` - e.g: military training zones, firing ranges.
 * `dam` - polygon, line
 * `dike`
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -997,6 +997,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `nightclub`
 * `notary`
 * `nursing_home` - _See planned bug fixes in [#1085](https://github.com/tilezen/vector-datasource/issues/1085)._
+* `obelisk` - A tall structure, usually a monument or memorial (so see also `monument` and `memorial`).
 * `observatory`
 * `office` - An office which didn't match a more specific kind.
 * `offshore_platform`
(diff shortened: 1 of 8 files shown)


yaml/pois.yaml
```diff
@@ -1576,11 +1593,27 @@ filters:
...
       kind: {col: highway}
+  # memorial plaques more specific than plain memorials
+  - filter:
+      any:
+        - {historic: memorial, memorial: plaque}
+        - historic: memorial_plaque
+    min_zoom: 16
+    output:
+      <<: *output_properties
+      kind: plaque
```
nvkelso: I'm fine with this generic kind value of `plaque` (versus `memorial_plaque`).

yaml/landuse.yaml
```diff
@@ -756,6 +771,55 @@ filters:
       <<: *output_properties
       kind: beach
       tier: 5
+      # whitelist most common surface values for beaches. i'm not quite sure i'd
+      # call surface=grass a beach, but it's in the data..?
+      kind_detail:
+        case:
+          - when: {surface: [grass, gravel, pebbles, pebblestone, rocky, sand]}
```
nvkelso: Looks like my favorite British English-ism https://wiki.openstreetmap.org/wiki/Tag:natural=shingle never factors in here (guess it would be rocky), looking at TagInfo.

yaml/roads.yaml
```diff
@@ -78,6 +78,14 @@ global:
       sidewalk: {col: sidewalk}
       sidewalk_left: {col: 'sidewalk:left'}
       sidewalk_right: {col: 'sidewalk:right'}
+      cutting:
+        case:
+          - when: {cutting: ['yes', 'right', 'left', 'both']}
```
nvkelso: Can this work more like bike lanes – where it's yes, right, left and if both then it'd just be yes? (Please verify, this is from memory.)

yaml/roads.yaml
```diff
@@ -78,6 +78,14 @@ global:
...
+            then: true
+      embankment:
+        case:
+          - when: {embankment: ['yes', 'right', 'left', 'two_sided', 'both']}
```
nvkelso: Can this work more like bike lanes – where it's yes, right, left and if both then it'd just be yes? (Please verify, this is from memory.)

docs/layers.md
```diff
@@ -621,11 +628,11 @@ The value of the OpenStreetMap `wall` tag. Common values include `brick`, `castl
 
 ##### Wetland `kind_detail` values:
 
-The value of the OpenStreetMap `wetland` tag. Common values include `bog`, `fen`, `mangrove`, `marsh`, `reedbed`, `saltmarsh`, `string_bog`, `swamp`, `tidalflat`, and `wet_meadow`.
+The value of the OpenStreetMap `wetland` tag. If available, value will be one of: `bog`, `fen`, `mangrove`, `marsh`, `mud`, `reedbed`, `saltern`, `saltmarsh`, `string_bog`, `swamp`, `tidalflat`, `wet_meadow`.
```
nvkelso: New to me! https://en.wikipedia.org/wiki/Saltern

docs/layers.md
```diff
@@ -853,6 +861,7 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `consulting`
 * `convenience`
 * `copyshop` - A shop offering photocopying and printing services.
+* `cosmetics` - A shop selling cosmetics.
```
nvkelso: Since we also say chemist sells "cosmetics" let's say for this one:
`A specialty shop selling cosmetics.`

yaml/pois.yaml
```diff
@@ -2100,6 +2138,18 @@ filters:
       <<: *output_properties
       kind: sanitary_dump_station
 
+  - filter: {highway: elevator}
+    min_zoom: 16
```
nvkelso: To `min_zoom` 17, please. This is an "indoor" feature.

yaml/pois.yaml
```diff
@@ -1576,11 +1593,27 @@ filters:
...
+    output:
+      <<: *output_properties
+      kind: plaque
+  # plain memorial
   - filter: {historic: memorial}
     min_zoom: 17
     output:
       <<: *output_properties
       kind: memorial
+  # obelisk - but not a monument or a memorial
```
nvkelso: I think obelisk should be sorted above plaque and memorial – it's the thing visible from far while the others are generally attached to an obelisk?

integration-test/1425-more-osm-features.py
```diff
@@ -0,0 +1,1377 @@
...
+
+        z, x, y = (16, 19304, 24371)
+
+        self.generate_fixtures(
+            # https://www.openstreetmap.org/way/360547782
+            dsl.way(360547782, dsl.tile_box(z, x, y), {
+                'natural': u'beach',
+                'source': u'openstreetmap.org',
+                'surface': u'gravel',
+                # fake name to get the POI to appear
```
nvkelso: Nice ;)

nvkelso: This is true for pois, but for landuse the polygon would still show up regardless of name being present?

integration-test/1425-more-osm-features.py
```diff
@@ -0,0 +1,1377 @@
...
+    def test_elevator_way(self):
+        import dsl
+
+        z, x, y = (16, 19300, 24647)
+
+        self.generate_fixtures(
+            # https://www.openstreetmap.org/way/510502210
+            dsl.way(510502210, dsl.tile_box(z, x, y), {
+                'highway': u'elevator',
+                'level': u'(-1,-2)',
```
nvkelso: Someday we'll do indoor, but not today!

yaml/pois.yaml
```diff
@@ -497,6 +498,16 @@ filters:
       <<: *output_properties
       kind: danger_area
       tier: 3
+  # bunker
+  - filter: {military: bunker}
+    min_zoom: 16
```
nvkelso: Can this be 16 if with a name and 18 if no-name?

yaml/landuse.yaml
```diff
@@ -489,6 +493,17 @@ filters:
       <<: *output_properties
       kind: glacier
       tier: 3
+  # reef
+  - filter: {natural: reef}
```
nvkelso: Odd question – but since this is underwater should this go into the **water** layer instead of **landuse**?

yaml/roads.yaml
zerebubuth: I was just setting `cutting` to `true` if it matched any of those (or omitting if it didn't match). I've updated in 8c79ac2 to make the values `yes`, `left` or `right`.

Looking at the properties for `cycleway`, it looks like we export a `cycleway_both` property, so we don't collapse that down onto `cycleway`. Perhaps that's something we should come back to on a ☔ (although now I suppose it'd be a breaking change, we'd need to wait for v2.0).

docs/layers.md
zerebubuth: Good point, thanks. Fixed in fe1cbf0.

yaml/pois.yaml
zerebubuth: OK. Updated in 1e6b874.

yaml/pois.yaml
zerebubuth: OK. Done in bfb655a. Moved the `monument`/`memorial` info to `kind_detail` and updated documentation and added tests.

integration-test/1425-more-osm-features.py
zerebubuth: Yup, the fake name is only needed for POIs. I've split the test in 4f20bd3 between `pois` and `landuse`, with the fake name only on the `pois` test.

yaml/landuse.yaml
zerebubuth: Sure, that makes sense. Moved to `water` layer in 973963c.

zerebubuth:
@nvkelso I think I addressed all the issues. Would you like to give it a quick look over again, just to be sure?

yaml/pois.yaml
```diff
@@ -1139,6 +1156,17 @@ filters:
     output:
       <<: *output_properties
       kind: motorway_junction
+  # obelisk - note that this takes precedence over monument and memorial!
+  - filter: {man_made: obelisk}
+    min_zoom: { clamp: { min: 15, max: 17, value: { sum: [ { col: zoom }, 2.24 ] } } }
```
nvkelso: For the Washington Monument in DC, if something is "tall", let's make it zoom 14. (greater than 20 meters?)

If it's medium tall then zoom 15? (greater than 10 meters?)

zerebubuth: Done in b90875b. Also pushed the priority of obelisks above artworks, as some appear to also have `tourism=artwork` on them, but obelisk seems more specific - especially for the >10m ones.

Issue #1557, opened 2018-08-22
Pacific ocean broken at zoom 5


zerebubuth:
* **What did you see?**

<img width="1678" alt="screen shot 2018-08-22 at 17 53 27" src="https://user-images.githubusercontent.com/271360/44478071-6f436c80-a634-11e8-9e56-dbf1b6d50f3c.png">

* **What were you expecting to see out of a Torquay hotel bedroom window? Sydney Opera House, perhaps? The Hanging Gardens of Babylon? Herds of wildebeest sweeping majestically -?**

Don't be silly. I expect to be able to see the sea.

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

Pacific ocean, west of 90W and in the whole northern hemisphere at (nominal) zoom 5.

nvkelso:
Verified fix on dev, closing.

PR #1558, opened 2018-08-22
Make ocean polygons valid

zerebubuth:
Run ST_MakeValid on geometries after mercator projection in ne_10m_ocean table.

The missing water described in #1557 was because the polygon in `ne_10m_ocean` was not valid, and was being rejected by the processing pipeline.

Connects to #1557.


nvkelso:
Let's keep this change just for ocean polygons, but does seem a useful general pattern for any polygon data we bring in (from Natural Earth or elsewhere, like country polygons).

Issue #1555, opened 2018-08-20
Be more robust to None in road network code.


zerebubuth:
Backtrace:

```
Traceback (most recent call last): 
   File "/usr/src/tilequeue/tilequeue/command.py", line 2287, in tilequeue_meta_tile_low_zoom    formatted_tiles, _ = processor.process_tiles() 
   File "/usr/src/tilequeue/tilequeue/process.py", line 757, in process_tiles    cut_coords, self.buffer_cfg, self.output_calc_mapping 
   File "/usr/src/tilequeue/tilequeue/process.py", line 498, in process_coord    output_calc_spec) 
   File "/usr/src/tilequeue/tilequeue/process.py", line 398, in process_coord_no_format    unpadded_bounds) 
   File "/usr/src/tilequeue/tilequeue/process.py", line 84, in _postprocess_data    layer = fn(ctx) 
   File "/usr/src/vector-datasource/vectordatasource/transform.py", line 7235, in road_networks    shape, props, fid = fn(shape, props, fid, zoom) 
   File "/usr/src/vector-datasource/vectordatasource/transform.py", line 7203, in _fixup_country_specific_networks    network, ref = logic.fix(network, ref) 
   File "/usr/src/vector-datasource/vectordatasource/transform.py", line 5977, in _normalize_pl_netref    if ref.startswith('A'): 
 AttributeError: 'NoneType' object has no attribute 'startswith'
```

This results from assuming that `ref` is a `str` without checking. In all of this code, either `ref` or `network` could be `None`. Therefore, we need to check that it's not `None` first (ideally `isinstance`, but practically if it's truthy it's going to be a string).

nvkelso:
@zerebubuth Is this issue good to close? Is there a good place on the map to look for verification?

zerebubuth:
Yes, I think so. We were missing a lot of tiles, including `3/4/2` and `3/5/1`, both of which seem to be OK in the current build:

<img width="1680" alt="screen shot 2018-09-05 at 09 43 52" src="https://user-images.githubusercontent.com/271360/45082067-736d9080-b0f0-11e8-90b6-d7346ca1ea63.png">

PR #1563, opened 2018-08-24
Fix bug in Spanish network normalization

zerebubuth:
Where we might call `len()` on `None`.

Continuation of #1555.

### DIFF
--- a/integration-test/1555-be-more-robust-to-none.py
+++ b/integration-test/1555-be-more-robust-to-none.py
@@ -36,3 +36,29 @@ class RobustToNoneTest(FixtureTest):
                 'id': 367163748,
                 'network': 'PL:motorway',
             })
+
+    def test_es_ref_no_digits(self):
+        import dsl
+        z, x, y = (16, 32360, 24673)
+        self.generate_fixtures(
+            dsl.is_in('ES', z, x, y),
+            dsl.way(1, dsl.tile_diagonal(z, x, y), {
+                'highway': 'motorway',
+                'source': 'openstreetmap.org',
+            }),
+            dsl.relation(1, {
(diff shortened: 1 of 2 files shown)


integration-test/1555-be-more-robust-to-none.py
```diff
@@ -36,3 +36,29 @@ def test_fake_rel_with_ref_none(self):
...
+                'type': 'route',
+                'source': 'openstreetmap.org',
+            }, ways=[1]),
+        )
+
+        # can't figure out network or shield_text for this!
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 1,
+                'network': type(None),
```
iandees: It's kinda weird to have `type(None)` here rather than just `None`.

vectordatasource/transform.py
```diff
@@ -5419,23 +5419,23 @@ def _normalize_cn_netref(network, ref):
...
     num_digits = 0
-    for i in xrange(0, len(num)):
-        if num[i].isdigit():
-            num_digits += 1
-        else:
-            break
+    if num:
+        num = num.lstrip('-')
+
+        for i in xrange(0, len(num)):
```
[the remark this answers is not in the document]
zerebubuth: Good point! Thanks for spotting that. Fixed in caadbdf.

integration-test/1555-be-more-robust-to-none.py
zerebubuth: Yup, it's totally weird. It's to work around the previous meaning of `None` as "anything", while abusing the type matching to match `None` because it's the only value of `NoneType`. I've added #1564 to follow up - hopefully we can make the "wildcards" in the property matching clearer.

Issue #1559, opened 2018-08-23
Missing fallback country codes on roads


zerebubuth:
We expect roads in countries to be given a "fallback" network code equal to the ISO country code if we don't know anything more about the network than which country it's in, but this isn't happening in some places, e.g: Norway and Canada.

nvkelso:
Verified missing fallback 2-char country code network values are now working in Canada, Norway, and Australia, closing.

PR #1566, opened 2018-08-24
Fix missing fallback network value for roads

zerebubuth:
Roads in Norway and Canada weren't being assigned `network: NO` or `network: CA` respectively if we couldn't figure out any other information about the road in question. This meant they weren't being shown as country-specific defaults.

Fixed this with a generic fallback for roads where the relation doesn't provide `network`, and a specific fix in Canada's case where we were assigning `network = None` explicitly from the tags.

Connects to #1559.


zerebubuth:
Great, thanks!

PR #1583, opened 2018-08-29
Fix tests

zerebubuth:
Fix integration tests by:

* Inlining the fixture data for many test cases where it was easy to do.
* Manual conversion of a few more tests.
* Upload fixture cache to Nextzen tile assets bucket and point to it.

The branch name is now misleading! Not all tests have been converted to generative, however only 61 (of 232) remain.


zerebubuth:
Last test run time was 3m47s :stopwatch:

integration-test/1016-missing-localized-names.py
```diff
@@ -1,14 +1,13 @@
...
...bc2 \u0baf\u0bcb\u0bb0\u0bcd\u0b95\u0bcd', u'name:ca': u'Nova York', u'ref:fips': u'36', u'name:ka': u'\u10dc\u10d8\u10e3-\u10d8\u10dd\u10e0\u10d9\u10d8', u'name:haw': u'Nuioka', u'old_name:vi': u'N\u1eefu \u01af\u1edbc', u'wikipedia': u'en:New York', u'source': u'openstreetmap.org', u'boundary': u'administrative', u'name:zh': u'\u7ebd\u7ea6\u5dde', u'name:nah': u'Yancu\u012bc York', u'admin_level': u'4', u'name:lv': u'\u0145ujorka', u'name:ja': u'\u30cb\u30e5\u30fc\u30e8\u30fc\u30af\u5dde', u'name:chr': u'\u13c4\u13f2\u13a9', u'name:he': u'\u05e0\u05d9\u05d5 \u05d9\u05d5\u05e8\u05e7', u'is_in:country_code': u'US', u'name:ga': u'Nua-Eabhrac', u'name:la': u'Novum Eboracum', u'name:uk': u'\u041d\u044c\u044e-\u0419\u043e\u0440\u043a', u'name:bg': u'\u041d\u044e \u0419\u043e\u0440\u043a', u'name:sr': u'\u040a\u0443\u0458\u043e\u0440\u043a', u'name:nv': u'N\xed\xedy\xf3\xf3 Hahoodzo', u'name:el': u'\u039d\u03ad\u03b1 \u03a5\u03cc\u03c1\u03ba\u03b7', u'name:eo': u'Nov-Jorkio', u'name:en': u'New York', u'name': u'New York', u'name:gd': u'Eabhraig Nuadh', u'ref': u'NY', u'name:gv': u'York Noa', u'name:yi': u'\u05e0\u05d9\u05d5 \u05d9\u05d0\u05e8\u05e7', u'name:es': u'Nueva York'}))  # noqa
```
nvkelso: What does `# noqa` mean here in the comment (and every other fixture added)?

integration-test/__init__.py
```diff
@@ -57,7 +57,7 @@
 # caching them for everyone to reuse.
 FIXTURE_CACHE = environ.get(
     'FIXTURE_CACHE',
-    'http://s3.amazonaws.com/mapzen-tiles-assets/integration-test-fixtures')
+    'http://s3.amazonaws.com/nextzen-tile-assets/integration-test-fixtures')
```
nvkelso: 👍

integration-test/1016-missing-localized-names.py
zerebubuth: It's how we tell [`flake8` to ignore the line](http://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors), otherwise these all end up much longer than the permitted line width and the test fails `flake8` before it has even been run!

I'd love to reformat these so that they are OK (e.g: as I did in the [manually-edited island tests](https://github.com/tilezen/vector-datasource/pull/1583/commits/3cc51e4b049cbd2a08d39744cbbc7dd70885fa18)), but all of the auto-formatting tools I've tried have produced results that look weird to me. And reformatting by hand is a fair bit of work.

Issue #1274, opened 2017-06-19
Complex building shapes that should not extrude


nvkelso:
* **What did you see?** base building of Fernsehturm in Berlin extrudes as tall as central tower
* **What did you expect to see?** Only the tower should extrude up and up
* **What map location are you having problems with?** 
    * http://www.openstreetmap.org/relation/4065110#map=19/52.52087/13.40957

From @zerebubuth:

> We should review whether we should ignore any height tagging (or add `extrude: false`) on any building outline which has `building:part`s. We already do some of this analysis server-side
to assign the `root_id` to all `building:part`s, so we could extend the logic at that point to add/remove tags from the root building.
> 
> We should also do some analysis first to see if if it's common to treat the building footprint as a "bounding volume", and in how many cases we'd end up missing bits of the building if we ignored the footprint.

I would prefer we stripped off the `height` info (before adding a new `extrude: false` property) – or just set the exported `height` to be `0`.

nvkelso:
This is also happening at SF Ferry Building.

Issue #1032, opened 2016-09-09
Add root_id for features within building relations


nvkelso:
Picking up where https://github.com/tilezen/vector-datasource/issues/653 left off...

We want to add `root_id` for `building` and `building_part` features within building relations. 

@meetar pointed out the Bank of England building at http://tangrams.github.io/explorer/#17.0/51.5143/-0.0890/osm_relation/true:

<img width="571" alt="screen shot 2016-08-26 at 3 38 26 pm" src="https://cloud.githubusercontent.com/assets/459970/18018178/7c7743fe-6ba3-11e6-8762-82e7edc579e1.png">

In our tiles, the inner and outer shapes here seem to be split into two features, correlating to 
https://www.openstreetmap.org/way/4959510 and https://www.openstreetmap.org/relation/553472, but with no `root_id` or other node which would associate them with each other.

@zerebubuth followed up with details:

Looks like there's a whole mess of things going on with the BoE building:

* [Relation 553472](https://www.openstreetmap.org/relation/553472) is a multipolygon relation, with `building=yes, amenity=bank` and `osm2pgsql` puts it in the polygons table.
* [Way 4959510](https://www.openstreetmap.org/way/4959510) is an outer of the above relation, but since it has different tags from its relation, (`building=public`), `osm2pgsql` cautiously retains it. This means there's a second polygon in the polygons table with the same outline as relation 553472 but without the holes.
* [Relation 2917677](https://www.openstreetmap.org/relation/2917677) is also present and although it doesn't have any tags itself, `osm2pgsql` processes it to "steal" the tags from its outer ([way 220644664](https://www.openstreetmap.org/way/220644664)), so it also ends up as a building.

I think we have two different issues here:

1. Relation 553472 and way 4959510 are both included as buildings due to a combination of tagging and how `osm2pgsql` interprets that tagging.
2. Relation 2917677 is an additional building on a similar footprint to relation 553472, which seems like it should be linked to the latter although it's not a `building:part`, and there's no relation linking them or containing them both (e.g: a "site" relation).

If this isn't a widespread problem, I'd be tempted to call it a data issue rather than introduce a special case to handle it. Although it would be nice to be able to present problems like this in a way that makes it easy for people to find and fix.

nvkelso:
Similar to https://github.com/tilezen/vector-datasource/issues/348 for guessing a POIs area by intersecting it with the building footprint.

PR #1689, opened 2018-10-17
Building merging

zerebubuth:
Better building merging, resulting in fewer polygons at zooms 14 & 15 (nominal). In summary, the buildings layer size is drastically reduced:

Tile coordinate (512 size) | Size before (bytes) | Size after (bytes) | Reduction
--- | --- | --- | ---
`13/3033/4647` | 442,333 | 2,950 | 99.3%
`14/6067/9294` | 587,743 | 67,008 | 88.6%
`15/12134/18589` | 384,337 | 234,313 | 39.0%

## Zoom 14

Ignore the grey area to the south and west; that's just because I didn't render those tiles. The visual output is very similar to the style before, as we are now dropping buildings which weren't rendered anyway.

z14-after

## Zoom 15

At this zoom, we're now doing _much_ more merging, and in most dense areas we now merge into single "city block" polygons.

z15-after

## Zoom 16

Zoom 16 is almost unfiltered, but we do now drop a few of the very small polygons.

z16-after

# Changes

1. Building size selections for zoom now more closely match the criteria in the style.
2. Merging was previously done by slice, which effectively was a random subset of the polygons. This is now done by quad tree decomposition, so nearby polygons are grouped together and are much more likely to merge.
3. Building heights are now quantized at 10m intervals rather than 5m at zooms 14 & 15, and at 20m intervals for zoom 13.
4. We drop more properties on buildings (`short_name`, `osm_relation`) to enable better merging.
5. We drop small courtyards and other inners within buildings.

Overall, this leads to much less detail in the buildings layer, but this detail isn't really visible at zooms 14 and 15 anyway. The plus side is that we now include many more "city block" polygons to give better visual texture to cities.


nvkelso:
Nice! Huge file savings :)

For zoom 15 can I see a variant where we don’t combine the city buildings
in a “block”?

I do like the effect, but it tends to deemphasize the real world large
buildings. Are the real big buildings still notes so they could be called
out?

For zoom 16 historically we didn’t do dropping at this max zoom, can you
add details on this change? I’m curious how it effects the flower sculpture
building near the ferry building in SF and tiny backyard sheds in Daily
City south of SF.

zerebubuth:
> For zoom 15 can I see a variant where we don’t combine the city buildings in a “block”?

Sure! But I'm not sure what you want to see. Some options:

* A version where we do no merging at all?
* A version where we merge without buffering?
* A version where we quantize the heights at finer granularity (which will prevent some merging)?

> I do like the effect, but it tends to deemphasize the real world large buildings. Are the real big buildings still notes so they could be called out?

Personally, I think that's a good trade-off for getting a better visual sense of the building density.

Perhaps we could de-emphasise the "city blocks" in the style - for example, to use a less contrasting outline colour for buildings less than 50m in height or similar. Here's the view at zoom 15.9, and the tall buildings still stand out a lot from the background.

z15 9

> For zoom 16 historically we didn’t do dropping at this max zoom, can you add details on this change?

The changes should only have affected the `min_zoom` allocation, and in fact should result in more stuff at z16 instead of z17:

```diff
   - &z16_area_volume
-      any:
-        way_area: { min: 25 }
-        volume: { min: 50000 }
+      way_area: { min: 10 }
```

I'll go and check why there was a change in size at z16 - I wonder if perhaps I did the old/new test with different dates.

queries.yaml
```diff
@@ -975,19 +958,77 @@ post_process:
...
         - name
+        - short_name
         - addr_housenumber
         - addr_street
-      max_merged_features: 1000
+        - osm_relation
+      # NOTE: max_merged_features is set to keep the time taken for geometry
+      # merging down, as it seems to go up with the square of the number of
+      # features.
+      max_merged_features: 2000
```
rmarianski: Wouldn't this mean that the size of the tile would stay larger than we'd like though? And would we want to try another strategy in these cases, like drop some buildings from even appearing that don't rank high enough (not sure this is possible), or increase the buffer size we are using? Maybe it's smarter to leave it and just track this somehow so after runs we can see which tiles triggered the max?

nvkelso:
Ah, that’s a good tip Re 50 meter height, that’s probably enough - good
demo image! I’d probably go farther and remove extrusion on those entirely
and stack under roads and then it’s fine client side.

Can you add screenshot sequences for SF and DC and Paris, please? They have
different urban patterns.

I wonder if for zoom 15 we should do larger dilation on the buffer to
reduce exterior details, and use large px value for dropper inners?

queries.yaml
```diff
@@ -975,19 +958,77 @@ post_process:
...
+
+  # drop small inners in buildings after we do the merge, too! because the merge
+  # process may also have created small inners.
+  - fn: vectordatasource.transform.drop_small_inners
+    params:
+      end_zoom: 15
+      source_layers: [buildings]
+      pixel_area: 0.25
```
rmarianski: Great idea!

queries.yaml
```diff
@@ -975,19 +958,77 @@ post_process:
...
+  # assign `scale_rank` - AFTER merging, as this changes the area and volume
+  - fn: vectordatasource.transform.csv_match_properties
+    resources:
+      matcher:
+        type: file
+        init_fn: vectordatasource.transform.CSVMatcher
+        path: spreadsheets/scale_rank/buildings.csv
+    params:
+      source_layer: buildings
+      target_value_type: int
```
rmarianski: Nice catch here.

queries.yaml
```diff
@@ -975,19 +958,77 @@ post_process:
...
+      start_zoom: 13
+      end_zoom: 14
+      where: scale_rank > 2
+
+  - fn: vectordatasource.transform.drop_features_where
+    params:
+      source_layer: buildings
+      start_zoom: 14
+      end_zoom: 15
+      where: scale_rank > 3
```
rmarianski: This makes sense to drop afterwards, but is it crazy to initially drop some of the smaller buildings too to cut down on the candidates we need to consider for merging? Or is this actually counterproductive because it might cut down on the merging that we can actually do? Does the drop smaller inners handle this well enough anyway?

vectordatasource/transform.py
```diff
@@ -2956,6 +2959,11 @@ def _thaw(thing):
 
 
 def quantize_val(val, step):
+    # special case: if val is very small, we don't want it rounding to zero, so
+    # round the smallest values up to the first step.
+    if val < step:
+        return int(step)
+
```
rmarianski: Great catch!

vectordatasource/transform.py
```diff
@@ -2976,27 +2988,281 @@ def _merge_lines(linestring_shapes):
...
+    This method is recursive, and will bottom out after 5 levels deep, which
+    might mean that sometimes more than shapes_per_merge items are merged at
+    once.
+    """
+
+    assert isinstance(merger, RecursiveMerger)
+
+    # don't keep recursing. if we haven't been able to get to a smaller number
+    # of shaped by 5 levels down, then perhaps there are particularly large
```
rmarianski: `# of shapes`

vectordatasource/transform.py
```diff
@@ -3078,6 +3349,11 @@ def merge_building_features(ctx):
     if end_zoom is not None and zoom >= end_zoom:
         return None
 
+    # this formula seems to give a good balance between larger values, which
+    # merge more but can merge everything into a blob if too large, and small
+    # values which retain detail.
+    tolerance = min(5, 0.4 * tolerance_for_zoom(zoom))
```
rmarianski: :zap:

bcamper:
Excited to see more work here!

What's up with the (over-)simplification on that z15.9 screen? Existing or unrelated?

screen shot 2018-10-17 at 12 01 44 pm

And I'm sorry you remind me again about "nominal" zoom :) Is that the "display" zoom, or the underlying 512px tile zoom?

My main interest here is ensuring that we have the full detail (for some definition) at the max tile zoom (sounds like that is addressed in comments).

But to confirm, this would drastically reduce the buildings in a z14 view like this, right?
https://tangrams.github.io/tangram/#14.080000000000013/40.669452032003285/-73.97878168006201
tangram-1539792843724

nvkelso:
> What's up with the (over-)simplification on that z15.9 screen? Existing or unrelated?

Oops. That's an new error we should fix in this PR.

> But to confirm, this would drastically reduce the buildings in a z14 view like this, right?

Yes, it's just too much data and file size at zoom 14, even after doing the generalization we see in zoom 15 example by combining the buildings in a city block.

zerebubuth:
> What's up with the (over-)simplification on that z15.9 screen? Existing or unrelated?

I think that's partly existing, although highlighted by the larger number of visible buildings now:

z15 9-comparison

However, there are some differences: for example, the two "H" shaped buildings above and to the right of the bend in Avenida Professor Alfonso Bovero (above where "Bovero" is in the right image). The new tile has a mangled version and the master version looks better. There are a bunch of places where they look about as bad as eachother, though.

bcamper:
Yep, I've liked having higher building density available (and I think this is still considerably more at z15 than Mapbox?), but totally agree tile size needs to come down for typical base map cases (though may indicate the future value of a couple different tile flavors with more/less feature density, or you know, true "raw" tiles...). I am interested to also see if there is a noticeable improvement in tile build time from reduced feature processing (even in the cases where they are not displayed)! Let me know when an endpoint is available :)

zerebubuth:
I re-ran the last run with a clean working copy. Looks like I must have had some local mods I'd forgotten about, because the results are different. (Although I'm going to check again, to make sure this is reproducible).

I'll also do some runs in SF, Washington DC & Paris.

Tile coordinate (512 size) | Size before (bytes) | Size after (bytes) | Reduction
--- | --- | --- | ---
`13/3033/4647` | 442,333 | 3,012 | 99.3%
`14/6067/9294` | 587,743 | 71,634 | 87.8%
`15/12134/18589` | 384,337 | 388,012 | -1.0% (i.e: it's larger)

## Zoom 14 ##

z14-clean-pr-181023

## Zoom 15 ##

z15-clean-pr-181023

## Zoom 16 ##

z16-clean-pr-181023

zerebubuth:
# San Francisco

San Francisco does pretty well. Like São Paulo, there's a lot of "blocks" which can be merged at zoom 15 nominal.

Overall results:

Tile coordinate (512 size) | Nominal zoom | Size before (bytes) | Size after (bytes) | Reduction
-- | -- | -- | -- | --
`13/1310/3165` | 14 | 136,234 | 12,221 | 91%
`14/2620/6331` | 15 | 368,384 | 52,801 | 86%
`15/5241/12663` | 16 | 154,484 | 154,410 | 0%

## Zoom 14

z14-sf

## Zoom 15

z15-sf

## Zoom 16

z16-sf

# Washington DC

Downtown DC has a _lot_ of big buildings, most of which don't need any kind of merging. Outside those blocks of large buildings, the other blocks also seem to merge well, until we reach the suburbs.

Overall results:

Tile coordinate (512 size) | Nominal zoom | Size before (bytes) | Size after (bytes) | Reduction
-- | -- | -- | -- | --
`13/2343/3134` | 14 | 65,353 | 8,357 | 87%
`14/4687/6268` | 15 | 189,200 | 24,003 | 87%
`15/9375/12536` | 16 | 235,878 | 237,319 | -1%


## Zoom 14

z14-dc

## Zoom 15

z15-dc

## Zoom 16

z16-dc

# Georgetown Interlude

Just outside DC, in the evocatively named Foxhall Village, there are a lot of detached houses, which don't get merged at all. However, it looks like we're still getting a decent reduction in size (about 80% rather than 90%, but coming from a lower base), perhaps due to more aggressive geometry simplification.

The lack of individual large buildings means the buildings layer for this tile is almost empty at zoom 14 nominal.

Overall results:

Tile coordinate (512 size) | Nominal zoom | Size before (bytes) | Size after (bytes) | Reduction
-- | -- | -- | -- | --
`13/2341/3133` | 14 | 90,342 | 1,981 | 98%
`14/4683/6266` | 15 | 96,979 | 20,881 | 78%
`15/9367/12533` | 16 | 82,509 | 82,482 | 0%

## Zoom 15

z15-dc-georgetown

## Zoom 16

z16-dc-georgetown

# Paris

Paris seems to have a lot of buildings with inner courtyards or other non-rectangular features which result in many inners in the merged blocks at zoom 15 nominal. Perhaps the tolerance for dropping small inners needs to be increased to remove most of these?

Overall results:

Tile coordinate (512 size) | Nominal zoom | Size before (bytes) | Size after (bytes) | Reduction
-- | -- | -- | -- | --
`13/4149/2817` | 14 | 253,303 | 17,329 | 93%
`14/8298/5635` | 15 | 273,873 | 62,551 | 77%
`15/16597/11271` | 16 | 207,522 | 208,308 | 0%

## Zoom 14

z14-paris

## Zoom 15

z15-paris

## Zoom 16

z16-paris

nvkelso:
Overall this feels like a good improvement and I think we should go forward with it as-is... **after confirming** we're not regressing zoom 16 as the Ferry Building in SF is missing it's tower and DC buildings look generalized... and file a followup issue to drop small courtyards at zoom 15.

For Brazil zoom 15, it still feels like we're cluttering up the tile with "fake" low rise buildings:

image

In San Francisco also at zoom 15: I see some merging... but overall too many small features:

image

**At zoom 16 in SF we seem to have lost the tower detail on the Ferry Building right along the water?!** Please ensure there's not an off-by-one problem here before merging.

image

In DC at zoom 15 it is aggregating row houses like I'd expect, so that's a vote to keep it as proposed in this PR:

image

Zoom 16 in DC also feels over generalized, did that change in this PR?

image

Zoom 15 in Georgetown does look nice!

image

Zoom 15 in Paris feels like we could stand dropping small inner courtyards, but otherwise looks good:

image

queries.yaml
```diff
@@ -975,19 +958,77 @@ post_process:
   - fn: vectordatasource.transform.drop_properties_with_prefix
     params: {prefix: mz_}
 
+  # drop small inners in buildings before we do the merge.
+  - fn: vectordatasource.transform.drop_small_inners
+    params:
+      end_zoom: 15
+      source_layers: [buildings]
+      pixel_area: 0.25
```
nvkelso: Could make this 0.5 or 1? But let's track that with a new issue and merge this as-is.

queries.yaml
nvkelso: How is this different than line 962? We should only do this procedure once.

queries.yaml
nvkelso: It's more "coherent" to operate on the entire urban fabric... but agree it can result in large aggregate polygons that would otherwise have been dropped as smaller inputs. Generally they'll still have low height and can be filtered client-side, or we could later drop them server-side if we need more file size savings.

vectordatasource/transform.py
nvkelso: Good catch of typo

zerebubuth: Yup, well spotted, thanks! Fixed in 937023cb.

zerebubuth:
> Investigate if we're dropping small inner courtyards twice now

We are, deliberately. We have one pass of dropping inners before the merge and another after the merge. The first is just to make life simpler for the merge; some things which were multipolygons are now simple polygons, which is much easier (and computationally cheaper) for the dilate / buffer algorithm to handle. The pass after merging is to because the merging process itself sometimes creates courtyards when merging adjacent buildings (e.g: neighbouring buildings like `[]` get merged to `□`).

nvkelso:
And a closer read of the comments say that, sorry for the noise!

zerebubuth:
Follow-up for a re-visit to drop more courtyards in #1696.

zerebubuth:
PS: We also talked about the Ferry Building, and how the tower isn't visible in the new tiles. This is because the building footprint has `height=83.1`, which is the height of the whole building including the tower, so it covers it up. The building used to have `height=16.1`, but this has been more recently edited to split it into a `building:part` with that height.

We've seen a similar issue before with the Fernsehturm in Berlin #1274 and a sort-of similar issue with the Bank of England #1032. And it looks like the plan would be to add the `root_id` on the building as well, and strip off the height. Although we might want to check first to see that the `building:part`s cover the footprint of the building.

Not adding a follow-up for this, as I think #1274 has it covered.

Issue #1687, opened 2018-10-11
Fewer places at low zoom.


zerebubuth:
The `0/0/0` tile contains:

* 91 countries,
* 744 regions,
* 25 localities.

We couldn't possibly render all of these in a single tile, so we should remove some of them. The properties on these features are the major contributor to size of tiles at zooms 3 and lower.

nvkelso:
From https://github.com/tilezen/vector-datasource/issues/1684:

Let's use the min_zoom on country, region, and locality features from Natural Earth to drop labels.

In particular country labels don't all need to be shown at low zooms, just a few handful. There's a **label min_zoom** separate from feature min_zoom for this.

nvkelso:
For 1/41/-74...

New:

![image](https://user-images.githubusercontent.com/853051/49972101-7e9d3d00-fee5-11e8-9b37-4cb0e7ea5f59.png)

Old:

![image](https://user-images.githubusercontent.com/853051/49972127-85c44b00-fee5-11e8-94be-cefa291c14c9.png)

nvkelso:
**0/43/-73**

For zoom 0 we're still including too many countries, I'll file followup issue. Seems like we're doing 512 min_zoom math instead of 256 math? Like at zoom 0 we should only see features with min_zoom < 1, but many 1.x graded features are included.

New:
![image](https://user-images.githubusercontent.com/853051/49972295-10a54580-fee6-11e8-98d7-8dc8e84df513.png)


Old:
![image](https://user-images.githubusercontent.com/853051/49972286-07b47400-fee6-11e8-8baa-fa0a3b3b50f6.png)

![image](https://user-images.githubusercontent.com/853051/49972360-40544d80-fee6-11e8-9a52-125b52dbaacb.png)
`

nvkelso:
Generally this is working to significantly reduce file size... but we're still over stuffed. Followup in https://github.com/tilezen/vector-datasource/issues/1729.

For zoom 0 we're still including too many countries. The same is true for zooms 1 and 2. Addressing this will have significant file size savings.

Seems like we're doing 512 min_zoom math instead of 256 math? Like at zoom 0 we should only see features with min_zoom < 1, but many 1.x graded features are included.

New:
![image](https://user-images.githubusercontent.com/853051/49972295-10a54580-fee6-11e8-98d7-8dc8e84df513.png)


Old:
[...]

Issue #1643, opened 2018-09-28
Add option to strip out all but core set of name translations


nvkelso:
We have ~21 core Tilezen languages but there are many more than that in OSM.

There's also a subset of "UN" languages to think about:

- `name:en`  // English
- `name:es`  // Spanish
- `name:fr`  // French
- `name:de`  // German
- `name:ru`  // Russian
- `name:zh`  // Chinese
- `name:ar`  // Arabic
- `name:pt`  // Portuguese

Add option to strip out all but core set of name translations

PR #1693, opened 2018-10-25
Use the NE min zoom to cull places at low zooms.

zerebubuth:
We were already using the NE min zoom as the min_zoom property, but had failed to actually act on that by removing items outside of the zoom range.

Overall file size reduction (NOTE: most of the very low zoom labels come from the `earth` layer - things like continent labels):

Tile | Previous places size (bytes) | New places layer size (bytes) | Percentage improvement
-- | -- | -- | --
0/0/0 | 518,111 | 0 | 100%
1/0/0 | 139,980 | 0 | 100%
1/0/1 | 42,043 | 0 | 100%
1/1/0 | 266,788 | 0 | 100%
1/1/1 | 83,228 | 0 | 100%
2/0/0 | - | - | -
2/0/1 | 33,015 | 14,536 | 56%
2/0/2 | 20,091 | 1,979 | 90%
2/0/3 | - | - | -
2/1/0 | 4,326 | 4,091 | 5%
2/1/1 | 178,337 | 44,582 | 75%
2/1/2 | 58,487 | 20,020 | 66%
2/1/3 | - | - | -
2/2/0 | 2,145 | 0 | 100%
2/2/1 | 410,712 | 110,756 | 73%
2/2/2 | 83,789 | 32,137 | 62%
2/2/3 | - | - | -
2/3/0 | 827 |   | 100%
2/3/1 | 155,139 | 44,066 | 72%
2/3/2 | 44,228 | 21,546 | 51%
2/3/3 | - | - | -

## Zoom 0

It's all continent and ocean labels, no places visible.

z0

## Zoom 1

It's all continent and ocean labels, no places visible.

z1

## Zoom 2

I was expecting _some_ places to be visible at this zoom level, as some are present in the tiles. Perhaps the label collision prevents any from appearing?

z2

## Zoom 3

Looks like a pretty good selection of capitals, major cities and country labels.

z3

## Zoom 4

Nice overview of US states and major cities. Do we have the same state data for Canada - given that we show the state borders, it might be nice to label them too? Perhaps a :umbrella: issue?

z4

### DIFF
--- a/integration-test/1687-fewer-places-at-low-zoom.py
+++ b/integration-test/1687-fewer-places-at-low-zoom.py
@@ -1,0 +1,45 @@
+# -*- encoding: utf-8 -*-
+from . import FixtureTest
+
+
+class LowZoomPlacesTest(FixtureTest):
+
+    def test_zoom_1(self):
+        import dsl
+
+        z, x, y = (3, 7, 3)
+
+        self.generate_fixtures(
+            dsl.way(607976629, dsl.tile_centre_shape(z, x, y), {
+                "min_zoom": 1,
+                "__ne_max_zoom": 10,
+                "__ne_min_zoom": 3,
+                "area": 0,
+                "place": "country",
+                "name": "Guam",
+                "population": 185427,
+                "source": "openstreetmap.org",
+            }),
+        )
+
+        # should exist at zoom 3 (the min zoom from NE)
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 607976629,
+                'kind': 'country',
+                'name': 'Guam',
+            })
+
+        # should exist at zoom 2 (one past the min zoom)
+        self.assert_has_feature(
+            z-1, x//2, y//2, 'places', {
+                'id': 607976629,
+                'kind': 'country',
+                'name': 'Guam',
+            })
+
+        # should not exist at zoom 1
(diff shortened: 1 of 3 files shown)


zerebubuth:
Connects to #1687.

Note that tile `2/2/1` is still pretty huge - 110,756 bytes. However most (69%) of this is properties, which will get better once we implement #1643.

nvkelso:
Here's what I see in Natural Earth for these zooms:

screen shot 2018-10-25 at 09 54 31

screen shot 2018-10-25 at 09 54 47

screen shot 2018-10-25 at 09 55 01

screen shot 2018-10-25 at 09 55 35

nvkelso:
Also, your zoom 2 looks pretty weird!?

nvkelso:
The likely reason you're not seeing Canada region labels at zoom 3 is they are set to 3.5 and Tangram only does the evaluation at zoom 3 (and 3.5 !< 3 so it filters them out). Or your PR / existing code filters them.

zerebubuth:
Yup, zoom 2 was messed up because I was using the Go version of `tapalcatl`, which doesn't understand the "max zoom packing" feature/hack that we added to `tapalcatl-py`. Moving to the latter, I now see:

z2-again

Which is has a bunch of cities on it. :tada: 

(Although no space for London? What's up with that? :scream_cat:)

Issue #1642, opened 2018-09-25
Drop all_networks and all_shield_texts from low and mid zooms


nvkelso:
To decrease file size let's drop `all_networks` and `all_shield_texts` from low and mid-zooms when there's often not room to display this information. 

- highway: less than zoom 10
- major_road: less than zoom 12
- other kinds: less than zoom 14

NOTE: MVT format currently doesn't include the array objects while GeoJSON and TopoJSON do.

nvkelso:
No obvious regressions (in MVT) here, closing.

(Would later yield savings if we write out nested props as strings into MVT like some software does.)

PR #1694, opened 2018-10-26
Drop the all_networks and all_shield_texts at mid zooms

zerebubuth:
Drop the `all_networks` and `all_shield_texts` properties from roads at various zooms. Connects to #1642.


zerebubuth:
> Any file size difference to note?

Finally got around to measuring this on a zoom range over San José, but sadly there wasn't as much change as I'd hoped: About a 0.4% saving on the uncompressed GeoJSON tile. Since we don't export list-valued properties in MVT, there was no change in size there.

Issue #1695, opened 2018-10-28
Include turning_circle(, turning_loop, mini_roundabout)


westnordost:
In OSM, there are 
* [`highway=turning_circle`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dturning_circle)
* [`highway=turning_loop`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dturning_loop)
* [`highway=mini_roundabout`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmini_roundabout)

These features all have in common that they can be represented as a circular slub in the road-way at the node position where it is tagged.

On the default mapnik map, the first two look like this.
![](https://wiki.openstreetmap.org/w/images/8/8e/Mapping-Features-Turning-Circle.png)

In the vector-datasource though, they are not included, so it cannot be rendered. I think it would make sense to include it in one way or another because it is useful to see if one can easily turn at the end of dead end street or not.

nvkelso:
Sounds good! We already have `mini_roundabout` in the **pois** layer: https://github.com/tilezen/vector-datasource/blob/767cc230410996a3974b146fa621ec2ab99e8751/yaml/pois.yaml#L1354-L1358

Can you submit a PR for the other two? Probably a min_zoom of `16` since they are terminal and not in normal traffic flow? Since they don't have names you'd need to also extend the list of highway PONI's in:

https://github.com/tilezen/vector-datasource/blob/767cc230410996a3974b146fa621ec2ab99e8751/yaml/pois.yaml#L191-L192

westnordost:
I do not fully understand the configuration file, but I'll try. Not sure if I have to do more than this:
https://github.com/tilezen/vector-datasource/pull/1697

nvkelso:
PR looks great! We can continue work thru the test part more in that PR together.

We're working on the v1.6 milestone and this change would go out in the next Nextzen prod build once that finalizes (but may be a month or two).

nvkelso:
This all should probably be at **zoom 17** not 16 or (and mini_roundabouts at 17 not 15, to match OSM style). They'd still be overstuffed into tiles at zoom 16, though.

18/37.75586/-122.47092

![image](https://user-images.githubusercontent.com/853051/49976314-42260d00-fef6-11e8-9e17-96f8efaeb722.png)

19/37.77639/-122.44749

[...]

PR #1697, opened 2018-10-29
Add turning circles and turning loops to pois.yaml

westnordost:
Implements #1695

- [x] Update tests
- [x] Update docs

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1160,6 +1160,8 @@ To resolve inconsistency in data tagging in OpenStreetMap we normalize several o
 * `travel_agency`
 * `travel_agent`
 * `tree`
+* `turning_circle`
+* `turning_loop`
 * `tyres` - A shop selling car tyres or tires.
 * `university`
 * `variety_store`
--- a/integration-test/1695-turning-circles-turning-loops.py
+++ b/integration-test/1695-turning-circles-turning-loops.py
@@ -13,7 +13,7 @@ class TurningCirclesAndLoops(FixtureTest):
         self.assert_has_feature(
             16, 32703, 21771, 'pois',
             {'id': 106186562, 'kind': 'turning_circle'})
-            
+
     def turning_loop(self):
         self.generate_fixtures(dsl.point(4260010359, (8.43452, 49.4596352),
             {u'source': u'openstreetmap.org', u'highway': u'turning_loop'}))
--- a/integration-test/1695-turning-circles-turning-loops.py
+++ b/integration-test/1695-turning-circles-turning-loops.py
@@ -7,16 +7,20 @@ from . import FixtureTest
 class TurningCirclesAndLoops(FixtureTest):
 
     def turning_circle(self):
-        self.generate_fixtures(dsl.point(106186562, (-0.3544854, 51.5785667),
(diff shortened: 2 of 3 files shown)


nvkelso:
Nice! Don't need to do migrations anymore. See some of the recent PRs for generative test, or use the older format which is still supported (eg with an OSM id).

westnordost:
Not sure what tests I would need to update.

nvkelso:
You'll need to add a test. Generally we name the test file after the issue number and a descriptive string and place into the **integration-test/** dir (like `1695-turning-circles-turning-loops.py`): 

- https://github.com/tilezen/vector-datasource/blob/master/integration-test/1186-the-pois-with-no-name.py

Then you'd have something like, but modify so it covers both new kinds, and make sure that's the right coordinate for the zoom at lat/lng position as I hacked it together for you.

```
# -*- encoding: utf-8 -*-
from shapely.wkt import loads as wkt_loads
import dsl
from . import FixtureTest


# test turning circles and ...
class TurningCirlce(FixtureTest):

    def turning_circle(self):
        # comment
        self.generate_fixtures(dsl.way(1589837084, wkt_loads('POINT (-104.819282920265 39.61691951649848)'), {u'source': u'openstreetmap.org', u'highway': u'turning_circle'}))  # noqa

        self.assert_has_feature(
            16, 10461, 25315, 'pois',
            {'id': 1589837084})
```

westnordost:
I do have a problem with this test setup, it seems to test much more than it should:
- It appears as if an actual OSM element is tested here, but I think this is not the case. (Still I entered a "real" one now, inclusive the tile numbers, meh)
- both the position and the tile numbers are irrelevant for the test and just adds noise + difficulty to implement (the implementer needs to find out in which tile number a given position is)

Ideally, I would want to do something like this:
```python
    def turning_circle(self):
        self.assert_convert_tags_at_zoom_level(
                16,
                {'highway': 'turning_circle'},
                {'kind': 'turning_circle'}
        )
```

nvkelso:
I agree that the XY part of the ZXY coordinate shouldn't need to be manually noted by default... but we've run into cases in the past where it's important to specify the exact tile, especially for non-point features. In any case the layer also has to be specified because some features show up in 2 layers not just one.

westnordost:
Sure, correct assignments of geometry to tiles should also have its test, but I'd say it should be a designated one. But anyway, that was only a side-note as a suggestion for improvement.
Regarding this PR, is there anything left to do?

nvkelso:
@westnordost Circle says something is busted in the checks, looks like it's all line length and whitespace issues:

```
#!/bin/bash -eo pipefail
. ~/env/bin/activate
find . -not -path '*/.eggs/*' -not -path '*/data/*' -name '*.py' | xargs flake8
./integration-test/1695-turning-circles-turning-loops.py:10:80: E501 line too long (144 > 79 characters)
./integration-test/1695-turning-circles-turning-loops.py:15:1: W293 blank line contains whitespace
./integration-test/1695-turning-circles-turning-loops.py:17:80: E501 line too long (140 > 79 characters)
Exited with code 123
```

Please clean those up and then if Circle passes we should be all good to go.

westnordost:
I can do that, though you should know that I copied the style as I saw it in other tests.

PR #1699, opened 2018-10-31
Don't include buffered land at zooms it's not used in

zerebubuth:
The admin boundaries processing only starts at zoom 8, so at best it would be superfluous to include buffered land before that. As it turns out, this meant we were including the buffered land polygons in the output tiles at all lower zooms.

This drastically reduces the size of the boundaries layer for low zoom tiles:

Zoom | X | Y | Size before (bytes) | Size after (bytes) | Size reduction
-- | -- | -- | -- | -- | --
6 | 18 | 24 | 5,599 | 1,653 | 70%
5 | 9 | 12 | 6,964 | 1,585 | 77%
4 | 4 | 6 | 22,244 | 9,195 | 59%
3 | 2 | 3 | 53,328 | 6,446 | 88%
2 | 1 | 1 | 102,871 | 13,097 | 87%
1 | 0 | 0 | 148,128 | 20,722 | 86%
0 | 0 | 0 | 230,801 | 10,700 | 95%

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -146,7 +146,7 @@ sources:
   # note: this goes into the _boundaries_ layer.
   buffered_land:
     - template: buffered_land.jinja2
-      start_zoom: 0
+      start_zoom: 8
 
 layers:
   water:


nvkelso:
LGTM but don't we drop buffered land completely in a later transform? Like it should never export in any zoom tiles?

zerebubuth:
> don't we drop buffered land completely in a later transform?

That's the case with the admin areas that we use for country code assignment, but sadly not the case for the buffered land.

Issue #1683, opened 2018-09-29
Strip names off most boundary lines at low and mid-zooms


nvkelso:
We include default and localized `name:left` and `name:right` properties on boundary lines when the aren't used for labeling until later zooms. This bloats file size.

There are also hints in Natural Earth about the zoom they should start being labeled we can leverage.

nvkelso:
SEO: boundaries layer

We could also follow the lead in https://github.com/tilezen/vector-datasource/pull/1693 and drop features and names (name and name:*) based on the NE min_zoom and min_label?

nvkelso:
Looking around Switzerland I still see a lot of name variants that aren't labeled but there in the data, but overall this is working (as proved by the reduction in the boundary layer size) so I think it's just a parameter tuning opportunity in the next release, with possible incorporation of the Natural Earth zooms as it's not necessary to include both the label centroid and the line names at most zooms together.

8/47.132/8.432

![image](https://user-images.githubusercontent.com/853051/49976997-49024f00-fef9-11e8-908d-7fb53a71dbb4.png)

Like:

![image](https://user-images.githubusercontent.com/853051/49977023-66cfb400-fef9-11e8-8712-33eed68f6a2c.png)

nvkelso:
This is mostly working, closing.

PR #1700, opened 2018-11-02
Strip long names from boundaries.

zerebubuth:
This strips names off boundaries between zooms 8 and 10 when they're too long to be reasonably rendered in the tile, based on some assumption about the scale at which we'll be rendering text. This can help remove many names on very short segments of boundary that we'd never have been able to render anyway.

Additionally, merges boundary lines. We weren't doing this previously, so would have a few fragmented boundaries which were duplicating properties and inflating feature count.

Some quick tests show we're reducing GeoJSON boundaries layer size by around 23% on tile `8/134/89` and MVT boundaries layer size by 9% (because MVT is already deduplicating the name values there's less opportunity for us to save.)

One observation: Many of the names which are using a lot of bytes are not long in terms of number of rendered characters, but the characters take up more space encoded as UTF-8. It would be interesting to see how many of these we could compress away as transliterations, but that would require client-side support.

Connects to #1683.

### DIFF
--- a/integration-test/976-fractional-pois.py
+++ b/integration-test/976-fractional-pois.py
@@ -16,9 +16,10 @@ class FractionalPois(FixtureTest):
             'https://www.openstreetmap.org/relation/61320',
         ], clip=self.tile_bbox(9, 150, 192, padding=2))
 
+        # NOTE: might not have an ID if it has been merged
         self.assert_has_feature(
             9, 150, 192, 'boundaries',
-            {'min_zoom': 8, 'id': -224951,
+            {'min_zoom': 8,
              'source': 'openstreetmap.org',
              'name': 'New Jersey - New York'})
 
--- a/integration-test/992-boundaries-min_zoom-and-name.py
+++ b/integration-test/992-boundaries-min_zoom-and-name.py
@@ -127,13 +127,15 @@ class BoundariesMinZoomAndNameNe(FixtureTest):
 class BoundariesMinZoomAndNameOsm(FixtureTest):
     def test_region_boundary_zug_luzern(self):
         # Switzerland region HAS name, OpenStreetMap
+        # do this at z12, as the boundary between Zug and Luzern is quite
+        # short, and we want enough space to label.
         self.load_fixtures([
             'http://www.openstreetmap.org/relation/1686447',
             'http://www.openstreetmap.org/relation/1685677',
-        ], clip=self.tile_bbox(8, 133, 89))
+        ], clip=self.tile_bbox(12, 2144, 1438))
 
         self.assert_has_feature(
-            8, 133, 89, 'boundaries',
+            12, 2144, 1438, 'boundaries',
             {'kind': 'region', 'name': 'Zug - Luzern'})
 
     def test_region_boundary_salzburg_tirol(self):
--- a/integration-test/992-boundaries-min_zoom-and-name.py
+++ b/integration-test/992-boundaries-min_zoom-and-name.py
@@ -125,7 +125,20 @@ class BoundariesMinZoomAndNameNe(FixtureTest):
 
 
 class BoundariesMinZoomAndNameOsm(FixtureTest):
-    def test_region_boundary_zug_luzern(self):
+    def test_region_boundary_zug_luzern_z8(self):
+        # Switzerland region HAS NO name, OpenStreetMap
+        self.load_fixtures([
+            'http://www.openstreetmap.org/relation/1686447',
+            'http://www.openstreetmap.org/relation/1685677',
+        ], clip=self.tile_bbox(8, 133, 89))
+
+        # test that the regional boundary is present at zoom 8, although it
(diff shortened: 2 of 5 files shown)


integration-test/976-fractional-pois.py
```diff
@@ -16,9 +16,10 @@ def test_state_boundary(self):
             'https://www.openstreetmap.org/relation/61320',
         ], clip=self.tile_bbox(9, 150, 192, padding=2))
 
+        # NOTE: might not have an ID if it has been merged
```
nvkelso: 👍 thanks for adding this note!

integration-test/992-boundaries-min_zoom-and-name.py
```diff
@@ -127,13 +127,15 @@ def test_10m_regions(self):
...
+        # do this at z12, as the boundary between Zug and Luzern is quite
+        # short, and we want enough space to label.
         self.load_fixtures([
             'http://www.openstreetmap.org/relation/1686447',
             'http://www.openstreetmap.org/relation/1685677',
-        ], clip=self.tile_bbox(8, 133, 89))
+        ], clip=self.tile_bbox(12, 2144, 1438))
 
         self.assert_has_feature(
-            8, 133, 89, 'boundaries',
```
nvkelso: I think we should still keep the zoom 8 test but for `kind: region` only as that proves we are taking the OSM data correctly at zoom 8.

queries.yaml
```diff
@@ -592,11 +592,29 @@ post_process:
...
+      start_zoom: 8
+      end_zoom: 15
+
+  # drop labels on boundaries which are too short to render.
+  - fn: vectordatasource.transform.drop_names_on_short_boundaries
+    params:
+      source_layer: boundaries
+      start_zoom: 8
+      end_zoom: 11
+      factor: 11
```
nvkelso: Please add comment about `factor`, what it does, and it's units.

vectordatasource/transform.py
```diff
@@ -1959,6 +1959,93 @@ def admin_boundaries(ctx):
...
 
+def _unicode_len(s):
+    if isinstance(s, str):
+        return len(s.decode('utf-8'))
+    elif isinstance(s, unicode):
+        return len(s)
+    return None
+
+
+def _delete_labels_longer_than(max_label, props):
```
nvkelso: is `max_label` = `max_length`, both terms are mentioned here but I think only one is used in practice.

vectordatasource/transform.py
```diff
@@ -1959,6 +1959,93 @@ def admin_boundaries(ctx):
...
+                shape_length = label_shape.length
+            else:
+                # get the longest section to see if that's labellable - if
+                # not, then none of the sections could have a label and we
+                # can drop the names.
+                shape_length = max(part.length for part in label_shape)
+
+            # maximum number of characters we'll be able to print at this
+            # zoom.
+            max_label = int(shape_length / tolerance)
```
nvkelso: should this be called `max_length` (or even `max_chars`)? 

Please add comment about what units these are in... logical pixels? meters? characters?

vectordatasource/transform.py
```diff
@@ -1959,6 +1959,93 @@ def admin_boundaries(ctx):
...
+    factor = params.optional('factor', typ=(int, float), default=10.0)
+
+    layer = _find_layer(ctx.feature_layers, layer_name)
+    zoom = ctx.nominal_zoom
+
+    if zoom < start_zoom or \
+       (end_zoom is not None and zoom >= end_zoom):
+        return None
+
+    tolerance = factor * tolerance_for_zoom(zoom)
```
nvkelso: Please add comment for what type of units this results in?

integration-test/992-boundaries-min_zoom-and-name.py
zerebubuth: Good idea. Fixed in fb5565c.

queries.yaml
zerebubuth: Added comment and changed to a more descriptive name in a75fd59.

vectordatasource/transform.py
zerebubuth: Well spotted, thanks! I think that was a typo / thinko. Fixed in a75fd59.

vectordatasource/transform.py
zerebubuth: I've changed this to `max_label_chars`, added some comments and made other variable names more descriptive in a75fd59 - hopefully that makes it clear enough?

vectordatasource/transform.py
zerebubuth: Changed this to `meters_per_letter`, since that's what it is - Mercator meters per letter / character / grapheme cluster in the string. The factor was roughly how many pixels to expect the width of each letter to be, I've found that 11 is a pretty good value by experimentation, but we should see what the results of a build look like and tweak it.

Issue #1227, opened 2017-04-27
Improve road/line merging II


nvkelso:
Following up from improvements in https://github.com/tilezen/vector-datasource/issues/1191...

nvkelso:
@bcamper said in https://github.com/tilezen/vector-datasource/issues/1191#issuecomment-296692145:

Here's what I've found (using the new debug stats in https://github.com/tangrams/tangram/pull/544) for a high-complexity NYC view at z12 (#12/40.68636570811375/-73.94691467285234):

*Note: line geometry counts here are the number of **GL triangles** Tangram had to build to render the features (lines decomposed into triangles, etc.), not the number of features in the source data - but for line features, they should scale linearly together.*

**Tangram master / prod tiles**
- Total line geometries: 258,232
- roads 222,512
  - highways 23,810
  - major roads 39,304
  - minor roads **148,756**
  - railways 7,812

**Tangram master / dev tiles**
- Total line geometries: 226,079 **(-13% from prod)**
- roads 190,294
  - highways 17,884
  - major roads 21,056
  - minor roads **144,110**
  - railways 4,556

**Tangram merge-multilines / dev tiles**
- This branch does additional client-side line merging + simplification
- Total line geometries: 156,359 **(additional -30% from dev tiles, -39% from current prod)**
- roads 121,198
  - highways 15,212
  - major roads 20,374
  - minor roads **80,086**
  - railways 3,168

**Summary**
It looks like the dev tiles are effectively reducing the numbers of highways, major roads, and railways (these are just the top sub-layers), but aren't really affecting the number of *minor roads* (only 148k -> 144k in this view) - which are the vast majority at this zoom. 

For reference, the client-side code I'm using is [here](https://github.com/tangrams/tangram/blob/5bb881d29a56e8eab59ed173cec294d87b424b40/src/sources/data_source.js#L106-L191) (sorry, it's a bit more complex than it needs to be due to some premature optimization to avoid reallocating/copying LineStrings that never need to be modified). It has two steps:

1. `mergeContiguousMultiLineString()`: iterates through the LineStrings in a MultiLineString, and merges any LineStrings where the last and first coordinates are equal (iterates in their natural order, so only helps if contiguous segments also happen to be sequential in the feature data!).
[...]

Issue #1331, opened 2017-07-21
Consider dropping access=yes and related tags


nvkelso:
Picking up from https://github.com/tilezen/vector-datasource/pull/1328 it looks like we allow `access=*` and `motor_vehicle =*` at early zooms. We should either consider dropping `access=no` related features at low zooms, or at least drop `access=yes` properties from features at all (or just low) zooms to promote more feature merging to achieve smaller tile size and better label rendering. 

#### access=* related tags we use now:

- [ ] `access=*`
- [ ] `surface=*`
- [ ] `vehicle=*`
- [ ] `motor_vehicle=*`
- [ ] `bicycle=*`
- [ ] `foot:*`
- [ ] `horse=*`
- [ ] `ski=*`
- [ ] `snowmobile=*` (we [reference](https://github.com/tilezen/vector-datasource/blob/71e9e131aa7044b4cb1b31163a9931054e425809/yaml/roads.yaml#L366) snowmobile=designated but don't seem to actually export the `snowmobile` tag now, a bug?
- [ ] `snowshoe=*`

#### network related:

- [ ] cycleway network at low and mid zooms (should mostly just be the boolean `is_bicycle_related`)
- [ ] but networks (network, shields) at low and mid zooms (should just be `is_bus_route`)

matkoniecz:
>  We should either consider dropping access=no related features at low zooms

Probably also access=private.

Note that access tag may be overridden - for example [highway=track + access=no + bicycle=designated] should not be displayed later than [highway = cycleway].

matkoniecz:
motor_vehicle=yes in many cases is thoroughly useless and may be safely dropped ( https://josm.openstreetmap.de/browser/josm/trunk/data/validator/unnecessary.mapcss#L2 )

[motor_vehicle=yes][vehicle!=no][access!=no][bicycle_road!=yes][highway =~ /motorway|trunk|primary|secondary|tertiary|unclassified|residential|service|living_street/]

(used to appearing very often, still appears after years of JOSM validator rule - http://overpass-turbo.eu/s/qx5 , not sure about effects).

nvkelso:
@matkoniecz Thanks for the tips and background :)

nvkelso:
At 11/37.9675/-122.4409

![image](https://user-images.githubusercontent.com/853051/48571221-2dd0ff00-e8bb-11e8-87dd-d24e38e4ae98.png)

nvkelso:
[...]

PR #1703, opened 2018-11-09
Improve road merging

zerebubuth:
Connects to #1227.

There were two main issues with the road network:

1. The merging we were already doing would generally not merge across a 4-way junction. The new code tries to merge continuous lines having a similar bearing at the junction together. This results in fewer points.
2. Most geometry ops (including union and intersection/clipping) seem to make MultiLineStrings "simple" - which means that sibling lines within the MultiLineString do not cross and only meet each other at their endpoints. Therefore, clipping the road network to the tile boundary undoes all the work in the junction merging step. Instead, the merging step will create several MultiLineStrings, each one containing only the subset of LineStrings which don't intersect any sibling. This does mean we end up duplicating feature properties, but overall the saving is still worthwhile.

For `12/1206/1540`, a tile containing many `kind: minor_road`s over Brooklyn, these changes resulted in a 11.7% reduction in the GeoJSON layer size, 6.7% reduction in the MVT layer size and 28.9% reduction in the number of geometry commands (i.e: points) associates with `minor_road` features.

There's still plenty of scope for improving tile size, as there are many properties that are probably superfluous and could be removed at this zoom level. The image below shows the tile with a dot at the first and last vertex for each line, coloured uniquely by properties:

image

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -1127,6 +1127,15 @@ post_process:
       source_layer: roads
       start_zoom: 8
       end_zoom: 16
+      # setting the following will try to merge linestrings across junctions
+      # (i.e: more than 2 roads meeting at a point) where the angle between
+      # roads at that point is less than 5 degrees.
+      merge_junctions: true
+      merge_junction_angle: 5.0
+      # setting the following will cause lines, or parts of multi-lines,
+      # shorter than 0.1px at nominal zoom to be dropped.
+      drop_short_segments: true
+      drop_length_pixels: 0.1
 
   - fn: vectordatasource.transform.simplify_layer
     params:
--- a/queries.yaml
+++ b/queries.yaml
@@ -1132,10 +1132,6 @@ post_process:
       # roads at that point is less than 5 degrees.
       merge_junctions: true
       merge_junction_angle: 5.0
-      # setting the following will cause lines, or parts of multi-lines,
-      # shorter than 0.1px at nominal zoom to be dropped.
-      drop_short_segments: true
-      drop_length_pixels: 0.1
 
   - fn: vectordatasource.transform.simplify_layer
     params:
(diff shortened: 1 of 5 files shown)


queries.yaml
```diff
@@ -1122,19 +1122,47 @@ post_process:
...
+  # probably aren't visible anyway.
+  - fn: vectordatasource.transform.merge_line_features
+    params:
+      source_layer: roads
+      start_zoom: 8
+      end_zoom: 16
+      # setting the following will cause lines, or parts of multi-lines,
+      # shorter than 0.1px at nominal zoom to be dropped.
+      drop_short_segments: true
+      drop_length_pixels: 0.1
```
nvkelso: Thanks for making this configurable. Did you try a larger value like 0.2 or 0.5?

zerebubuth: The issue that I saw in a few places was that of very small segments, often less than a (Mercator) metre. These seem to have been created by tiny gaps between adjacent landuse areas (which give the roads a `landuse_kind`, meaning they're merged separately).

Getting rid of these almost-points probably doesn't affect the appearance of the tile in this case, but I'm worried that it might in other places. I think if we find we have more substantial fragments of road, we might want to try either dropping the attributes which are preventing them from being merged, or altering the algorithms (e.g: intercut) so as not to create them in the first place.

Otherwise, we risk creating little gaps in the road network, and in some places where the attributes change in a short distance (e.g: bridge, bus routes, cycleways, surface, etc...) these little gaps might add up to something visible.

nvkelso: Sounds reasonable. Merge ahead!

bcamper:
OMG return of the line merging!!! 👏👏👏  Let me know when there is a place I can try this out and I will do a comparison to see how it affects Tangram geometry count and memory usage.

zerebubuth:
I think we'll see an even bigger improvement after #1331. There's a lot of missed opportunities for merging in the image above, even on minor roads which we've stripped the names from already.

Issue #1598, opened 2018-09-17
Traffic / congestion layer information


musculman:
It would be great if for the new version there would be the possibility of providing traffic / congestion information in a separate layer.

It should support at least showing information for both directions of traffic flow, considering countries with right-hand and left-hand traffic.

As a first proposal, it could contain a copy of the geometries from the transportation (roads) layer with extended attributes with the information for traffic, such as the level of congestion (low, medium, high, severe) and incident (closure of roads, accidents and such).

That way, users could get the data from the `traffic` layer or from the `roads` one. Probably the `traffic` one could be retrieved separately, as it is highly dynamic and usually it's not worth caching it in the CDN for example as it changes every 1-3 minutes.

nvkelso:
Hi @musculman. The team has talked about this before by adding an [OSMLR](https://mapzen.com/blog/osmlr-2nd-technical-preview/) (which is similar to [OpenLR](http://www.openlr.org/)) type ID to the **roads** layer so any type of linear interpolation can be applied. Could either be a **new property** on **roads** layer features, or a (also) subset of roads into a new **traffic** "skeleton" layer. The later evolution of OSMLR that has continued past Mapzen is called [SharedStreets](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md).

But this is all academic until one has traffic data to add, and there's not an open source for that. Please link to it if you have one! If this is for an internal data source you have, could you share a data sample with documentation here? That would help a lot to further the discussion.

Agree traffic data should be retrieved separately (list of key, values pairs per tile, like: road ID, traffic "color") from the geometry to improve performance (should be joined at run-time client-side). Tuning the time-to-live with respect to how fast the data can be processed and distributed is always a production challenge ;)

nvkelso:
Another use of SharedStreets (OSMLR) IDs would be to power easy local government data > Tilezen **roads** layer matching to power apps like: https://www.remix.com/streets.

musculman:
You are right, I was thinking of an internal data source that we would like to use.
Unfortunately, I do not know of an open source traffic information that could be used.

Regarding a data sample, here is an uncompressed tile that we have generated when testing out our prototypes:

```
layers {
  name: "flow"
  features {
    id: 0
    tags: 0
    tags: 0
    tags: 1
    tags: 1
    type: LINESTRING
    geometry: 9
    geometry: 4236
    geometry: 4604
    geometry: 10
    geometry: 906
    geometry: 1467
[...]

PR #1705, opened 2018-11-15
Adding Traffic Flow and Traffic Incident Layers

conor-ettinoffe-here:
As discussed here: https://github.com/tilezen/vector-datasource/issues/1598, proposing documentation changes to add the first version of two new optional layers for representing Traffic Flow and Traffic Incidents.


docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
+
+#### `optional` properties:
+The following properties should only be included when **roads** layer features doesn't include `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md). In that case the traffic data should be delivered without geometry and run-time linked with the **roads** layer. But baring that, the following `optional` properties enable sizing and layering of the traffic overlay with respect to the roads network:
+
+* `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md), or..
+* `road_kind`: [kind of the road](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#road-properties-common)
+* `is_bridge`: set to `true` when the linear is a bridge
+* `is_tunnel`: set to `true` when the linear is a tunnel
+* `ramp` or `is_link`: set to `true` when the linear is a slip-road
```
nvkelso: Tilezen considers these to be the same thing as `is_link`, you can add ramp to the description though.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `stationary`: Stationary traffic, congestion
+* `none`: No traffic flow due to blockage or closure
+
+#### `common-optional` properties:
+* `source`: provider of traffic information
+* `speed`: speed in km/h
+* `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
+
+#### `optional` properties:
+The following properties should only be included when **roads** layer features doesn't include `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md). In that case the traffic data should be delivered without geometry and run-time linked with the **roads** layer. But baring that, the following `optional` properties enable sizing and layering of the traffic overlay with respect to the roads network:
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `unknown`: traffic status unknown
+* `free`: Free flowing, not disturbed, traffic
+* `minor`: Minor traffic
+* `slow`: Slowly moving traffic
+* `queuing`: The traffic is in queues but still moves slowly
+* `stationary`: Stationary traffic, congestion
+* `none`: No traffic flow due to blockage or closure
+
+#### `common-optional` properties:
+* `source`: provider of traffic information
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* Geometry types: `line`
+
+#### `common` properties:
+* `id`: Unique traffic event ID. Can be referenced when checking for updated traffic information for specified event
+* `kind`: the **severity** of the flow information as seen by traffic provider to indicate the traffic "color"
+* `kind_detail`: the **congestion** level of traffic flow, with 0.0 representing completely free flowing traffic to 1.0 completely congested traffic
+* `min_zoom`: this value is derived from combination of `road_kind`, `kind`, and `kind_detail` values. A suggestion for which zoom to draw a feature. The value is a float
+
+#### Possible `kind` values for traffic flow:
+* `unknown`: traffic status unknown
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
 * `yellowgreen`
 
+## Traffic Flow Layer
+This **optional** layer is meant to be updated every few minutes, due to the highly dynamic nature of the data. Consideration should be given to rendering the features from this layer linked to the **roads** layer when linear referencing is available, or as an overlay
+
+* Layer name: `traffic_flow`
+* Geometry types: `line`
+
+#### `common` properties:
+* `id`: Unique traffic event ID. Can be referenced when checking for updated traffic information for specified event
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
 * `yellow`
 * `yellowgreen`
 
+## Traffic Flow Layer
+This **optional** layer is meant to be updated every few minutes, due to the highly dynamic nature of the data. Consideration should be given to rendering the features from this layer linked to the **roads** layer when linear referencing is available, or as an overlay
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `queuing`: The traffic is in queues but still moves slowly
+* `stationary`: Stationary traffic, congestion
+* `none`: No traffic flow due to blockage or closure
+
+#### `common-optional` properties:
+* `source`: provider of traffic information
+* `speed`: speed in km/h
+* `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
+
+#### `optional` properties:
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+
+* `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md), or..
+* `road_kind`: [kind of the road](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#road-properties-common)
+* `is_bridge`: set to `true` when the linear is a bridge
+* `is_tunnel`: set to `true` when the linear is a tunnel
+* `ramp` or `is_link`: set to `true` when the linear is a slip-road
+
+## Traffic Incident Layer
+This **optional** layer is meant to be updated every few minutes, due to the highly dynamic nature of the information contained within this layer. Consideration should be given to rendering the features from this layer either linked to the **roads** layer for linear geometries or as an overlay, and always as an overlay for point geometries
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+
+* Layer name: `traffic_incidents`
+* Geometry types: `line`, `point`
+
+#### `common` properties:
+* `id`: Unique traffic event ID. Can be referenced when checking for updated traffic information for specified event
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `kind`: type of the incident
+* `min_zoom`: this value is derived from the internal `warning_level` used for the severity of the incidents which corresponds to a list of values: low (3), minor (2), major (1), critical (0). A suggestion for which zoom to draw a feature. The value is a float
+
+#### Possible values for the incident `kind` are:
+* `accident`: there has been a collision
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `construction`: building or roadworks are taking place
+* `disabled_vehicle`: a vehicle is unable to move and is obstructing the road
+* `mass_transit`: a large amount of people are migrating from one location to another
+* `planned_event`: an organised event is taking place causing disruption
+* `road_hazard`: there are dangerous objects on the surface of the road
+* `weather`: weather conditions are causing disruptions
+* `other`: an incident not explainable with the labels above has occurred
+
+#### `common-optional` properties:
+* `source`: provider of traffic information
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+* `start_time`: the time the incident begins/begun as [unix time](https://en.wikipedia.org/wiki/Unix_time). At least one of the `start_time` or `stop_time` must be set
+* `stop_time`: the time the incident ends/ended as [unix time](https://en.wikipedia.org/wiki/Unix_time). At least one of the `start_time` or `stop_time` must be set
+* `title`: a short description of the incident, localized, such as `title:en` for English and `title:de` for German. Could be used for a title of a pop-up shown in the screen
+* `description`: potentially long description and comment on the incident. Localized, like `title` mentioned above
+
+#### `optional` properties:
+The following properties should only be included when **roads** layer features doesn't include `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md). In that case the traffic data should be delivered without geometry and run-time linked with the **roads** layer. But baring that, the following `optional` properties enable sizing and layering of the traffic overlay with respect to the roads network:
```
nvkelso: Github markdown needs an extra return after the ### line.

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+
+* `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md), or..
+* `road_kind`: [kind of the road](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#road-properties-common)
+* `is_bridge`: set to `true` when the linear is a bridge
+* `is_tunnel`: set to `true` when the linear is a tunnel
+* `ramp` or `is_link`: set to `true` when the linear is a slip-road
+
+## Traffic Incident Layer
```
nvkelso: `## Traffic Incidents` (remove the layer bit, make plural)

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
 * `yellow`
 * `yellowgreen`
 
+## Traffic Flow Layer
```
nvkelso: `## Traffic Flow` (remove the layer bit)

docs/layers.md
```diff
@@ -2108,6 +2108,80 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+This **optional** layer is meant to be updated every few minutes, due to the highly dynamic nature of the data. Consideration should be given to rendering the features from this layer linked to the **roads** layer when linear referencing is available, or as an overlay
+
+* Layer name: `traffic_flow`
+* Geometry types: `line`
+
+#### `common` properties:
+* `id`: Unique traffic event ID. Can be referenced when checking for updated traffic information for specified event
+* `kind`: the **severity** of the flow information as seen by traffic provider to indicate the traffic "color"
+* `kind_detail`: the **congestion** level of traffic flow, with 0.0 representing completely free flowing traffic to 1.0 completely congested traffic
```
nvkelso: Let's move `congression` down to `common-optional` properties section and call it `congestion` not `kind_detail`, to go along with `speed` down there.

docs/layers.md
conor-ettinoffe-here: I have removed `is_link` from the description to explain this is for slip-roads/ramps.

conor-ettinoffe-here: Aha makes sense - so basically we just need the first description. I will update with a new patch for this. Thanks @zerebubuth!

docs/layers.md
```diff
@@ -2108,6 +2108,90 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+#### `common-optional` properties:
+
+* `source`: provider of traffic information
+* `congestion`: the level of traffic flow, with 0.0 representing completely free flowing traffic to 1.0 completely congested traffic
+* `speed`: speed in km/h
+* `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
+
+#### `optional` properties:
+
+The following properties should only be included when **roads** layer features doesn't include `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md). In that case the traffic data should be delivered without geometry and run-time linked with the **roads** layer. But baring that, the following `optional` properties enable sizing and layering of the traffic overlay with respect to the roads network:
```
zerebubuth: You say that the following properties should only be included when the **roads** layer features don't include `linear_ref_id`, but then the first list item is `linear_ref_id` (although it says "or.." at the end).

This confused me. Do you think it might be clearer if split into two lists, e.g:

> If the **roads** layer features include `linear_ref_id`, then traffic flow layer features may include the following, in which case the traffic data should be delivered without geometry and run-time linked with the **roads** layer.
>
> * `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md)
>
> However, if the **roads** layer does not include `linear_ref_id`, then it should be delivered with geometry and the following properties to enable sizing and layering of the traffic overlay with respect to the roads network:
>
> * `road_kind`: [kind of the road](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#road-properties-common)
> ...etc...

?

docs/layers.md
```diff
@@ -2108,6 +2108,90 @@ Transit lines may have their colours mapped onto one of these CSS colours. The i
...
+
+* `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md), or..
+* `road_kind`: [kind of the road](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#road-properties-common)
+* `is_bridge`: set to `true` when the linear is a bridge
+* `is_tunnel`: set to `true` when the linear is a tunnel
+* `is_link`: set to true when the linear is a slip-road
```
zerebubuth: It can be helpful for people writing styles to have [a `sort_rank` property](https://github.com/tilezen/vector-datasource/blob/935a6ee920464379275ace3d29c3ef17ec94a255/docs/layers.md#feature-ordering) to guide the draw order of features. Of course, any style can override or ignore this, but it often greatly simplifies the work of writing styles if a default is provided. Tilezen includes one in the [**roads** layer common properties](https://github.com/tilezen/vector-datasource/blob/935a6ee920464379275ace3d29c3ef17ec94a255/docs/layers.md#road-properties-common), and I think it would be helpful to do the same if the **traffic flow** layer includes geometry.

docs/layers.md
zerebubuth: Great, thanks!

docs/layers.md
conor-ettinoffe-here: With regards to styling the traffic flow - one will be provided with both the raw traffic congestion levels (in the `congestion`) field and the _suggested_ traffic congestion levels from the `kind` values. 

Therefore, the ordering can be made from the order of the `kind` values where `unknown` would be drawn first and `none` last - as it is the most severe. Also, if one wishes to have more granular control over the ordering, the style may also be made from the congestion level where 0.0 be drawn first and 1.0 last. Up to the stylist in this case. What do you think?

docs/layers.md
conor-ettinoffe-here: Split these out :)

docs/layers.md
zerebubuth: I agree that there's enough information for the stylist to do anything that they want. My point was that it's very helpful to the stylist to have that pre-encoded in a `sort_rank` property that's compatible with all the other layers. That way, it's easier to get started with a style that looks right, and we don't have to include a bunch of "polyfill" code to get a sensible-looking map.

docs/layers.md
```diff
@@ -2140,9 +2140,12 @@ This **optional** layer is meant to be updated every few minutes, due to the hig
 
 #### `optional` properties:
 
-The following properties should only be included when **roads** layer features doesn't include `linear_ref_id`: identifier link to a linear references system, eg [SharedStreet](https://github.com/sharedstreets/sharedstreets-ref-system/blob/master/OSMLR.md). In that case the traffic data should be delivered without geometry and run-time linked with the **roads** layer. But baring that, the following `optional` properties enable sizing and layering of the traffic overlay with respect to the roads network:
+If the roads layer features include linear_ref_id, then traffic flow layer features may include the following, in which case the traffic data should be delivered without geometry and run-time linked with the roads layer.
```
zerebubuth: Just a nit-pick, I think it would be clearer that we're referring to a property name if `linear_ref_id` were marked-up with backticks. It seems to have emerged as a convention, although probably isn't 100% observed everywhere in this document.

conor-ettinoffe-here: Lucky you mentioned, I needed to update the Incidents with the separated paragraphs too

docs/layers.md
conor-ettinoffe-here: Okay, this makes sense if all other layers can also use this feature. Would you agree that we can make this one optional for the traffic layer?

zerebubuth: Great, thanks! Yes, it can be optional.

Would it also be useful for linear features in the **traffic_incidents** layer too, or are they intended to be rendered as points or above road labels?

PR #1707, opened 2018-11-19
Drop bicycle network and shield text to match style.

zerebubuth:
Drop RCN from 12 down to 0, LCN from 15 down to 0, and any network on track paths from 12 down to zero. These properties inhibit some degree of merging at mid zooms.

[Context](https://github.com/tilezen/vector-datasource/issues/1331#issuecomment-439132593).

Connects to #1331. (Note that this is one part of many - submitting a PR for each to avoid it becoming unreadably large.)

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -991,30 +991,52 @@ post_process:
       properties:
         - walking_network
         - walking_shield_text
-  # drop RCN network & shield text below zoom 13.
+  # drop RCN network & shield text below zoom 11, and shield text at one more
+  # zoom.
+  # https://github.com/tilezen/vector-datasource/pull/1707#discussion_r236522771
   - fn: vectordatasource.transform.drop_properties
     params:
       source_layer: roads
       start_zoom: 0
-      end_zoom: 13
+      end_zoom: 11
       properties:
         - bicycle_network
         - bicycle_shield_text
       where: >-
         bicycle_network == 'rcn'
-  # drop LCN network & shield text below zoom 16. this is earlier than a
-  # previous drop of the all_*, so make sure we clear those out too.
   - fn: vectordatasource.transform.drop_properties
     params:
       source_layer: roads
       start_zoom: 0
-      end_zoom: 16
+      end_zoom: 12
+      properties:
+        - bicycle_shield_text
+      where: >-
+        bicycle_network == 'rcn'
+  # drop LCN network & shield text below zoom 14 and shield text only at
+  # one more zoom level. (should be present at zoom 15).
+  # https://github.com/tilezen/vector-datasource/pull/1707#discussion_r236523895
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
       properties:
         - bicycle_network
         - bicycle_shield_text
       where: >-
         bicycle_network == 'lcn'
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 15
+      properties:
+        - bicycle_shield_text
+      where: >-
+        bicycle_network == 'lcn'
   # drop any bicycle network & shield text on tracks below zoom 13.
+  # https://github.com/tilezen/vector-datasource/pull/1707#discussion_r236524127
   - fn: vectordatasource.transform.drop_properties
     params:
       source_layer: roads
--- a/queries.yaml
+++ b/queries.yaml
@@ -821,12 +821,20 @@ post_process:
       start_zoom: 0
       end_zoom: 14
       properties:
-        - all_bicycle_networks
-        - all_bicycle_shield_texts
         - all_walking_networks
         - all_walking_shield_texts
         - all_bus_networks
         - all_bus_shield_texts
+  # drop all_bicycle_* until the max zoom:
+  # https://github.com/tilezen/vector-datasource/pull/1707#discussion_r236522176
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 16
+      properties:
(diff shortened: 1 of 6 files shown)


zerebubuth:
Follow-up: On `12/1206/1540`, which is basically Brooklyn, the number of `minor_road` features is 38% lower after the change. This means a lot more merging is going on. The layer as a whole is 8% smaller.

nvkelso:
Wow, that’s great!

Can you stat here the tile so I can take a look, please?

zerebubuth:
For the `roads` layer features, the counts by kind:

Kind | Before | After | Reduction
-- | -- | -- | --
ferry | 6 | 6 |  
highway | 59 | 59 |  
major_road | 570 | 568 | 0%
minor_road | 313 | 200 | 36%
path | 322 | 313 | 3%
rail | 14 | 14 |  

Which is what we expected: A big reduction in the number of `minor_road` kind features and some small reductions in other types. This doesn't have such a huge impact on the bytes per feature:

Kind | Type | Before | After | Reduction
-- | -- | -- | -- | --
ferry | geom_cmds | 154 | 154 |  
ferry | metadata | 24 | 24 |  
ferry | properties | 178 | 178 |  
highway | geom_cmds | 2649 | 2649 |  
highway | metadata | 240 | 240 |  
highway | properties | 1451 | 1422 | 2%
major_road | geom_cmds | 8170 | 8168 | 0%
major_road | metadata | 2281 | 2273 | 0%
major_road | properties | 14785 | 13576 | 8%
minor_road | geom_cmds | 20101 | 19919 | 1%
minor_road | metadata | 1262 | 814 | 35%
minor_road | properties | 7240 | 4000 | 45%
path | geom_cmds | 5671 | 5652 | 0%
path | metadata | 1291 | 1255 | 3%
path | properties | 6994 | 6777 | 3%
rail | geom_cmds | 1201 | 1201 |  
rail | metadata | 59 | 59 |  
rail | properties | 243 | 243 |  
  |   | 73994 | 68604 | 7%

Note that nothing has really changed for any highway kind features. The changes in the size of the properties is likely to be due to re-arrangement of the string table, leading to small changes to the length of the varints used to encode the key and value indexes.

For the layer as a whole, including the string table:

  | Before | After | Reduction
-- | -- | -- | --
Features | 73994 | 68604 | 7%
Metadata | 16 | 16 | 0%
String Table | 12631 | 12326 | 2%
TOTAL | 86641 | 80946 | 7%

(PS: Looks like I got my maths a little wrong before - the reductions are 35% and 7% for the minor roads count and layer size, rather than 38% and 8% - perhaps I summed up the counts and byte sizes together by mistake or something.)

The GeoJSON for the `roads` layer is [here in a gist](https://gist.github.com/zerebubuth/3b87f5805c2e98dfe29d4dd45f4664bc) and it looks like this (coloured by `kind`, `kind_detail` and `landuse_kind`):

image

queries.yaml
```diff
@@ -985,6 +985,42 @@ post_process:
...
+  # previous drop of the all_*, so make sure we clear those out too.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 16
+      properties:
+        - bicycle_network
+        - bicycle_shield_text
+        - all_bicycle_networks
```
nvkelso: Dropping `all_bicycle_networks` and `all_bicycle_shield_texts` all the time until the max zoom would be best (and not limited to just `lcn`).

queries.yaml
```diff
@@ -985,6 +985,42 @@ post_process:
...
         - walking_network
         - walking_shield_text
+  # drop RCN network & shield text below zoom 13.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 13
+      properties:
+        - bicycle_network
```
nvkelso: Reviewing https://github.com/tangrams/walkabout-style/blob/gh-pages/layers/bike-interlay.yaml I see that the network IS referenced like `bicycle_network: [icn,ncn,rcn]` at lower zooms, but `rcn` is ignored at super low zooms like `filter: { bicycle_network: rcn, $zoom: [8,9,10] }`:

https://github.com/tangrams/walkabout-style/blob/gh-pages/layers/bike-interlay.yaml#L1788-L1797

So it would be:

- okay to drop `bicycle_network` from 0 to 11 (or is that 10, I get confused)
- okay to drop `bicycle_shield_text` from 0 to 12 (one more zoom than `bicycle_network` is dropped)
- see other comment about always dropping the `all*` variants until the max zoom) even for `rcn`

queries.yaml
```diff
@@ -985,6 +985,42 @@ post_process:
...
+        - all_bicycle_networks
+        - all_bicycle_shield_texts
+      where: >-
+        bicycle_network == 'lcn'
+  # drop any bicycle network & shield text on tracks below zoom 13.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 13
```
nvkelso: For reference this comes from:

https://github.com/tangrams/walkabout-style/blob/gh-pages/layers/bike-interlay.yaml#L569-L573

queries.yaml
```diff
@@ -985,6 +985,42 @@ post_process:
...
+      properties:
+        - bicycle_network
+        - bicycle_shield_text
```
nvkelso: Should this include the `all*` variants, too?

nvkelso:
Look at the sample GeoJSON file we can probably also drop `bicycle` until zoom 13:

https://tangrams.github.io/walkabout-style/#13/37.8177/-122.5060

nvkelso:
Please also make sure https://github.com/tilezen/vector-datasource/pull/1707#issuecomment-441925279 is tracked in an issue somewhere.

Issue #1553, opened 2018-08-03
Support truck (HGV) routing properties


nvkelso:
**tl;dr**

See [PR #1709](https://github.com/tilezen/vector-datasource/pull/1709) for documentation on what was actually decided. 

I'm fine if this is for zooms 14+ only at first? We should generally follow the stripping we added in v1.6.0 depending on the kind of road...

- [ ] Add `low_emission_zone` polygons to **landuse** layer.
- Add properties on existing line geoms in the **roads** layer:
    - [ ] `hgv`: optional property indicating general truck heavy goods vehicle truck access. See below for list of values.
        values whitelist: `no`, `designated`, `destination`, `delivery`, `local` and `agricultural`.
    - [ ] `hgv_restriction`: optional property indicating limitations to heavy goods vehicle truck access. See below for list of values. Available on both point and line geometries. See also `hgv_restriction_shield_text`.
        values include: `weight` (kg), `height` (cm), `length` (cm), `width` (cm), `wpa` (weight per axle, in kg), `kpra` (king pin to rear axle leght, in cm), `hazmat`, `other` and `multiple` if more than one.
    - [ ] `hgv_restriction_shield_text`: optional and paired with `hgv_restriction` points with values like `5.1m`. Because the units are different per restriction an abbreviation should be provided. Values in meters can be specified with one decimal precision but value of 5.0m should be given as 5m.
    - [ ] `toll`: optional `boolean` value indicating whether a fee must be paid by general traffic to travel the feature.
    - [ ] `toll_hgv`: optional `boolean` value when tool only applies to certain classes of vehicles like heavy good vehicle trucks.
- Add new `hgv_restriction` kind points in the **roads** layer, with following properties:
    - [ ] `hgv`: optional property indicating general truck heavy goods vehicle truck access. 
        values whitelist: `no`, `designated`, `destination`, `delivery`, `local` and `agricultural`.
    - [ ] `hgv_restriction`: optional property indicating limitations to heavy goods vehicle truck access. See below for list of values. Available on both point and line geometries. See also `hgv_restriction_shield_text`.
        values include: `weight` (kg), `height` (cm), `length` (cm), `width` (cm), `wpa` (weight per axle, in kg), `kpra` (king pin to rear axle leght, in cm), `hazmat`, `other` and `multiple` if more than one.
[...]

PR #1709, opened 2018-11-20
Adds hgv (heavey good vehicle) roads properties and related environmental_zone landuse

musculman:
Connects with #1553 to document heavy good vehicle (truck) properties [edits by @nvkelso]

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -511,6 +511,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `dog_park`
 * `embankment` - A raised area of land, usually to carry a road or railway.
 * `enclosure`
+* `low_emission_zone` - An area beloging to a low emission zone, such as the (London Low Emission Zone)[https://en.wikipedia.org/wiki/London_low_emission_zone]. (See also)[https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dlow_emission_zone].
 * `farm`
 * `farmland`
 * `farmyard`
@@ -1249,7 +1250,7 @@ Common values include: `drinkable_powered`, `drinkable_manual`, `drinkable_no_pu
 ![image](images/mapzen-vector-tile-docs-roads.png)
 
 * Layer name: `roads`
-* Geometry types: `line`
+* Geometry types: `line`, `point`
 
 More than just roads, this OpenStreetMap and Natural Earth based transportation layer includes highways, major roads, minor roads, paths, railways, ferries, and ski pistes matching the selection found in High Road. Sort them with `sort_rank` to correctly represent layered overpasses, bridges and tunnels. Natural Earth roads at zooms < 8 and OpenStreetMap at zooms 8+. See zoom ranges section below for more information per kind.
 
@@ -1317,6 +1318,8 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `description`: OpenStreetMap features
 * `distance`: ski pistes from OpenStreetMap
 * `embankment`: If the road or railway is on an embankment the value will be one of `yes`, `left` or `right` depending on whether the embankment is on both sides, the left side or the right side, respectively.
+* `hgv`: optional property indicating general truck heavy goods vehicle truck access. See below for `kind_detail` list.
+* `hgv_restriction`: optional property indicating limitations to heavy goods vehicle truck access. See below for `kind_detail` list.
 * `motor_vehicle`: OpenStreetMap features
 * `operator`: OpenStreetMap features
 * `piste_difficulty`: ski pistes from OpenStreetMap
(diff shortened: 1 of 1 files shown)


docs/layers.md
```diff
@@ -1395,6 +1400,11 @@ Railway `service` values are:
 
 **Piers** and **quays** start showing up at zoom 13+ with `kind_detail` values of `pier` and `quay`, respectively. If mooring information is available, the `mooring` property will be one of `no`, `yes`, `commercial`, `cruise`, `customers`, `declaration`, `ferry`, `guest`, `private`, `public`, `waiting`, `yacht` or `yachts`.
 
+#### Roads hgv `kind_detail` values
+
+For `hgv`the `kind_detail` values are `no`, `designated`, `destination`, `delivery`, `local` and `agricultural`.
+For `hgv_restriction` the `kind_detail` values are `weight`, `height`, `length`, `width`, `wpa` (weight per axle), `kpra` (king pin to rear axle leght, in cm.), `hazmat`, `other` and `multiple` if more than one.
```
nvkelso: Since people always ask, let's specify units as (kg) and (m) next to each value, please...

docs/layers.md
```diff
@@ -1995,6 +2005,15 @@ Alpha-2 code | Country
...
 `ZW` | Zimbabwe
 
+#### Road transportation `kind` values (points):
+
+* `hgv_restriction`: indicate point location for access limitations for heavy goods vehicle trucks. See also `hgv_restriction_shield_text`.
+* `hgv_restriction_shield_text`: paired with `hgv_restriction` points with values like `5.1m`. Because the units are different per restriction an abbreviation should be provided. Values in meters can be specified with one decimal precision but value of 5.0m should be given as 5m.
+
+#### Roads hgv `kind_detail` values
+
+For `hgv_restriction` the `kind_detail` values are `weight`, `height`, `length`, `width`, `wpa` (weight per axle), `kpra` (king pin to rear axle leght, in cm.), `hazmat`, `other` and `multiple` if more than one.
```
nvkelso: Same comment about including units, please.

docs/layers.md
```diff
@@ -511,6 +511,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `dog_park`
 * `embankment` - A raised area of land, usually to carry a road or railway.
 * `enclosure`
+* `environmental_zone` - An area beloging to a protected environmental area, such as the (London Loz Emission Zone)[https://en.wikipedia.org/wiki/London_low_emission_zone].
```
nvkelso: Maybe there's a better term for this like `restricted_driving_zone`?

docs/layers.md
```diff
@@ -1317,6 +1318,8 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `description`: OpenStreetMap features
 * `distance`: ski pistes from OpenStreetMap
 * `embankment`: If the road or railway is on an embankment the value will be one of `yes`, `left` or `right` depending on whether the embankment is on both sides, the left side or the right side, respectively.
+* `hgv`: optional property indicating general truck heavy goods vehicle truck access. See below for `kind_detail` list.
+* `hgv_restriction`: optional property indicating limitations to heavy goods vehicle truck access. See below for `kind_detail` list.
```
nvkelso: Please also include here (and not below):

- `hgv_restriction_shield_text` - optional property for `hgv_restriction` kind points with values like `5.1m`. Because the units are different per restriction an abbreviation should be provided. Values in meters can be specified with one decimal precision (but value of 5.0m should be given as 5m).

nvkelso:
Had a quick look now, will have a deeper look Tuesday with a fresh set of eyes...

docs/layers.md
musculman: Will post an update with it

docs/layers.md
musculman: Not sure about that, as in general these zones are either `low_emission_zone` or `zero_emission_zone` [see here](https://en.wikipedia.org/wiki/Low-emission_zone).
That is why we opted for `environmental_zone` as more neutral as it is an area with environmental restrictions, but I get the point that perhaps it is not descriptive enough.
Should we just go with `low_emission_zone` as in [OSM](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dlow_emission_zone)?

docs/layers.md
musculman: Why here? I put it below with the other `point` feature description. Because we have a set of values for the lines and another set for the points. Or would you rather have all defined together? So, merging the section " Road transportation `kind` values (points):" into this one?

docs/layers.md
musculman: Will do as well

docs/layers.md
nvkelso: Yes, let's go with OSM's `low_emission_zone` terminology.

nvkelso:
@musculman Please take one last look with my edits today before we merge. Thanks for your contribution to Tilezen!

docs/layers.md
musculman: Good, this is the term I used in my updated commit :)

docs/layers.md
musculman: With your last commits from today now I see what you meant by this comment :)
I thought it was more clear to have it separated, but I am obviously fine to have it together with the additional clarification you added as "this is also for points" and such.
All clear.

musculman:
> @musculman Please take one last look with my edits today before we merge. Thanks for your contribution to Tilezen!

All fine from my side!

nvkelso:
🎉

PR #1710, opened 2018-11-22
Drop many "detail" tags on minor roads at mid zooms. 

zerebubuth:
NOTE: This follows on from #1707 and #1708, so those should have been merged before this.

On minor roads (only - this doesn't apply to major roads or highways), drop the following from zoom 15 down:

* `colour`
* `motor_vehicle`
* `operator`
* `route`
* `route_name`
* `state`
* `symbol`
* `type`

And the following from zoom 14 down:

* `surface`

And the following from zoom 13 down:

* `cutting`
* `embankment`

This helps a lot with road merging, as the fewer distinct combinations of road properties there are, the more roads can be merged into one another.

Also, drop most values of the `access` property from zoom 13 down, keeping `access=no`, `access=private` and `access=official` (on all road types). The `no` and `private` values are important for knowing which streets are inaccessible, but many other distinctions are unimportant at this zoom level, where streets might be a few pixels long, e.g: `access=customers` (if styled at all) can be made prominent as the view is zoomed in a bit more.

Further, we merge some `access` values together; `emergency`, `military`, `restricted` and `forestry` all map to `access=official`, which indicates that there is some kind of restriction, but allows for geometry merging and value compression through re-use.

Finally, drop `landuse_kind` on roads on zoom < 13 where it's one of `residential` or `industrial`. These are very common landuse kinds, but aren't strongly styled. At zoom 13 they're getting to be a fraction of the size of a tile and in my opinion of limited usefulness.

These measures together produced a 14% reduction in the number of `minor_road` features on `12/1206/1540`, reflected in a similar drop in the size of the metadata and properties for the MVT tile. Overall, the `roads` layer was 4% smaller (about 3kB for this tile).

Connects to #1331.


queries.yaml
```diff
@@ -881,6 +881,73 @@ post_process:
...
+      where: >-
+        kind == 'minor_road'
+  # keep "surface" until zoom 15.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 15
+      properties:
+        - surface
```
nvkelso: It looks like surface is used to set "brown" color on line and text halo in Walkabout at some zooms, but looks like only for `kind` = **path** so we're good?

https://github.com/tangrams/walkabout-style/blob/gh-pages/layers/bike-interlay.yaml#L705-L708

queries.yaml
```diff
@@ -881,6 +881,73 @@ post_process:
...
         (kind == 'major_road' and zoom <  7)
+  # drop these "detail" tags to get better merging at zoom < 16
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 16
+      properties:
+        - colour
+        - motor_vehicle
```
nvkelso: It looks like `motor_vehicle` = **no** could be used for `minor_road` starting at zoom 12:

https://github.com/tangrams/walkabout-style/blob/gh-pages/walkabout-style.yaml#L2294-L2304

queries.yaml
```diff
@@ -881,6 +881,73 @@ post_process:
...
+        kind == 'minor_road'
+  # drop to get better merging at mid zooms.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
+      properties:
+        - cutting
+        - embankment
```
nvkelso: 👍

queries.yaml
```diff
@@ -881,6 +881,73 @@ post_process:
...
+      where: >-
+        kind == 'minor_road'
```
nvkelso: Could probably be for any kind (not just `minor_road`)...

queries.yaml
```diff
@@ -881,6 +881,73 @@ post_process:
...
+      properties:
+        - surface
+      where: >-
+        kind == 'minor_road'
+  # drop to get better merging at mid zooms.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 14
```
nvkelso: 👍

queries.yaml
```diff
@@ -985,6 +1052,42 @@ post_process:
       properties:
         - walking_network
         - walking_shield_text
+  # drop RCN network & shield text below zoom 13.
```
nvkelso: This PR has commits from the other PR, which has requested changes...

queries.yaml
```diff
@@ -881,6 +881,73 @@ post_process:
...
+        - "no"
+        - private
+        - official
+      # some values are less important, but belong to a category that might
+      # still be important. for example, access=military at mid or low zooms
+      # might effectively be described as "there's a road here, but you probably
+      # don't want to go there", while still not being quite the same thing as
+      # access=private or access=no, which are both "you definitely don't want
+      # to go there".
+      remap:
```
nvkelso: 👍

queries.yaml
```diff
@@ -821,12 +821,20 @@ post_process:
...
       end_zoom: 14
       properties:
-        - all_bicycle_networks
-        - all_bicycle_shield_texts
         - all_walking_networks
         - all_walking_shield_texts
         - all_bus_networks
         - all_bus_shield_texts
+  # drop all_bicycle_* until the max zoom:
+  # https://github.com/tilezen/vector-datasource/pull/1707#discussion_r236522176
```
nvkelso: Ah, I didn't expand the diff enough in the earlier review. It's probably better to be more consistent and just drop all the `all*` in the same zoom range. I'd be fine making that end zoom 15 and call it good?

zerebubuth: It's a bit of a corner-case thing, since we only export `all_*` properties on JSON tiles. I like them for multi-shielding, but AFAIK, no one is using it for that and it does affect merging... So perhaps this is something that we come back to once we're into multi-shielding territory?

nvkelso: Sure

zerebubuth:
> Worth pointing out that this reverts #1252

It doesn't fully revert it - we are now dropping surface for `minor_road` only. So it's still available on other road kinds, including paths.

matkoniecz:
I see that it is too late for a comment. Is it OK to open a new issue if it will break my map?

zerebubuth:
Yes, please do! Apologies if the change breaks your map - this is part of an effort to reduce tile size, but we don't want to break anything.

We've been trying to find places where we've got unused amounts of detail in the feature properties. Please let us know what properties you are using and we'll try to accommodate them.

matkoniecz:
Currently, I am using `surface` tag across all zoom levels to decide rendering for road (rather than typical using `highway` tag as - for my private cycling oriented rendering - that is why I made #1252).

But maybe for low zoom levels it is safe anyway to assume that displayed roads are well paved ones or to switch from cycling oriented display of surface to overview map.

nvkelso:
@matkoniecz can you remind us of the public URL or app that shows off your cycling map? If the YAML is available that'd be a bonus, can be shared privately :)

matkoniecz:
https://github.com/matkoniecz/Zazolc (personal fork of StreetComplete).

matkoniecz:
Style is at https://github.com/matkoniecz/Zazolc/blob/fork/app/src/main/assets/map_theme/fork-style.yaml

zerebubuth:
I notice that you [use 3 categories of surface](https://github.com/matkoniecz/Zazolc/blob/304342dfafbf1876b7faf9ebcc2c34cfd2b53ef5/app/src/main/assets/map_theme/fork-style.yaml#L189-L212). Would it be OK if we remapped to those values (or perhaps `paved`, `compacted`, `unpaved` rather than `good`, `medium`, `bad`), and perhaps set some defaults (e.g: `kind in (highway, major_road, minor_road)` ⇒ `paved`, maybe `kind == path` ⇒ `unpaved`)? Hopefully that would give you enough information to render the style that you want while reducing the number of distinct values for data in the tiles?

nvkelso:
To make it easier to follow along just in this issue:

```
            # paved
            good_surface: 
                filter: {surface: [asphalt, paved, wood, metal, tartan, metal_grid]}

            # compacted
            medium_surface:
                filter: {surface: [concrete, paving_stones, compacted, sett]}

            # unpaved
            bad_surface:
                filter: {surface: [unpaved, dirt, earth, fine_gravel, grass, grass_paver, gravel,
                ground, mud, pebblestone, salt, sand, woodchips, clay, cobblestone,
                cobblestone:flattened, concrete:lanes, concrete:plates, artificial_turf, decoturf]}
                draw:
```

@zerebubuth are you proposing to remap **just at low zooms** but the full surface value would still be present at high zooms (my preference, and we've talked about doing the same for landuse before)? What zoom would the cutoff be? Right now the dropping property cutoff is zoom 15 ([link](https://github.com/tilezen/vector-datasource/pull/1710/files#r236525948)).

    - FYI: In Walkabout `surface`: **compacted** is used [as earlier](https://github.com/tangrams/walkabout-style/blob/gh-pages/layers/bike-interlay.yaml#L1189-L1192) as zoom 8 on paths.

The full and raw values would still be available past that zoom, so I think these 3 category mappings (`paved`, `compacted`, `unpaved`) are fine, even for cyclist since they could zoom in and see full details.

I don't like inventing default surface data on roads that are missing the data. It's better for a style to just have a "no data" symbolization when the data isn't there.

Should we drop `surface: paved` all together from low zooms so it's implied? I think it's better to include it (so the set is complete and represents the true data), unless it balloons the tile size.

nvkelso:
(and it's almost time to create a new issue for this!)

nvkelso:
Interesting. A case of "go big, or go home, there is no middle way", lol.

I think we should do the mapping into those 3 (plus null) general values at all low zooms for all **roads** layer features.

But let's try only backfilling the values for `minor_road` features at zooms 13 and 14 like you describe. Good idea!

nvkelso:
Continued the discussion into https://github.com/tilezen/vector-datasource/issues/1716.

Issue #1702, opened 2018-11-08
Remove point labels from water layer for lines


rmarianski:
While working on https://github.com/tilezen/vector-datasource/issues/1135, we noticed that we are generating label points for line features. We should remove these label points from the tiles, but only for lines in the water layer.

PR #1712, opened 2018-11-27
Remove water point labels

zerebubuth:
Don't generate water layer label points from linear features.

Connects to #1702. See also https://github.com/tilezen/tileops/pull/42.


zerebubuth:
I don't think this should require the changes in tileops - filed a follow-up for us to move the configuration around: https://github.com/tilezen/tilequeue/issues/358.

Issue #1477, opened 2018-02-13
Water layer are too big at zoom 8 (zoom 7 at 512 tile size)


nvkelso:
tl;dr

- [ ] water layer **min_zoom** to d.d rounding (instead of d.dd), so `1.1` instead of `1.11`.
- [ ] water layer simplify start of zoom 8 config.
- [ ] Consider dropping names from small and medium sized water bodies to improve their chance of being merged.
- [ ] Drop tiny and small water features at mid zooms
- ~~[ ] roads layer **drop properties** at low zooms for `surface` (#1197) and `access` (#1328) tags to the roads. EG we forgot to [drop them at low zooms like we do for other tags](https://github.com/tilezen/vector-datasource/blob/b3d1dcb6ee17f31c7fc72706877e566ba4931e33/queries.yaml#L611-L630).~~ https://github.com/tilezen/vector-datasource/issues/1331

===

* **What did you see?** Tile size is too large at zoom 8 (zoom 7 at 512 tile size). 
  - With compression in TopoJSON: 2x file size increase
  - With uncompressed GeoJSON: 7/37/48 tile is 13.5 mb to 3.6 mb! (or 3.75x as large!!!)
* **What did you expect to see?** Smaller tile size, more generalized geoms, less features
* **What map location are you having problems with?** 
  -  https://tile.nextzen.org/tilezen/vector/v1/512/all/7/37/48.json?api_key=zxy
  -  https://tile.mapzen.com/mapzen/vector/v1/512/all/7/37/48.json?api_key=zxy
  -  https://tile.nextzen.org/tilezen/vector/v1/512/all/8/74/97.json?api_key=zxy
  -  https://tile.mapzen.com/mapzen/vector/v1/512/all/8/74/97.json?api_key=zxy
* **Screenshot?** _Props for animated gifs._

We did change the NE to OSM switchover in https://github.com/tilezen/vector-datasource/pull/1359. 

So it's expected the feature count be different...

![image 3](https://user-images.githubusercontent.com/853051/36172888-1ce69970-10bc-11e8-901b-f8c1ca947dad.png)

![pasted image at 2018_02_13 12_25 pm 1](https://user-images.githubusercontent.com/853051/36172889-1cfbee4c-10bc-11e8-831d-e84e00bcb5ce.png)

But it seems there are to many features / they have too many names, and their geometries aren't simplified.

At zoom 7 at 512 px tile size in Nextzen build (showing no geom simplification):

[...]

PR #1714, opened 2018-11-28
Water layer too big

zerebubuth:
Incorporating the area look up table from https://github.com/tilezen/vector-datasource/issues/1477#issuecomment-442269890 and a "curve fitted" pixel-area name rule, we can reduce the number of water features (which weren't getting rendered anyway) in the tile. Additionally, we were missing simplification for water and earth layers at zoom 8 (nominal). With those changes, a zoom 8 (nominal) 512px tile at `7/30/43` looks like this (previous values in red, behind, layer in front in blue is new):

image

On tile `7/30/43`, which is pretty much a worst-case scenario and has thousands of tiny lakes, we get a 97% reduction in total size (from 2.4MB to 62kB). The reduction in the number of features is massive - partly because we're being more selective about the features at this zoom and partly because we're merging a lot more lakes that we're able to strip the names off:

Kind | Before | After | Reduction
-- | -- | -- | --
lake | 1 | 0 | 100%
riverbank | 115 | 2 | 98%
water | 2858 | 40 | 98%

The bytes per kind shows similar reduction:

Kind | Before | After | Reduction
-- | -- | -- | --
lake | 141 | 0 | 100%
riverbank | 90992 | 1888 | 98%
water | 2396628 | 60309 | 97%


zerebubuth:
Connects to #1477.

yaml/water.yaml
```diff
@@ -139,39 +160,39 @@ filters:
       kind: ocean
     table: ne
   - filter: {featurecla: Alkaline Lake}
-    min_zoom: 0
+    min_zoom: *water_polygon_min_zoom
```
nvkelso: For Natural Earth features we should just trust their `min_zoom` data property.

Here and elsewhere...

yaml/water.yaml
```diff
@@ -18,6 +18,27 @@ globals:
...
+      table:
+        # limit show smaller landuse areas to higher zooms
+        - [ 0, 40000000000 ]
+        - [ 2, 20000000000 ]
+        # some weird natural earth scale set transition
+        - [ 3, 80000000000 ]
+        - [ 4,  5000000000 ]
+        - [ 5,   700000000 ]
+        - [ 6,   500000000 ]
+        - [ 7,   160000000 ]
```
nvkelso: Suggest removing lines for zooms 0-7 as those are from Natural Earth and should be taken from the data, including the junk comment copy-pasted from Walkabout for `# limit show smaller landuse areas to higher zooms`.

yaml/water.yaml
```diff
@@ -18,6 +18,27 @@ globals:
...
+        - [ 8,    20000000 ]
+        - [ 9,     5000000 ]
+        - [ 10,    1000000 ]
+        - [ 11,     200000 ]
+        - [ 12,      50000 ]
+        - [ 13,      20000 ]
+      default: 14
```
nvkelso: This seems to now show tiny lakes and swimming pools at zoom 14 instead of zoom 16 max? Although in practice browsing around Cupertino (link below) most swimming pools were getting zoom 14.xxx incorrectly (and even showing before the surrounding house footprints!)

Suggest adding zoom 14, 15, 16, and default 17 instead:

```
        - [ 14,      2000 ]
        - [ 15,      1000 ]
        - [ 16,      400 ]
      default: 17
```

I've been browsing around Cupertino to gain some of these numbers:

- https://tangrams.github.io/walkabout-style/#17/37.32740/-122.04532

queries.yaml
```diff
@@ -161,21 +161,21 @@ layers:
       - vectordatasource.transform.add_id_to_properties
       - vectordatasource.transform.detect_osm_relation
       - vectordatasource.transform.remove_feature_id
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
```
nvkelso: All the min_zooms are ints in the area transform below... although switching the NE values to be data driven may sometimes introduce some float values with tenths. So this doesn't hurt, but it probably doesn't help as much now that we're switching to the area lookup table.

queries.yaml
```diff
@@ -626,6 +626,20 @@ post_process:
...
+    params:
+      source_layer: water
+      start_zoom: 0
+      end_zoom: 16
+      properties: [name]
+      geom_types: [Polygon, MultiPolygon]
+      # NOTE: the 270 factor comes from curve fitting the area table from the
+      # walkabout style, with 270 being the min, ranging up to 1070.
+      where: >-
+        area < 270.0 * pixel_area
```
nvkelso: 💯

yaml/water.yaml
nvkelso: (Should add comment to that effect... &/or keep filters as backstop against future OSM / NE zoom transition changes?)

yaml/water.yaml
zerebubuth: Yup, good point. Does this YAML pre-date NE v4? I think we might have been relying on the filter which removes features smaller than a pixel a little too much. Fixed in 67aa978edf2298a6028cdbab8ab762aaf2f59f6a.

yaml/water.yaml
zerebubuth: Sounds good. Clipped to 7 and quickly tested locally. Looks okay to me. Fixed in 67aa978edf2298a6028cdbab8ab762aaf2f59f6a.

yaml/water.yaml
zerebubuth: Well spotted, thanks! Fixed in 91c05f0.

Issue #988, opened 2016-08-24
Add label collision rank for pois and other layer


nvkelso:
When two different kinds features of equal min_zoom collide, which should win?

I have a big spreadsheet of relative ranks between `kind` & `kind_detail` values (as part of the category work). This would remove a good bit of hack logic from the stylesheet, and provide much more fidelity than we have currently there.

_Since the discussion is so long, the related PR is: https://github.com/tilezen/vector-datasource/pull/1793._

zerebubuth:
Can we use `sort_key` for this?

nvkelso:
For pois and places that could work.

But we already have `sort_key` on features like roads and even some building and water features that that means z-stacking order. Because those features also need the new collision values we should probably be consistent everywhere. 

Looking at Tangram this is called "priority" so I propose `priority_key` (though maybe we should call them both `sort_rank` and `priority_rank` in keeping with `scalerank` (or `scale_rank`).

nvkelso:
Complete mapping:

- Existing **Tilezen** `sort_key` sets **Tangram** [order](https://mapzen.com/documentation/tangram/draw/#order). Proposed Tilezen change is `sort_rank` (considered a few other terms like order_rank, z_depth_rank, z_rank, depth_rank).
- Proposed **Tilezen** `priority_rank` would set **Tangram** [priority](https://mapzen.com/documentation/tangram/draw/#priority).

nvkelso:
Subtask of https://github.com/tilezen/vector-datasource/issues/1351.

nvkelso:
There's a ton of client side logic to set these for POIs layer in particular, moving into v1.7 milestone.

nvkelso:
We can sorta use the sort_rank, except the landuse labels should (a) generally collide out the road labels / that can change by zoom and (b) we need to set them for **places** and **pois** layers, so let's do some research...

Generally the collision should be driven by min_zoom (or min_label) and then tie break with the relative ranking of the features. 

@zerebubuth can you generate a CSV that has the following columns, filled out for all the kinds?

layer | kind | label_placement | min_zoom | sort_rank
----- | ---- | ---------------- | ---------- | ----------
water     | water | true               | 3.0             | 200
earth     | cliff | true               | 8.0             | 227
landuse | park | true               | 15.0            | 156
buildings | address | true               | 16.0           | 475
boundaries | region | null               | 16.0           | 250
roads | highway | null               | 16.0           | 444
[...]

PR #1719, opened 2018-12-05
Updating Traffic Incidents

conor-ettinoffe-here:
- Add warning_level to be available
- Adding traffic_flow to show the effect of the incident has had on the flow of traffic

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -2167,7 +2167,8 @@ This **optional** layer is meant to be updated every few minutes, due to the hig
 
 * `id`: Unique traffic event ID. Can be referenced when checking for updated traffic information for specified event
 * `kind`: type of the incident
-* `min_zoom`: this value is derived from the internal `warning_level` used for the severity of the incidents which corresponds to a list of values: low (3), minor (2), major (1), critical (0). A suggestion for which zoom to draw a feature. The value is a float
+* `warning_level`: the severity of the incident that has occured, with three possible values: low (3), minor (2), major (1) and critical (0).
+* `min_zoom`: this value is derived from the `warning_level`. A suggestion for which zoom to draw a feature. The value is a float
 
 #### Possible values for the incident `kind` are:
 
@@ -2203,6 +2204,16 @@ However, if the roads layer does not include `linear_ref_id`, then it should be 
 * `is_tunnel`: set to `true` when the linear is a tunnel
 * `is_link`: set to true when the linear is a slip-road
 * `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
+* `traffic_flow`: the flow of traffic due to the occurance of the Traffic Incident, see description below
+
+#### Traffic Incidents `traffic_flow` values:
+* `unknown`: traffic status unknown
+* `free`: Free flowing, not disturbed, traffic
+* `minor`: Minor traffic
+* `slow`: Slowly moving traffic
+* `queuing`: The traffic is in queues but still moves slowly
+* `stationary`: Stationary traffic, congestion
+* `none`: No traffic flow due to blockage or closure
 
 ## Water
 


nvkelso:
Regarding `warning_level`, I don't think we should include this as it's **routing** focused not _map display_ focused. It should inform the `min_zoom`, and even a possible [collision](https://github.com/tilezen/vector-datasource/issues/988) rank. Please make a stronger case for it if you want it.

Previous discussion around this from Oct. 5th:

- https://github.com/tilezen/vector-datasource/issues/1598#issuecomment-425464977
- https://github.com/tilezen/vector-datasource/issues/1598#issuecomment-427267666
- https://github.com/tilezen/vector-datasource/issues/1598#issuecomment-427376549

I think we're already setting the **kind_detail** of `traffic_flow` features in the **traffic_flow** layer to those values. So again I'm not sure why this change is necessary.  Please provide more information.

- https://github.com/tilezen/vector-datasource/blob/ecf199a94c7ef1510040a2742041b51065fe1a4c/docs/layers.md#possible-kind-values-for-traffic-flow

conor-ettinoffe-here:
> Regarding `warning_level`, I don't think we should include this as it's **routing** focused not _map display_ focused. It should inform the `min_zoom`, and even a possible [collision](https://github.com/tilezen/vector-datasource/issues/988) rank. Please make a stronger case for it if you want it.
> 
> Previous discussion around this from Oct. 5th:
> 
> * [#1598 (comment)](https://github.com/tilezen/vector-datasource/issues/1598#issuecomment-425464977)
> * [#1598 (comment)](https://github.com/tilezen/vector-datasource/issues/1598#issuecomment-427267666)
> * [#1598 (comment)](https://github.com/tilezen/vector-datasource/issues/1598#issuecomment-427376549)
> 
> I think we're already setting the **kind_detail** of `traffic_flow` features in the **traffic_flow** layer to those values. So again I'm not sure why this change is necessary. Please provide more information.
> 
> * https://github.com/tilezen/vector-datasource/blob/ecf199a94c7ef1510040a2742041b51065fe1a4c/docs/layers.md#possible-kind-values-for-traffic-flow

Hi @nvkelso,

Within the discussion it was said "warning_level could be used to filter out incidents based on zoom level, so I think that using that for min_zoom would work fine". But one could also want to use the `warning_level` for other purposes such as displaying more dangerous incidents larger than other for instance.

Regarding the addition of `traffic_flow` it doesn't necessarily need to be of that name, but what I was trying to express with this field is the effect that the Incident has on the traffic around it, for example. There can be an accident, this could simply be slowing down the traffic flow because it has blocked one lane - or it could be blocking the whole road, thus freezing the flow of traffic.

nvkelso:
I can see adding `warning_level` as an **optional** property (not common, not common-optional). Really that's a design decision that's individual to a map style implementation and entirely dependent on veracity of that data (which you have well covered ;). Please update the PR.

But I continue to view `traffic_flow` property in this context as whole routing focused and inappropriate for map display.

conor-ettinoffe-here:
> I can see adding `warning_level` as an **optional** property (not common, not common-optional). Really that's a design decision that's individual to a map style implementation and entirely dependent on veracity of that data (which you have well covered ;). Please update the PR.

Moved the warning level to optional :)

> But I continue to view `traffic_flow` property in this context as whole routing focused and inappropriate for map display.

I've removed the field. Will need to see if there's another way to fulfil such a use case. I can see how you see this as routing focused, the use case one is trying to fulfill with including such information is to display the effect of the incident.
For example, a construction icon with a black background when its a fully disruptive construction site (No traffic), or a green background when its not obstructing traffic, or yellow when its causing some disruptions.

nvkelso:
> > But I continue to view `traffic_flow` property in this context as whole routing focused and inappropriate for map display.
> 
> I've removed the field. Will need to see if there's another way to fulfil such a use case. I can see how you see this as routing focused, the use case one is trying to fulfill with including such information is to display the effect of the incident.
> For example, a construction icon with a black background when its a fully disruptive construction site (No traffic), or a green background when its not obstructing traffic, or yellow when its causing some disruptions.

That sounds like a very technical display of data rather than for a general audience ;) 

My guess is the combination of `min_zoom` for these features in this layer, and also displaying with the **traffic_flow** lines from the other layer will mostly achieve what you want. But I'm happy to revisit in the next release. Let's get this merged now!

Issue #1588, opened 2018-08-30
Set sort_order for quay, wharf, other new landuse kinds


nvkelso:
* **What did you see?** Quay has sort_rank of 11.
* **What did you expect to see?** Should be higher value, that's just a default which sorts it almost at the bottom of the stack (in this case under industrial).
* **What map location are you having problems with?** https://www.openstreetmap.org/way/602259367#map=19/51.52862/-0.24830

QA followup for https://github.com/tilezen/vector-datasource/issues/1423.

nvkelso:
Among others:

- [ ] airfield
- [ ] danger_area 
- [ ] naval_base
- [ ] port_terminal
- [ ] ferry_terminal
- [ ] container_terminal
- [ ] shipyard
- [ ] wharf
- [ ] quay
- [ ] boatyard
- [ ] range

nvkelso:
I spot checked these, they are all working.

Issue #1574, opened 2018-08-24
Set better sort_rank on embankment lines


nvkelso:
* **What did you see?** Embankments are hidden by other landuse polygons
* **What did you expect to see?** They should sit above polygons.
* **What map location are you having problems with?** https://www.openstreetmap.org/way/245501242#map=19/51.54168/-0.00664

nvkelso:
QA followup on https://github.com/tilezen/vector-datasource/issues/1425.

nvkelso:
Verified.

Issue #1569, opened 2018-08-24
Wetland should sort above water


nvkelso:
it has sort_key of 35 now, ocean is 200 should be above that and any other water.

https://www.openstreetmap.org/way/61555323#map=18/51.89562/1.26762

nvkelso:
Also checkout `mud`

nvkelso:
`Wetland` works, though `mud` is fishy (followup in https://github.com/tilezen/vector-datasource/issues/1753).

PR #1720, opened 2018-12-06
Add sort keys for "natural" landuse kinds

zerebubuth:
Insert `natural_wood`, `natural_forest` and `natural_park` into the middle of the sort ranking, bumping up all the values above them.

Connects to #1096.

16 total new:

- [x] natural_wood
- [x] natural_forest
- [x] natural_park
- [x] airfield
- [x] danger_area
- [x] naval_base
- [x] port_terminal
- [x] ferry_terminal
- [x] container_terminal
- [x] shipyard
- [x] wharf
- [x] quay
- [x] boatyard
- [x] range
- [x] wetland
- [x] embankment (lines)


nvkelso:
Please update for the following issues, too:
 
- #1569 Wetland should sort above water bug priority medium
- #1574 Set better sort_rank on embankment lines bug priority medium
- #1588 Set sort_order for quay, wharf, other new landuse kinds bug priority medium

zerebubuth:
Added all those others - except for `boatyard` and `range`. Those are POIs only, and we don't generate landuse polygons for them at the moment. Would you like me to add landuse polygons for those, or are they fine as POIs?

nvkelso:
Ooops, yes those should also be landuse polygons. Good catch!

spreadsheets/sort_rank/water.csv
```diff
@@ -1,16 +1,16 @@
 kind,boundary,sort_rank
-river,,201
```
nvkelso: What does the `-` mean here?

spreadsheets/sort_rank/landuse.csv
```diff
@@ -18,85 +19,97 @@ groyne,*,225
...
+taxiway,*,72
+aerodrome,*,70
+airfield,*,69
+recreation_track,*,68
+caravan_site,*,67
+recreation_ground,*,66
+quay,*,65
+wharf,*,65
+port_terminal,*,64
+ferry_terminal,*,64
```
nvkelso: Can't share sort values otherwise it creates flickr no Tangram ES... so this and quay and possibly others need to offset away from each other.

spreadsheets/sort_rank/landuse.csv
```diff
@@ -1,4 +1,5 @@
 kind,geometrytype,sort_rank
+embankment,*,274
```
nvkelso: Seems like this should go below power_line, more just under fence?

spreadsheets/sort_rank/water.csv
zerebubuth: It means `boundary` is not set. The empty match in `river,,201` translates to `kind == 'river' and boundary == ''`, which never matched (`boundary` is either `None` or `True`), so all water features were getting a `sort_rank` of 200.

spreadsheets/sort_rank/landuse.csv
zerebubuth: I've re-arranged them in ffe23bd to make them unique. Also added a test, so that we'll get a test failure if a duplicate is added in future.

spreadsheets/sort_rank/landuse.csv
zerebubuth: OK. Fixed in ffe23bd.

nvkelso:
The following two kind values are new to the landuse layer and need to be documented:

- `boatyard`
- `range`

Issue #1716, opened 2018-12-03
Reduce surface tag precision at mid-zooms to 3 values


nvkelso:
_Originally posted by @zerebubuth in https://github.com/tilezen/vector-datasource/pull/1710#issuecomment-443200236_

In https://github.com/tilezen/vector-datasource/pull/1710 we stripped all `surface` properties at mid- zooms. Instead we should re-map surface tags at mid-zooms to still promote more merging, but keep some data fidelity.

It turns out @matkoniecz  makes use of those tags in https://github.com/tilezen/vector-datasource/pull/1710#issuecomment-442541000:

- https://github.com/matkoniecz/Zazolc (personal fork of StreetComplete).
- Style is at https://github.com/matkoniecz/Zazolc/blob/fork/app/src/main/assets/map_theme/fork-style.yaml

The full and raw values would still be available past that zoom (like 13 or 15), so I think these 3 category mappings (`paved`, `compacted`, `unpaved`) are fine at mid-zooms, even for cyclist since they could zoom in and see full details.

For just `minor_roads` we'd also backfill missing `surface` tags as `paved` to promote more merging (rather than leaving it null). But for other road kinds we'd leave null as null.

```
            # paved
            good_surface: 
                filter: {surface: [asphalt, paved, wood, metal, tartan, metal_grid]}

            # compacted
            medium_surface:
                filter: {surface: [concrete, paving_stones, compacted, sett]}

            # unpaved
            bad_surface:
                filter: {surface: [unpaved, dirt, earth, fine_gravel, grass, grass_paver, gravel,
                ground, mud, pebblestone, salt, sand, woodchips, clay, cobblestone,
                cobblestone:flattened, concrete:lanes, concrete:plates, artificial_turf, decoturf]}
                draw:
```

nvkelso:
Verified.

PR #1723, opened 2018-12-10
Reduce surface tag precision

zerebubuth:
Reduce the number of different values for the `surface` tag on roads layer features below the zooms we typically stop naming the feature. This should allow for some information about surface finish, particularly important for walking and cycling maps, without leading to a combinatoric explosion of all the different properties. This should help effective merging at zoom levels where we don't really care about individual features (because they're just "background texture" or because they are too small to be individually noticeable).

Connects to #1716.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1305,7 +1305,7 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `all_bus_networks` and `all_bus_shield_texts`: All of the bus and trolley-bus routes of which this road is a part, and each corresponding shield text. See `bus_network` and `bus_shield_text` below. **Note** that these properties will not be present on MVT format tiles, as we cannot currently encode lists as values.
 * `bus_network`: Note that this is often not present for bus routes / networks. This may be replaced with `operator` in the future, see [issue 1194](https://github.com/tilezen/vector-datasource/issues/1194).
 * `bus_shield_text`: Contains text intended to be displayed on a shield related to the bus or trolley-bus network. This is the value from the `ref` tag and is _not_ guaranteed to be numeric, or even concise.
-* `surface`: Common values include `asphalt`, `unpaved`, `paved`, `ground`, `gravel`, `dirt`, `concrete`, `grass`, `paving_stones`, `compacted`, `sand`, and `cobblestone`. `cobblestone:flattened`, `concrete:plates` and `concrete:lanes` values are transformed to `cobblestone_flattened`, `concrete_plates` and `concrete_lanes` respectively.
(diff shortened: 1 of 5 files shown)


nvkelso:
Addresses everything but:

> For just minor_roads we'd also backfill missing surface tags as paved to promote more merging (rather than leaving it null). But for other road kinds we'd leave null as null.

zerebubuth:
> For just minor_roads we'd also backfill missing surface tags as paved

Yup, I had second thoughts about this, so didn't initially implement it. I've added it (although disabled) in 1f1b9a3.

I ran a small sample of tiles in a zoom sequence roughly over Brooklyn:

Z | X | Y | Nominal Zoom | master | PR 1723 | PR 1723 + backfill | Reduction (master → PR 1723) | Reduction (+backfill)
-- | -- | -- | -- | -- | -- | -- | -- | --
8 | 75 | 96 | 9 | 165512 | 164972 | 164972 | 0.33% | 0.00%
9 | 150 | 192 | 10 | 222914 | 222452 | 222452 | 0.21% | 0.00%
10 | 301 | 385 | 11 | 137376 | 138059 | 138059 | -0.50% | 0.00%
11 | 603 | 770 | 12 | 112679 | 114868 | 114868 | -1.94% | 0.00%
12 | 1206 | 1540 | 13 | 73156 | 74589 | 73656 | -1.96% | 1.25%
13 | 2412 | 3080 | 14 | 70081 | 70945 | 71385 | -1.23% | -0.62%
14 | 4825 | 6161 | 15 | 23749 | 23749 | 23749 | 0.00% | 0.00%
15 | 9650 | 12323 | 16 | 17252 | 17252 | 17252 | 0.00% | 0.00%
16 | 19300 | 24646 | 16 | 5848 | 5848 | 5848 | 0.00% | 0.00%

Note that the sizes here are for roads layer only (although roads are often the largest single layer at mid zooms).

Keeping the simplified `surface` property has a slightly positive effect at zooms 8 & 9 because we're applying the simplification to all road kinds, whereas before we were only dropping `surface` on `minor_road`. Zooms 12 through 14 see a small (>2%) increase in layer size due to the reduced opportunities for merging. Zoom 15 and up is unchanged, as we still keep the full range of values.

Backfilling the default `surface: paved` at zoom 12 has the expected effect: it recovers some of the lost ground by providing better merging opportunities. At zoom 13, however, it makes the layer slightly _bigger_ because it takes a byte each for the index of the key and value that we're adding to those features' properties.

For the basic PR without backfilling, the property indices take up 11,927 bytes across 561 `minor_road` features. With backfilling we're able to merge down to 548 features - but the property indices are now 12,629 bytes. The average bytes per feature increased by almost 2, as we'd expect, but this pushes the overall size unexpectedly higher.

It looks like the benefits of backfilling might be minor at best. We'd get better results if we dropped `surface: paved` and made that an explicit default. However, dropping properties that exist feels much more invasive than backfilling defaults, and I'm not sure we want to do it. Perhaps in v2 we can switch to having an explicit default of `paved`.

nvkelso:
I've filed https://github.com/tilezen/vector-datasource/issues/1724 for the v2 breaking change.

Agree we should not implement the `surface` backfilling right now in this PR. Take it out entirely versus commenting it out and then we're good to merge?

zerebubuth:
> Take it out entirely versus commenting it out and then we're good to merge?

Sounds good. Removed in c842849.

Issue #1715, opened 2018-12-03
Slimming down boundaries


zerebubuth:
Since https://github.com/tilezen/tilequeue/pull/354 fixed a bug, we've been pulling in far more boundaries, median increase in _layer_ size of 40%. Note that the boundaries layer is generally fairly small, and by stripping long names off (#1683) we've reduced layer size significantly at lower zooms. The increase mainly affects mid zooms 9-11, although also zoom 5 (nominal).

We annotate the boundaries with the left and right OSM IDs of the relations they delineate, but these are probably not used in styling. We could remove them to save a significant amount of space in tiles with a lot of boundaries. In addition, this would open up opportunities for merging which are not available at the moment as the left/right ID pair is nearly-unique.

As well, or instead, we could consider showing detail boundaries (`kind: locality`) at higher zooms (currently start at 10). The problem is that countries vary wildly in the density of their localities, so it's possible that a good value for the `min_zoom` in one country wouldn't be appropriate for another. Annotating with the country code and clipping might be an option, although complex.

zerebubuth:
E.g: `9/285/169` (512 tile, so that's zoom 10), near Warsaw, has a lot of very detailed localities. For example, [Robercin](https://www.openstreetmap.org/relation/4258233).

![image](https://user-images.githubusercontent.com/271360/49398875-8ae00800-f737-11e8-8cb1-2b57e981db57.png)

nvkelso:
1. I'm fine always stripping off the `id` from OSM and NE at low zooms (say anything that's zoom 0-12), and possibly all the way up to the maximum zoom of 15 if we really needed it (keeping `id` at the max zoom), for all **boundaries** layer features.

2. Are we currently stripping the labels off the `region` lines per country as part of https://github.com/tilezen/vector-datasource/pull/1700 and https://github.com/tilezen/vector-datasource/issues/1683? That also might help?

3. Regarding stripping names off early `locality` **boundaries** lines, yes! I don't need to see those labels until zoom 13+.

4. In addition, it looks like we start including `locality` boundaries lines at zoom 10 (the same zoom as counties). We can drop `locality` min_zoom to 11, that'll save generally on file size.

**Back to locality names...**

Here's [zoom 12](http://tangrams.github.io/bubble-wrap/#12/37.8668/-122.3561), the point label is enough, along with boundary line (in tiles but not shown in Bubble Wrap).

[...]

PR #1725, opened 2018-12-11
Slimming down boundaries

zerebubuth:
Remove `id:left` and `id:right` properties at mid and low zooms (z<13) which are almost-unique and prevent effective merging. Push `min_zoom` for `locality` boundaries from 10 to 11.

On a test tile, `9/285/169`, which is a pretty extreme example, we get a layer size reduction of 70% (from 78kB down to 23kB) from the `id:left` and `id:right` drop, and a _further_ 83% reduction (from 23kB down to 4kB) by not including `locality` at this zoom level. Overall, the layer is 91% smaller!

Connects to #1715.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -141,6 +141,8 @@ Combination of OpenStreetMap administrative boundaries (zoom >= 8) and Natural E
 
 #### Boundaries properties (common):
 
+* `name`: A suggested label, formed from the left and right region names, if available. If the name appears to be too long to be rendered on the geometry at a particular zoom it may be omitted.
+* `id`: Identifier for the feature, only provided at zoom 13+.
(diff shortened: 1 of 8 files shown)


docs/layers.md
```diff
@@ -151,8 +149,10 @@ Combination of OpenStreetMap administrative boundaries (zoom >= 8) and Natural E
 
 #### Boundaries properties (common optional):
 
-* `id:left`: For the relation on the left side of the boundary line.
-* `id:right`: For the relation on the right side of the boundary line.
+* `name`: A suggested label, formed from the left and right region names, if available. If the name appears to be too long to be rendered on the geometry at a particular zoom it may be omitted.
```
nvkelso: The `name` and `id` are still common properties, we just strip them at low zooms. I like the new description text, but please move them back to the `common` section. (This will better match roads and other layer documentation.)

zerebubuth: OK, cool. Fixed in 4a01b1d.

Issue #1638, opened 2018-09-24
Modify no-name PONI zooms


nvkelso:
```
        #most POI icons without name should be hidden, except these
        no-name:
            filter: { name: false }
            draw:
                mapzen_icon_library:
                    visible: false
            # eg park polygons in SF that have area but no name, sigh
            no-name-no-area:
                filter:
                    kind: true
                    name: false
                draw:
                    mapzen_icon_library:
                        visible: false
            sports-pitch:
                filter:
                    $zoom: { min: 17 }
                    kind: [grass, pitch]
                    kind_detail: [baseball, basketball, football, hockey, soccer, tennis]
                draw:
                    mapzen_icon_library:
                        visible: global.icon_visible_poi_landuse_e
                        priority: 69
            z17:
                filter:
                    $zoom: { min: 17 }
                    kind: [parking]
                draw: { mapzen_icon_library: { visible: global.icon_visible_poi_landuse_e } }
            z18:
                # for different styles, different sets
                # since [picnic_site] doesn't have bubble-wrap icon, it's omitted here
                filter:
                    $zoom: { min: 18 }
                    kind: [drinking_water, information, playground, toilets, traffic_signals]
                draw: { mapzen_icon_library: { visible: global.icon_visible_poi_landuse_e } }
            z19:
                filter:
                    $zoom: { min: 19 }
                    kind: [bicycle_parking]
                draw: { mapzen_icon_library: { visible: global.icon_visible_poi_landuse_e } }
```

zerebubuth:
Please tell me if I'm getting this wrong, but it looks like that `no-name-no-area` subfilter is redundant? We should be setting a `kind` property on every feature, so it doesn't add anything over the top-level `no-name` rule?

nvkelso:
Yes, I agree with that.

nvkelso:
Verified.

PR #1727, opened 2018-12-12
Modify no-name PONI zooms

zerebubuth:
Push down min_zoom on some POIs with no names (PONIs).

- *Sports pitches:* Drop the `min_zoom` to 17 if there's no name and the sport is whitelisted.
- *Parking:* Nothing to do, we had already changed it independently to zoom 17.
- *Drinking water:* Added to PONIs, output only at zoom 18 (not sure there are many named?)
- *Information:* Drop `min_zoom` to 18 if there's no name.
- *Playgrounds:* Added to PONIs, drop `min_zoom` to 18 if there's no name.
- *Toilets:* Nothing to do, had already been changed to 18.
- *Traffic signals:* Dropped to `min_zoom` 18 - pretty sure these don't usually get named.
- *Bicycle parking:* Dropped to 19 when there's no name.

Connects to #1638.

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -29,6 +29,37 @@ global:
     tram_routes: {col: tram_routes}
   - &health_facility_type_kind_detail
     kind_detail: {col: 'health_facility:type'}
+  # common properties for {bi|motor}cycle parking
+  - &cycle_parking_properties
+    access: {col: access}
+    operator: {col: operator}
+    capacity: {col: tags->capacity}
+    covered:
+      case:
+        - when:
+            covered: 'yes'
+          then: true
+        - when:
+            covered: 'no'
+          then: false
+    fee:
+      case:
+        - when:
+            fee: true
+            not:
+              fee: ['no', 'Free', 'free', '0', 'No', 'none']
+          then: true
+      # TODO tests expect this to be false and not omitted, is that what we want?
+        - else: false
+    cyclestreets_id: {col: tags->cyclestreets_id}
+    maxstay: {col: tags->maxstay}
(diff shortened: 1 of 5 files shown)


yaml/pois.yaml
```diff
@@ -29,6 +29,37 @@ global:
...
+            covered: 'no'
+          then: false
+    fee:
+      case:
+        - when:
+            fee: true
+            not:
+              fee: ['no', 'Free', 'free', '0', 'No', 'none']
+          then: true
+      # TODO tests expect this to be false and not omitted, is that what we want?
```
nvkelso: yes

yaml/pois.yaml
```diff
@@ -826,15 +858,29 @@ filters:
...
     output:
       <<: *output_properties
       kind: pitch
       kind_detail: {col: sport}
       tier: 6
+  # pitch - only show no-name pitches for some smaller set of sports, and at a
+  # higher zoom.
+  - filter:
+      leisure: pitch
+      sport: [baseball, basketball, football, hockey, soccer, tennis]
```
nvkelso: The kind_detail were only listed in the example because that's what we have icons for... but I don't think that needs to be limited in the data like this. I'd remove this line and change the comment above.

yaml/pois.yaml
zerebubuth: Sounds good. Removed in 5265c9d.

yaml/pois.yaml
zerebubuth: OK, removed the filter and adjusted tests in 5265c9d.

Issue #1729, opened 2018-12-13
Further places layer plastic surgery


nvkelso:
_Originally posted by @nvkelso in https://github.com/tilezen/vector-datasource/issue_comments#issuecomment-447149749_

For zoom 0 we're still including too many countries. The same is true for zooms 1 and 2. Addressing this will have significant file size savings.

Seems like we're doing 512 min_zoom math instead of 256 math? Like at zoom 0 we should only see features with min_zoom < 1, but many 1.x graded features are included.

New:
![image](https://user-images.githubusercontent.com/853051/49972295-10a54580-fee6-11e8-98d7-8dc8e84df513.png)

Old:
![image](https://user-images.githubusercontent.com/853051/49972286-07b47400-fee6-11e8-8baa-fa0a3b3b50f6.png)

![image](https://user-images.githubusercontent.com/853051/49972360-40544d80-fee6-11e8-9a52-125b52dbaacb.png)
`

Poland is included at zoom 2, but it's min_zoom is 3.

![image](https://user-images.githubusercontent.com/853051/49972840-ba390680-fee7-11e8-910b-f17bc33dd465.png)

nvkelso:
Probably this line can change:

https://github.com/tilezen/vector-datasource/blob/master/vectordatasource/transform.py#L8132

>             if min_zoom is not None and min_zoom <= nominal_zoom + 1:

to:

>            if min_zoom is not None and min_zoom <= nominal_zoom:

zerebubuth:
> Seems like we're doing 512 min_zoom math instead of 256 math? Like at zoom 0 we should only see features with min_zoom < 1

Our 512px tiles are designed to contain the same content as the 4 `z+1` 256px tiles which cover the same area. Therefore, the nominal zoom (i.e: style zoom) of a 512px tile is `z+1`. So the nominal zoom of the `0/0/0` 512px tile is 1.

Also, because clients typically overzoom a zoom `N` tile for zooms `<N+1`, we designed the tile to contain things with a min zoom up to `N+1`. Therefore, the `0/0/0` tile would contain, for example, features with a `min_zoom` of 1.9 because the `0/0/0` tile is the one being displayed when the nominal zoom is 1.9.

I'm happy to make a change as suggested (basically dropping a whole bunch of features needed for overzooming), but we should be aware that it'll be a major change compared with how we've done things in the past. Although it might not matter, if clients aren't evaluating fractional zooms anyway.

nvkelso:
[...]

PR #1734, opened 2018-12-14
Exclude more features from places layer at low zooms

nvkelso:
Connects to https://github.com/tilezen/vector-datasource/issues/1729 to limit **places layer** features to only >= +0.5 of the nominal zoom instead of >= +1 to reduce amount of features (and reduce file size, translations multiply!), especially at low zooms.

This better matches how Natural Earth is curated starting in **version 4.1** for `min_zoom`.

For example, this will remove the 1.7 features above from the zoom 1 tile, but they'll show up in zoom 2 tile and collide out min_zoom 2.0 features – good).

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -8129,7 +8129,7 @@ def min_zoom_filter(ctx):
         for feature in features:
             _, props, _ = feature
             min_zoom = props.get('min_zoom')
-            if min_zoom is not None and min_zoom <= nominal_zoom + 0.5:
+            if min_zoom is not None and min_zoom < nominal_zoom + 1:
                 new_features.append(feature)
 
         layer['features'] = new_features
@@ -8150,7 +8150,10 @@ def tags_set_ne_min_max_zoom(ctx):
     for _, props, _ in layer['features']:
         min_zoom = props.pop('__ne_min_zoom', None)
         if min_zoom is not None:
-            props['min_zoom'] = min_zoom
+            if math.ceil(min_zoom) == math.trunc(min_zoom) + 1:
+                props['min_zoom'] = math.ceil(min_zoom)
+            else:
+                props['min_zoom'] = min_zoom
 
         max_zoom = props.pop('__ne_max_zoom', None)
         if max_zoom is not None:
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -8152,8 +8152,10 @@ def tags_set_ne_min_max_zoom(ctx):
     for _, props, _ in layer['features']:
         min_zoom = props.pop('__ne_min_zoom', None)
         if min_zoom is not None:
-            if math.ceil(min_zoom) == math.trunc(min_zoom) + 1:
-                props['min_zoom'] = math.ceil(min_zoom)
+            # don't overstuff features that are in the long tail of won't display
+            # but make their min_zoom consistent with when they show in tiles
+            if ceil(min_zoom) == trunc(min_zoom) + 1:
+                props['min_zoom'] = ceil(min_zoom)
             else:
                 props['min_zoom'] = min_zoom
 
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -3,7 +3,6 @@
 
 from collections import defaultdict, namedtuple
 from math import ceil
-from math import trunc
 from numbers import Number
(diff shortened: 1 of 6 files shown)


vectordatasource/transform.py
```diff
@@ -8129,7 +8129,7 @@ def min_zoom_filter(ctx):
         for feature in features:
             _, props, _ = feature
             min_zoom = props.get('min_zoom')
-            if min_zoom is not None and min_zoom <= nominal_zoom + 1:
+            if min_zoom is not None and min_zoom <= nominal_zoom + 0.5:
```
zerebubuth: Rather than change the zoom range we include in the tile (i.e: `N` to `N+1`), I think it would be better to change the min zoom we're applying from NE. For example, changing [the setting code](https://github.com/tilezen/vector-datasource/blob/1f74fe90aacfa9e6a8201dff94440eaeedb2b2c0/vectordatasource/transform.py#L8150-L8157) to assign `props['min_zoom'] = min_zoom + 0.5` and similar for `max_zoom`. The two methods should be mathematically identical.

nvkelso: Hmm, I don't like changing the `min_zoom` values to be inconsistent with Natural Earth – mostly because its had for me to wrap my head around as the author of Natural Earth! – but also because it's it's just values in the x.6 to x.9 range that should be ceiling'd up to x+1.0 to be more consistent?

Either way the places layer should be < nominal_zoom +1 not <= to avoid the Poland example?

nvkelso: Attempting fix in https://github.com/tilezen/vector-datasource/pull/1734/commits/9914c2733f102b0be2e77c09edc6b86799d8dda9.

nvkelso:
@zerebubuth Ready for another look.

vectordatasource/transform.py
```diff
@@ -8129,7 +8131,7 @@ def min_zoom_filter(ctx):
         for feature in features:
             _, props, _ = feature
             min_zoom = props.get('min_zoom')
-            if min_zoom is not None and min_zoom <= nominal_zoom + 0.5:
+            if min_zoom is not None and min_zoom < nominal_zoom + 1:
```
zerebubuth: :+1: thanks!

vectordatasource/transform.py
```diff
@@ -8150,7 +8152,13 @@ def tags_set_ne_min_max_zoom(ctx):
     for _, props, _ in layer['features']:
         min_zoom = props.pop('__ne_min_zoom', None)
         if min_zoom is not None:
-            props['min_zoom'] = min_zoom
+            # don't overstuff features into tiles when they are in the
+            # long tail of won't display, but make their min_zoom
+            # consistent with when they actually show in tiles
+            if ceil(min_zoom) == trunc(min_zoom) + 1:
```
zerebubuth: I think this is the same as "if min_zoom is an integer", in which case the `if/else` isn't necessary and we can just write `props['min_zoom'] = ceil(min_zoom)` for everything (because `ceil(x) = x` for integers).

zerebubuth: Something like: `props['min_zoom'] = ceil(min_zoom) if min_zoom % 1 > 0.5 else min_zoom`

Which basically says "if the fractional part is more than half round it up, otherwise keep it the same". If you prefer, it might be more readable as:

```python
if min_zoom % 1 > 0.5:
  min_zoom = ceil(min_zoom)
props['min_zoom'] = min_zoom
```

nvkelso: Fix in https://github.com/tilezen/vector-datasource/pull/1734/commits/ec0584455fcc8fd17a6f5b4ffe0425bc6ab65d69.

nvkelso:
🎉 Huzzah! Merging.

yaml/water.yaml
```diff
@@ -215,7 +215,7 @@ filters:
   #     - {kind: ocean}
   #   table: shp
   - filter: {meta.source: shp}
-    min_zoom: 0
```
zerebubuth: Ooops, I think this might have been the polygon, not the ocean label....

nvkelso: oh, shoot!

nvkelso: ah: `meta.source: shp` is for the **openstreetmapdata.com** source.

nvkelso: Looks like they are all nodes:

image

Issue #1636, opened 2018-09-24
Hide early no-area garden, university, and add allotments


nvkelso:
```
            hide-until-z16-no-area:
                filter:
                    area: false
                    kind: [allotments, college, garden, university]
                    $zoom: { max: 16 }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
Seems like we don't have **allotments** in the POIs layer, but do in Landuse layer. We should add to POIs layer. (We already have **garden** in both places.)

- allotment node: https://www.openstreetmap.org/node/1271465901
- allotment way: https://www.openstreetmap.org/way/32055218

**Garden** node:

- SF botanical garden has garden nodes within garden way, example node: https://www.openstreetmap.org/node/2969748430.

Seems like **college** nodes already default to zoom 16 so no-op.

But **university** nodes default to zoom 15 now, should go to 16.

- https://www.openstreetmap.org/node/4628353540

nvkelso:
The `allotments` examples in the PR are showing up as **landuse** `label_placement` points instead? That's not a ship block, but we should investigate and possibly file issue for next milestone.

![image](https://user-images.githubusercontent.com/853051/50331641-32896400-04b4-11e9-9f97-5d000a8a6a46.png)

The SF garden example point still has min_zoom 16 not 17.

Perhaps this isn't on **staging**?

zerebubuth:
Yup, nothing merged more recently than 2018-12-14 will be on staging, sorry.

nvkelso:
Verified on staging.

PR #1742, opened 2018-12-19
Hide early gardens, allotments, university

zerebubuth:
Hide early point (no area) gardens, universities. Add allotments as a POI type (was already landuse).

Gardens without an area should already have been allocated a [`min_zoom` of 16](https://github.com/tilezen/vector-datasource/blob/fac4f71a526f4eada6144f08783b8d2e3d98c589/yaml/pois.yaml#L853).

Connects to #1636.

### DIFF
--- a/integration-test/1636-hide-early-no-area-garden.py
+++ b/integration-test/1636-hide-early-no-area-garden.py
@@ -1,0 +1,97 @@
+# -*- encoding: utf-8 -*-
+from . import FixtureTest
+
+
+class HideEarlyNoAreaGardenTest(FixtureTest):
+
+    def test_allotments_node(self):
+        import dsl
+
+        z, x, y = (16, 32683, 21719)
+
+        self.generate_fixtures(
+            # https://www.openstreetmap.org/node/1271465901
+            dsl.point(1271465901, (-0.4660196, 51.7557019), {
+                'landuse': u'allotments',
+                'name': u'Midland Hill Allotments',
+                'source': u'openstreetmap.org',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'pois', {
+                'id': 1271465901,
+                'kind': u'allotments',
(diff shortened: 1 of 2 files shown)


yaml/pois.yaml
```diff
@@ -835,7 +835,14 @@ filters:
   ############################################################
   # TIER 6
   ############################################################
-  # allotments - no POI
+  # allotments
+  - filter:
+      landuse: allotments
+    min_zoom: 16
```
nvkelso: Gardens and allotments are basically the same, they should share the same zoom:

>     min_zoom: { min: [ { max: [ 12, { col: zoom }, *tier6_min_zoom ] }, 16 ] }

Or better yet just expand the existing garden filter, like I'll comment?

yaml/pois.yaml
```diff
@@ -548,7 +548,7 @@ filters:
       tier: 3
   # university
   - filter: {amenity: university}
-    min_zoom: { min: [ { max: [ { sum: [ { col: zoom }, 2.55 ] }, *tier3_min_zoom ] }, 15 ] }
+    min_zoom: { min: [ { max: [ { sum: [ { col: zoom }, 2.55 ] }, *tier3_min_zoom ] }, 16 ] }
```
nvkelso: Just noting for review that **college** was already set to default `min_zoom: 16` so no action was needed there.

integration-test/1636-hide-early-no-area-garden.py
```diff
@@ -0,0 +1,97 @@
...
+                'source': u'openstreetmap.org',
+            }),
+        )
+
+        # should have point in POIs
+        self.assert_has_feature(
+            z, x, y, 'pois', {
+                'id': 32055218,
+                'kind': u'allotments',
+                'min_zoom': 16,
```
nvkelso: Should be large enough test to see it show up in zoom 15.

Issue #1632, opened 2018-09-24
Hide early cafe, restaurant to zoom 15


nvkelso:
```
            hide-until-z15-any:
                filter:
                    $zoom: { max: 15 }
                    kind: [cafe, restaurant]
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
Example for **restaurant**:

- https://www.openstreetmap.org/way/409101877#map=19/40.80551/-124.17063

Same applies to **cafe**, but couldn't find easy example so just dupe that way and change tag for test?

PR #1748, opened 2018-12-20
Hide early cafe, restaurant

zerebubuth:
And biergarten, pub, bar, nightclub, fast_food - they're all in the same group and seemed "similar enough" to restaurants & cafes that we wouldn't want them appearing at z9 if restaurants were waiting until z15.

Connects to #1632.


nvkelso:
LGTM, but please hold merging.

Issue #1631, opened 2018-09-24
Hide early post_office


nvkelso:
```
            hide-until-z14-any:
                filter:
                    kind: [natural_forest, natural_park, natural_wood, post_office]
                    $zoom: { max: 14 }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
**post_office** that's getting 13.9 min_zoom now:

- https://www.openstreetmap.org/way/28842845#map=19/37.86867/-122.27000

nvkelso:
These are all **landuse** layer features (not POIs) and can be ignored:

- `natural_forest`
- `natural_park`
- `natural_wood`

zerebubuth:
Here's the breakdown of zooms assigned to all those `amenity` features in California, according to my local database:


Amenity | z6 | z7 | z8 | z9 | z10 | z11 | z12 | z13 | z14 | z15
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
bank |   |   |   | 3 | 31 | 252 | 980 | 35 | 3 |  
cinema |   |   | 2 | 4 | 67 | 64 | 27 | 1 |   |  
courthouse |   |   |   | 4 | 26 | 68 | 10 | 2 |   |  
embassy |   |   |   |   |   | 4 | 3 | 1 |   |  
fire_station |   | 1 | 3 | 9 | 94 | 210 | 514 | 59 | 4 |  
fuel | 2 | 3 | 3 | 4 | 34 | 232 | 937 | 889 | 45 | 4
library |   |   |   | 1 | 72 | 278 | 179 | 29 | 3 |  
police |   |   |   | 6 | 32 | 115 | 60 | 21 | 3 |  
post_office |   |   | 1 | 5 | 37 | 149 | 132 | 53 | 4 |  
theatre |   |   | 2 | 1 | 42 | 154 | 132 | 42 | 11 | 3

I'll tweak it a little to give the following distribution, chosen to try and ease in the largest examples at 14:


Amenity | z14 | z15 | z16
-- | -- | -- | --
bank | 61 | 622 | 621
cinema | 96 | 59 | 10
courthouse | 63 | 40 | 7
embassy |   | 4 | 4
fire_station | 157 | 353 | 384
fuel | 93 | 256 | 1804
library | 146 | 290 | 126
police | 61 | 116 | 60
post_office | 83 | 150 | 148
theatre | 88 | 157 |  142

[...]

PR #1751, opened 2018-12-20
Hide early post office and other POIs.

zerebubuth:
Clamp zoom to 14-16 and adjust zoom offset to balance the zoom at which POIs start to appear so that it's spread across 14-16.

Connects to #1631.


nvkelso:
LGTM, but hold merging.

Issue #1749, opened 2018-12-20
Too many nature_reserve at zoom 10


nvkelso:
* **What did you see?** I can see POIs at zoom 10, when the landuse sibling isn't visible until much later.
* **What did you expect to see?** Probably just do what the polygon is doing.
* **What map location are you having problems with?** 

- https://tangrams.github.io/explorer/#10.0/42.3961/-71.1365/kind/nature_reserve
- https://tangrams.github.io/explorer/#10.0/51.4352/0.0492/kind/nature_reserve

* **Screenshot?** _Props for animated gifs._

![image](https://user-images.githubusercontent.com/853051/50301365-f74b4e80-043b-11e9-96bb-635cc8a09542.png)
![image](https://user-images.githubusercontent.com/853051/50301369-faded580-043b-11e9-97db-e45e74442b91.png)

nvkelso:
Related to: https://github.com/tilezen/vector-datasource/issues/1609

nvkelso:
QA verified.

Issue #1609, opened 2018-09-24
Many small parks and nature_reserves labeled as pois at zoom 10


nvkelso:
10/37.5331/-482.0698

![image](https://user-images.githubusercontent.com/853051/45982350-4c2e2300-c00d-11e8-85a0-12033fcfffd4.png)

Too small, icon covers it up completely, more appropriate for zoom 12:

- https://www.openstreetmap.org/way/43449405#map=12/37.4491/-121.8550

![image](https://user-images.githubusercontent.com/853051/45982395-7bdd2b00-c00d-11e8-89d4-1c9f7f6a8802.png)

Big enough for icon + polygon to show, but really show at 11:

- https://www.openstreetmap.org/way/39339860#map=11/37.3993/-121.8040

![image](https://user-images.githubusercontent.com/853051/45982423-91525500-c00d-11e8-8857-182de21735af.png)

nvkelso:
Seems to be related to:

```
  - filter:
      any:
        leisure: park
        landuse: park
        boundary: national_park
    min_zoom: { min: [ { max: [ 9, { sum: [ { col: zoom }, 2 ] }, *tier2_min_zoom ] }, 14 ] }
    output:
      <<: *output_properties
      kind: park
      tier: 2
```

Which uses:

```
  - &tier2_min_zoom
    lookup:
      key: { col: way_area }
      op: '>='
      table:
        - [ 4, 1000000000 ]
        - [ 5, 1000000000 ]
        - [ 6,  150000000 ]
        - [ 7,  100,000,000 ]
        - [ 8,   10,000,000 ]   # 15,396,834 san bruno at zoom 10
        - [ 9,    5,000,000 ]   #4,451,500 alum rock park at zoom 11
        - [ 10,   1,000,000 ]    # 3,807,010 ed lavin park at zoom 12
                                           # same for John McClaren at 2,311,825 to zoom 12
        - [ 11,    500,000 ]     #    627,235 mount sutro at zoom 13
                                           # same for Glen Park Canyon at 415,605 to zoom 13
        - [ 12,    250,000 ]
        - [ 13,    100,000 ]
        - [ 14,     50000 ]
        - [ 15,      2000 ]
      default: 16
```

Maybe to:

```
  - &tier2_min_zoom_small_parks
    lookup:
      key: { col: way_area }
      op: '>='
      table:
        - [ 8,   160000000 ]
        - [ 9,    40000000 ]
        - [ 10,   10000000 ]
[...]

PR #1752, opened 2018-12-20
Reduce number of nature reserve POIs at zoom 10.

zerebubuth:
Use same tier 2 zoom range for nature reserve POIs as the landuse polygon features.

Connects to #1749.


nvkelso:
This should be consolidated for general fix in https://github.com/tilezen/vector-datasource/issues/1609.

zerebubuth:
I added some stuff to deal with #1609.

yaml/pois.yaml
```diff
@@ -951,15 +965,6 @@ filters:
     output:
       <<: *output_properties
       kind: battlefield
-  # park w/ US Parks Service operator
```
nvkelso: Was this a catch-all for any leisure or landuse park before, but with a bad comment?

nvkelso: Yes, on line 432 in this diff.

Issue #1747, opened 2018-12-20
Add public_transport=station, railway=halt as a synonym for station


zerebubuth:
E.g: https://www.openstreetmap.org/node/2160213344

Context: https://github.com/tilezen/vector-datasource/pull/1741#issuecomment-448758869

nvkelso:
QA verified.

PR #1756, opened 2018-12-27
Add public_transport=station as a synonym for railway=station.

zerebubuth:
When there's also a `rail`, `light_rail` or `railway` tag, in case a more general-purpose public transport station is tagged this way. For example, a bus station without any rail interconnect.

Connects to #1747.

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -233,7 +233,7 @@ global:
         - natural: [ cave_entrance, peak, volcano, geyser, hot_spring, rock, saddle,
             stone, spring, tree, waterfall ]
         - power: [ pole, tower, generator ]
-        - public_transport: [ platform, station, stop_area ]
+        - public_transport: [ platform, stop_area ]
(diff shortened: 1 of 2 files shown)


yaml/pois.yaml
```diff
@@ -233,7 +233,7 @@ global:
         - natural: [ cave_entrance, peak, volcano, geyser, hot_spring, rock, saddle,
             stone, spring, tree, waterfall ]
         - power: [ pole, tower, generator ]
-        - public_transport: [ platform, stop_area ]
+        - public_transport: [ platform, station, stop_area ]
```
nvkelso: big stations with POIs visible at zoom 10 should have a name, please remove station here.

nvkelso: Fixed in https://github.com/tilezen/vector-datasource/pull/1756/commits/628980cd96284e46f76594d0c132f746f34f3f0c.

Issue #1630, opened 2018-09-24
Hide early landmark, museum, prison


nvkelso:
```
            hide-until-z12-any:
                filter:
                    kind: [landmark, museum, prison]
                    $zoom: { max: 12 }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
Looks like we mod the min to be 12 instead of 0 here?

```
  - filter: {amenity: prison}
    min_zoom: { clamp: { min: 0, max: 15, value: { sum: [ { col: zoom }, 2.55 ] } } }
    output:
      <<: *output_properties
      kind: prison
  - filter: {tourism: museum}
    min_zoom: { clamp: { min: 0, max: 16, value: { sum: [ { col: zoom }, 1.43 ] } } }
    output:
      <<: *output_properties
      kind: museum
  - filter: {historic: landmark}
    min_zoom: { clamp: { min: 0, max: 15, value: { sum: [ { col: zoom }, 1.76 ] } } }
    output:
      <<: *output_properties
      kind: landmark
```

nvkelso:
- Prison currently at zoom 9!
    https://www.openstreetmap.org/relation/3955540#map=16/40.7920/-73.8823
- Museum that is < 12
    https://www.openstreetmap.org/relation/2778854#map=16/48.8664/2.3152
- Landmark that is < 12
    https://www.openstreetmap.org/way/27104863#map=19/37.80592/-122.42315

nvkelso:
QA verified.

PR #1757, opened 2018-12-27
Hide early prisons, museums, landmarks until z12.

zerebubuth:
Connects to #1630.


integration-test/1630-hide-early-landmark.py
```diff
@@ -0,0 +1,103 @@
...
+                'wikidata': 'Q120119',
+                'wikipedia': 'en:Rikers Island',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'pois', {
+                'id': 3955540,
+                'kind': 'prison',
+                'min_zoom': 12,
```
nvkelso: Should be `13`, please.

yaml/pois.yaml
```diff
@@ -1119,17 +1119,17 @@ filters:
       <<: *output_properties
       kind: mall
   - filter: {amenity: prison}
-    min_zoom: { clamp: { min: 0, max: 15, value: { sum: [ { col: zoom }, 2.55 ] } } }
+    min_zoom: { clamp: { min: 12, max: 15, value: { sum: [ { col: zoom }, 2.55 ] } } }
```
nvkelso: Should be min: `13`, please.

nvkelso: - `prison` is important, but shouldn't be so prominent – I think zoom 13 would be good, and same zoom we limit hospitals to (which are often bigger). Here's SF JV hall:
    image
    and the 1st prison in California is only good for zoom 13: https://www.openstreetmap.org/way/304320226
    same for New York City's finest at zoom 13: https://www.openstreetmap.org/relation/3955540

Issue #1627, opened 2018-09-24
Modify zoom of bicycle, theatre, water_tower


nvkelso:
tl;dr

Only bicycle, theatre, and water_tower are relevant here. See comment below.

```
            hide-until-z17-no-area:
                filter:
                    area: false
                    $zoom: { max: 17 }
                    # note: atm and drinking_water appear hear and in area versions
                    kind: [accountant, administrative, advertising_agency, animal, architect, association, atm, bakery, bed_and_breakfast, bicycle, bicycle_parking, bicycle_rental_station, books, butcher, car, car_repair, chalet, clothes, company, computer, consulting, convenience, doityourself, drinking_water, dry_cleaning, educational_institution, emergency_phone, employment_agency, estate_agent, fashion, financial, florist, foundation, gift, government, greengrocer, guest_house, hairdresser, hostel, hotel, insurance, it, jewelry, lawyer, mast, memorial, mobile_phone, motel, newspaper, ngo, notary, optician, parking, pet, physician, playground, political_party, post_box, religion, research, subway_entrance, tax_advisor, telecommunication, telephone, theatre, therapist, toilets, traffic_signals, travel_agent, water_tower]
                draw:
                    mapzen_icon_library:
                        visible: false

```

nvkelso:
Change these, for no-area (node) case:

- **bicycle** shop (zoom 15 now, might interfere with Walkabout bike overlay, consider 16 instead?), really this should be an area throttle, similar to cafe?
    - bike node, to zoom 17: https://www.openstreetmap.org/node/2484293076#map=19/37.77108/-122.43223
    - bike way, to zoom 16?: https://www.openstreetmap.org/way/260354461#map=16/37.7771/-122.4077
    - bike way, to zoom 15? https://www.openstreetmap.org/way/264534357#map=15/37.7665/-122.4527

Maybe:

```
  - filter: {shop: bicycle}
    min_zoom: { clamp: { min: 15, max: 17, value: { sum: [ { col: zoom }, 2.5 ] } } }
    output:
      <<: *output_properties
      kind: bicycle
```

- **theatre** from zoom 16 now, but it already has an area throttle, so peal off to it's own? 
    - node, zoom 17?: https://www.openstreetmap.org/node/358805392#map=17/37.78302/-122.41223
    - medium way, zoom 16?: https://www.openstreetmap.org/way/35115840#map=16/37.7793/-122.4146
[...]

PR #1758, opened 2018-12-28
Modify zoom bicycle, theatre, water tower

zerebubuth:
Modify the `min_zoom` of bicycle shops and theatres based on area, and water towers based on height.

Notes:

1. I swapped the z15/16 examples for bicycle shops, as SF Bike Connection is larger than American Cyclery Too.
2. I used a different example for a short water tower with a height. Your example had a height of 9m, but I thought it would provide better test coverage to have one in the 10-20m range.

Connects to #1627.

### DIFF
--- a/integration-test/1627-modify-zoom-bicycle-theatre-water-tower.py
+++ b/integration-test/1627-modify-zoom-bicycle-theatre-water-tower.py
@@ -94,3 +94,90 @@ class BikeShopTest(FixtureTest):
                 # min_zoom between 16 and 17, less than the node at z17
                 'min_zoom': lambda z: 16 <= z < 17,
             })
+
+
+class TheatreTest(FixtureTest):
+
+    def test_theatre_node(self):
+        import dsl
+
+        z, x, y = (16, 10483, 25329)
+
+        self.generate_fixtures(
+            # https://www.openstreetmap.org/node/358805392
+            dsl.point(358805392, (-122.411371, 37.782168), {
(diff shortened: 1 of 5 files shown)


integration-test/1627-modify-zoom-bicycle-theatre-water-tower.py
```diff
@@ -0,0 +1,253 @@
...
+                'website': 'http://bikeconnection.net/',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'pois', {
+                'id': 260354461,
+                'kind': 'bicycle',
+                # should be in z15 tile, so min_zoom between 15 and 16
+                'min_zoom': lambda z: 15 <= z < 16,
```
nvkelso: 👍

integration-test/1627-modify-zoom-bicycle-theatre-water-tower.py
```diff
@@ -0,0 +1,253 @@
...
+                'source': 'openstreetmap.org',
+                'wikidata': 'Q7103971',
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'pois', {
+                'id': 35115840,
+                'kind': 'theatre',
+                'min_zoom': lambda z: 16 <= z < 17,
```
nvkelso: There's probably more work to do here with respect to building volume or other indicators – as this is one of the more significant theatres in SF. But this is good for now.

Issue #1625, opened 2018-09-24
Hide early parking when no area, show parking areas earlier


nvkelso:
```
            hide-until-z18-no-area:
                filter:
                    $zoom: { max: 18 }
                    area: false
                    kind: [bench, parking, waste_basket]
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
**Bench** and **waste_basket** are already `min_zoom` 18.

**Parking**:

- node, zoom 18: https://www.openstreetmap.org/node/1706619902#map=19/37.77619/-122.44515
- way huge, zoom 15: https://www.openstreetmap.org/way/32928608#map=15/37.7293/-122.4853
- way medium, zoom 16: https://www.openstreetmap.org/way/162093420#map=16/37.7286/-122.4760
- way small, zoom 17: https://www.openstreetmap.org/way/506510951#map=17/37.73077/-122.48156

nvkelso:
Related to https://github.com/tilezen/vector-datasource/pull/1727/files.

nvkelso:
QA verified.

PR #1760, opened 2018-12-28
Estimate capacity for parking lots and garages.

zerebubuth:
If the feature provides a `capacity` tag, then use that as long as it's a valid integer (some aren't, such as `capacity=limited`). If that doesn't work, try to estimate the number of spaces from the area and, where possible, the number of levels. From fitting the area to features with a capacity, it looks like 46 square Mercator "meters" isn't too far off a reasonable value per parking space.

Once we've estimated the capacity, use that to figure out what the min zoom should be.

Connects to #1625.

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1051,3 +1051,31 @@ BEGIN
   RETURN 17;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+-- return the capacity of a parking lot, or estimate it from the area and building levels.
+CREATE OR REPLACE FUNCTION tz_estimate_parking_capacity(capacity TEXT, parking TEXT, levels TEXT, way_area REAL)
+RETURNS INTEGER AS $$
+DECLARE
+  levels_int INTEGER;
+  spaces_per_level INTEGER;
+BEGIN
+  -- if the capacity is set, then use that.
+  IF capacity ~ '^[0-9]+$' THEN
+    RETURN capacity::integer;
+  END IF;
+  -- otherwise, try to use the information we have to guess the capacity
+  spaces_per_level := (way_area / 46.0)::integer;
+  levels_int := CASE
+    WHEN levels ~ '^[0-9]+$' THEN levels::integer
+    WHEN parking = 'multi-storey' THEN 2
+    ELSE 1
+  END;
+  -- if we get a silly answer, don't set that - just return NULL to indicate
+  -- that we're unsure.
+  IF levels_int * spaces_per_level > 0 THEN
+    RETURN levels_int * spaces_per_level;
+  ELSE
+    RETURN NULL;
+  END IF;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -956,12 +956,6 @@ filters:
           - [ 16,  100 ]
           - [ 17,   10 ]
         default: 18
-
-
- - [ 15, 1000 ]    # was 350
-           - [ 16,   200 ]    # was 130
-           - [ 17,      45 ]    # no change, seems fine
-
     output:
       <<: *output_properties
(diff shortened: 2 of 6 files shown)


data/functions.sql
```diff
@@ -1051,3 +1051,31 @@ BEGIN
   RETURN 17;
 END;
 $$ LANGUAGE plpgsql IMMUTABLE;
+
+-- return the capacity of a parking lot, or estimate it from the area and building levels.
+CREATE OR REPLACE FUNCTION tz_estimate_parking_capacity(capacity TEXT, parking TEXT, levels TEXT, way_area REAL)
```
nvkelso: This is smart!

data/functions.sql
```diff
@@ -1051,3 +1051,31 @@ BEGIN
...
+DECLARE
+  levels_int INTEGER;
+  spaces_per_level INTEGER;
+BEGIN
+  -- if the capacity is set, then use that.
+  IF capacity ~ '^[0-9]+$' THEN
+    RETURN capacity::integer;
+  END IF;
+  -- otherwise, try to use the information we have to guess the capacity
+  spaces_per_level := (way_area / 46.0)::integer;
```
nvkelso: Just for sanity check:
https://ag.tennessee.edu/cpa/Information%20Sheets/CPA%20222.pdf

1 acre = 4046 sq meters; 

> If a field that is 180 feet by 242 feet (approximately 1 acre) is designed with six rows of parking spaces with each parking space being approximately 10 feet by 18 feet and the traffic lanes are 24 feet wide, approximately 150 spaces can be designed. 

So that's 4046 sq meters / 150 spaces = 27 sq meters per space for a parking garage type situation. 

> Many fields used for parking are not flat, not perfectly square and do not accommodate lined parking spaces. Accounting for these and other imperfections in farm fields that are often used for customer parking, many on-farm marketers often estimate that they can park between 80 and 100 automobiles per acre. 

So that's 4046 sq meters / 89 spaces = 45 sq meters per space for bigger surface parking lot situation.

nvkelso: So if it's a parking garage can fit up to 2x as many cars as surface lot (even more if at diagonal, which is rare), all things considered. 

But the point is just to scale it based on something, and this is a reasonable :)

yaml/pois.yaml
```diff
@@ -946,7 +946,15 @@ filters:
   ############################################################
   # parking
   - filter: {amenity: parking}
-    min_zoom: 17
+    min_zoom:
+      lookup:
+        key:  { call: { func: tz_estimate_parking_capacity, args: [ { col: capacity }, { col: 'parking' }, { col: 'building:levels' }, { col: 'way_area' } ] } }
+        op: '>='
+        table:
+          - [ 15, 350 ]
```
nvkelso: Perhaps this should be even bigger now that we have the capacity measurements?

```
           - [ 15, 1000 ]    # was 350
           - [ 16,   200 ]    # was 130
           - [ 17,      45 ]    # no change, seems fine
```

Looking at Disneyland in California:
17/33.80043/-117.91088

image

There are also some absurdly large parking garages (maybe OSM tagging issues?) in Los Vegas on the Strip:
15/36.1038/-115.1634

But I think still zoom 15 is good first zoom for those (and likely client side don't show the name).

image

nvkelso: The parking garages in Los Vegas are tagged correctly, our building logic just isn't exporting parking_garage as the kind_detail correctly.

data/functions.sql
zerebubuth: Yup, there's a wide range in the existing data. At the moment, the capacity estimate probably has some pretty large error bars. We can fix it so that `parking != surface` implies 27 square meters instead of 45, but we should also include a projection correction. Because `way_area` is in square Mercator meters, parking lots will seem higher capacity at higher latitudes. Follow-up in #1768.

Issue #1624, opened 2018-09-24
Hide early bike parking, car sharing


nvkelso:
```
            hide-until-z19-any:
                filter:
                    $zoom: { max: 19 }
                    kind: [bicycle_parking, burial_vault, car_sharing]
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
- **car_sharing**
    https://www.openstreetmap.org/node/1409407263

nvkelso:
QA verified.

PR #1761, opened 2018-12-28
Hide bicycle parking and car sharing POIs until z19.

zerebubuth:
Connects to #1624.

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -1637,7 +1637,7 @@ filters:
     output:
       <<: *output_properties
       kind: car_sharing
-- filter:
+  - filter:
       barrier: [block, bollard, cattle_grid]
     min_zoom: 16
     output:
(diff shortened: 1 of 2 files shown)


yaml/pois.yaml
```diff
@@ -1619,11 +1619,17 @@ filters:
...
   - filter:
-      amenity: [bus_station, car_rental, car_sharing, recycling, shelter]
+      amenity: [bus_station, car_rental, recycling, shelter]
     min_zoom: 16
     output:
       <<: *output_properties
       kind: {col: amenity}
+  - filter:
+      amenity: car_sharing
+    min_zoom: 19
```
nvkelso: We should treat this more like we do `bicycle_parking`, also to abide by SemVer so the default for names only changes one zoom (16 to 17), but then the no-names can go to zoom 19.

```
  - filter:
      amenity: bicycle_parking
      name: true
    min_zoom: 17
    output:
      <<: *output_properties
      <<: *cycle_parking_properties
      kind: bicycle_parking
  - filter:
      amenity: bicycle_parking
    min_zoom: 19
    output:
      <<: *output_properties
      <<: *cycle_parking_properties
      kind: {col: amenity}
```

Issue #1623, opened 2018-09-24
Hide early military


nvkelso:
```
            # hide based on kind and area filters
            military-early:
                filter:
                    kind: [military]
                    any:
                        # show labels for smaller landuse areas at higher zooms
                        - { $zoom: [8], area: { max: 50000000 } }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
Another one where the icon completely covers the feature at zoom 9, shouldn't show up until zoom 11:

![image](https://user-images.githubusercontent.com/853051/46221141-18027d00-c301-11e8-9525-7af4177ef3c3.png)

nvkelso:
https://www.openstreetmap.org/way/44018189#map=11/38.2479/-122.7766

nvkelso:
Looks like the area breaks in https://github.com/tilezen/vector-datasource/issues/1609 would solve this problem.

nvkelso:
Also verify against this view of London / southern England:

![image](https://user-images.githubusercontent.com/853051/51057732-4b5cd500-159b-11e9-9fa9-9416436b4e97.png)

nvkelso:
QA verified.

PR #1762, opened 2018-12-28
Hide early military POIs.

zerebubuth:
Connects to #1623.


nvkelso:
Added a zoom 8 test, merging.

Issue #1622, opened 2018-09-24
Hide early university, college


nvkelso:
```
            university-early:
                filter:
                    kind: [university, college]
                    any:
                        # show labels for smaller landuse areas at higher zooms
                        - { $zoom: [11], area: { max: 2000000 } }
                        - { $zoom: [12], area: { max: 500000 } }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
- **university**
    - shouldn't be visible at all before zoom 10 (zoom 10 okay), now it's available earlier (should change)
       https://www.openstreetmap.org/relation/309918#map=10/36.6534/-121.7544
       ![image](https://user-images.githubusercontent.com/853051/49908125-f740c280-fe2d-11e8-988b-ae0f22267941.png)
    - should be zoom 11: https://www.openstreetmap.org/way/29268613#map=15/37.4305/-122.1685
      ![image](https://user-images.githubusercontent.com/853051/49907992-71247c00-fe2d-11e8-8d09-9ac57354c958.png)
    - should be zoom 12: https://www.openstreetmap.org/way/124842735#map=17/37.64142/-122.10823
     ![image](https://user-images.githubusercontent.com/853051/49907931-315d9480-fe2d-11e8-8f91-33361e23e347.png)
- **college**
    - to zoom 12 (not 11): https://www.openstreetmap.org/way/31486857#map=12/37.8411/-122.1091
       ![image](https://user-images.githubusercontent.com/853051/49908379-168c1f80-fe2f-11e8-8791-ea210901ceb7.png)
    - don't see an easy zoom 11 example...

nvkelso:
QA verified.

PR #1763, opened 2018-12-28
Hide early universities, colleges.

zerebubuth:
Connects to #1622.


yaml/pois.yaml
```diff
@@ -562,14 +562,34 @@ filters:
...
+          - [ 11,  5000000 ]
+          - [ 12,  1000000 ]
+          - [ 13,   500000 ]
+          - [ 14,   500000 ]
+          - [ 15,   100000 ]
+        default: 16
     output:
       <<: *output_properties
-      kind: university
+      kind: { col: amenity }
```
nvkelso: This is a no-op, and the old way is more legible.

yaml/pois.yaml
```diff
@@ -562,14 +562,34 @@ filters:
...
+    min_zoom:
+      lookup:
+        key: { col: way_area }
+        op: '>='
+        table:
+          - [ 10, 10000000 ]
+          - [ 11,  5000000 ]
+          - [ 12,  1000000 ]
+          - [ 13,   500000 ]
+          - [ 14,   500000 ]
```
nvkelso: Zoom 14 has same area threshold as zoom 13? Seems different than `tier3_min_zoom` breaks (below), which are also weird.

```
  - &tier3_min_zoom
    lookup:
      key: { col: way_area }
      op: '>='
      table:
        - [ 8,  10000000 ]
        - [ 9,   5000000 ]
        - [ 10,  1000000 ]
        - [ 11,   500000 ]
        - [ 12,   500000 ]
        - [ 13,   100000 ]
        - [ 14,    50000 ]
        - [ 15,     2000 ]
      default: 16
```

nvkelso: Comparing tier3 and tier4, I think the following cuts are more consistent with old behavior?

```
- [ 10, 10000000 ]
- [ 11,  1000000 ]
- [ 12,   500000 ]
- [ 13,   100000 ]
- [ 14,    50000 ]
- [ 15,     2000 ]
```

yaml/pois.yaml
```diff
@@ -562,14 +562,34 @@ filters:
...
       tier: 3
   # college
   - filter: {amenity: college}
-    min_zoom: { min: [ { max: [ { sum: [ { col: zoom }, 2.35 ] }, *tier3_min_zoom ] }, 16 ] }
+    min_zoom:
+      lookup:
+        key: { col: way_area }
+        op: '>='
+        table:
+          - [ 12,   600000 ]
```
nvkelso: Similar comment to https://github.com/tilezen/vector-datasource/pull/1763/files#r244405058, though this one used a different _scale factor_ of **2.35** versus **2.55** so maybe that's why the values are less as proposed?

nvkelso: I suspect we can just reuse:

```
- [ 12,   500000 ]
- [ 13,   100000 ]
- [ 14,    50000 ]
- [ 15,     2000 ]
```

yaml/pois.yaml
zerebubuth: Ooops, yeah. Thanks for spotting that. I had tried to merge the `university` and `college` parts, but it turned out to not be possible with the area thresholds we wanted and I must have forgotten to put that back the way it was.

yaml/pois.yaml
zerebubuth: Cool, thanks. Updated in 5af8d31.

Yup, the `tier3_min_zoom` was probably where I started from, and the repeated 500,000 was copy pasta from there. As written, it does mean that nothing in tier 3 will get a `min_zoom` of 12. Would we want to change that to 200,000 perhaps?

yaml/pois.yaml
zerebubuth: OK, thanks. Updated in 5af8d31.

yaml/pois.yaml
nvkelso: Yes, I'd be fine changing tier3 filter generically to `200,000` as you propose (as well).

nvkelso:
Fix looks good, and I've filed https://github.com/tilezen/vector-datasource/issues/1769 for the general tier 3 fix.

Issue #1621, opened 2018-09-24
Hide early school, kindergarten


nvkelso:
```
            school-early:
                filter:
                    kind: [school, kindergarten]
                    any:
                        # show labels for smaller landuse areas at higher zooms
                        - { $zoom: [13], area: { max: 100000 } }
                        - { $zoom: [14], area: { max: 50000 } }
                        - { $zoom: [15], area: { max: 10000 } }
                        - { $zoom: [16], area: { max: 5000 } }
                        - { $zoom: [17], area: { max: 2000 } }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
This is already using **tier6_min_zoom**, which seems like okay logic, but let's split it off to follow the above guidelines instead, and only start at zoom 13, default to 18 when no area.

**School**, by zoom:

- https://www.openstreetmap.org/way/39335805#map=13/37.7776/-122.4917
- https://www.openstreetmap.org/way/226324839#map=14/37.74764/-122.48013
- https://www.openstreetmap.org/way/162470880#map=15/37.75786/-122.50194
- https://www.openstreetmap.org/way/162470889#map=16/37.7614/-122.5022
https://www.openstreetmap.org/node/358803287#map=18/37.76986/-122.44519

**kindergarden** is same general logic, single test should suffice:

- https://www.openstreetmap.org/way/256908810#map=18/37.78281/-122.42791

nvkelso:
QA verified.

PR #1764, opened 2018-12-28
Hide early schools and kindergartens.

zerebubuth:
Alterations:

1. The example schools for zoom 13 and 14 ([George Washington High](https://www.openstreetmap.org/way/39335805) and [Abraham Lincoln High](https://www.openstreetmap.org/way/226324839)) have very similar areas, and the 14 example (Abe Lincoln) was actually slightly smaller. Swapped it out for [McKinleyville Middle](https://www.openstreetmap.org/way/295324170) instead.
2. The [zoom 16 example](https://www.openstreetmap.org/way/162470889) is a park. Swapped it for a similar sized school, [Bessie Carmichael Elementary](https://www.openstreetmap.org/way/353732824).

Connects to #1621.

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -926,11 +926,24 @@ filters:
       kind: playground
       tier: 6
   # school
-  - filter: {amenity: school}
-    min_zoom: { min: [ { max: [ { sum: [ { col: zoom }, 2.3 ] }, *tier6_min_zoom ] }, 15 ] }
+  - filter:
+      amenity:
+        - school
+        - kindergarten
+    min_zoom:
+      lookup:
+        key: { col: way_area }
+        op: '>='
+        table:
+          - [ 13,  100000 ]
+          - [ 14,   50000 ]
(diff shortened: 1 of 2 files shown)


yaml/pois.yaml
```diff
@@ -926,11 +926,24 @@ filters:
...
+      lookup:
+        key: { col: way_area }
+        op: '>='
+        table:
+          - [ 13,  100000 ]
+          - [ 14,   50000 ]
+          - [ 15,   10000 ]
+          - [ 16,    5000 ]
+          - [ 17,    2000 ]
+        default: 18
```
nvkelso: Our SemVer says we can only change a full zoom default at a time to be minor version increment. So let's make the default zoom 17 and remove the zoom 17 area filter?

nvkelso: Resolved via https://github.com/tilezen/vector-datasource/pull/1764/commits/5b1002d33bb145dde6de984b5638db53ae83d465.

Issue #1619, opened 2018-09-24
Hide some earlier golf courses


nvkelso:
```
            golf_course-early:
                filter:
                    kind: golf_course
                    any:
                        - { $zoom: [12], area: { max: 2500000 } }
                draw:
                    mapzen_icon_library:
                        visible: false
```

nvkelso:
Looks like we could use the new area filters in https://github.com/tilezen/vector-datasource/issues/1609 for this...

nvkelso:
QA verified.

PR #1765, opened 2018-12-28
Hide early golf courses.

zerebubuth:
Connects to #1619.


yaml/pois.yaml
```diff
@@ -626,7 +626,14 @@ filters:
       tier: 4
   # golf_course
   - filter: {leisure: golf_course}
-    min_zoom: { min: [ { max: [ 12, { sum: [ { col: zoom }, 2 ] }, *tier4_min_zoom ] }, 14 ] }
+    min_zoom:
+      lookup:
+        key: { col: way_area }
+        op: '>='
+        table:
+          - [ 12, 2000000 ]
```
nvkelso: This was `2500000` in the scene file / issue, but 2 million also seems reasonable unless proven otherwise in QA review round.

Issue #874, opened 2016-06-29
Automate tilejson generation


rmarianski:
The tilejson tends to [fall out of date](https://github.com/tilezen/vector-datasource/issues/872) as we make changes to layers. Ideally, we'd be able to create an automated process to keep them fresh.

nvkelso:
Should we include more property keys? We have a ton, but only list 1 up to 6 for roads in this config file.

rmarianski:
We can try to infer them from the yaml file, but it won't be exactly right. Some queries include properties directly, and we have transforms that update the properties too.

We had also previously discussed being able to generate some documentation and these tilejson files from a single source. Or maybe there's a way that we can annotate the documentation itself to allow us to extract the relevant data to generate the tilejson.

rmarianski:
Thinking about this a little bit more, we should be able to generate this by annotating the yaml and python transform configuration. Additionally, it might be worthwhile to convert the documentation to a template, and have the lists of possible kind values be injected in from this generated list.

zerebubuth:
I agree - we can be more explicit about the values we're including, or the lists of keys we expect from the database. That, plus documentation strings and "for more information" links should be able to provide us pretty comprehensive docs in both HTML / Markdown and TileJSON.

zerebubuth:
Since #1771, this might be within reach? Worth another look soon?

PR #1771, opened 2019-01-07
Add script to output all the kinds in the YAML files.

zerebubuth:
It will also read the `sort_rank` spreadsheet CSVs and output the sort rank information where available.

The output is a table of the layer, kind, start zoom and sort rank (or `None` where those last two aren't available). The "start zoom" is the minimum `min_zoom` at which the feature could possibly appear. For some features where we use an area-based scaling rule, this could theoretically be zero if the feature were large enough - but this doesn't happen in practice. However, it indicates that we could need to put a `clamp` on the range to prevent them from showing up at unexpectedly low zooms. The start zoom can be `None` if the information is obtained externally, e.g: from an external table such as WOF or NE.

Additionally, did some cleanup to the YAML files:

1. The WOF layers didn't explicitly mention their `placetype`, so that has been whitelisted now. In practice, these were the only `placetype`s values in the database, but it's helpful to be able to know that by looking at the YAML and not having to dig into the database.
2. There were many instances of a `tags->` prefix on column names, which is a holdover from when we used to have OSM tables which were partially column-based and partly hstore-based. These aren't needed any more, so I removed them.

Related to https://github.com/tilezen/vector-datasource/issues/988#issuecomment-450453906.

### DIFF
--- a/scripts/all_the_kinds.py
+++ b/scripts/all_the_kinds.py
@@ -179,7 +179,6 @@ def parse_item(layer_name, item, sort_rank):
             all_kinds = parse_filter_for_col(kind['col'], item['filter'])
 
         elif kind.keys() == ['case']:
-            # TODO: figure this out
             all_kinds = parse_case_for_kinds(kind['case'])
 
         else:
--- a/scripts/all_the_kinds.py
+++ b/scripts/all_the_kinds.py
@@ -1,0 +1,288 @@
+from collections import namedtuple
+
+
+# we don't have data to evaluate the function, so this just looks up the
+# minimum possible value the function could return.
+KNOWN_FUNCS = {
+    'mz_calculate_path_major_route': 8,
+    'mz_get_min_zoom_highway_level_gate': 14,
+    'tz_looks_like_service_area': 13,
+    'tz_looks_like_rest_area': 13,
+    'mz_calculate_ferry_level': 8,
+}
+
+
+# structures to return data from the parsing functions. basically a key->value
+# mapping where the keys are (layer, kind) tuples and the values are the other
+# stuff we parsed out of the various YAML/CSV files.
+KindKey = namedtuple('KindKey', 'layer kind')
+KindInfo = namedtuple('KindInfo', 'min_zoom sort_rank')
+
+
+# parses the values for a column out of a filter definition. this is used when
+# the output defines "kind: {col: some_column}", and we want to find out what
+# values "some_column" could have.
+#
+# this is most often seen in things such as:
+#
+#  - filter:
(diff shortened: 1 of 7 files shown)


scripts/all_the_kinds.py
```diff
@@ -0,0 +1,288 @@
...
+#    output:
+#      <<: *output_properties
+#      kind: {col: amenity}
+#
+# where we've white-listed the amenity values in the filter, so we know all the
+# possible values that the kind can have.
+#
+# theoretically, this can be arbitrarily complex, but it turns out that in all
+# the cases we have at the time of writing, the whitelist is available at the
+# top level of the filter! lucky us :-)
```
rmarianski: :zap:

scripts/all_the_kinds.py
```diff
@@ -0,0 +1,288 @@
...
+
+    if isinstance(kind, (str, unicode)):
+        all_kinds = [kind]
+
+    elif isinstance(kind, dict):
+        if kind.keys() == ['col']:
+            all_kinds = parse_filter_for_col(kind['col'], item['filter'])
+
+        elif kind.keys() == ['case']:
+            # TODO: figure this out
```
rmarianski: this comment still apply?

scripts/all_the_kinds.py
```diff
@@ -0,0 +1,288 @@
...
+# where we've white-listed the amenity values in the filter, so we know all the
+# possible values that the kind can have.
+#
+# theoretically, this can be arbitrarily complex, but it turns out that in all
+# the cases we have at the time of writing, the whitelist is available at the
+# top level of the filter! lucky us :-)
+def parse_filter_for_col(col, filter_defn):
+    if col.startswith('tags->'):
+        raise AssertionError("Output kind column starts with deprecated "
+                             "'tags->' prefix: %r" % (col,))
```
rmarianski: I see we've decided to bite the bullet here! 🎉

scripts/all_the_kinds.py
zerebubuth: Nope! Well spotted, thanks. Removed in 4977583.

scripts/all_the_kinds.py
zerebubuth: The test suite seems large enough now that I thought we'd probably catch a regression. There's a lot of noise from the change, but at least it's not part of a "feature PR".

zerebubuth:
> Not sure what the original motivation here was, but is it possible to push some of this information into the tilejson?

That's a great idea. And, as it turns out, something we've been wanting to do for a while: #874. We should take another look at doing it.

And, if we put the description in the YAML, perhaps we could generate a chunk of the docs automatically too?

Issue #1738, opened 2018-12-17
Drop name, name:left, name:right from locality boundary lines at zoom 11 and 12


nvkelso:
Related to followup in https://github.com/tilezen/vector-datasource/issues/1737 from https://github.com/tilezen/vector-datasource/pull/1725.

* **What did you see?** Lots of name properties still on locality lines in boundaries layer at zoom 11 and 12.
* **What did you expect to see?** No name properties on locality lines at zoom 11 and 12, just one big multi-line-string ;)
* **What map location are you having problems with?** 11/37.8791/-122.3927
* **Screenshot?** 
![image](https://user-images.githubusercontent.com/853051/50115812-55d6b980-01fd-11e9-8398-3ae7b4b03626.png)

nvkelso:
QA verified.

PR #1774, opened 2019-01-07
Drop names from locality boundaries at z11 & z12.

zerebubuth:
Connects to #1738.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -616,6 +616,15 @@ post_process:
         - "id:left"
         - "id:right"
 
+  # drop any name on locality boundaries at zooms 11 and 12 (min two zooms).
+  - fn: vectordatasource.transform.drop_names
+    params:
+      source_layer: boundaries
+      start_zoom: 11
+      end_zoom: 13
+      where: >-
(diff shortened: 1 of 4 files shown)


queries.yaml
```diff
@@ -616,6 +616,15 @@ post_process:
         - "id:left"
         - "id:right"
 
+  # drop any name on locality boundaries at zooms 11 and 12 (min two zooms).
+  - fn: vectordatasource.transform.drop_names
```
nvkelso: nit: consider adding a note here saying that the data comes in at zoom 11 in the boundaries.yaml and changes should be coordinated, and add note over there as well?

zerebubuth: Cool, good idea. Added in 7c0a1d8.

(Alternatively, we could just have the `start_zoom` here be 0, and then it wouldn't matter if we changed the locality `min_zoom` to 10... but that means running this filter at 10 zoom levels where there isn't any data for it to process, which seems wasteful - so instead just sticking with pointing the problem out and requiring human attention to make sure it doesn't get out of sync.)

Issue #1730, opened 2018-12-14
Further water layer name dropping at low zooms


nvkelso:
Following up from https://github.com/tilezen/vector-datasource/issues/1477 and https://github.com/tilezen/vector-datasource/pull/1714/files, and based on original zoom & area table in https://github.com/tilezen/vector-datasource/issues/1477#issuecomment-442269890.

We can drop even more water layer labels derived from polygons at zooms 0-5.

Zooms 6+ seems mostly okay. But I don't understand how that works... seems like we drop water labels based on a width / area treshhold – where does that happen?

But there's also funk with how the Natural Earth names are brought in at low zooms which make it look better than it should? Example at bottom.

**Need to change so:** 

- [ ] No lake labels at zoom 0-3
- [ ] Then zoom 4 and 5 have different threshold:
- [x] Then zoom 6+ no change? Might need zoom 6,7 for Natural Earth of the area threshold is OSM only?

```
                        - { $zoom: { min: 4 },  area: { min: 40000000000 } }
                        - { $zoom: { min: 5 },  area: { min: 10000000000 } }
```

The original in https://github.com/tangrams/walkabout-style/blob/gh-pages/walkabout-style.yaml#L1258-L1279

```
        water-area-labels:
            visible: global.text_visible_water_labels
            filter:
                all:
                    - name: true
                    - $geometry: point
                    - not: { kind: [riverbank, dock, playa] }
                    - area: true
                any:
                    # show labels for smaller landuse areas at higher zooms
                    - { $zoom: { min: 5 },  area: { min: 10000000000 } }
                    - { $zoom: { min: 6 },  area: { min: 5000000000 } }
                    - { $zoom: { min: 7 },  area: { min: 400000000 } }
                    - { $zoom: { min: 8 },  area: { min: 200000000 } }
                    - { $zoom: { min: 9 },  area: { min: 100000000 } }
                    - { $zoom: { min: 10 }, area: { min: 10000000 } }
                    - { $zoom: { min: 11 }, area: { min: 4000000 } }
                    - { $zoom: { min: 12 }, area: { min: 750000 } }
                    - { $zoom: { min: 13 }, area: { min: 100000 } }
                    - { $zoom: { min: 14 }, area: { min: 50000 } }
                    - { $zoom: { min: 15 }, area: { min: 10000 } }
                    - { $zoom: { min: 16 } }
```

Desired at zoom 4:

[...]

PR #1775, opened 2019-01-07
Further water layer name dropping

zerebubuth:
Drop names on lakes at zooms 0-3. Use manual thresholds for area cut-offs for water feature labels at zooms 4+.

Also: add support to `scripts/mktest.py` to make (fake) polygon geometry for Natural Earth tests.

Connects to #1730.


queries.yaml
```diff
@@ -650,19 +650,42 @@ post_process:
...
+
   # before extracting the label placement points, remove small water areas which
   # are probably too small to label.
   - fn: vectordatasource.transform.drop_properties
     params:
       source_layer: water
-      start_zoom: 0
+      start_zoom: 4
       end_zoom: 16
       properties: [name]
```
nvkelso: Does this drop all name related properties (like `name:en`, `name:de`, `name:fr`, & etc)?

queries.yaml
```diff
@@ -650,19 +650,42 @@ post_process:
...
+    params:
+      source_layer: water
+      start_zoom: 0
+      end_zoom: 4
+      properties: [name]
+      geom_types: [Polygon, MultiPolygon]
+      where: >-
+        kind == 'lake'
+
   # before extracting the label placement points, remove small water areas which
```
nvkelso: I find this comment confusing, perhaps reword to:

>   # before extracting the label placement points, remove names 
>   # from small water areas which are probably too small to label.

queries.yaml
```diff
@@ -650,19 +650,42 @@ post_process:
       start_zoom: 8
       end_zoom: 15
 
+  # no lake labels at zoom 0-3:
+  # https://github.com/tilezen/vector-datasource/issues/1730
+  - fn: vectordatasource.transform.drop_properties
```
nvkelso: Should this be `drop_properties_with_prefix`?

queries.yaml
nvkelso: Seems like we should use: `drop_properties_with_prefix` instead?

queries.yaml
zerebubuth: Yup, sorry. I've expanded the comment and added context in 349cffa. Please take a look and let me know if that makes it clearer.

queries.yaml
zerebubuth: This only drops the `name`, which is sufficient to prevent a label placement from being generated. We only generate those where there is a `name` property. However, it does leave the `name:*` and other name-like properties on the polygon, which is a waste of space.

I added a function specifically for names (and related variants such as `official_name`) in #1774, and I've re-used that here in 01eeaf1 and added test to make sure we clear out the translations from the original polygon too.

queries.yaml
zerebubuth: Yup, see 01eeaf1 - we've now got a function specifically for dropping name properties.

Issue #1596, opened 2018-09-17
Junk healthcare values


nvkelso:
In the POIs YAML it looks like we have a bad copy-paste job for some values:

* **What did you see?**  `- filter: {healthcare: [bank, blood, chiropractor, donation, hospice, midwife, midwife_occupational, optometrist_paediatrics, physiotherapist_physiotherapy, podiatrist_psychotherapist, rehabilitation, speech, therapist, therapist]}`
* **What did you expect to see?** Values that exist IRL and in https://taginfo.openstreetmap.org/keys/healthcare#values.

zerebubuth:
Ooops! I guess we're not testing enough of those values. Should we treat this as a regression rather than leaving to `1.6`?

nvkelso:
@zerebubuth I think most of those are "new" in v1.5 so no? 

These look like they're covered in other v1.5 filters:

  - filter: {healthcare: [blood_donation, blood_bank]}

The rest are only covered in this filter. In any event, not enough to hold up the release and we can sort it out in v1.6.

PR #1784, opened 2019-01-11
Unscramble healthcare whitelist.

zerebubuth:
Also add tests.

Connects to #1596.


zerebubuth:
Ooops, good point! I thought we already had them, but it seems not. Fixed in 8c525b4.

Issue #1253, opened 2017-05-16
Salt marsh properties missing from wetlands in landuse layer


nvkelso:
* **What did you see?**

Lots of island / park labels near JFK airport in New York City floating on the water.

* **What did you expect to see?**

These islands are parks and [wetland=saltmarsh](http://wiki.openstreetmap.org/wiki/Tag:wetland%3Dsaltmarsh
) which we're missing from our landuse layer.

Looks like this is an issue with Mapzen house styles where we do include `wetland` in the landuse layer, but we could do a better job indicating what type of wetland it is (maybe with `kind_detail`?).

In many cases wetland are over land, but in some cases wetland is over water.

Including the extra wetland properties could help symbolize.

* **What map location are you having problems with?** 

New York City area

- https://www.openstreetmap.org/way/50977027#map=17/40.59625/-73.83117
- https://www.openstreetmap.org/way/50976996#map=16/37.4562/-122.0959

* **Screenshot?**

http://tangrams.github.io/walkabout-style/#13/40.6149/-73.8263

<img width="947" alt="screen shot 2017-05-16 at 09 48 36" src="https://cloud.githubusercontent.com/assets/853051/26118004/2f8a5aa6-3a1d-11e7-809c-5ec54c0c1176.png">

nvkelso:
Looks like 1/2 million features tagged with `wetland=*`:

- http://wiki.openstreetmap.org/wiki/Key:wetland

nvkelso:
Wetland `kind_detail` value to whitelist: 

- bog, swamp, marsh, reedbed, wet_meadow, tidalflat, mangrove, saltmarsh, string_bog, saltern, mud, fen

nvkelso:
QA verified.

PR #1785, opened 2019-01-11
Add wetland POIs with kind_detail.

zerebubuth:
Wetland landuse areas already existed, but for many of the examples in #1253 were being overridden by the `boundary=national_park`. I've swapped that around, so that `natural=wetland` matches first (as long as it doesn't have a `protect_class` or `protect_title` that would make us think it's a national park), and added it as a POI type to match.

Connects to #1253.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1253,6 +1253,7 @@ Common values include `pit_latrine`, ` flush`, ` chemical`, ` pour_flush`, ` buc
 Common values include: `drinkable_powered`, `drinkable_manual`, `drinkable_no_pump`, `drinkable`, `not_drinkable_powered`, `not_drinkable_manual`, `not_drinkable_no_pump`, `not_drinkable`.
 
 #### Wetland `kind_detail` values
+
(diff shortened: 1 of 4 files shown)


docs/layers.md
```diff
@@ -1251,6 +1252,9 @@ Common values include `pit_latrine`, ` flush`, ` chemical`, ` pour_flush`, ` buc
 
 Common values include: `drinkable_powered`, `drinkable_manual`, `drinkable_no_pump`, `drinkable`, `not_drinkable_powered`, `not_drinkable_manual`, `not_drinkable_no_pump`, `not_drinkable`.
 
+#### Wetland `kind_detail` values
+The value of the OpenStreetMap `wetland` tag. If available, value will be one of: `bog`, `fen`, `mangrove`, `marsh`, `mud`, `reedbed`, `saltern`, `saltmarsh`, `string_bog`, `swamp`, `tidalflat`, `wet_meadow`.
```
nvkelso: Nit: markdown now requires an extra line whitespace between headers and body text.

nvkelso: Fixed in https://github.com/tilezen/vector-datasource/pull/1785/commits/83eecabe8e9c12b58bc683e8e9ca75ed781bd29b.

Issue #1781, opened 2019-01-10
Match OSM Carto landcover coverage


nvkelso:
For reference, OSM Carto has low zoom and high zoom landuse definitions:

- https://github.com/gravitystorm/openstreetmap-carto/blob/3cbd1112b9d82f361ca95e660e64b2b63f1cfce2/project.mml#L82-L165

### High zoom (10+):

- aeroway IN
    - [x] apron
    - [x] aerodrome
- amenity IN
    - [ ] bicycle_parking
    - [ ] motorcycle_parking
    - [x] university
    - [x] college
    - [x] school
    - [ ] taxi
    - [x] hospital
    - [ ] kindergarten
    - [x] grave_yard
    - [x] prison
    - [x] place_of_worship
    - [ ] clinic
    - [x] ferry_terminal
    - [ ] marketplace
    - [ ] community_centre
    - [ ] social_facility
    - [ ] arts_centre
    - [ ] parking_space
    - [ ] bus_station
    - [ ] fire_station
    - [ ] police
    - [x] parking
- healthcare IN
    - [ ] clinic
    - [x] hospital
- landuse IN
    - [x] quarry
    - [ ] vineyard
    - [x] orchard
    - [x] cemetery
    - [x] residential
    - [ ] garages
    - [x] meadow
    - [x] grass
    - [x] allotments
    - [x] forest
    - [ ] farmyard
    - [x] farmland
    - [ ] greenhouse_horticulture
    - [x] recreation_ground
    - [x] village_green
    - [x] retail
    - [x] industrial
    - [x] railway
    - [x] commercial
    - [ ] brownfield
    - [ ] landfill
    - [ ] construction
    - [x] plant_nursery
    - [ ] religious
- leisure IN
    - [x] swimming_pool (in **water** layer)
    - [x] playground
    - [x] park
    - [ ] recreation_ground
    - [x] common
    - [x] garden
    - [x] golf_course
    - [ ] miniature_golf
    - [x] sports_centre
    - [x] stadium
    - [x] pitch
    - [ ] ice_rink
    - [x] track
    - [x] dog_park
    - [ ] fitness_station
- man_made IN
    - [x] works
    - [x] wastewater_plant
    - [x] water_works
- natural" IN
    - [x] beach
    - [ ] shoal
    - [ ] heath
    - [ ] grassland
    - [x] wood
    - [ ] sand
    - [x] scree
    - [ ] shingle
    - [ ] bare_rock
    - [x] scrub
[...]

Issue #1779, opened 2019-01-08
Remap landuse kinds at mid and low zooms to improve merging


nvkelso:
Essentially https://github.com/tilezen/vector-datasource/pull/1723 (reduce surface tag precision in **roads** layer, but for **landuse** layer.

We're adding even more landcover polygons via https://github.com/tilezen/vector-datasource/issues/1259. We should endever to not increase the file size a huge amount. 

I propose this simplification start be zooms 0 to  12 inclusive (so zoom 13 would have the "raw" values). This also corresponds roughly with where raster landcover sources max out on their spatial fidelity. 

Related:

- #1721 to reduce landuse size at mid-zooms
- #1033 to drop IDs at low zooms in landuse layer

nvkelso:
Expanding on the list in https://github.com/tilezen/vector-datasource/issues/1259:

- **urban_area** = 20 (`developed` in Marble Cutter, propose reusing `urban_area` which is a Natural Earth-ism we already have at low zooms. Alternatively: `residential`?) 
    landuse=[residential,farmyard,industrial,cemetery,commercial,retail,construction,brownfield,recreation_ground,garages,religious,highway,static_caravan,school,churchyard,greenery]
- **barren** = 30 (use the Marble Cutter-ism, no good OSM-ism?
    natural=[cliff,scree,beach,shingle,fell]
    landuse=quarry
- **forest** = 40 (Marble Cutter-ism is equivelent to landuse=forest in OSM, good)
    natural=wood
    landuse=[forest,logging,conservation,plantation]
- **scrub** = 50 
    natural=scrub
- **grassland** = 70 (use OSM's `grassland` instead of Marble Cutter's `herbaceous`)
    natural=[grassland,heath]
    landuse=[grass,meadow,village_green]
    landcover=grass
- **farmland**  = 80  (use OSM's `farmland` instead of Marble Cutter's `cultivated`)
    landuse=[farmland,orchard,vineyard,greenhouse_horticulture,farm,aquaculture,plant_nursery,salt_pond,field,allotments,animal_keeping,flowerbed,livestock]
- **wetland** = 90 (use OSM's `wetland` instead of Marble Cutter's `wetlands`, note plural)
    natural=[wetland,mud]
- **glacier** = 100 (both are same)
    natural=glacier
- **desert** = 110  (¿both are same?)
    natural=[bare_rock,sand,desert]

nvkelso:
Working

PR #1790, opened 2019-01-14
Remap landuse kinds at mid and low zooms.

zerebubuth:
Remap `kind` of various `landuse` layer features to give better opportunity for mid and low zoom merging.

Many of the kinds specified in #1779 depend on #1781, and will have to be implemented as part of #1781 when it gets done later.

Additionally, `natural=cliff` is in the `earth` layer, not the `landuse` layer. Because it wouldn't be merged anyway, I left it out. I think if we want to move that to the `landuse` layer (which we might if we want to merge with `landuse=quarry`), we probably have to add that to the v2 wishlist?

Connects to #1779.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -335,6 +335,60 @@ post_process:
       start_zoom: 4
       properties: [barrier]
 
+  # coarsen the landuse kinds at mid & low zooms to help improve merging.
+  #
+  # do this _early_; before we assign sort_rank and to avoid inconsistency when
+  # we join (e.g: roads) onto landuse and wouldn't want the road landuse_kind to
+  # be different from the polygon in case that makes it the wrong colour.
+  #
+  # **NOTE** WE'RE CHANGING THE KINDS HERE!
+  - fn: vectordatasource.transform.remap
+    params:
+      layer: landuse
+      start_zoom: 0
+      end_zoom: 13
+      property: kind
+      # note that we don't need to remap things to themselves, e.g:
+      # kind=desert stays as-is.
+      remap:
+        # barren
+        quarry: barren
+        scree: barren
+        shingle: barren
+
+        # desert
+        bare_rock: desert
+        sand: desert
+
+        # farmland
+        allotments: farmland
+        farm: farmland
+        orchard: farmland
+        plant_nursery: farmland
+        vineyard: farmland
+
+        # forest
+        natural_wood: forest
+        wood: forest
+
+        # grassland
+        grass: grassland
+        heath: grassland
+        meadow: grassland
+        village_green: grassland
+
+        # urban_area
+        cemetery: urban_area
+        commercial: urban_area
+        industrial: urban_area
+        recreation_ground: urban_area
+        residential: urban_area
+        retail: urban_area
+        school: urban_area
+
+        # wetland
+        mud: wetland
+
   # sort key
   - fn: vectordatasource.transform.csv_match_properties
     resources:
--- a/queries.yaml
+++ b/queries.yaml
@@ -361,7 +361,6 @@ post_process:
         sand: desert
 
         # farmland
-        allotments: farmland
         farm: farmland
         orchard: farmland
         plant_nursery: farmland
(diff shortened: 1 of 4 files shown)


queries.yaml
```diff
@@ -335,6 +335,60 @@ post_process:
...
+        quarry: barren
+        scree: barren
+        shingle: barren
+
+        # desert
+        bare_rock: desert
+        sand: desert
+
+        # farmland
+        allotments: farmland
```
nvkelso: While I think technically correct as-is, alternatively `village_green` and `allotments` could both be remapped to `urban_area` if that promotes more merging? Would require data stats...

queries.yaml
```diff
@@ -335,6 +335,60 @@ post_process:
...
+  #
+  # do this _early_; before we assign sort_rank and to avoid inconsistency when
+  # we join (e.g: roads) onto landuse and wouldn't want the road landuse_kind to
+  # be different from the polygon in case that makes it the wrong colour.
+  #
+  # **NOTE** WE'RE CHANGING THE KINDS HERE!
+  - fn: vectordatasource.transform.remap
+    params:
+      layer: landuse
+      start_zoom: 0
```
nvkelso: Generally more consistent to start at zoom 0+, but we've modified some other configs to be 9+ only. Should we do the same here?

(We show `national_park` earlier (not covered here), but I think most everything else – especially this type of landcover remapping – starts at zoom 9.)

zerebubuth: The zoom ranges in the YAML files allow `residential` down to zoom 4 and `natural_wood` / `wood` down to 6, and a bunch of others down to 8. Whether any are actually that large is questionable, and I'm running a query to find out - although it may take some time to run.

We start bringing in OSM landuse starting at z4+, so we could change it to that, but at the moment any higher `start_zoom` would potentially mean stuff gets "un-generalised" when zooming out. Also, we've had a problem in the past with lack of coordination between `start_zoom` when we've changed the zoom that we bring stuff in at, or start simplifying, so I'm in two minds about whether to be "minimal" and cover only the zoom ranges at which we have data, or "defensive against bugs" by covering the whole zoom range down to zero.

queries.yaml
zerebubuth: Sure, we can map it however you want! The [original request](https://github.com/tilezen/vector-datasource/issues/1779#issuecomment-452452746) was for:

> - **grassland** = 70 (use OSM's `grassland` instead of Marble Cutter's `herbaceous`)
>    - ...
>    - other landuse=[grass,village_green]
> - **farmland**  = 80  (use OSM's `farmland` instead of Marble Cutter's `cultivated`)
>    - ...
>    - other landuse=[...,allotments,...]

Allotments are usually close to (sub)urban areas, so I can see that they would probably merge with nearby `residential` areas. `village_green`, I'm not so sure about - it'll certainly be near houses, but whether there's enough houses for someone to have bothered drawing a `residential` area... I'm not sure.

nvkelso: Changed in https://github.com/tilezen/vector-datasource/pull/1790/commits/6c711e2a9805361692c50098ceaf25a6175af12d.

(This one is more to taste, thanks for the change!)

queries.yaml
nvkelso: I'm fine leaving it 0.

nvkelso:
> Additionally, natural=cliff is in the earth layer, not the landuse layer. Because it wouldn't be merged anyway, I left it out. I think if we want to move that to the landuse layer (which we might if we want to merge with landuse=quarry), we probably have to add that to the v2 wishlist?

Let's just leave `cliff` where it is in the earth layer. I think we only bring in lines there, so it doesn't actually matter?

nvkelso:
And so this Github PR is easier to read...

@zerebubuth responded to https://github.com/tilezen/vector-datasource/pull/1790#discussion_r247681531 in https://github.com/tilezen/vector-datasource/pull/1790#discussion_r247914382:

> The zoom ranges in the YAML files allow residential down to zoom 4 and natural_wood / wood down to 6, and a bunch of others down to 8. Whether any are actually that large is questionable, and I'm running a query to find out - although it may take some time to run.

zerebubuth:
Additional follow-up in #1795.

Issue #1721, opened 2018-12-07
☂ Reduce landuse layer size at mid-zooms


nvkelso:
Similar to https://github.com/tilezen/vector-datasource/issues/1716, but for the **landuse** layer's landcover related tags.

* **What did you see?** At mid-zooms where we source **landuse** layer contents from OSM from ~ zoom 8 to zoom 13 we see relatively large file size % from the landuse layer compared to other layers.
* **What did you expect to see?** Less file size, by dropping properties and reducing kind precision for "green" and other `natural=*` "landcover" related landuse kinds. We've also talked about "buffer, merge, shrink" like we now do with the **buildings** layer.

Possibly also related:

- Adding grass green areas: https://github.com/tilezen/vector-datasource/issues/1259
- Reducing kind precision: https://github.com/tilezen/vector-datasource/issues/1779
- Demoting some landuse down: https://github.com/tilezen/vector-datasource/issues/1794
- Unifying wood versus forest: https://github.com/tilezen/vector-datasource/issues/1397
- TagInfo: https://taginfo.openstreetmap.org/keys/natural?filter=ways#values

nvkelso:
Followup for farmland (and possibly orchard) at zoom 9 here: https://github.com/tilezen/vector-datasource/issues/1813.

nvkelso:
Farmland is working now, closing :)

Issue #1698, opened 2018-10-30
Too many service areas on the map


nvkelso:
* **What did you see?** Lots of junk service areas in NYC and DC
* **What did you expect to see?** Only service areas with large areas should be visible earlier. The ones without area (or very small area) are likely tagging errors in OSM. For no area (or tiny area) make the `min_zoom` 16 instead? There might be a related kind here for rest stop so investigate that, too.
* **What map location are you having problems with?** 

Bad:
![image](https://user-images.githubusercontent.com/853051/47757350-31317d00-dc63-11e8-86b2-5bac0aa02157.png)

Good:
![image](https://user-images.githubusercontent.com/853051/47757353-3abae500-dc63-11e8-9f2a-082d4fb4d056.png)

http://tangrams.github.io/bubble-wrap/#12/40.7175/-73.9495

![image](https://user-images.githubusercontent.com/853051/47757301-f3ccef80-dc62-11e8-80f1-bbe7b0109265.png)

nvkelso:
Doesn't look like this is on staging yet.

nvkelso:
Verified fix on staging:

![image](https://user-images.githubusercontent.com/853051/50115479-60dd1a00-01fc-11e9-852c-e2a04c65af89.png)

And reviewed other known good rest areas / service areas and they still show up.

PR #1791, opened 2019-01-15
Merge polygons in landuse layer.

zerebubuth:
This merges polygons in the landuse layer to improve tile size. Two different types of merging are used; at higher zooms we try to merge into as large a multipolygon as possible and at mid and low zooms we also do a dilate/erode step to try and blend together adjacent polygons. This can be extremely effective when merging large polygons separated by small spaces, such as paths within a forest.

For example, tile `9/263/166` near [Groningen, the Netherlands](https://tile.nextzen.org/preview.html#lat=53.2143&lng=5.9917&z=10) originally had a 663kB `landuse` layer, but this reduced to 84kB, a drop of 87%. The total tile size went from 859kB to 280kB, a drop of 67%. Of the remaining size,  155kB is roads.

The vast majority of the features in this tile came from `kind: grass` polygons:

original grass polygons

It looks as if each individual field has been mapped as a separate feature (or rather, [imported](https://wiki.openstreetmap.org/wiki/3dShapes)). There are many small gaps between adjacent polygons, often for [small water courses or paths](https://www.openstreetmap.org/way/64306504). This means it's hard enough to merge the geometries - and on top of that, the different sizes of adjacent features means that wildly different `min_zoom` values get set.

Re-using the buffering merge implemented earlier for buildings in #1698 and allowing different `min_zoom`s to be merged together (it's generally purely area-based, unlike buildings), gives:

merged grass polygons

Which still shows a lot of detail in the boundary, but has almost entirely removed small gaps and voids.

The irony is that the `grass` landuse kind doesn't even seem to be rendered by Bubble Wrap at zoom 10, so here's a screenshot of a slightly modified style:

bubble wrap with extra grass

Connects to #1721.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -1387,7 +1387,7 @@ post_process:
   - fn: vectordatasource.transform.drop_small_inners
     params:
       end_zoom: 14
-      source_layers: [landuse, water]
+      source_layers: [water]
       pixel_area: 0.1
 
   # drop this layer before simplify_and_clip - we don't want it in the output,
@@ -1421,12 +1421,64 @@ post_process:
       where: >-
         landuse_kind in ('residential', 'industrial')
 
-  # merge any remaining polygons by their properties.
+  # drop osm_relation tag, used to indicate that shapes came from OSM relations.
+  # however, it interferes with merging.
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: landuse
+      start_zoom: 4
+      end_zoom: 15
+      properties:
+        - osm_relation
+
+  # drop names on landuse areas at mid and low zooms. it interferes with
+  # merging, and anything that we would have labelled would have become a
+  # label by now anyway (see handle_label_placement around line 716-ish).
+  - fn: vectordatasource.transform.drop_names
+    params:
+      source_layer: landuse
+      start_zoom: 4
+      end_zoom: 13
+
+  # drop small inners before merge, as not having to consider these can speed up
+  # the buffer & clipping algorithms.
+  - fn: vectordatasource.transform.drop_small_inners
+    params:
+      end_zoom: 15
+      source_layers: [landuse]
+      pixel_area: 0.8
+
+  # merge at mid and low zooms, buffering to try and eliminate small gaps
+  # between adjacent polygons.
   - fn: vectordatasource.transform.merge_polygon_features
     params:
       source_layer: landuse
       start_zoom: 4
       end_zoom: 13
+      merge_min_zooms: true
+      buffer_merge: true
+      # this formula seems to give a good balance between larger values, which
+      # merge more but can merge everything into a blob if too large, and small
+      # values which retain detail.
+      buffer_merge_tolerance: >-
+        min(150, 0.6 * tolerance_for_zoom)
+
+  # merge at high-ish zooms without buffering - but we still want to merge
+  # multiple small polygons together to get file size savings.
+  - fn: vectordatasource.transform.merge_polygon_features
+    params:
+      source_layer: landuse
+      start_zoom: 13
+      end_zoom: 15
+      merge_min_zooms: true
+
+  # drop small inners after merge, as small inners could have been introduced
+  # by the merge algorithm.
+  - fn: vectordatasource.transform.drop_small_inners
+    params:
+      end_zoom: 15
+      source_layers: [landuse]
+      pixel_area: 0.8
 
   # first, merge linestrings together where the properties are the same, and
   # make sure we merge across junctions. this will create a set of non-simple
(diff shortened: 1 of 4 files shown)


queries.yaml
```diff
@@ -1421,12 +1421,64 @@ post_process:
...
+    params:
+      source_layer: landuse
+      start_zoom: 4
+      end_zoom: 13
+
+  # drop small inners before merge, as not having to consider these can speed up
+  # the buffer & clipping algorithms.
+  - fn: vectordatasource.transform.drop_small_inners
+    params:
+      end_zoom: 15
```
nvkelso: Just curious, why is this 1 more zoom at 15 (versus 14) than in v1.6?

vectordatasource/transform.py
```diff
@@ -3618,9 +3628,19 @@ def merge_polygon_features(ctx):
...
+            'tolerance_for_zoom': tfz,
+        })
+    else:
+        tolerance = tfz
+
     def _props_pre((shape, props, fid)):
         # drop area while merging, as we'll recalculate after.
         props.pop('area', None)
+        if merge_min_zooms:
+            props.pop('min_zoom', None)
```
nvkelso: What is the `min_zoom` of the resulting polygon, I'm reading this as it's stripped off? 

Could it be the min of the input `min_zoom` values instead?

nvkelso: (Perhaps this is answered in the new `_props_merge` function below... but can you walk thru an example, please?

nvkelso:
I made a GIF from your two frames:

landuse_merge

vectordatasource/transform.py
zerebubuth: Yup, that's exactly what's happening. When `_merge_features_by_property` runs, it calls our `_props_pre` function for each feature, and whatever we return is the "key" used to group features to be merged. We `pop` the `min_zoom` off the properties so that the merge algorithm will merge adjacent polygons with different `min_zoom` - which happens all the time in the patchwork of highly variable-sized fields in the 3dshapes import.

However, `_merge_features_by_property` (now) also keeps the original properties, and passes the list of those into our [`_props_merge` function](https://github.com/tilezen/vector-datasource/pull/1791/files/268f1e8d661ffa7d13b7245986c8f4804db4ab7d#diff-ccacdffc74e3f0be743bf4d7d6a1bb9cR3652) below, and the merged feature gets assigned whatever we return. The `_props_merge` implementation I've used here takes the minimum of all the `min_zoom`s, as this will guarantee that we see the merged feature at least at all the zooms of the biggest feature we queried from the database or RAWR tile.

The step-by-step run-through might be something like:

1. We call `merge_polygon_features` on a layer with two features, `[(shape1, props1, fid1), (shape2, props2, fid2)]`.
2. `merge_polygon_features` calls `_merge_features_by_property`.
3. `_merge_features_by_property` calls back into `merge_polygon_features`' `_props_pre` for each feature, where we strip the min zoom. In this example, `props1.pop('min_zoom') == props2.pop('min_zoom') == props`.
4. `_merge_features_by_property` stores the features indexed by that result, so we get something like: `{props: [(shape1, props1), (shape2, props2)]}`
5. `_merge_features_by_property` tries to merge `shape1` and `shape2`, getting `shape_merged`
6. `_merge_features_by_property` calls back into `merge_polygon_features`' `_props_merge` with `[props1, props2]`, and we take the min of all the `min_zoom` parameters, returning `props_merged`.
7. `_merge_features_by_property` calls back into `merge_polygon_features`' `_props_post`, where it resets the `area`  on `props_merged` (and is why we don't bother with that in step 6).
8. `_merge_features_by_property` returns a new list of features: `[(shape_merged, props_merged, some_fid)]`, where `some_fid` is one of the `fids` of one of the merged features.

queries.yaml
zerebubuth: I found that we still had some tiles at zoom 13 & 14 which had quite a lot of `grass` landuse. In the particular case I was looking at, they were all suburban lawns and, while the polygons couldn't be merged very well, they benefited from being merged into the same MultiPolygon.

So I added the [second `merge_polygon_features`](https://github.com/tilezen/vector-datasource/pull/1791/files/268f1e8d661ffa7d13b7245986c8f4804db4ab7d#diff-fc49e5c256610bc0b32a7de5a0520835R1468), which does a simpler merge step at zooms 13 & 14. I extended the range of the `drop_small_inners` to cover those zooms too, just to make sure we would be dropping any courtyards or cut-outs in those lawn features, or fields elsewhere.

For the lawns, I didn't get as substantial savings (only about 30% smaller for the layer - 15% smaller for the whole tile), and I figured we could just look at it when we have a global build and figure out whether this merging is doing what we want.

PR #1792, opened 2019-01-16
All the kind details

zerebubuth:
Extends `scripts/all_the_kinds.py` with a `--kind-details` option to enumerate all the `kind_detail` values as well as `kind`. This can now print out a table of all the `kind`, `kind_detail`, `min_zoom` and `sort_rank` - an example can be seen in [`all_kind_detail.md`](https://gist.github.com/zerebubuth/b39a460ce260c90eff3f587bfb290f2b).

In order to be able to enumerate all the values, it was necessary to whitelist all the `kind_detail` values. These either came from the existing documentation or Taginfo. I've added a test for this, to make sure that any `kind` or `kind_detail` we add in the future is also fully enumerable.


vectordatasource/meta/kinds.py
```diff
@@ -0,0 +1,473 @@
...
+            if result is None:
+                sort_rank_val = None
+            else:
+                sort_rank_key, sort_rank_val = result
+                assert sort_rank_key == 'sort_rank'
+                sort_rank_val = int(sort_rank_val)
+
+            info = KindInfo(parse_start_zoom(item['min_zoom']), sort_rank_val)
+
+            values[key] = info
```
rmarianski: Should some of this be predicated on the `include_kind_detail` parameter? If not, to me it looks like that isn't used and maybe the function shouldn't take it?

zerebubuth: Yes, well spotted, thanks!

That's definitely a bug. I think it should have aggregated by `kind_detail`, but obviously I forgot to do that... oops. Hopefully it's fixed in 6ff5b58 - would appreciate it if you could take a look and let me know what you think.

rmarianski: 👍

PR #1793, opened 2019-01-16
Add collision rank

zerebubuth:
Adds a method for assigning collision ranks to features.

### Method 

The idea of a collision rank is that it's a unique integer assigned to everything that gets labelled. See #988 for more background. This adds a new resource, `spreadsheets/collision_rank.yaml`, which is a YAML-encoded list of filters. The syntax of the filters is the same as the `kind` & `min_zoom` YAML in the `yaml/` directory (hopefully this minimises the amount of duplication, and means we can share bug fixes and/or new features).

Each feature is checked against each filter in turn, and is assigned the `collision_rank` corresponding to the matching index. Indices start at 1, are incremented by one for each filter, and can have "reserved blocks" which skip the index counter onwards. For example, if `collision_rank.yaml` contains:

```yaml
- kind: foo
- kind: bar
- _reserved: {from: 10, to: 20}
- kind: baz
```

Then a feature with:

*  `kind: foo` would get `collision_rank: 1`,
*  `kind: bar` would get `collision_rank: 2`,
*  `kind: baz` would get `collision_rank: 21`.

Note that indices 3-9 are unused in this example, but if we added an entry after `bar`, then it would be assigned 3. The code asserts that indices are unique, so we can't accidentally overflow the reserved blocks. Both the `from` and `to` values are inclusive, hence `baz` is assigned the next index, 21.

### Why not use a CSV

We've been managing `sort_rank` with CSVs, but the issue is that we often don't care about the exact value of the `sort_rank`, but want to easily insert values and re-order the following lines. Having the `sort_rank` explicitly in the file means that any such re-arrangement has a very large diff, and is hard to check visually. This is an experiment to see if we can make something that still gives the features we want, but without the drawbacks of the CSV.

Additionally, the CSVs tended to get mangled by Excel when loading and/or saving, as we were using a bunch of special characters to indicate things such as "any" `*` or "must be null" `-` .

Finally, CSV files can't contain comments, but YAML ones can. This will make future maintenance easier, as we can comment `# this reserved block for X, Y or Z`.

Connects to #988.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -1595,6 +1595,15 @@ post_process:
 
   # collision rank
   - fn: vectordatasource.transform.add_collision_rank
+    params:
+      where: >-
+        layer_name == 'pois' or
+        _has_name or
+        ref is not None or
+        shield_text is not None or
+        bicycle_shield_text is not None or
+        bus_shield_text is not None or
+        walking_shield_text is not None
     resources:
       ranker:
         type: file
--- a/queries.yaml
+++ b/queries.yaml
@@ -1592,3 +1592,11 @@ post_process:
   - fn: vectordatasource.transform.max_zoom_filter
     params:
       layers: [places]
+
+  # collision rank
+  - fn: vectordatasource.transform.add_collision_rank
+    resources:
+      ranker:
+        type: file
+        init_fn: vectordatasource.transform.load_collision_ranker
+        path: spreadsheets/collision_rank.yaml
--- a/spreadsheets/collision_rank.yaml
+++ b/spreadsheets/collision_rank.yaml
@@ -1,6 +1,943 @@
-- layer: earth
-  kind: continent
-- layer: water
-  kind: ocean
-- layer: places
-  kind: country
+# reserved for UX/UI
+- _reserved: {from: 1, to: 100}
+# reserved for house theme overlays
+- _reserved: {from: 101, to: 201}
+# reserved for data overlays
+- _reserved: {from: 202, to: 302}
+- {layer: 'earth', kind: 'continent'}
+- {layer: 'water', kind: 'ocean'}
+- {layer: 'places', kind: 'country'}
+# reserved for population_rank
+- _reserved: {from: 306, to: 325}
+- {layer: 'places', kind: 'locality'}
+# reserved for population_rank; guess the pop from kind_detail ala
+# _default_min_zoom_for_place_kind() : city, farm, hamlet, isolated_dwelling,
+# locality, scientific_station, town, village,
+- _reserved: {from: 327, to: 346}
+- {layer: 'pois', kind: 'national_park'}
+- {layer: 'places', kind: 'region'}
+# reserved for population_rank
+- _reserved: {from: 349, to: 368}
+- {layer: 'water', kind: 'sea'}
+- {layer: 'pois', kind: 'nature_reserve'}
+- {layer: 'water', kind: 'lake'}
+# for shields & etc
+- _reserved: {from: 372, to: 382}
+- {layer: 'roads', kind: 'highway', kind_detail: 'motorway'}
+# for shields & etc
+- _reserved: {from: 384, to: 394}
+# for shields & etc
+- _reserved: {from: 395, to: 405}
+- {layer: 'roads', kind: 'highway', kind_detail: 'motorway_link'}
+# for shields & etc
+- _reserved: {from: 407, to: 417}
+# for shields & etc
+- _reserved: {from: 418, to: 428}
+- {layer: 'roads', kind: 'highway', kind_detail: 'trunk'}
+# for shields & etc
+- _reserved: {from: 430, to: 440}
+# for shields & etc
+- _reserved: {from: 441, to: 451}
+- {layer: 'roads', kind: 'highway'}
+# for shields & etc
+- _reserved: {from: 453, to: 463}
+- {layer: 'pois', kind: 'forest'}
+- {layer: 'earth', kind: 'island'}
+- {layer: 'water', kind: 'water'}
+- {layer: 'pois', kind: 'protected_area'}
+- {layer: 'water', kind: 'bay'}
+- {layer: 'water', kind: 'fjord'}
+- {layer: 'water', kind: 'strait'}
+- {layer: 'water', kind: 'playa'}
+- {layer: 'pois', kind: 'aerodrome'}
+- {layer: 'pois', kind: 'military'}
+- {layer: 'pois', kind: 'naval_base'}
+- {layer: 'pois', kind: 'airport'}
+- {layer: 'pois', kind: 'airfield'}
+- {layer: 'pois', kind: 'park'}
+- {layer: 'pois', kind: 'volcano'}
+- {layer: 'pois', kind: 'wood'}
+# reserve for kind_details
+- _reserved: {from: 480, to: 490}
+- {layer: 'pois', kind: 'peak'}
+- {layer: 'water', kind: 'river'}
+# for shields & etc
+- _reserved: {from: 493, to: 503}
+- {layer: 'transit', kind: 'train'}
+# for shields & etc
+- _reserved: {from: 505, to: 515}
+- {layer: 'pois', kind: 'beach'}
+# reserve for kind_details
+- _reserved: {from: 517, to: 527}
+- {layer: 'pois', kind: 'university'}
+- {layer: 'pois', kind: 'battlefield'}
+# for shields & etc
+- _reserved: {from: 530, to: 540}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'primary'}
+# for shields & etc
+- _reserved: {from: 542, to: 552}
+# for shields & etc
+- _reserved: {from: 553, to: 563}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'primary_link'}
+# for shields & etc
+- _reserved: {from: 565, to: 575}
+# for shields & etc
+- _reserved: {from: 576, to: 586}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'secondary'}
+# for shields & etc
+- _reserved: {from: 588, to: 598}
+# for shields & etc
+- _reserved: {from: 599, to: 609}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'secondary_link'}
+# for shields & etc
+- _reserved: {from: 611, to: 621}
+# for shields & etc
+- _reserved: {from: 622, to: 632}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'tertiary'}
+# for shields & etc
+- _reserved: {from: 634, to: 644}
+# for shields & etc
+- _reserved: {from: 645, to: 655}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'tertiary_link'}
+# for shields & etc
+- _reserved: {from: 657, to: 667}
+# for shields & etc
+- _reserved: {from: 668, to: 678}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'trunk'}
+# for shields & etc
+- _reserved: {from: 680, to: 690}
+# for shields & etc
+- _reserved: {from: 691, to: 701}
+- {layer: 'roads', kind: 'major_road', kind_detail: 'trunk_link'}
+# for shields & etc
+- _reserved: {from: 703, to: 713}
+# for shields & etc
+- _reserved: {from: 714, to: 724}
+- {layer: 'roads', kind: 'major_road'}
+# for shields & etc
+- _reserved: {from: 726, to: 736}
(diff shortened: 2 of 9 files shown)


nvkelso:
👀

rmarianski:
LGTM so far!

nvkelso:
@zerebubuth Please update the example config file based on the table in https://github.com/tilezen/vector-datasource/issues/988#issuecomment-455013270.

zerebubuth:
Updated the config in 961030e and added tests in eec237e. I think this is ready for a proper review!

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
+# reserved for UX/UI
+- _reserved: {from: 1, to: 100}
+# reserved for house theme overlays
+- _reserved: {from: 101, to: 201}
+# reserved for data overlays
+- _reserved: {from: 202, to: 302}
+- {layer: 'earth', kind: 'continent'}
```
nvkelso: `$layer` instead, since `layer` is a valid (OSM) property?

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'water', kind: 'ocean'}
+- {layer: 'places', kind: 'country'}
+# reserved for population_rank
+- _reserved: {from: 306, to: 325}
```
nvkelso: I think all the reserved lines from here down should be reserved counts instead of numerical ranges. Here this would be `reserved: { count: 10 }`

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
+# reserved for UX/UI
+- _reserved: {from: 1, to: 100}
```
nvkelso: Can just be 1 to 99

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+# reserved for house theme overlays
+- _reserved: {from: 101, to: 201}
```
nvkelso: Can be 100 to 199

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+# reserved for data overlays
+- _reserved: {from: 202, to: 302}
```
nvkelso: Can be 200 to 299.

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'earth', kind: 'continent'}
+- {layer: 'water', kind: 'ocean'}
```
nvkelso: Do we need to quote the `layer` value, `kind` value & etc?

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+# reserve for kind_details
+- _reserved: {from: 4438, to: 4448}
+- {layer: 'pois', kind: 'fire_hydrant'}
+- {layer: 'pois', kind: 'gas_canister'}
+- {layer: 'pois', kind: 'level_crossing'}
+- {layer: 'pois', kind: 'love_hotel'}
+- {layer: 'pois', kind: 'post_box'}
+- {layer: 'pois', kind: 'power_pole'}
+- {layer: 'pois', kind: 'street_lamp'}
+- {layer: 'pois', kind: 'traffic_signals'}
```
nvkelso: I got `traffic_signals` one wrong... let's move it just above the `atm` row, please.

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'pois', kind: 'bookmaker'}
+- {layer: 'pois', kind: 'bureau_de_change'}
+- {layer: 'pois', kind: 'car_parts'}
+- {layer: 'pois', kind: 'donation'}
+- {layer: 'pois', kind: 'hanami'}
+- {layer: 'pois', kind: 'healthcare_alternative'}
+# reserve for kind_details
+- _reserved: {from: 4083, to: 4103}
+- {layer: 'pois', kind: 'it'}
+- {layer: 'pois', kind: 'midwife_occupational'}
```
nvkelso: A few of these medical related kinds still need to be cleaned up.

nvkelso: See https://github.com/tilezen/vector-datasource/pull/1784/files.

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'pois', kind: 'wilderness_hut'}
+- {layer: 'pois', kind: 'nursing_home'}
+# reserve for kind_details
+- _reserved: {from: 2392, to: 2412}
+- {layer: 'pois', kind: 'telescope'}
+- {layer: 'landuse', kind: 'city_wall'}
+- {layer: 'water', kind: 'riverbank'}
+- {layer: 'earth', kind: 'cliff'}
+- {layer: 'earth', kind: 'arete'}
+- {layer: 'boundaries', kind: 'maritime'}
```
nvkelso: This was in the CSV list output, but does it get transformed into `maritime_boundary` = True somewhere else in the code?

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'pois', kind: 'motorcycle'}
+- {layer: 'pois', kind: 'newsagent'}
+- {layer: 'pois', kind: 'newspaper'}
+- {layer: 'pois', kind: 'ngo'}
+- {layer: 'pois', kind: 'notary'}
+- {layer: 'pois', kind: 'optician'}
+- {layer: 'pois', kind: 'paediatrics'}
+# reserve for kind_details
+- _reserved: {from: 3610, to: 3630}
+- {layer: 'pois', kind: 'optometrist_paediatrics'}
```
nvkelso: Separate these into 2 rows: 

- optometrist
- paediatrics

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+# reserve for kind_details
+- _reserved: {from: 3632, to: 3652}
+- {layer: 'pois', kind: 'perfumery'}
+- {layer: 'pois', kind: 'petting_zoo'}
+- {layer: 'pois', kind: 'physician'}
+# reserve for kind_details
+- _reserved: {from: 3656, to: 3676}
+- {layer: 'pois', kind: 'physiotherapist_physiotherapy'}
```
nvkelso: Rename to just `physiotherapy` (as `physiotherapist` is covered next few lines down).

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- _reserved: {from: 3365, to: 3385}
+- {layer: 'pois', kind: 'architect'}
+- {layer: 'pois', kind: 'art'}
+- {layer: 'pois', kind: 'association'}
+- {layer: 'pois', kind: 'aviary'}
+- {layer: 'pois', kind: 'baby_hatch'}
+- {layer: 'pois', kind: 'bakery'}
+- {layer: 'pois', kind: 'beauty'}
+- {layer: 'pois', kind: 'bed_and_breakfast'}
+- {layer: 'pois', kind: 'blood'}
```
nvkelso: Remove `blood`, it was a jumbled version of `blood_bank`, which we have immediately below.

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'buildings', kind: 'building'}
+# reserve for kind_details
+- _reserved: {from: 3873, to: 4073}
+- {layer: 'pois', kind: 'atv'}
+- {layer: 'pois', kind: 'bicycle_rental_station'}
+- {layer: 'pois', kind: 'boat_storage'}
+- {layer: 'pois', kind: 'bookmaker'}
+- {layer: 'pois', kind: 'bureau_de_change'}
+- {layer: 'pois', kind: 'car_parts'}
+- {layer: 'pois', kind: 'donation'}
```
nvkelso: Remove `donation`, it was jumbled up in the health care funk.

spreadsheets/collision_rank.yaml
```diff
@@ -0,0 +1,943 @@
...
+- {layer: 'pois', kind: 'sanitary_dump_station'}
+- {layer: 'pois', kind: 'scuba_diving'}
+- {layer: 'pois', kind: 'snowmobile'}
+- {layer: 'pois', kind: 'social_facility'}
+# reserve for kind_details
+- _reserved: {from: 3795, to: 3815}
+- {layer: 'pois', kind: 'speech_therapist'}
+# reserve for kind_details
+- _reserved: {from: 3817, to: 3837}
+- {layer: 'pois', kind: 'speech'}
```
nvkelso: Remove, from jumbled health care PR.

spreadsheets/collision_rank.yaml
nvkelso: In this case, both of the real values are covered elsewhere... so this line should be removed.

spreadsheets/collision_rank.yaml
nvkelso: (This and the next few suggestions will make documentation cleaner.)

queries.yaml
```diff
@@ -1592,3 +1592,11 @@ post_process:
   - fn: vectordatasource.transform.max_zoom_filter
     params:
       layers: [places]
+
+  # collision rank
+  - fn: vectordatasource.transform.add_collision_rank
```
nvkelso: This will always result in collision_rank being added.

We probably only need to include collision rank when:

- Always for the POIs layer
- Only in the other layers when feature has one of `name`, `shield_text`, or `ref` (since we haven't converted some refs to shield_text)?

Perhaps this is an over optimization...

spreadsheets/collision_rank.yaml
zerebubuth: Good point. Fixed in 87241c9.

spreadsheets/collision_rank.yaml
zerebubuth: Support for reserved counts added in 82d74dc.

spreadsheets/collision_rank.yaml
zerebubuth: I changed the ranges to "round numbers" in 82d74dc.

spreadsheets/collision_rank.yaml
zerebubuth: OK, fixed in 82d74dc.

spreadsheets/collision_rank.yaml
zerebubuth: OK, fixed in 82d74dc.

spreadsheets/collision_rank.yaml
zerebubuth: Yes, good point. Fixed in 87241c9.

spreadsheets/collision_rank.yaml
zerebubuth: Moved in ba8e791.

spreadsheets/collision_rank.yaml
zerebubuth: Fixed in cfe0df9.

spreadsheets/collision_rank.yaml
zerebubuth: Yes! Well spotted, thanks. Fixed in 998d0b3 and tests added.

spreadsheets/collision_rank.yaml
zerebubuth: Fixed in cfe0df9.

spreadsheets/collision_rank.yaml
zerebubuth: Fixed in cfe0df9.

spreadsheets/collision_rank.yaml
zerebubuth: Fixed in cfe0df9.

spreadsheets/collision_rank.yaml
zerebubuth: Fixed in cfe0df9.

spreadsheets/collision_rank.yaml
zerebubuth: Fixed in cfe0df9.

queries.yaml
zerebubuth: Good point. Yes, it's probably better not to add a `collision_rank` for things which we don't think will get labelled. In 8e8a6e8 I added a filter to only add `collision_rank` if one or more of the following were true:

* It's in the `pois` layer.
* It has a name-like thing (including `name:*` language translations, `short_name`, etc...)
* It has a `ref`
* It has a `shield_text` (or `bus_shield_text`, `bicycle_shield_text`, etc...)

PR #1796, opened 2019-01-18
Add HGV properties

zerebubuth:
Add:

* `low_emission_zone` polygons in `landuse`.
* `hgv_restriction` for `roads` as documented, with a value of `weight`, `height`, `width`, `length`, `wpa`, `hazmat` or `multiple`.
* `hgv_restriction_shield_text` for `roads` as documented, with a number + unit quantity where available (i.e: for all except `hazmat`).
* `hgv` restriction on `roads`; one of `agricultural`, `delivery`, `designated`, `destination`, `local`, `no`.
* `toll` and `toll_hgv` on `roads`; `true` when present, omitted default of `false`.

Notes:

* The documentation was originally confused about whether units for length should be cm or metres, kg or metric tonnes. I've settled on metres and tonnes, as these produce shield text closer to signage actually seen in Real Life, such as 10t or 3m, and also shorter shield text (1.5t vs 1,500kg).
* The documentation describes a `kpra` restriction, but I wasn't able to find a commonly-used OSM tag which corresponds to that, so I've left it out. If we find one, we can add it in a follow-up.
* I've added the restriction and shield text for `roads` layer line geometries, but not yet points. We might need to give some more thought to that, since the points at the end of the road ought to represent the addition of restrictions in the direction of travel and therefore we need to consider the restrictions _between_ two roads. In other words, we wouldn't want to insert a point a the start of a road with the same or less restrictive restriction than the previous one. This requirement to consider connectivity is not simple with our current architecture, as we'd have to reconstruct it from the `planet_osm_ways` table - something we currently do to figure out if a gate is on a road, but gates are (comparatively) rare and working out which are the ingress nodes onto a way is significantly more complex.

Connects to #1553.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1468,7 +1468,7 @@ The `hgv` property indicates general truck heavy goods vehicle truck access, val
 
 **Roads `hgv_restriction` values:**
 
-For `hgv_restriction` property indicates general truck heavy goods vehicle truck access restrictions, values include: `weight` (kg), `height` (cm), `length` (cm), `width` (cm), `wpa` (weight per axle, in kg), `kpra` (king pin to rear axle leght, in cm), `hazmat`, `other` and `multiple` if more than one.
+For `hgv_restriction` property indicates general truck heavy goods vehicle truck access restrictions, values include: `weight` (metric tonnes), `height` (metres), `length` (metres), `width` (metres), `wpa` (weight per axle, in metric tonnes), `kpra` (king pin to rear axle leght, in metric tonnes), `hazmat` (true if restricted, otherwise omitted), `other` and `multiple` if more than one.
 
 
 #### Roads layer network values
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -520,7 +520,6 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 * `dog_park`
 * `embankment` - A raised area of land, usually to carry a road or railway.
 * `enclosure`
-* `low_emission_zone` - An area beloging to a low emission zone, such as the (London Low Emission Zone)[https://en.wikipedia.org/wiki/London_low_emission_zone]. (See also)[https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dlow_emission_zone].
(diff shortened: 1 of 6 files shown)


integration-test/1553-hgv-properties.py
```diff
@@ -81,7 +81,8 @@ def _check_restriction(self, tags, restriction=None, shield_text=None):
             'highway': 'unclassified',
             'source': 'openstreetmap.org',
         })
-        self.generate_fixtures(dsl.way(1, dsl.tile_diagonal(z, x, y), all_tags))
+        self.generate_fixtures(
+            dsl.way(1, dsl.tile_diagonal(z, x, y), all_tags))
```
rmarianski: ❄️

docs/layers.md
```diff
@@ -1468,7 +1468,7 @@ The `hgv` property indicates general truck heavy goods vehicle truck access, val
 
 **Roads `hgv_restriction` values:**
 
-For `hgv_restriction` property indicates general truck heavy goods vehicle truck access restrictions, values include: `weight` (kg), `height` (cm), `length` (cm), `width` (cm), `wpa` (weight per axle, in kg), `kpra` (king pin to rear axle leght, in cm), `hazmat`, `other` and `multiple` if more than one.
+For `hgv_restriction` property indicates general truck heavy goods vehicle truck access restrictions, values include: `weight` (metric tonnes), `height` (metres), `length` (metres), `width` (metres), `wpa` (weight per axle, in metric tonnes), `kpra` (king pin to rear axle leght, in metric tonnes), `hazmat` (true if restricted, otherwise omitted), `other` and `multiple` if more than one.
```
nvkelso: It's fine to change the units on this, but let's coordinate with @musculman.

zerebubuth: Yup! Good idea.

I don't consider this to be a _change_ per se, since it was previously documented as cm & kg for `hgv_restriction` and metres & tonnes for `hgv_restriction_shield_text`. I followed the documentation for `hgv_restriction_shield_text`, and the diff above merely updates the conflicting documentation for `hgv_restriction` to match.

Issue #1794, opened 2019-01-16
Demote some early landcover down a few zooms


nvkelso:
* **What did you see?** Some landuse polygons are rediculously early, which bloats tile size and makes merging harder. Connects with https://github.com/tilezen/vector-datasource/issues/1721.
* **What did you expect to see?** Only shown at relevant zooms.
* **What map location are you having problems with?** Global!
* **Screenshot?** _Props for animated gifs._

[Jump to suggested fixes »](https://github.com/tilezen/vector-datasource/issues/1794#issuecomment-454936064)

nvkelso:
Table from @zerebubuth:


kind | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
national_park | 436 | 436 | 591 | 591 | 663 | 1230 | 1433 | 2023 | 2277 | 2595 | 2842 | 3112 | 3730 | 5091 | 5091
battlefield | 2 | 2 | 4 | 4 | 4 | 15 | 22 | 47 | 64 | 97 | 119 | 143 | 208 | 436 | 436
nature_reserve |   | 1470 | 1470 | 4005 | 4846 | 13325 | 17729 | 33136 | 41763 | 50877 | 62749 | 71571 | 99968 | 106675 | 106675
park |   | 642 | 642 | 1658 | 1922 | 4457 | 6193 | 16064 | 26243 | 44641 | 89994 | 145287 | 609038 | 813963 | 813963
protected_area |   | 299 | 299 | 609 | 669 | 1292 | 1671 | 2865 | 3540 | 4273 | 5273 | 6012 | 8566 | 9447 | 9447
forest |   | 179 | 179 | 376 | 446 | 1359 | 1940 | 296179 | 487838 | 660059 | 968297 | 1309198 | 3213533 | 3769716 | 3769716
residential |   |   |   | 53 | 109 | 8517 | 26403 | 197933 | 368297 | 600812 | 1068240 | 1578688 | 4140345 | 5018540 | 5018540
glacier |   |   |   |   |   | 6165 | 9074 | 19391 | 25045 | 31542 | 35301 | 38317 | 42862 | 43309 | 43309
[...]

Issue #1178, opened 2017-03-16
Bring piers in earlier


meetar:
The buildings on these piers show up before the piers do:

https://tangrams.github.io/explorer/#15.2/32.7165/-117.1731/kind/pier

Can we bring piers in at zoom 13, to see those features earlier? (Possibly by giving piers the same area thresholds as buildings?)

jschleuss:
+1

Our sister publication in San Diego would like them to show up earlier. Contact me if you have questions.

nvkelso:
Hi Jon :)

Can you confirm you want the pier landuse layer polygons earlier, not the
pier lines in the roads layer?

Looks like those are toggled by area right now, we could look into moving
them into a different zoom tier. Do they want the labels to show up earlier
for some of these, too? (Is the label more important or same importance as
the polygon?)

Are there any other attributes on the piers that tells us one is more
important than another? Like something to indicate the cruise ship terminal
is more “interesting” than say a industrial pier?

jschleuss:
Yessir!

We're looking for the pier landuse to show up earlier. It looks like the largest piers show up first (maybe at z13) and then more at z14 and z15. Labels of the piers aren't important to us at the present moment. 

This is all for our mapmaker, so don't let us rule the conversation. But some folks in San Diego said it's a little weird to have a map of Downtown SD and not see those piers. I've updated our YAML file https://github.com/datadesk/web-map-maker/issues/59

I can't see anything that would make some piers more important than others. It didn't even occur to me until they mentioned it. LA's piers are puny by comparison. Thanks for looking into this!

![piers](https://cloud.githubusercontent.com/assets/695934/24023735/8d3f2e42-0a6b-11e7-83b2-63089b0a041a.gif)

nvkelso:
Great GIF, thanks for the explanation. I agree they should show earlier,
will take some weeks to get a fix out.

zerebubuth:
Which zoom level should these be visible until? The smallest of the main piers in the GIF are:

* [Broadway Pier](http://www.openstreetmap.org/way/275609725#map=11/32.7217/-117.1836), which has area 17,855 and currently ends up with a `min_zoom` of 15. Should that be 11 instead?
* The unnamed [pier at the end of West G Street](http://www.openstreetmap.org/way/275609722#map=11/32.7362/-117.1458) has area 4,734 and also currently a `min_zoom` of 15. Should that also be 11, or would such a small polygon be almost invisible at that zoom?

[...]

PR #1797, opened 2019-01-22
Demote early landcover

zerebubuth:
Make changes to the start zoom of various features as described in https://github.com/tilezen/vector-datasource/issues/1794#issuecomment-454936064.

However:

* I didn't change the `min_zoom` of piers, as we deliberately included them at lower than usual zooms as part of #1178. If you think we should override that, then let me know; it's a simple change.
* The `camp_site` features were area-graded, but the majority are too large for the minimum of the range (previously 13+). At a "one pixel" zoom, the majority would come in at zoom 9. It feels like `camp_site` is similar to `caravan_site`, so for the moment I've given them the same zoom range. Let me know if you'd prefer something different. 

Connects to #1794.


integration-test/1794-demote-early-landcover.py
```diff
@@ -0,0 +1,319 @@
...
+        self._check_min_zoom({'leisure': 'dog_park'}, 'dog_park', 13)
+
+    def test_cinema(self):
+        self._check_min_zoom({'amenity': 'cinema'}, 'cinema', 13)
+
+    def test_water_park(self):
+        self._check_min_zoom({'leisure': 'water_park'}, 'water_park', 13)
+
+    def test_harbour(self):
+        self._check_min_zoom({'landuse': 'harbour'}, 'harbour', 13)
```
nvkelso: I see a test for `harbour`, but looks like we need to modify the data clamp to 13 (from 9)? Discuss.

zerebubuth: The test was just to check - `harbour` was already [clamped to z13](https://github.com/tilezen/vector-datasource/pull/1797/files/1c15d82b1c14d8718d6658c73879e8099eb39978#diff-a2a00baa9c8d3ef5107119155e3872ceR777).

zerebubuth:
> Nit: Looks like `dog_park` still needs to be clamped down to 13 from 9.

Yup, it was already [clamped to 13](https://github.com/tilezen/vector-datasource/pull/1797/files#diff-a2a00baa9c8d3ef5107119155e3872ceR1221).

Issue #1806, opened 2019-01-24
0/0/0 tile has clipped water content


nvkelso:
* **What did you see?** The 0/0/0 tile is missing water content at the north side.
* **What did you expect to see?** Water all the way north.
* **What map location are you having problems with?** 
* **Screenshot?** _Props for animated gifs._

![image](https://user-images.githubusercontent.com/853051/51713702-ba98e700-1fe7-11e9-8f90-223f1828fb85.png)

[1-0-0.mvt.zip](https://github.com/tilezen/vector-datasource/files/2793953/1-0-0.mvt.zip)
[0-0-0.mvt.zip](https://github.com/tilezen/vector-datasource/files/2793954/0-0-0.mvt.zip)

nvkelso:
Looks good!

PR #1807, opened 2019-01-25
Clip to Mercator world bounds before projecting shapefiles.

zerebubuth:
Previously, we had set an option to make `ogr2ogr` drop points which could not be projected into Mercator coordinates. This generally worked fine, and we ended up with output that looked correct.

However, the 1:110M oceans polygon was being truncated because there wasn't a point at exactly the corner of the Mercator world, and dropping the out-of-range points left a bit of a gap.

We can fix this by clipping (i.e: intersecting) with the Mercator world bounds before projecting. This should ensure that points are within the destination projection without the need to drop them. It's effectively inserting a point at the corner of the Mercator world.

Connects to #1806.


zerebubuth:
NOTE to self: We'll need to regenerate the assets bundle after this is merged.

Issue #1552, opened 2018-08-03
Support alternate points of view in boundaries layer


nvkelso:
Natural Earth supports the concept of "points of view" (POV) to indicate the claimants for different boundary disputes. 

Right now Tilezen supports just the default "defacto" point of view, but sometimes that deviates from US State Department and United Nations views. And showing maps of India, Pakistan, China, and other areas require special treatments. 

Tilezen should support both a default POV (now) and additional POV for country and disputed boundary lines (possibly also region but Natural Earth data doesn't support that now).

musculman:
When looking at this feature it would probably make sense to think about how to solve the fact that you would need to give different geometries depending on the `view` used.

In particular, we should not only think about a particular border being listed as belonging to two (or more) countries, but also the fact that these countries might have different understanding on where the actual border is.

Therefore, it has two main use cases:
- pure disputed borders: if `view` of Country A is used, the border of country A is here, if `view` of country B is used, the border is not here but somewhere else.
- different border depending on the `view`. As opposed to the previous case in which the border is either shown as disputed (dashed) or belonging to country A or B; in this case, the border is not there and it is somewhere else, changing the actual shape of the country.

Looking at a possible implementation for a border with an example, could be:

- `is_disputed_by`=A
- `is_accepted_by`=B
This border will be displayed in country B but not in country A.

- `is_disputed_by`=A
This border will be displayed in all views, but not in country A.

Additional attributes worth considering are:
- `claimed by` : an attribute that indicates which country claims a cartographic administrate polygon or linear administrative boundary in a disputed area.
- `controlled by`: in disputed areas which country is, according to international agreements, in control of the area.
- `line of control`: usually marks the internationally accepted boundaries, but are located in areas under dispute. Some countries might claim large territory, but de-facto control only up to this line.

zerebubuth:
As an alternative way of looking at things; borders (national or otherwise) are largely fictitious. In some cases, there is a physical barrier marking the boundary. This makes it easy to map and draw - the big fence or wall is the border. Where there is no physical barrier, the border is by nature more abstract and porous, so it matters less exactly where it is drawn.

Some countries mandate, sometimes by law, that maps used within their jurisdiction are drawn to support their opinions (and often, ideology). Therefore, rather than calling it a `view`, we could call it a `mandatory_delusion` where it differs from the ground truth (_de facto_) feature.

The logic for drawing might look something like:

```python
ground_truth = feature.properties.get('ground_truth', 'draw')
[...]

PR #1809, opened 2019-01-31
Alternate viewpoints for country boundaries.

zerebubuth:
Adds alternate viewpoints about country boundaries following #1552. This means adding a `kind:XX` where a particular country has a different opinion about the `kind` of a boundary from the default. This should be used in the same way as `name:XX` translations. A new `kind`, `unrecognized`, indicates that the boundary should not be shown in that country's viewpoint.

Connects to #1552.


yaml/boundaries.yaml
```diff
@@ -95,53 +149,25 @@ filters:
...
+        - 'International boundary (verify)'
+        - 'Lease limit'
+        - 'Line of control (please verify)'
+        - 'Overlay limit'
+        - 'Unrecognized'
+        - 'Breakaway'
+        - 'Claim boundary'
+        - 'Elusive frontier'
+        - 'Reference line'
     min_zoom: 0
```
nvkelso: Please look up the `min_zoom` from the data.

nvkelso:
This looks good to me so far, just the one comment about min_zoom... and please add a test for the transform of say `kind:in` along the India, Pakistan border somewhere (on data fixture).

Issue #1814, opened 2019-02-04
Aerodrome polygons over aeroways


bcamper:
* **What did you see?**

`aerodrome` polygons have a higher `sort_rank` than `aeroway` features (runways, taxiways, etc.). This causes the aeroways to be hidden.

* **What did you expect to see?**

Aeroways rendered over aerodromes.

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

#14.750063185324565/40.64191771698365/-73.77976573047053

* **Screenshot?** _Props for animated gifs._

See JFK airport (verified at SFO as well):

Default view:
![jfk-aerodrome](https://user-images.githubusercontent.com/16733/52185721-5b5d8280-27f0-11e9-94eb-52aa0315118f.png)

With landuse layer disabled, to show aeroways:
![jfk-runways](https://user-images.githubusercontent.com/16733/52185725-66181780-27f0-11e9-8a34-bfe45449ddee.png)

nvkelso:
Fixed:

![image](https://user-images.githubusercontent.com/853051/52752946-922e5800-2fa9-11e9-8856-efb7cfa76eb9.png)

![image](https://user-images.githubusercontent.com/853051/52752960-a1150a80-2fa9-11e9-9a0b-da2cdb1db04d.png)

PR #1818, opened 2019-02-04
Aeroways in roads layer should sort under aerodrome landuse polygons.

zerebubuth:
Connects to #1814.

### DIFF
--- a/integration-test/1814-aerodrome-should-be-under-aeroways.py
+++ b/integration-test/1814-aerodrome-should-be-under-aeroways.py
@@ -1,0 +1,50 @@
+# -*- encoding: utf-8 -*-
+from . import FixtureTest
+
+
+class AerodromeSortTest(FixtureTest):
+
+    def test_sort_order(self):
+        import dsl
+
+        z, x, y = (16, 0, 0)
+
+        poly = dsl.tile_box(z, x, y)
+        line = dsl.tile_diagonal(z, x, y)
+
+        def _src(props):
+            p = {'source': 'openstreetmap.org'}
+            p.update(props)
+            return p
+
+        self.generate_fixtures(
+            dsl.way(1, poly, _src({'aeroway': 'aerodrome'})),
+            dsl.way(2, poly, _src({'aeroway': 'runway'})),
+            dsl.way(3, poly, _src({'aeroway': 'taxiway'})),
+            dsl.way(4, poly, _src({'aeroway': 'apron'})),
+            dsl.way(5, line, _src({'aeroway': 'runway'})),
+            dsl.way(6, line, _src({'aeroway': 'taxiway'})),
+        )
+
+        landuse_kinds = {}
+        with self.features_in_tile_layer(z, x, y, 'landuse') as features:
+            for feature in features:
+                kind = feature['properties']['kind']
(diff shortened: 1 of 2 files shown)


zerebubuth:
> I suspect the issue here is the `runway` polygons seem to be missing (!?)

I checked out JFK [with the data layer on](https://www.openstreetmap.org/way/510553024#map=17/40.64659/-73.76501&layers=D) and I don't see any `runway` polygons - looks like the runways and taxiways are lines-only.

> the `aerodrome` don't seem to have a good `sort_rank` value anymore w/r/t to the other features (74 versus 61 and 62).

Would you prefer that I brought the `aerodrome` landuse `sort_rank` down to 60 instead of changing the `roads` layer `sort_rank`s?

nvkelso:
## About the missing runway features that used to be in Tilezen builds. 

It looks like ~60 people have been retagging hundreds of `aeroway=runway` polygons in OSM with `area:aeroway=runway`(and taxiway as well)! We'll need to update our 2 filters for those to allow both.

- https://www.openstreetmap.org/way/167318897/history was recently retagged like this.

## Sort order needs to change

**roads:**

These lines should sort below the polygon representation of `runway`, `apron`, and `taxiway`, but above `aerodrome`. 

This will resolve @bcamper's original bug report ("missing" lines, covered by the larger airport AOI), and the more accurate polygon representations to visually mask the thinner and possibly stroked lines by default – an solution we developed at Mapzen for the basemap designs.

_to_
```
*,*,aeroway,runway,*,*,*,76
*,*,aeroway,taxiway,*,*,*,75
```

_from_
```
*,*,aeroway,runway,*,*,*,62
*,*,aeroway,taxiway,*,*,*,61
```

**landuse:**

_to:_
```
runway,*,79
apron,*,78
taxiway,*,77
# runway lines at 76
# taxiway lines 75
aerodrome,*,74
airfield,*,73
```

from:
```
runway,*,78 
apron,*,77
taxiway,*,76
# we had a gap of 2 once here, something got mangled along the way to get it to be only 1
aerodrome,*,74
airfield,*,73
```

I suspect when we added more landuse in the last few releases we add them "at the bottom", but didn't also adjust the related roads layer sort values. Seems like a good reason to switch these "spreadsheets" from CSV to YAML so we can add explicate gaps and comments to future self to avoid regression again.

nvkelso:
YAML conversion tracked for later work in https://github.com/tilezen/vector-datasource/issues/1823.

integration-test/1814-aerodrome-should-be-under-aeroways.py
```diff
@@ -62,3 +62,34 @@ def _src(props):
             aerodrome_rank = landuse_kinds['aerodrome']
             self.assertTrue(aerodrome_rank < roads_kinds['runway'])
             self.assertTrue(aerodrome_rank < roads_kinds['taxiway'])
+
+            # however, we also want the roads to be _under_ the landuse
+            # polygons for the same type, due to the styling adding a
+            # casing which looks weird.
+            self.assertLess(roads_kinds['runway'], landuse_kinds['runway'])
```
nvkelso: 💯

zerebubuth:
OK, I think I updated everything. Please take a look and let me know.

Issue #1826, opened 2019-02-07
Default country labels to zoom 6 instead of 1 for no-NE matches


nvkelso:
Baby steps towards https://github.com/tilezen/vector-datasource/issues/1815, by hiding the junk down to zoom 6 instead of defaulting to 1 when we can't find a match in Natural Earth.

In the image below **Transnistria** and **Nagorno-Karabakh** are more disputed territories than countries. 

![image](https://user-images.githubusercontent.com/853051/52189804-cce90100-27ef-11e9-9120-e8ffb04a72b9.png)

nvkelso:
This is working :)

![image](https://user-images.githubusercontent.com/853051/52752849-2ba93a00-2fa9-11e9-9704-3cbfbdd82a8f.png)

![image](https://user-images.githubusercontent.com/853051/52752869-382d9280-2fa9-11e9-9ab8-a38356759546.png)

![image](https://user-images.githubusercontent.com/853051/52752893-509dad00-2fa9-11e9-8d2e-f5e75ae1cb91.png)

PR #1827, opened 2019-02-07
Push countries and regions without NE min zoom down to 6.

zerebubuth:
For countries and regions which aren't matched to a Natural Earth record, push the min zoom down to 6.

Connects to #1826.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -8314,13 +8314,16 @@ def tags_set_ne_min_max_zoom(ctx):
                 min_zoom = ceil(min_zoom)
             props['min_zoom'] = min_zoom
 
-        elif props.get('kind') in ('country', 'region'):
+        elif props.get('kind') == 'country':
             # countries and regions which don't have a min zoom joined from NE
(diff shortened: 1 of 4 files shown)


vectordatasource/transform.py
```diff
@@ -8314,6 +8314,14 @@ def tags_set_ne_min_max_zoom(ctx):
                 min_zoom = ceil(min_zoom)
             props['min_zoom'] = min_zoom
 
+        elif props.get('kind') in ('country', 'region'):
+            # countries and regions which don't have a min zoom joined from NE
+            # are probably either vandalism or unrecognised countries. either
+            # way, we probably don't want to see them at zoom, which is lower
+            # than most of the curated NE min zooms. see issue #1826 for more
+            # information.
+            props['min_zoom'] = max(6, props['min_zoom'])
```
nvkelso: Nit: if region let's do 8 instead.

zerebubuth: Fixed in 9938a5d1.

PR #1829, opened 2019-02-07
add road_kind_detail to traffic layers, so possible traffic line offseting can be done properly

zaczkows:
add missing road's kind_detail, so all possible road types can be represented


nvkelso:
Agree this is necessary and was missing in the original documentation PR. Thanks for addressing in this PR!

PR #1830, opened 2019-02-08
Add root_id to Traffic Incident & Traffic Flow Layers

conor-ettinoffe-here:
One of our engineers has identified the following use case leading to this proposal:

In a case where lets say 3 road segments are of different road_kind, The tile should have 3 different features. Using the incident ID as a feature ID breaks the requirement that all features inside a tile must have unique ID. Thus, we decided that in this particular case that is quite rare but still possible, we will assign unique feature id for each feature and add a common incident_id field for all of them.
23ca38d


docs/layers.md
```diff
@@ -2327,6 +2327,7 @@ This **optional** layer is meant to be updated every few minutes, due to the hig
 
 #### `optional` properties:
 
+* `incident_id`: used to identify an incident that is common across multiple features. E.g. Where `road_kind` may differ across features but affected by a common incident
```
nvkelso: I think this should be moved to the `optional` section in the ** traffic_flow** layer section above (instead of **traffic_incidents** layer here)?

But does this happen in practice? Is there documentation you can share for the source data schema, please?

This reminds me of the `root_id` we already have for building some pois (like transit stations). 

To be consistent with that, we could say:

- `root_id` - an identifier which can be used to link or group together multiple traffic flow features which are related by being part of a common traffic incident split across several kinds of roads.

conor-ettinoffe-here: Hi Nathaniel,

> This reminds me of the `root_id` we already have for building some pois (like transit stations).
> 
> To be consistent with that, we could say:
> 
> * `root_id`: an identifier which can be used to link or group together multiple traffic flow features which are related by being part of a common traffic incident split across several kinds of roads.

I will update the docs accordingly.

> I think this should be moved to the `optional` section in the ** traffic_flow** layer section above (instead of **traffic_incidents** layer here)?

This particular case happens for the line incidents, it can also happen to traffic flow so I will add root_id that layer too.

> But does this happen in practice? Is there documentation you can share for the source data schema, please?

I talked with the engineer who found this case, he mentioned it occurs very rarely but can happen. Please check the diagram below that describes where root_id is required:
image

So here's an example where the link L2 may be a different road_kind and therefore they will have different IDs but the same root_id.

nvkelso: So you'd only ever be joining data onto the `root_id` (nea `incident_id`)? And the basic `id` here is what then? Does your system have a unique incident+road_segment ID that can be used for the `id` instead and for the join?

nvkelso: I should clarify that the `id` property here is NOT the same as the MVT feature `id`:

- The MVT feature id is required to be unique, but we don't set that in Tilezen
- Tilezen's `id` property is not required to be unique for a single tile – sometimes multiple features do share the same ID.

So perhaps this PR is moot and it's okay to close this PR?

(In any event, let's continue this discussion for v1.8 milestone as v1.7 is closing today.)

conor-ettinoffe-here: Hi Nathaniel. Indeed we were using the MVT ID creating the problem we were facing - but we can actually use the standard id field from Tilezen. I will close this topic.

PR #1831, opened 2019-02-08
Adding Express Lane flag to Traffic

conor-ettinoffe-here:
- Adding optional flag to identify whether Traffic Flow or Incident information is related to a Express Lane linear.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -2288,7 +2288,7 @@ However, if the roads layer does not include `linear_ref_id`, then it should be 
 * `is_bridge`: set to `true` when the linear is a bridge
 * `is_tunnel`: set to `true` when the linear is a tunnel
 * `is_link`: set to true when the linear is a slip-road
-* `is_express_lane`: set to `true` when the linear is an HOV (High-Occupancy Vechicle) lane.
+* `is_express_lane`: set to `true` when the flow refers to an express lane.
 
 ## Traffic Incidents
 
@@ -2342,7 +2342,7 @@ However, if the roads layer does not include `linear_ref_id`, then it should be 
 * `is_tunnel`: set to `true` when the linear is a tunnel
 * `is_link`: set to true when the linear is a slip-road
 * `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
-* `is_express_lane`: set to `true` when the linear is an HOV (High-Occupancy Vechicle) lane.
+* `is_express_lane`: set to `true` when the incident affects an express lane.
 
 ## Water
 
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -2342,7 +2342,7 @@ However, if the roads layer does not include `linear_ref_id`, then it should be 
 * `is_tunnel`: set to `true` when the linear is a tunnel
 * `is_link`: set to true when the linear is a slip-road
 * `drives_on_left`: set to `true` when the country drives on the left (e.g. in the United Kingdom)
-* `is_hov_lane`: set to `true` when the linear is or the incident affects a HOV (High-Occupancy Vechicle) express lane
+* `is_hov_lane`: set to `true` when the linear is, or the incident affects, a HOV (High-Occupancy Vechicle) express lane
 
 
 ## Water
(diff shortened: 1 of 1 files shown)


docs/layers.md
```diff
@@ -2288,6 +2288,7 @@ However, if the roads layer does not include `linear_ref_id`, then it should be
 * `is_bridge`: set to `true` when the linear is a bridge
 * `is_tunnel`: set to `true` when the linear is a tunnel
 * `is_link`: set to true when the linear is a slip-road
+* `is_express_lane`: set to `true` when the linear is an HOV (High-Occupancy Vechicle) lane.
```
nvkelso: Why is this necessary, can you explain your rational with an example, please? 

We currently don't include `hov` tags on the **roads** layer features, so seems slightly odd to include it here in the **traffic_flow** and **traffic_incidents** layers. While there has been discussion to add lane tags into the tiles, and then hov could be added as part of that, I'm still not sure why it's necessary – except for HD mapping, which Tilezen currently isn't really geared for.

If we were to do this...

The OSM-ism we'd follow is `hov` instead of `express`, please say `is_hov` instead.

See also:

- https://wiki.openstreetmap.org/wiki/Key:hov

Visually would you sort the HOV traffic over the non-HOV traffic flow lines? Would you want a way to turn off the non-HOV traffic lines? This starts to get tricky fast.

docs/layers.md
```diff
@@ -2341,6 +2342,7 @@ However, if the roads layer does not include `linear_ref_id`, then it should be
 * `is_tunnel`: set to `true` when the linear is a tunnel
 * `is_link`: set to true when the linear is a slip-road
 * `drives_on_left`: set to `true` when the country drives on the left, e.g. In the U.K
+* `is_express_lane`: set to `true` when the linear is an HOV (High-Occupancy Vechicle) lane.
```
nvkelso: See above, would be `is_hov`.

But please explain why HOV tag is needed here at all. Plus we're talking about points here, right? So the comment would also need to update.

docs/layers.md
conor-ettinoffe-here: Hi Nathaniel, 
I got confused (I'm British/Irish) and misunderstood what an Express Lane is. So updating the patch with a new commit for the explanations.

However this still won't satisfy this comment:
> While there has been discussion to add lane tags into the tiles, and then hov could be added as part of that, I'm still not sure why it's necessary – except for HD mapping, which Tilezen currently isn't really geared for.

Indeed this is lane-specific information, but basically we'd like to be able to have this tag so that the client is aware that the Traffic Flow/Incident information is affecting the Express Lane on a particular road.

docs/layers.md
conor-ettinoffe-here: Please see my comment above :)

conor-ettinoffe-here:
> I updated this to the OSM-ism `is_hov_lane`, reworded the docs, and LGTM for merging.

Hi @nvkelso,

I misunderstood and didn't actually mean HOV, rather the [express lane](https://en.wikipedia.org/wiki/Local-express_lanes) thus, why I changed the name of the field and the explanation after the comment I made
> I got confused (I'm British/Irish) and misunderstood what an Express Lane is. So updating the patch with a new commit for the explanations.
 
So, could we please keep it as such `is_express_lane`?

nvkelso:
@conor-ettinoffe-here I remember seeing followup about this over the weekend, but I can't find now. Can you summarize with a comment here, please?

conor-ettinoffe-here:
> @conor-ettinoffe-here I remember seeing followup about this over the weekend, but I can't find now. Can you summarize with a comment here, please?

Hi Nathaniel, its okay no longer a point :) I basically was saying I meant Express Lane and not HOV - however according to OSM HOV is a synonym of express lane:

> [W] High-occupancy vehicle lanes (sometimes also called carpool lanes, commuter lanes, express lanes, transit lanes). The number of such lanes could be tagged using lanes:hov=*.

So all is clear now. Thanks!

nvkelso:
Okay, sounds good! I agree it's a little confusing, but all is well that ends well :)

PR #1842, opened 2019-02-15
Fix links for low emission zone

musculman:
- [x] Update tests (N/A)
- [x] Update docs


nvkelso:
Good catch!

nvkelso:
@zerebubuth That reminds me the OSM polygon is busted now. Let's try to fix that up so the next build has good sample data for the new feature.

PR #1844, opened 2019-02-15
Faster "perform SQL updates"

zerebubuth:
There's a few things going on here:

* Execute one `UPDATE` per table, rather than several. This should reduce the number of passes over the table, especially since the lines and polygons tables touched every row when setting the label placement point.
* Shard the updates 4x over `osm_id` for the OSM tables. This helps a lot when the `UPDATE` is CPU-bound, rather than disk-bound, as it seems to be when we're calculating a lot of `ST_PointOnSurface()` and min zoom `pl/pgsql` functions.
* Move index creation into one file per index, to be executed in parallel. PostgreSQL allows any number of index creations to run concurrently, and these can be very long-running. Also remove `ANALYZE` from `apply-*.sql`, so it can run after index creation.


data/apply-planet_osm_line.sql
```diff
@@ -5,82 +5,15 @@ BEGIN
...
+    mz_transit_level = mz_calculate_min_zoom_transit(planet_osm_line.*),
+    mz_water_min_zoom = mz_calculate_min_zoom_water(planet_osm_line.*),
+    mz_label_placement = ST_PointOnSurface(way)
   WHERE
-    mz_earth_min_zoom IS NOT NULL OR
-    mz_landuse_min_zoom IS NOT NULL OR
-    mz_road_level IS NOT NULL OR
-    mz_transit_level IS NOT NULL OR
-    mz_water_min_zoom IS NOT NULL;
+    SHARDING;
```
nvkelso: `WHERE SHARDING;` is new to me. Is it a Postgres-ism, or related to the script changes in `data/perform-sql-updates.sh`? Might be worth an inline comment?

data/perform-sql-updates.sh
```diff
@@ -38,10 +38,41 @@ echo "done."
...
-psql $PSQLOPTS  $@ -f apply-planet_osm_polygon.sql &
-psql $PSQLOPTS  $@ -f apply-planet_osm_line.sql &
-psql $PSQLOPTS  $@ -f apply-planet_osm_point.sql &
+psql $PSQLOPTS  $@ -e -f apply-updates-non-planet-tables.sql &
+
+# use postgres' own estimate of the percentile breakdown of the osm_id column to
+# guide the distribution of jobs, so hopefully they end up mostly evenly sized.
+for tbl in polygon line point; do
+    sql_script="apply-planet_osm_${tbl}.sql"
+    for pct in 25 50 75; do
```
nvkelso: Is the machine we run this on always 4 cores?

data/apply-planet_osm_line.sql
zerebubuth: It's nothing in PostgreSQL, sadly. It comes from [a hack](https://github.com/tilezen/vector-datasource/pull/1844/files#diff-ef923fe585ff36710968cbfbe143fdd6R53) where I'm replacing `SHARDING` with the range of `osm_id` to shard over. I would have used a proper Jinja2 template, but that was going to introduce a lot of incidental complexity.

I'm not totally sure whether an inline comment would work there, or if it would have weird side-effects... I'll have to test and make sure.

data/perform-sql-updates.sh
zerebubuth: Ideally, it would be related to the number of cores on the PostgreSQL server, not the machine we're running the script on. However, there's a couple of things that make matching the number of statements to the range of queries hard:

1. Doing this in shell script is _possible_, but makes me uncomfortable. I find that getting shell code correct is a lot harder than it looks. Going to fixed concurrency seemed like a good compromise. We can go to variable concurrency, but I'd want to switch all of this into Python, or something else that has less magic than shell.
2. I'm not aware of any call in PostgreSQL to tell us how many cores the server has, which is the important number to match. We could pass that in as an argument, but again; complexity.

Overall, it seemed to me like a good compromise to get something working quickly. If we want it to be more readable and more robust, then I can take a look at doing that next week?

nvkelso: I'm fine with hardcoding it now... just as long as we're getting good use out of the PostggresSQL Server we generally run this on. If that's 4 cores, so be it! :)

data/apply-planet_osm_line.sql
zerebubuth: Added comments in e684cd8, looks like they're correctly preserved through all the string mangling we're doing.

zerebubuth:
I've tested this out, and it seems to work - the `min_zoom` and indexing step was down from 54h before to 20h on this one test. Please have a look and let me know what you think.

data/perform-sql-updates.sh
nvkelso: In the comment, please also note this is configured for a 4 core PostgreSQL machine and would need to be adopted to other cores?

zerebubuth: Good idea. Fixed in dbebede.

zerebubuth:
> Is it a pain to name it something like `$SHARDING` and replace on that?

`$SHARDING` might be tricky (and fragile) since `$` is a special character in `sed`. How about `{{SHARDING}}` in a061e45? Hopefully that makes us think "templating".

PR #1846, opened 2019-02-26
[docs] HGV time restrictions

nzjony:
This is a request to add time based hgv (heavy good vehicle) restrictions. This is used both for routing (can the truck actually pass through this road segment if it leaves at a given time) and for visualisation (why has this truck been routed this way and not with a more direct route).

- Such restrictions can be limited to a specific day or days.
- Sometimes there is more than one time duration, for example around school start and end times.
- There may also be a reference just to night / day time and no specific time.
- Format is easily convertible to local time formats.


nvkelso:
@nzjony Interesting. Generally we don't include this level of fidelity in vector tiles, but let's keep talking.

Can you provide a fully fleshed out example in the new docs, please? MVT in particular doesn't support array objects, so are you proposing to use OSM's `;` separator character between array elements as a text string, or?

nvkelso:
@nzjony ping

nzjony:
hey @nvkelso, sorry for the delay, I got pulled onto another topic :(, so I have been unable to follow up with this.

What do you mean by "new docs", is there another location for such a change?

I will try to get some data to give some examples, but unfortunately I will likely be delayed because of a change of priorities, will get back to you.

nvkelso:
Thanks, I understand shifting priorites. 

For your PR, can you update the new **Roads `hgv_time_restrictions` values:** section to have an example value in the specific case, not the abstract case?

PR #1847, opened 2019-02-26
Fix slow road merging

zerebubuth:
The tile `9/454/201` was taking a _huge_ amount of time to render - several hours. This is due to the changes I made in #1703 to prevent intersections between lines being re-added as part of MultiLineString operations. According to the OGC spec, LineStrings within a MultiLineString may only meet at points, but this means we end up with way more points in the output than we need to accurately describe the geometry. Therefore, we split them up into several features, where each MultiLineString in each feature only contains non-overlapping LineStrings.

The algorithm I was using to do this wasn't efficient - it simply looped over all the features looking for intersections with previous features. And it worked OK on cities with a lower road density and lots of mergeable cross-shaped junctions, such as New York.

However, in Tokyo, where the tile was taking hours to render, there are 400,000 lines in the tile and the road network is not a grid, meaning there's less opportunity for merging.

Therefore, I replaced the previous algorithm with one that is more efficient for larger numbers of tiles. It works by indexing all the shapes rather than doing direct comparison. For large inputs, the shapes are additionally sorted by bounding box area and the index split between small and large. This means that the small-to-small comparisons are done more quickly (any given two small shapes are much less likely to intersect than a small and a large, or a large and a large).

Finally, we have been stripping `name` off roads at low zooms, but were not stripping `name:*` translations or `official_name` and similar "name-like" properties. This PR adds an `all_name_variants` parameter to `drop_properties` which tells it to treat `name` as if it's all `name:*` and variants. One `drop_properties` which was only dropping `name` was converted to `drop_names` instead. Dropping these might (slightly) increase mergeability and will decrease the number of properties stored in the tile.

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -2403,9 +2403,12 @@ def drop_properties(ctx):
     """
 
     properties = ctx.params.get('properties')
+    all_name_variants = ctx.params.get('all_name_variants', False)
     assert properties, 'drop_properties: missing properties'
 
     def action(p):
+        if all_name_variants and 'name' in properties:
+            p = _remove_names(p)
         return _remove_properties(p, *properties)
 
     return _project_properties(ctx, action)
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -3949,6 +3949,107 @@ def _merge_junctions(features, angle_tolerance, simplify_tolerance):
     return new_features
 
 
+def _first_positive_integer_not_in(s):
+    """
+    Given a set of positive integers, s, return the smallest positive integer
+    which is _not_ in s.
+
+    For example:
+
+    >>> _first_positive_integer_not_in(set())
+    1
+    >>> _first_positive_integer_not_in(set([1]))
+    2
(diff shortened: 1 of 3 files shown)


zerebubuth:
Oh, I probably should have mentioned the times; previously, the `9/454/201` was taking several hours to generate - I don't know exactly how many because I gave up waiting. After this PR, it still takes 30 minutes, which isn't awesome but much better than it was.

vectordatasource/transform.py
```diff
@@ -4018,19 +4093,71 @@ def as_shape(self):
...
+    # which have fewer intersections towards the front of the array, so that
+    # they can be done more quickly.
+    def _bbox_area(geom):
+        minx, miny, maxx, maxy = geom.bounds
+        return (maxx - minx) * (maxy - miny)
+
+    # if there's a large number of geoms, switch to the split index and sort
+    # so that the spatially largest objects are towards the end of the list.
+    # this should make it more likely that earlier queries are fast.
+    if len(mls.geoms) > 15000:
```
rmarianski: Do we want to pull this value from config? Or is this a good number that's not likely to change?

Issue #1277, opened 2017-06-21
Add kind_detail for aerodrome


nvkelso:
Looks like there's disambiguation between types of airports:

- https://taginfo.openstreetmap.org/keys/aerodrome%3Atype#values

**Common values include:**

_We probably shouldn't include all these..._

- public
- private
- military
- military/public
- civil
- gliding
- club
- heliport
- glider
- regional
- airstrip
- international
- abandoned
- altiport
- private/public
- military/private

nvkelso:
Passing the kind_detail onto runway lines is covered in https://github.com/tilezen/vector-datasource/issues/1852.

nvkelso:
We still need to pass this to the POI.

zerebubuth:
> We still need to pass this to the POI.

Since we generate the POI label point and polygon from the same feature, and the rules for `kind_detail` look to be the same, I think this information will already be on the POI. We should check once we have the first proper v1.8 build to make sure, in case of bugs :beetle:

nvkelso:
Sadly not very many airports have this tagged, but it is working as designed per this PR. 

We should probably backfill this for `international` based on the presence of 3-char IATA codes, too. But for another issue in https://github.com/tilezen/vector-datasource/issues/1873 as the basic request here is working.

It is on the landuse:

![image](https://user-images.githubusercontent.com/853051/56513933-cb72cf00-64e8-11e9-9fa2-23c41a560df9.png)

It is passing to the POI:

![image](https://user-images.githubusercontent.com/853051/56513707-25bf6000-64e8-11e9-8d62-7b5da9ddd6c1.png)

I expect the runway lines to also have a `aerodrome_kind_detail` property via https://github.com/tilezen/vector-datasource/pull/1854:

![image](https://user-images.githubusercontent.com/853051/56513903-b72ed200-64e8-11e9-9207-c2862c3099ad.png)

PR #1851, opened 2019-03-13
Add whitelisted kind_detail to aerodromes.

zerebubuth:
Whitelisted values:

* `public`
* `private`
* `military/public`
* `airfield`
* `international`
* `regional`
* `gliding`

Also remapped `aeroway:type=military` to `kind=airfield`, same as `military=airfield`, as that seemed more appropriate. I left `military/public`, as I think that means an aerodrome shared between both.

Connects to #1277.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -482,7 +482,7 @@ _TIP: Some `landuse` features only exist as point features in OpenStreetMap. Fin
 
 #### Landuse `kind` values:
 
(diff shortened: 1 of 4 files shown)


zerebubuth:
> Should `military/public` be transformed to `military_public`?

Sure, although we usually do that to keys rather than values. Fixed in 0e99911.

Issue #1580, opened 2018-08-26
Distinguish military airfield from generic airfield


nvkelso:
Show as **military_airfield** - http://wiki.openstreetmap.org/wiki/Tag:military%3Dairfield (also landuse, and pois if has name)

- combo tagging, prefer `military_airfield`: https://www.openstreetmap.org/node/412044660#map=19/51.86662/-1.69666
- sole military airfield tagging: https://www.openstreetmap.org/node/4226596407#map=19/51.20595/-1.74485

Also ensure the polygon is tagged correct, not just the poi.

QA followup for https://github.com/tilezen/vector-datasource/issues/1423.

nvkelso:
Related: 

- Add kind_detail for aerodrome #1277

nvkelso:
Verified, closing.

Was:

![image](https://user-images.githubusercontent.com/853051/56514370-f14ca380-64e9-11e9-8335-fc197b4f7ea5.png)

Now:

![image](https://user-images.githubusercontent.com/853051/56514380-f7db1b00-64e9-11e9-8d71-3f315bd1fa82.png)

PR #1860, opened 2019-03-27
Prioritise military airfields over aerodromes.

zerebubuth:
For `pois` (both points and polygons), use `kind: military_airfield` instead of `kind: airfield` for these. 

Note: `landuse` still uses the `airfield` kind, with a `kind_detail: military`. Should that be changed to `kind: military_airfield` too, dropping the `kind_detail`?

Connects to #1580.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -665,6 +665,7 @@ At zoom 12 and below, we consolidate some landuse kinds to reduce the amount of 
 
 The current mappings are:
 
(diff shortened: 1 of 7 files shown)


zerebubuth:
Fixed the PR so that it's not renaming `kind: airfield` any more (i.e: no breaking changes). But kept the change in priority for `airfield` above `aerodrome`.

Issue #1590, opened 2018-08-30
Only set POI for harbour not landuse polygon


nvkelso:
* **What did you see?** landuse polygon out over the water, oops
* **What did you expect to see?** Only a POI label, not a landuse polygon when its a water feature.
* **What map location are you having problems with?** https://www.openstreetmap.org/way/3303804#map=16/51.4963/-2.7199

<img width="490" alt="screen shot 2018-08-29 at 23 39 27" src="https://user-images.githubusercontent.com/853051/44834105-fec4be80-abe4-11e8-95aa-cc9b7977eb73.png">

QA followup for https://github.com/tilezen/vector-datasource/issues/1423.

nvkelso:
Verified. POI still gets area, but no polygon drawn:

![image](https://user-images.githubusercontent.com/853051/56534396-6af68880-650e-11e9-9315-26e185cc6822.png)

PR #1863, opened 2019-03-28
When harbour landuse is also water, drop the polygon

zerebubuth:
When a polygon is tagged as both something that we would consider water and a habour (or port or terminal), then keep the label placement point in the `landuse` layer, but drop the polygon as it would cover the water polygon from the `water` layer.

Another way of doing this would be to subtract the `water` layer polygons from (a subset of) the `landuse` layer polygons.However, I though this might lead to more unexpected side-effects, so stuck with the more specific feature and tag-based system.

Connects to #1590.

### DIFF
--- a/yaml/landuse.yaml
+++ b/yaml/landuse.yaml
@@ -786,6 +786,11 @@ filters:
       kind: winter_sports
       tier: 4
   # port, ferry, container terminals
+  #
+  # if these are also water features, then we don't want to output a polygon
+  # for this. because we can't alter the geometry type at this stage, we handle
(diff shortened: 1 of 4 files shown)


yaml/landuse.yaml
```diff
@@ -797,6 +802,14 @@ filters:
     output:
       <<: *output_properties
       kind: {col: landuse}
+      tz_drop_polygon:
```
nvkelso: Let's just keep `mz` namespace, otherwise it becomes confusing with both? We've renamed it to "metazen" in WOF ;)

Issue #1838, opened 2019-02-14
Too many straight and bay labels visible at zoom 8


nvkelso:
Looks like a BC data import is messing with our area thresholds on these water labels. Everyone gets a water label!

![image](https://user-images.githubusercontent.com/853051/52753136-6364b180-2faa-11e9-90f2-657779c61f95.png)

nvkelso:
Verified:

![image](https://user-images.githubusercontent.com/853051/56534494-b01aba80-650e-11e9-837c-c2e8591fced6.png)

PR #1866, opened 2019-04-10
Limit number of bays in the tile to the largest 10.

zerebubuth:
This adds a `tile_kind_rank` on `bay`s, `strait`s and `fjord`s based on the area of the feature. Labels are only kept if they're in the top 10 for the metatile. For the area around Vancouver, this looks like it keeps enough labels, but they're not so overwhelming any more.

Connects to #1838.

### Zoom 8

before_z8

### Zoom 9

before_z9

### Zoom 8

pr_z8

### Zoom 9

pr_z9

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -2377,12 +2377,12 @@ Tilezen calculates the composite exterior edge for overlapping water polygons an
 #### Water `kind` values:
 
 * `basin` - polygon
-* `bay` - point, intended for label placement only
+* `bay` - point, intended for label placement only. With `tile_kind_rank`, see below.
 * `canal` - line
 * `ditch` - line
 * `dock` - polygon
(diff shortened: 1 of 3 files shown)


zerebubuth:
> ¿is the kind_tile_rank shared between those 3 kinds or separate ranks for each kind?

Shared between those 3, since it seemed like the context was "smaller coastal features". We use the same `min_zoom` calculation for all 3, so it seemed fair to compare them.

Easy to change, if you'd prefer it separate for all 3? I'm not sure how many tiles have significant numbers of more than one of these, so would the limit be 10 each, or 3-4 of each?

nvkelso:
I think it's fine to share them, the grouping is logically consistent.

Can you add docs that these kinds also have the new `kind_tile_rank`, please?

Issue #1808, opened 2019-01-28
Don't include deprecated WOF neigbourhoods (stale dump)


nvkelso:
* **What did you see?** `Tenderloin` neighbourhood label is visible in NYC.
* **What did you expect to see?** This label shouldn't be visible, it's been [deprecated in WOF](https://spelunker.whosonfirst.org/id/85869725/) upstream from 2018.
* **What map location are you having problems with?** Tenderloin label near Herald Square in New York City.
    - http://tangrams.github.io/bubble-wrap/#16/40.7518/-73.9849
* **Screenshot?** 
![image](https://user-images.githubusercontent.com/853051/51857742-dfdb6d00-22e7-11e9-9da4-b0cdc9b5dadb.png)

nvkelso:
![image](https://user-images.githubusercontent.com/853051/51857876-334dbb00-22e8-11e9-8303-8bca4d60fe85.png)

zerebubuth:
We do [support the `edtf:deprecated` tag](https://github.com/tilezen/tilequeue/blob/146dfbe0fcac645ffaf9ea9c79bfe90831f8d1df/tilequeue/wof.py#L329), so the problem could either be a bug or that the WOF dump we're using is older than the deprecation was added to the record.

nvkelso:
Where is the WOF dump configured?

zerebubuth:
It's just a [static dump](https://github.com/tilezen/tileops/blob/d7365f947156001328f4cd26753965126c14ee43/import/import_planet.sh#L169-L171) that we download and restore as part of `tileops`. Headers on the dump suggest it was last updated around 12 Apr 2018, although it's possible it might be older than that.

nvkelso:
We should update the dump, then? Or really try for a "last good" distribution from WOF on DB setup?

Was it on a regular chron before, but then switching to RAWR tiles we just dump'd instead of doing a live load off the WOF distributions on DB setup?

See also:

- https://dist.whosonfirst.org/sqlite/
- https://dist.whosonfirst.org/sqlite/whosonfirst-data-latest.db.bz2

zerebubuth:
Yes, we used to have an automatic update process for WOF which ran on a schedule, the same way we were updating with diffs from OSM. Setting up a database from scratch was still a manual and difficult process for both OSM and WOF. When we moved to the global build, we automated the OSM part of that, but left the WOF setup for later.

[...]

PR #1868, opened 2019-04-12
Update WOF import pipeline.

zerebubuth:
Previously, we would look at the metadata CSVs, load all the referenced GeoJSON files over HTTP, parse them and put them into PostgreSQL. The hosted `pgdump` version was just dumped from a manual import of this data.

Because we're now doing global builds, this data has grown stale. This change adds a script to download and parse the WOF "bundles", which are `tar.gz` files containing the GeoJSON files - so we download 4 files instead of thousands. Instead of loading this into a database, we dump the data out as a SQL file, ready to be imported at database setup time.

The SQL dump is put into the `shapefiles.tar.gz` versioned static data asset, similar to Natural Earth and the OSMData land/water polygons. This means it's stable across releases, but we can update it as part of a regular asset rebuild.

Connects to #1808.

### DIFF
--- a/data/Makefile-prepare-data.jinja2
+++ b/data/Makefile-prepare-data.jinja2
@@ -11,8 +11,8 @@ upload: shapefiles
 
 shapefiles: shapefiles.tar.gz
 
-shapefiles.tar.gz: {{ tgt_shapefile_zips }}
-	tar czf shapefiles.tar.gz {{ tgt_shapefile_zips }}
+shapefiles.tar.gz: {{ tgt_shapefile_zips }} wof_snapshot.sql
+	tar czf shapefiles.tar.gz $^
 
 download: {{ src_shapefile_zips }}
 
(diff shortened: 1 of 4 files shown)


data/Makefile-prepare-data.jinja2
```diff
@@ -11,8 +11,8 @@ upload: shapefiles
 
 shapefiles: shapefiles.tar.gz
 
-shapefiles.tar.gz: {{ tgt_shapefile_zips }}
-	tar czf shapefiles.tar.gz {{ tgt_shapefile_zips }}
+shapefiles.tar.gz: {{ tgt_shapefile_zips }} wof_snapshot.sql
+	tar czf shapefiles.tar.gz $^
```
zerebubuth: NOTE: the `$^` line noise means everything after the colon on the rule line, in this case the `tgt_shapefile_zips` and the `wof_snapshot.sql`.

Issue #1873, opened 2019-04-22
Backfill kind_detail on aerodromes based on iata codes


nvkelso:
In https://github.com/tilezen/vector-datasource/issues/1277 we added `kind_detail` for aerodrome landuse (AOI), pois, and roads (runway) layers features. But it turns out `aerodrome:type` isn't very consistently filled in. We want to use this property to filter features at low zooms – there are a medium amount of not international airports at the low and mid-zooms that are large but not meaningful to show that early. 

* **What did you see?** I saw `public` on SFO airport south of San Francisco, California, USA.
* **What did you expect to see?** I expected to see `international`, based on it having a 3-char IATA code (SFO)
* **What map location are you having problems with?** 15/37.6201/-122.3775
* **Screenshot?** 

It is on the landuse:

![image](https://user-images.githubusercontent.com/853051/56513933-cb72cf00-64e8-11e9-9fa2-23c41a560df9.png)

It is passing to the POI:

![image](https://user-images.githubusercontent.com/853051/56513707-25bf6000-64e8-11e9-8d62-7b5da9ddd6c1.png)

I expect the runway lines to also have a `aerodrome_kind_detail` property via https://github.com/tilezen/vector-datasource/pull/1854:

![image](https://user-images.githubusercontent.com/853051/56513903-b72ed200-64e8-11e9-9207-c2862c3099ad.png)

zerebubuth:
> I expected to see `international`, based on it having a 3-char IATA code (SFO)

Would you also expect to see `international` for [Arrabury Airport](https://www.openstreetmap.org/node/1042093163)? Although it has a 3-char IATA code (AAB), it looks like it only serves a [remote cattle station](https://en.wikipedia.org/wiki/Arrabury) and I'd be quite surprised if it serves any international routes. Although it does have 2 runways, from aerial imagery it looks like neither are paved.

Perhaps we could use this [list of international airports](https://en.wikipedia.org/wiki/List_of_international_airports_by_country) to override `public` → `international`?

nvkelso:
Huh, that's odd! Maybe it's got an IATA code because it's an emergency landing strips for international flights that need a quick landing spot in the middle of the outback?

Rather than relying on an external list... what about allowing the OSM tags to drive this?

[...]

PR #1878, opened 2019-04-30
Add Wikidata queries to static assets

zerebubuth:
[Wikidata](https://www.wikidata.org/wiki/Wikidata:Main_Page) can be a very useful source of extra information to extend the data coming from OpenStreetMap. In this case, we're extracting the [number of passengers](https://www.wikidata.org/wiki/Property:P3872) through an airport from Wikidata and will use that to categorise airports into major and minor.

The data is added as a new table, `wikidata`, indexed by the ID (i.e: `'Q[0-9]*'`) and providing `tags` which will be merged into the tags coming from OSM. Since we might do multiple queries, and these might return different tags on overlapping sets of Wikidata IDs, we merge all the Wikidata query results together. This means we should be careful not to use the same attribute names for more than one query.

Note: There's another PR coming which will do the merging of `tags`, categorisation of the aerodrome feature and bringing that data through into to the output tile.

Connects to #1873.

### DIFF
--- a/data/bootstrap.py
+++ b/data/bootstrap.py
@@ -1,6 +1,7 @@
 from jinja2 import Environment
 from jinja2 import FileSystemLoader
 import yaml
+import urllib
 
 with open('assets.yaml') as fh:
     asset_cfg = yaml.load(fh)
@@ -24,6 +25,7 @@ tgt_shapefile_zips = []
 tgt_shapefile_shps = []
 tgt_shapefile_wildcards = []
 cfg_shapefiles = asset_cfg['shapefiles']
+cfg_wikidata_queries = asset_cfg['wikidata-queries']
 
 # track these separately, as the same URL/zipfile can contain several sets
 # of shapefiles.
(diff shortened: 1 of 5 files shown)


data/bootstrap.py
```diff
@@ -96,6 +98,14 @@
 
     shapefiles.append(shapefile)
 
+WIKIDATA_BASE_QUERY = 'https://query.wikidata.org/sparql'
```
nvkelso: Can we cache the results and include in the asset bundle? Their Sparql endpoint is fairly reliable, but we shouldn't needlessly hit it again and again.

bcamper:
This is a pretty sweet addition for future data joins! 👏

data/bootstrap.py
zerebubuth: Yup, that's what this is doing. The queries get run once and the results converted to a SQL file when the `shapefiles.tar.gz` bundle is being built, and any tile builds will just use the static SQL file from the bundle.

Issue #1074, opened 2016-09-22
roads layer min_zooms at zoom 5 are mix of 3, 4 and 5 instead of just 5.x


nvkelso:
tl;dr

We only query NE data starting at zoom 5, but the **roads** `min_zoom` also includes z3 and z4 values.

Let's do what we did for NE water labels (https://github.com/tilezen/vector-datasource/pull/1812) and clamp the exported min_zoom to the tile's min_zoom.

We should set:

- 3 to 5.0
- 4 to 5.1
- 5 to 5.2

===

This may be happening in more layers than roads...
- **What did you see?** Roads receive min_zoom values of 3 and 4 at zoom 5, but are only included in tiles starting at zoom 5.
- **What did you expect to see?** Either include the roads earlier, or cap the min_zoom of the roads at layer's start_zoom of 5.
- **What map location are you having problems with?** 5/41.162/-100.195
- **Screenshot?** 

<img width="901" alt="screen shot 2016-09-21 at 23 15 35" src="https://cloud.githubusercontent.com/assets/853051/18738523/5c6bb0f0-8051-11e6-8f43-74f2f523b830.png">

<img width="904" alt="screen shot 2016-09-21 at 23 15 21" src="https://cloud.githubusercontent.com/assets/853051/18738522/5c42ba74-8051-11e6-9e6e-0eb8dc1ab3e6.png">

nvkelso:
Boundaries sees this to an extent on the transition from NE 1:50m data to 1:10m data at zoom 5 when countries like Russia suddenly pop in.

nvkelso:
This also happens for national_park features in pois layer.

PR #1881, opened 2019-05-01
Adjust NE road min_zoom remapping at zoom 5

nvkelso:
Connects with #1074.

- [x] Update tests
- [x] ~~Update docs~~ bug fix


nvkelso:
@zerebubuth Looks like the last Circle integration test failure is from a data figure (shp) where I updated the DBF to include the `min_zoom`, but perhaps I also need to publish a new asset bundle to S3 for it to complete the loop?

nvkelso:
Ready for review!

Issue #1280, opened 2017-06-23
Inconsistency between Natural Earth and OSM road classes


nvkelso:
I think the general `kind: highway` part of this was intentional, but possibly not the `type: Major Highway` > `kind: highway`. Generally speaking that should have gotten `kind: major_road, kind_detail: trunk`.

- https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L139-L148

```
  - filter:
      scalerank: true
      featurecla: Road
      type: [Major Highway, Beltway, Bypass]
    min_zoom: scalerank
    table: ne
    output:
      <<: *ne_properties
      kind: highway
      kind_detail: trunk
```

zerebubuth:
Just to be sure I've understood this correctly: We want the input features with `type` `Major Highway` to result in output features with `kind: major_road`? This is kinda confusing, since "major highway" does suggest a highway.

Should we also remove the `kind_detail: trunk` from the `Beltway` and `Bypass` output (assuming they're keeping the `kind: highway`)?

nvkelso:
The intent was to better match the motorway versus trunk "highway" transition between zoom 7 and 8 and OSM and Natural Earth (see images and link below). 

But you're right – we need to ALSO combine this with Natural Earth's `expressway` = 1 tag to get the desired effect. Good catch!

So something more like this (assuming we import the `expressway` boolean from NE)?

```
  - filter:
      scalerank: true
      featurecla: Road
      type: Major Highway
      expressway: 1
    min_zoom: scalerank
    table: ne
    output:
      <<: *ne_properties
      kind: highway
      kind_detail: motorway
  - filter:
      scalerank: true
      featurecla: Road
      type: [Beltway, Bypass]
      expressway: 1
    min_zoom: scalerank
    table: ne
    output:
      <<: *ne_properties
      kind: highway
      kind_detail: motorway
  - filter:
      scalerank: true
      featurecla: Road
      type: Major Highway
    min_zoom: scalerank
    table: ne
    output:
      <<: *ne_properties
      kind: major_road
      kind_detail: trunk
  - filter:
      scalerank: true
      featurecla: Road
      type: [Beltway, Bypass]
    min_zoom: scalerank
    table: ne
    output:
      <<: *ne_properties
      kind: major_road
      kind_detail: trunk
```

Here's a view of northern California:

http://tangrams.github.io/bubble-wrap/#8/39.637/-123.054

[...]

PR #1882, opened 2019-05-01
NE trunk should be major road, not highway

nvkelso:
Connects with #1280.

- [x] Update tests
- [x] ~~Update docs~~ (bug fix)


nvkelso:
Ready for review.

PR #1889, opened 2019-05-03
Update to new version of static assets bundle including wikidata.

zerebubuth:
Uploaded a new version of the static assets bundle which includes the `wikidata_snapshot.sql`.

Added a line to `import-shapefiles.sh` to also import the wikidata snapshot. Switched from the more readable `--no-create` option to `touch` to the more cryptic `-c` because the latter is supported by both the GNU (Linux) and Mac OS versions of `touch`, whereas the long option is only supported by the GNU version.


nvkelso:
Test failure is fixed by https://github.com/tilezen/vector-datasource/pull/1890 (unrelated to this PR).

Issue #1810, opened 2019-01-31
Add OSM alternate viewpoints for country boundaries


nvkelso:
In https://github.com/tilezen/vector-datasource/pull/1809 we added schema support for alternate points of view in the boundaries layer, and added data from Natural Earth. 

This issue tackles adding data parsing from OpenStreetMap, where the data for this is sparse today. And we'll need to figure out how to build polygons out of them so we can still do the left/right boundary labels.

zerebubuth:
I've been playing around with the data for this for a while, and it doesn't look like the data is even in a suitable state today to build sensible-looking lines, let alone polygons for left/right labels. For example, this is the "on the ground" view from OpenStreetMap:

![OTG view](https://user-images.githubusercontent.com/271360/53898422-eb7d2a80-402f-11e9-8448-36ac25ffbe3d.png)

And this is what the data indicates India's view should be:

![India view](https://user-images.githubusercontent.com/271360/53898496-149dbb00-4030-11e9-9f4e-de689fd70f43.png)

And what the data says Pakistan's view should be:

![Pakistan view](https://user-images.githubusercontent.com/271360/53898540-2e3f0280-4030-11e9-81ef-b4bb624f356e.png)

And what the data says China's view should be:

![China view](https://user-images.githubusercontent.com/271360/53898607-54fd3900-4030-11e9-9412-2d345101f95e.png)

## What's being shown?

This is looking at the `disputed_by` tag on the **way** and the `claimed_by` tag on **relations** that it is part of. Weird, I know, but the `disputed_by` tag on the relations means that the _claim_ is disputed - and dropping the way on that basis leads to loads of gaps. If a way is part of a country boundary relation (i.e: would appear in the "on the ground" view) then it's flagged "on the ground".

* If a way is in a relation `claimed_by` a view, it's displayed in red.
* If way is tagged as `disputed_by` a view, it's not shown.
* If neither of the above and it's flagged as "on the ground" then display in green.

## What's going on with the extra random lines?

Most of the line-of-control between India and Pakistan is tagged as `disputed_by=IN;PK`, but some segments aren't. Since the line-of-control is the "on the ground" view of the border between the two countries, these untagged segments get displayed in green even though they are disconnected from other parts of the border. Looks like a simple case of bad tagging - should be easy to correct.

Some other problems appear to stem from typos, for example this [alleged dispute between India and Switzerland (fix: `s/CH/CN/`)](https://www.openstreetmap.org/way/484553387).

[...]

PR #1895, opened 2019-05-13
Alternate viewpoints (v0.1)

zerebubuth:
Added support for alternate viewpoints of country borders.

Alternate viewpoints are used where two or more countries have different opinions about where their borders are. In many cases, the governments of these countries require that maps display their particular viewpoint, so it's very practically useful to have data in the tiles to support that.

The data output in the tiles will have the usual `kind` for _de-facto_ boundaries, but additional `kind:xx` data for the viewpoint of country XX (ISO 3166-1 alpha-2 code). For generally-recognised boundaries that XX thinks are interior to its territory, we set `kind = country`, `kind:xx = unrecognized_country`. For generally _not_ recognised boundaries that XX thinks are its borders, we set `kind = unrecognized_country` and `kind:xx = country`. Therefore, it's possible to recover the status of any given border by a similar method we use for translations; use `kind:yy` if it exists for the viewpoint YY, falling back to the regular `kind` if not.

This is done by looking at:

1. **Claim relations**. These give us the extra geometry which is generally unrecognised, but a country boundary for the viewpoint in the `claimed_by` tag or other viewpoints listed in the `recognized_by` tag. (NOTE: `recognized_by` support is still a TODO).
2. **Disputed ways**. Ways tagged `dispute(d)=yes` and with a list of viewpoints in `disputed_by` will "erase" those borders for the listed viewpoints. This is done by intersecting a buffered version of the ways with the polygonally-derived country borders.

Until now, we had only been including boundary features from the `planet_osm_polygon` table and nothing at all from the `planet_osm_line` table. However, the disputes and claims are linear features, so we needed to include those.

However, because `osm2pgsql` creates both linear and polygonal features for boundaries (ways and relations), then we must exclude all the linear features that are in `planet_osm_line` _except_ for the disputes and claims. This is further complicated because we change the geometry type during the query, dropping the polygon for its (oriented) boundary.
    
The way I've done this is to include a new flag, `mz_boundary_from_polygon`, which basically means "this was a polygon, but we transformed it", and this allows us to write YAML that works. However, it's kind of a horrible hack and I mildly regret doing it. A better alternative might have been to leave the polygons in the query (appropriately clipped to an expanded bounding box) and do the boundary transform at a later stage in the stack - however this might be left for a rainy day issue.

Things to note:

* The disputed ways must be part of the generally-recognised country borders. It's not possible to dispute a claim - nor is it necessary, since claims are only "visible" to the `claimed_by` and `recognized_by` viewpoints.
* Neither disputes nor claims go through the left/right matching process, so the `name` property on them is whatever the `name` tag was on the claim relation or disputed way.
* Since we don't build the claimed polygon, we aren't able to figure out what other features (internal borders, locality / place points, etc...) are in the claimed area, which limits our ability to automatically hide or adjust the `kind` of features in that area according to the claimant viewpoint.

Still TODO:

* Other admin boundaries (region, county, etc...)
* Support for `recognized_by`.

Connects to #1810.


nvkelso:
Nice! I see `mz_boundary_from_polygon` in a lot of the tests. Is that to make sure things are wired up for QA and then we'd drop that property before cutting the release?

integration-test/1482-maritime_boundary-buffered_land.py
```diff
@@ -4,54 +4,6 @@
 # Adds tests for OSM features (but not NE features)
 class MaritimeBoundary(FixtureTest):
 
-    def test_usa_canada_country_boundary(self):
```
nvkelso: Are these tests being replaced with something else? I see these IDs testing the `sort_key`, but not the `kind` later on in the test section?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -0,0 +1,153 @@
...
+                'kind': 'country',
+                'name:left': 'XX',
+                'name:right': 'YY',
+            })
+
+        # should get a section recognised only by XX
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'kind': 'unrecognized_country',
+                'kind:xx': 'country',
```
nvkelso: Please also add a `name` test?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -0,0 +1,153 @@
...
+            })
+
+        # should get a section recognised _except_ by XX
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'kind': 'country',
+                'kind:xx': 'unrecognized_country',
```
nvkelso: Please also add a `name` test?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -0,0 +1,153 @@
...
+                'dispute': 'yes',
+                'disputed_by': 'AA',
+                'name': 'BB claim',  # note: also CC claim?!
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'kind': 'unrecognized_country',
+                'kind:bb': 'country',
```
nvkelso: clarifying: do we document that we always normalize the 2-char "country" code to lowercase?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -0,0 +1,153 @@
...
+                'kind:cc': 'country',
+                # 'kind:dd': 'country',
```
nvkelso: TODO: This will be uncommented out once `recognized_by` is plumbed thru.

integration-test/1810-alternate-viewpoints.py
```diff
@@ -0,0 +1,153 @@
...
+            dsl.way(4, dsl.tile_diagonal(z, x, y), {
+                'dispute': 'yes',
+                'disputed_by': 'AA',
+                'name': 'BB claim',  # note: also CC claim?!
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'kind': 'unrecognized_country',
```
nvkelso: I wonder, if like the name fan out we did in https://github.com/tilezen/vector-datasource/pull/1454, if we should also indicate `kind:aa` and `kind:dd` = `unrecognized_claim`?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -0,0 +1,153 @@
...
+            dsl.way(3, dsl.tile_diagonal(z, x, y), {
+                'admin_level': '3',
+                'boundary': 'administrative',
+                'name': 'Region Name',
+                'type': 'boundary',
+            }),
+            dsl.way(4, dsl.tile_diagonal(z, x, y), {
+                'dispute': 'yes',
+                'disputed_by': 'AA',
+                'name': 'BB claim',  # note: also CC claim?!
```
nvkelso: Do we test for this and the later ways here?

integration-test/841-normalize-boundaries-kind.py
```diff
@@ -8,80 +8,233 @@ def test_aboriginal_lands_protected_area(self):
...
+        import dsl
+
+        z, x, y = 16, 10417, 25370
+
         # Relation: United States of America
-        self.load_fixtures([
-            'https://www.openstreetmap.org/relation/148838',
-        ], clip=self.tile_bbox(16, 10417, 25370, padding=2))
+        self.generate_fixtures(
+            # https://www.openstreetmap.org/relation/148838
```
nvkelso: This'll speed up the tests, lol

test/test_meta.py
```diff
@@ -154,25 +154,51 @@ def setUpClass(cls):
...
+        shape = LineString([(0, 0), (1, 1)])
+        props = {
+            'boundary': 'administrative',
+            'boundary:type': 'aboriginal_lands',
+            'admin_level': '2',
+        }
+        meta = make_test_metadata()
+        out_props = self.boundaries.fn(shape, props, None, meta)
+
+        # without the hack, shouldn't match (i.e: as if it were from
```
nvkelso: Please explain for our future selves the nature of this hack inline here in more detail (even if just `boundaries point-of-view hack`?

vectordatasource/transform.py
```diff
@@ -8906,6 +8927,132 @@ def remap_viewpoint_kinds(shape, props, fid, zoom):
...
+    from re import split
+
+    countries = []
+    candidates = split('[,;]', value)
+
+    for candidate in candidates:
+        # should have an ISO 3166-1 alpha-2 code, so should be 2 ASCII
+        # latin characters.
+        candidate = candidate.strip().lower()
+        if match('[a-z][a-z]', candidate):
```
nvkelso: Natural Earth has an `iso` "country code", which is used for disputed boundaries too. Special case that?

vectordatasource/transform.py
```diff
@@ -8906,6 +8927,132 @@ def remap_viewpoint_kinds(shape, props, fid, zoom):
...
+    zoom = ctx.nominal_zoom
+
+    if zoom < start_zoom or \
+       (end_zoom is not None and zoom >= end_zoom):
+        return None
+
+    # we buffer out the shape by a small amount so that we're more likely to
+    # get a clean cut against the boundary line. with lines intersecting
+    # against lines, we often see a sort of "dashed pattern" where numerical
+    # imprecision has meant two lines don't quite intersect.
```
nvkelso: add: `, which is solved by applying the small buffer.`?

vectordatasource/transform.py
```diff
@@ -8906,6 +8927,132 @@ def remap_viewpoint_kinds(shape, props, fid, zoom):
...
+    #
+    # tolerance for zoom is the length of 1px at 256px per tile, so we can take
+    # a fraction of that to get sub-pixel alignment.
+    buffer_distance = 0.1 * tolerance_for_zoom(zoom)
+
+    # first, separate out the dispute mask geometries
+    masks = []
+    other_features = []
```
nvkelso: `uncontested_features`? By the time I get to `if not masks:` below I'm not sure what `other_features` means, the name change would clarify that.

nvkelso: or `default_features` or? Since we're then applying the masks to the "other" features.

yaml/boundaries.yaml
```diff
@@ -95,16 +112,47 @@ filters:
       <<: *output_properties
       kind: {col: kind}
 
+  # OSM disputed boundaries
+  #
+  # NOTE: these come _first_ so that they match in preference to regular
+  # country borders. (this assumes disputed_by is never set on a whole
+  # country boundary -- perhaps true?)
```
nvkelso: These 2 are technically entirely disputed_by their neighbours, but instead tagged with claimed_by? How should this be setup / changed (data or processing)?

 **Taiwan:** https://www.openstreetmap.org/relation/449220
**Western Sahara:** https://www.openstreetmap.org/relation/5441968

vectordatasource/transform.py
```diff
@@ -8906,6 +8927,132 @@ def remap_viewpoint_kinds(shape, props, fid, zoom):
...
+    Unpack OSM "claimed_by" list into viewpoint kinds.
+
+    For example; "claimed_by=AA,BB,CC" should become "kind:aa=country,
+    kind:bb=country, kind:cc=country" (or region, etc... as appropriate for
+    the main kind, which should be "unrecognized_TYPE".
+    """
+
+    prefix = 'unrecognized_'
+    kind = props.get('kind')
+    claimed_by = props.get('claimed_by')
```
nvkelso: We don't want to export `claimed_by` in the properties, only adding new `kind:*` properties.

Where do we drop `claimed_by` property?

vectordatasource/transform.py
```diff
@@ -8906,6 +8927,132 @@ def remap_viewpoint_kinds(shape, props, fid, zoom):
...
+                cut_shape = shape.intersection(mask_shape)
+                cut_shape = _filter_geom_types(cut_shape, _LINE_DIMENSION)
+
+                shape = shape.difference(mask_shape)
+                shape = _filter_geom_types(shape, _LINE_DIMENSION)
+
+                if not cut_shape.is_empty:
+                    new_props = props.copy()
+                    for disputant in disputants:
+                        new_props['kind:' + disputant] = 'unrecognized_country'
```
nvkelso: We don't want to export `disputed_by` in the properties, only adding new `kind:*` properties.

Where do we drop `disputed_by` property?

nvkelso:
Great progress, this approach seems workable.

I left comments above, will review again once `Other admin boundaries (region, county, etc...)` and `Support for recognized_by` are added.

integration-test/1482-maritime_boundary-buffered_land.py
zerebubuth: I couldn't tell what they were testing, other than the maritime / non-maritime status that was replicated in the generative tests below. Since, as part of adapting the tests for the new line / polygon status, I was converting stuff to generative testing anyway, it seemed unnecessary to convert these when generative equivalents already existed.

zerebubuth:
> I see `mz_boundary_from_polygon` in a lot of the tests. Is that to make sure things are wired up for QA and then we'd drop that property before cutting the release?

No, it's an internal flag that's used to indicate a feature is the boundary line of an admin polygon, rather than from the lines table. I mentioned it in the PR description:

> The way I've done this is to include a new flag, `mz_boundary_from_polygon`, which basically means "this was a polygon, but we transformed it", and this allows us to write YAML that works. However, it's kind of a horrible hack and I mildly regret doing it.

Note that we drop all `mz_*` properties from features before formatting the tile, so this flag never appears in the output.

integration-test/1810-alternate-viewpoints.py
zerebubuth: Added `name` test in 67432dc4 (and also fixed bug in test).

integration-test/1810-alternate-viewpoints.py
zerebubuth: Added `name` test in 67432dc4 (and also fixed bug in test).

integration-test/1810-alternate-viewpoints.py
zerebubuth: Ooops, missed that, sorry. Updated the documentation in 4722e43e.

integration-test/1810-alternate-viewpoints.py
zerebubuth: I think the situation is a little bit different for boundaries. For the names, we were removing data that had come as input, which broke multi-level fallback. Single-level fallback worked just fine.

Is a situation in which we'd need to do multi-level fallback? In this test example, would we ever do a fallback sequence `kind:aa`, `kind:bb`, `kind`?

The `disputed_by = AA;DD` on the `claim` means that `AA` and `DD` dispute the claim, not whether they do or do not agree with the locations of the borders on the ground - I've assumed any viewpoint _not_ listed in `claimed_by` or `recognized_by` doesn't agree (and will default to `kind: unrecognized_*`).

When I first prototyped this, projecting the `disputed_by` from relations onto constituent ways produced borders which were `disputed_by` pretty much everyone and disappeared from the map. So I'd be very wary of using the `disputed_by` information from `claim` relations.

Is there a real-world example of where it's useful to have a multi-level fallback projected from the `claim` relation?

integration-test/1810-alternate-viewpoints.py
zerebubuth: I think it's already a case covered by the previous test but, just to be sure, I've added a further assertion in 68cbdbf1.

test/test_meta.py
zerebubuth: Good point. Added more comprehensive comment in 095be669.

vectordatasource/transform.py
zerebubuth: Sure. Added in 43fe2474.

vectordatasource/transform.py
zerebubuth: Re-worded the comment in 0be683fe, hope it's clearer now.

vectordatasource/transform.py
zerebubuth: I've refactored the whole thing in ed6f1363 and fc2a7ad6. Hopefully it's clearer now. I ended up splitting the features into `masks`, `boundaries` and putting anything non-matching straight into `new_features` to be returned, so the variable names should match up with their contents.

yaml/boundaries.yaml
zerebubuth: I think I was mistaken when I wrote that comment. I've added a test for this in c2efb38b, and it works, so I removed the comment and replaced it with better tagging recommendations.

For Western Sahara, the tagging looks good to me. I think :crossed_fingers: that it should work the next time we do a build.

For Taiwan, it's missing:

* A `boundary=claim` relation with `claimed_by=CN` and `admin_level=2`. I had a look around, and couldn't find one. It should contain the same ways as [relation 449220](https://www.openstreetmap.org/relation/449220), the country border.
* The ways making up the country border need `disputed=yes` and `disputed_by=CN` on them. I didn't audit all the ways, but a few tested at random (e.g: [this one](https://www.openstreetmap.org/way/53100715)) were missing those tags.

vectordatasource/transform.py
zerebubuth: Ah, we don't drop it - yet! Fixed in b7807eef.

vectordatasource/transform.py
zerebubuth: Ah, we don't drop it - yet! Fixed in b7807eef.

yaml/boundaries.yaml
nvkelso: Agree I need to add OSM tagging for Taiwan.

nvkelso:
I'd forgotten about the general drop all `mz_*` properties logic, makes sense now.

integration-test/1810-alternate-viewpoints.py
nvkelso: > Is a situation in which we'd need to do multi-level fallback? 

Generally countries are "aligned" with some another country in that they have a few opinions about their immediate neighbours, but lean on either ISO or the aligned country's foreign policy.

In practice someone developing in country X shouldn't need to know if Tilezen does or doesn't have data for that country's worldview (POV), but they should still indicate that as the primary POV, and ideally list out ISO or the other country Y (which we likely have in the data) as fallback, before falling back to defacto kind values.

> In this test example, would we ever do a fallback sequence kind:aa, kind:bb, kind?

I think it's more likely it'd be `kind:xx` (primary), `kind:aa` (secondary)

To be clear, I think the fallback only happens client side not server side.

> I wonder, if like the name fan out we did in #1454, if we should also indicate kind:aa and kind:dd = unrecognized_claim, since it's meaningful those countries have an opinion (versus generic default)?

Is that possible in current data structure, or would it cause problems like you described? I'm open to not doing the fan out.

zerebubuth: OK, cool. I added a test for this in 74b4b7fc, although the logic is getting a little bit complicated now!

zerebubuth:
> we can continue more work in additional PRs to make it easier to review?

Yes, sounds like a good idea!

Issue #1840, opened 2019-02-15
Add more kind:* POV from Natural Earth for low zooms and places layer


nvkelso:
_Originally posted by @nvkelso in https://github.com/tilezen/vector-datasource/issues/1552#issuecomment-463516835_

There's also work to add this in from the Natural Earth 1:50M and 1:110M scale sets (we've only added from 1:10M set in v1.7)... and from `populated_places` and `states_provinces` as some country and region capitals aren't recognized, and some regions aren't recognized.

This issue tracks integrating that data.

nvkelso:
Sister issue to OSM data in #1810.

nvkelso:
Nearterm, I'll prototype a sample data change in the 1:10m populated_placeas and region boundary lines, and same samples in the 1:110m and 1:50m scale sets to unblock Tilezen work. Then we'll do one other asset build once the final NE data is ready.

nvkelso:
Prototype data:

- [ne_v5.0.0-pre3-boundaries-pov.zip](https://github.com/tilezen/vector-datasource/files/3083549/ne_v5.0.0-pre3-boundaries-pov.zip)

Notes: 

- All themes included here have the additional POV columns with the same naming schema.
- Some themes, like 1:50m and 1:110m state and province boundary lines include the schema but none of the line features are disputed so those columns are entirely blank. But there are requests to add that data for India and China so let's plumb it in here to future proof.
- The 1:10m state and province boundary lines _does_ include sample data for India, China.
- The populated places in Taiwan include the placetype swapping (eg country capital goes to region capital, and region capitals go to generic localities)
- The 1:50m and 1:110m international boundaries files have been updated with sample data.
- More work remains to flesh out the column values completely, but these are good sample to unblock development.

Flagging @zerebubuth for integration work.

nvkelso:
Noting we should rename **kind** `unrecognized` to `unrecognized_country` so we can also support `uncrecognized_region`. 

Under our SemVer we can rename a recently added kind as a Patch release, but let's just bundle it in this minor update.

nvkelso:
I've done more data work and need to post a new preview ZIP.

nvkelso:
Here's v5.0.0-pre-4 with more detail filled out (not just samples) for the other themes, and some fixes in the 1:10m themes:

[...]

PR #1896, opened 2019-05-17
Add more viewpoint data from Natural Earth

zerebubuth:
Adds viewpoint data from the v5.0-pre4 version. This extends disputed information to admin-1 (state) boundaries and populated places. In the case of the boundary lines, the data is the same (`kind:xx` overrides) as for previous 1:10m disputed boundary work.
    
In the case of populated places, the dispute data is used to set overrides for the `country_capital` and `region_capital` properties, i.e: `country_capital:xx=True/False`.

Note: The new v5.0-pre4 data means a new asset bundle version `20190517` was created.

Connects to #1840.

### DIFF
--- a/data/assets.yaml
+++ b/data/assets.yaml
@@ -1,5 +1,5 @@
 bucket: tilezen-assets
-datestamp: 20190503
+datestamp: 20190517
 
 shapefiles:
 
@@ -85,19 +85,19 @@ shapefiles:
     prj: 4326
     tile: true
 
-  - name: ne_10m_populated_places
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_populated_places.zip
-    prj: 4326
-
-  - name: ne_110m_admin_0_boundary_lines_land
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
-    prj: 4326
-
-  - name: ne_50m_admin_0_boundary_lines_land
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/50m/cultural/ne_50m_admin_0_boundary_lines_land.zip
-    prj: 4326
-
-  # TEMPORARY DISABLE TO PULL IN v5-PRE2 DATA
+  # TEMPORARY DISABLE TO PULL IN v5-PRE4 DATA
+  #
+  # - name: ne_10m_populated_places
+  #   url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_populated_places.zip
+  #   prj: 4326
+  #
+  # - name: ne_110m_admin_0_boundary_lines_land
+  #   url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_boundary_lines_land.zip
+  #   prj: 4326
+  #
+  # - name: ne_50m_admin_0_boundary_lines_land
+  #   url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/50m/cultural/ne_50m_admin_0_boundary_lines_land.zip
+  #   prj: 4326
   #
   #- name: ne_10m_admin_0_boundary_lines_land
   #  url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_boundary_lines_land.zip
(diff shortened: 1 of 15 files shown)


zerebubuth:
Build failure was due to `botocore` requiring an old version of `urllib3` which is no longer installed. While we wait for `botocore` to relax their dependencies (see https://github.com/boto/botocore/issues/1733), we need to pin `urllib3` to an older version. As far as I can tell, `urllib3==1.24.2` is OK and doesn't seem to be subject to any disclosed vulnerability. Should be fixed in 1797a59c.

data/assets.yaml
```diff
@@ -85,54 +85,90 @@ shapefiles:
...
+  #   url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/50m/cultural/ne_50m_admin_1_states_provinces_lines.zip
+  #   prj: 4326
+  #   shapefile-name: ne_50m_admin_1_states_provinces_lines.shp
+
+  # - name: ne_10m_admin_1_states_provinces_lines
+  #   url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_1_states_provinces_lines.zip
+  #   prj: 4326
+
   - name: ne_10m_populated_places
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_populated_places.zip
```
nvkelso: Should we keep these source URLs but commented out (like above) so it's easy to add them back in once NE v5 is released?

docs/layers.md
```diff
@@ -740,7 +740,9 @@ Places with `kind` values of `continent`, `country`, with others added starting
 #### Place properties (common optional):
 
 * `country_capital`: a `true` value normalizes values between OpenStreetMap and Natural Earth for kinds of `Admin-0 capital`, `Admin-0 capital alt`, and `Admin-0 region capital`.
+* `country_capital:xx`: when present, either `true` or `false` to override the `country_capital` value for XX's viewpoint.
```
nvkelso: [docs] We should reference it uses the same points-of-view as `kind:xx`.

docs/layers.md
```diff
@@ -740,7 +740,9 @@ Places with `kind` values of `continent`, `country`, with others added starting
...
 * `region_capital`: a `true` value normalizes values between OpenStreetMap and Natural Earth for kinds of `Admin-1 capital` and `Admin-1 region capital`.
+* `region_capital:xx`: when present, either `true` or `false` to override the `region_capital` value for XX's viewpoint.
```
nvkelso: [docs] We should reference it uses the same points-of-view as `kind:xx`.

docs/layers.md
zerebubuth: Good idea! Fixed in 13257b22.

data/assets.yaml
zerebubuth: Yup, I kept the [whole block, but commented out](https://github.com/tilezen/vector-datasource/pull/1896/files#diff-211ab6b26f210fbe082be63a883f0a0cR91). That way it's easy to switch out the whole thing and comment out the new stuff, without having to comment / uncomment a bunch of lines individually.

integration-test/1810-alternate-viewpoints.py
```diff
@@ -264,3 +264,130 @@ def test_whole_claim(self):
...
+                'featurecla': default,
+                'fclass_iso': dispute,
+                'scalerank': 4,
+                'min_zoom': 4,
+                'source': 'naturalearthdata.com',
+            }),
+        )
+
+        expected['kind'] = 'locality'
+        self.assert_has_feature(z, x, y, 'places', expected)
```
nvkelso: I don't understand this test, can you walk me thru it?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -264,3 +264,130 @@ def test_whole_claim(self):
...
+        )
+
+    def test_country_region(self):
+        # region override disables country_capital and adds region_capital.
+        self._check(
+            'Admin-0 capital',
+            'Admin-1 capital',
+            {
+                'country_capital:iso': type(False),
+                'region_capital:iso': type(True),
```
nvkelso: Thanks for adding that.

integration-test/1810-alternate-viewpoints.py
nvkelso: Ah, I understand more now. Might be worth a comment above 298 `def _check` explaining the pattern since all the later tests use it.

integration-test/1810-alternate-viewpoints.py
```diff
@@ -264,3 +264,130 @@ def test_whole_claim(self):
...
+                'country_capital:cn': type(False),
+                'region_capital:cn': type(True),
+            })
+
+    def _check(self, default, dispute, expected):
+        import dsl
+
+        z, x, y = 16, 0, 0
+
+        self.generate_fixtures(
```
nvkelso: Why is this fixture needed?

queries/ne-boundaries-110m.jinja2
```diff
@@ -1,4 +1,4 @@
 {% extends 'ne.jinja2' %}
 {% block query %}
-{{ ne_boundaries_query(False, False, 'ne_110m_admin_0_boundary_lines_land') }}
+{{ ne_boundaries_query(False, True, 'ne_110m_admin_0_boundary_lines_land') }}
```
nvkelso: Good, we are updating the 110m file params here.

yaml/places.yaml
```diff
@@ -39,6 +39,42 @@ global:
         - [ 9, 9 ]
       default: 10
 
+  - &ne_country_capital [Admin-0 region capital, Admin-0 capital alt, Admin-0 capital]
```
nvkelso: Nit: format the same as next row?

`Admin-0 capital, Admin-0 capital alt, Admin-0 region capital`

yaml/places.yaml
```diff
@@ -39,6 +39,42 @@ global:
...
+  - &ne_region_capital [Admin-1 capital, Admin-1 region capital]
+
+  - &alternate_fclass
+    fclass_iso: { col: fclass_iso }
+    fclass_us: { col: fclass_us }
```
nvkelso: after ISO these should probably sort a-Z? 

(I regret not doing that in the SHP's DBF)

yaml/places.yaml
```diff
@@ -156,26 +192,29 @@ filters:
...
       <<: *output_properties
+      <<: *alternate_fclass
       kind: locality
       region_capital: true
     table: ne
   - filter: {featurecla: Populated place, scalerank: true}
     min_zoom: *ne_places_min_zoom
     output:
       <<: *output_properties
+      <<: *alternate_fclass
```
nvkelso: Good catch

data/assets.yaml
nvkelso: Ah, Github.com diff display not winning here. I see it now, tnx.

integration-test/1810-alternate-viewpoints.py
zerebubuth: Good idea, added in 4c43b1c2.

yaml/places.yaml
zerebubuth: Fixed in 04c96b34.

yaml/places.yaml
zerebubuth: Yup, makes them easier to read / find stuff. Fixed in c6dae6d7.

zerebubuth:
> We should plumb support thru for that (partly done?), but not critical for a staging build.

Should be done already. All the boundaries tables are set up for `fclass_*` processing. Once the data is non-empty, we should see that the next time we update the assets bundle and rebuild.

data/assets.yaml
zerebubuth: Yea, sorry. I guess that's the advantage of changing each line individually; the diff is much easier to read. :bowing_man:

nvkelso:
LGTM

PR #1897, opened 2019-05-24
Alternate viewpoints regions

zerebubuth:
This adds a few extra features missing from #1895:

1. Adds support for disputed/claimed region and county boundaries. This is similar to country boundary support, but with a different `admin_level`. Note that disputes will mark all intersecting boundaries as `unrecognized_*` in the disputants' viewpoints. If we find we need to selectively mark intersecting boundaries then we'll need to figure out a mechanism for indicating that in the data.
2. Adds support for `recognized_by` on claim relations, for where one viewpoint wants to include a claim made by a 3rd party. `recognized_by` should be a comma or semicolon delimited list of viewpoints, same as `disputed_by`.
3. Joins the Natural Earth `populated_places` data via Wikidata ID to include `fclass_*` on place points, which allows us to derive `country_capital:xx` and `region_capital:xx` overrides from NE data rather than (controversial) OSM tagging.

Connects to #1810 and #1840.


integration-test/1810-alternate-viewpoints.py
```diff
@@ -394,3 +393,147 @@ def test_none_none(self):
...
+                'type': 'boundary',
+                'mz_boundary_from_polygon': True,  # need this for hack
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'kind': 'region',
+                'kind:xb': 'unrecognized_region',
+                'name': 'XA region 2',
```
nvkelso: Why isn't this name: `XA internal region border` from the first way? 

Do these need to be linked up in relation or are the loose ways fine? Seems like we sometimes enforce the ways + relations in tests, but not here?

integration-test/1810-alternate-viewpoints.py
```diff
@@ -394,3 +393,147 @@ def test_none_none(self):
...
+                'type': 'boundary',
+                'mz_boundary_from_polygon': True,  # need this for hack
+            }),
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'boundaries', {
+                'kind': 'county',
+                'kind:xb': 'unrecognized_county',
+                'name': 'XA county 2',
```
nvkelso: Because of the different admin levels here, it looks like it takes the disputed from the `county` level (4), but the name from the localadmin level (6). Should it be the level 4 name instead?

vectordatasource/transform.py
```diff
@@ -8959,23 +8959,38 @@ def unpack_viewpoint_claims(shape, props, fid, zoom):
     For example; "claimed_by=AA,BB,CC" should become "kind:aa=country,
     kind:bb=country, kind:cc=country" (or region, etc... as appropriate for
     the main kind, which should be "unrecognized_TYPE".
+
+    Additionally, "recognized_by=XX,YY,ZZ" indicates that these viewpoints,
```
nvkelso: Most the data is `;` delimitated, should our examples (here and above) be as well?

vectordatasource/transform.py
```diff
@@ -8959,23 +8959,38 @@ def unpack_viewpoint_claims(shape, props, fid, zoom):
...
+        if recognized_by:
+            for viewpoint in _list_of_countries(recognized_by):
+                props['kind:' + viewpoint] = claimed_kind
+
     return (shape, props, fid)
 
 
 class _DisputeMasks(object):
     """
+    Creates a "mask" of polygons by buffering disputed border lines and
```
nvkelso: Thanks for adding this mini readme!

data/functions.sql
```diff
@@ -1100,14 +1100,45 @@ CREATE OR REPLACE FUNCTION extra_wikidata_properties(wikidata_id text)
...
+      'fclass_ps', p.fclass_ps,
+      'fclass_pt', p.fclass_pt,
+      'fclass_ru', p.fclass_ru,
+      'fclass_sa', p.fclass_sa,
+      'fclass_se', p.fclass_se,
+      'fclass_tr', p.fclass_tr,
+      'fclass_tw', p.fclass_tw,
+      'fclass_us', p.fclass_us,
+      'fclass_vn', p.fclass_vn
+    ]) INTO ne_tags FROM ne_10m_populated_places p
```
nvkelso: nit, in the raw_tiles PR we call `p` instead `pp`, and also proposed calling it `ne_pp`.

integration-test/1810-alternate-viewpoints.py
zerebubuth: We take the name from the relation (boundary or claim), rather than the way (dispute).

The `dsl.way` call is misleading; the generative fixtures used in the tests are injected as if they're the results of a query, and the boundary relation processing is done in the query, so although boundaries are relations (and polygons), the code post-query only sees line strings as if they came from ways.

Apologies - the whole query / RAWR tile input stack has become a mess as we've squeezed more features into it, so now things are not always what they appear (or what the function name suggests).

integration-test/1810-alternate-viewpoints.py
zerebubuth: `admin_level=6` is for counties. We use `admin_level=4` for regions.

data/functions.sql
zerebubuth: Changed in bb645779.

vectordatasource/transform.py
zerebubuth: Good point! I switched it to `;` in 2fa74855.

PR #1904, opened 2019-06-11
Add support for NaturalEarth region disputes & claims 

zerebubuth:
We added support for country disputes & claims in #1895 & #1896, and disputes about the status of capital cities in #1897. However, the support for region boundary lines wasn't fully implemented. This PR adds tests and changes to the kind remapping rules to support disputes & claims on regions tagged as NaturalEarth `featureclass` / `fclass_*` values.

Also, changed `kind: unrecognized` to `kind: unrecognized_country`, matching changes we'd previously made for OSM boundaries.

Connects to #1840 & #1810.

### DIFF
--- a/yaml/boundaries.yaml
+++ b/yaml/boundaries.yaml
@@ -274,6 +274,16 @@ filters:
   - filter:
       featurecla:
         - Unrecognized Admin-1 region boundary
+    min_zoom: *ne_region_boundaries_min_zoom
+    extra_columns: [scalerank]
+    output:
+      <<: *output_properties
+      <<: *ne_localized_kind_properties
+      kind: unrecognized_macroregion
+      kind_detail: '4'
+    table: ne
+  - filter:
+      featurecla:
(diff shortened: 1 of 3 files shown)


yaml/boundaries.yaml
```diff
@@ -264,6 +267,21 @@ filters:
     extra_columns: [scalerank]
     output:
       <<: *output_properties
+      <<: *ne_localized_kind_properties
       kind: region
       kind_detail: '4'
     table: ne
+  - filter:
+      featurecla:
+        - Unrecognized Admin-1 region boundary
```
nvkelso: `Unrecognized Admin-1 region boundary` should output `unrecognized_macroregion` instead?

(Not that the data exists for that right now, but for correctness's sake.)

zerebubuth: Good point. Fixed in 84c00c04. (Although we don't seem to have support for `macroregion` from OSM boundaries either, so there might be problems at the transition.)

Issue #1906, opened 2019-06-12
Missing capital POV from NE for region at high zooms


nvkelso:
* **What did you see?** Region capitals in Taiwan aren't getting any POV properties at high zooms.
* **What did you expect to see?** Those should be included by joining from OSM to NE via Wikidata_ID values. That seems to be working for country capitals, but not region capitals.
* **What map location are you having problems with?** Taiwan
* **Screenshot?** 

Low zoom from Natural Earth:
<img width="299" alt="Screen Shot 2019-06-12 at 16 51 38" src="https://user-images.githubusercontent.com/853051/59393948-9dfd1180-8d32-11e9-9d98-56729324cb1e.png">

High zoom from OSM missing NE POV properties:
<img width="293" alt="Screen Shot 2019-06-12 at 16 54 03" src="https://user-images.githubusercontent.com/853051/59393966-b53bff00-8d32-11e9-9c33-de2cf01b95d8.png">

nvkelso:
@zerebubuth can you take a look, please?

nvkelso:
OSM node: 

- https://www.openstreetmap.org/node/60655918
    - doesn't have `state_capital` property
    - does have Relation Taichung (2921154) (as `admin_centre`)

travisgrigsby:
@nvkelso can you list which of the properties that aren't making it in to the high zoom should be?  
I see `region_capital`, `region_capital:cn` that seem obviously missing and related, but then I also see a lot of name differences due to the different data sources providing the properties.  

One that may be important is, in the low zoom, the `name:zh` field has a value with 1 less char than in the high zoom.  Is this relevant? Anything else?

travisgrigsby:
A lot of this can be modeled off the description here: #2020 
Basically, add a join query like in `functions.sql`, call it to pull the data into the properties that vector datasource sees, then make sure they're exported (probably in `places.yaml`)

travisgrigsby:
Nm looks like this was already committed.

peitili:
does that mean we can close this issue?

travisgrigsby:
not sure yet!

nvkelso:
Yes, we can close this issue. It should have been closed in 2019 via https://github.com/tilezen/raw_tiles/pull/34 and https://github.com/tilezen/vector-datasource/pull/1907.

PR #1907, opened 2019-06-13
Backfill region capital status from Natural Earth

zerebubuth:
To figure out if a place (`city` or `town`) is a region or state capital, we have previously only looked at the OSM `state_capital` tag. However, some region capitals aren't tagged as such in OSM and this leads to changes across the boundary between Natural Earth and OSM data.

This patch back-fills the `region_capital` status from Natural Earth, so that if _either_ the OSM `state_capital` or the Natural Earth `featurecla` have appropriate values, then the `region_capital` flag will be set.

Connects to #1906.

### DIFF
--- a/yaml/places.yaml
+++ b/yaml/places.yaml
@@ -1,4 +1,7 @@
 global:
+  - &ne_country_capital [Admin-0 capital, Admin-0 capital alt, Admin-0 region capital]
+  - &ne_region_capital [Admin-1 capital, Admin-1 region capital]
+
   - &output_properties
     name: {col: name}
     source: {col: source}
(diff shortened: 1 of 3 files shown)


yaml/places.yaml
```diff
@@ -1,4 +1,7 @@
 global:
+  - &ne_country_capital [Admin-0 capital, Admin-0 capital alt, Admin-0 region capital]
```
nvkelso: Please use this for setting `ne_country_capital`, too :)

Great implementation!

zerebubuth: Good idea. Added in c87a6b0d.

Issue #598, opened 2016-03-13
Add `direction` to viewpoint POIs


nvkelso:
OSM allows the direction to be set on viewpoints, and Tangram can orient icons. 

* https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dviewpoint

nvkelso:
Looks like ~1000 of the 86,000 node features have direction, says TagInfo.

Direction values need to be normalized to degrees (°), sometimes they come in form of `north` or `N` or `NE`, see below.

Probably needs a TileStashe transform to ensure:

* Number should be in range of 0-359, where 0 is equal to north, using a clockwise rotation.
* Cardinal values (`N`, `W`, `S`, `E`, `NE`, `NW`, `SE`, and `SW`) are converted to their ° int values.

nvkelso:
Lookup for cardinal > degrees from OSM wiki:

![600px-angles](https://cloud.githubusercontent.com/assets/853051/14238385/89dfc74e-f9e5-11e5-9b1c-80d865bd4e20.png)

nvkelso:
Tangram supports setting an `angle` on icons now, let's add this to tiles!

/cc @dmvaldman.

dmvaldman:
Since Tangram support angles on icons, is the part that needs implementing on the Tangram side the conversion of cardinal values to degrees?

nvkelso:
I think that conversion should be on Tilezen, not Tangram. So as long as
Tangram supports sourcing the angle value from a feature property, &/or a
function, we should be good on your side.

nvkelso:
Here's a fun one, backward pointing stop signs:

- https://www.openstreetmap.org/node/3477023168

PR #1916, opened 2019-07-08
Add support for direction tag into viewpoint pois

rwrx:
Add support for direction tag into viewpoint pois.

Connects with https://github.com/tilezen/vector-datasource/issues/598.

PR #1921, opened 2019-08-09
[tests] POI direction calculation

nvkelso:
Connects with https://github.com/tilezen/vector-datasource/issues/598 to add tests for earlier PR.

- [x] Update tests
- [x] Update docs

### DIFF
--- a/vectordatasource/transform.py
+++ b/vectordatasource/transform.py
@@ -139,10 +139,7 @@ def _to_int_degrees(x):
     }
 
     # protect against bad cardinal notations
-    if x in cardinals:
-        return cardinals[x]
-
-    return None
+    return cardinals.get(x)
 
 
 def _coalesce(properties, *property_names):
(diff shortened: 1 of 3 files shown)


nvkelso:
@iandees @rmarianski I added tests and docs for the new `direction` property added in https://github.com/tilezen/vector-datasource/pull/1916 and added a few guards against bad data. Please have a 👀.

vectordatasource/transform.py
```diff
@@ -137,7 +138,11 @@ def _to_int_degrees(x):
         'west':  270, 'W': 270, 'WNW': 292, 'NW': 315, 'NNW': 337
     }
 
-    return cardinals[x]
+    # protect against bad cardinal notations
+    if x in cardinals:
```
iandees: This:

```python
if x in cardinals:
  return cardinals[x]

return None
```

Is equivalent to:

```python
return cardinals.get(x)
```

nvkelso: Fixed in https://github.com/tilezen/vector-datasource/pull/1921/commits/54b33711937fddde00a84322c739c9fbfe536b43.

PR #1922, opened 2019-08-15
changed protocol to https

MrKrisKrisu:


zerebubuth:
Thanks!

I'd like to merge this change, but as you can see from the CircleCI checks, it has some consequences for the tests. We're sorry; this isn't due to your patch. It's because a large number of tests rely on fetching external data from OSM or Overpass and the data has changed since the tests were written. We maintainers should change all the failing tests to include their own data (e.g: compare [old style](https://github.com/tilezen/vector-datasource/blob/deebbc80d91724f219c57964f7f27285802779c7/integration-test/526-inclusive-pois.py#L6-L9) and [new "generative" style](https://github.com/tilezen/vector-datasource/blob/deebbc80d91724f219c57964f7f27285802779c7/integration-test/1638-modify-poni-zooms.py#L7-L25)), or at least the 32 which are causing the checks for this PR to fail. @nvkelso - what do you think?

nvkelso:
@zerebubuth I think it's fine to merge as is and I'll file followup to update the tests (which I agree are from upstream data churn).

nvkelso:
Filed https://github.com/tilezen/vector-datasource/issues/1923 to track the generative test work.

PR #1925, opened 2019-09-27
Contours

rwrx:
Add contour lines layer.

It is also needed to modify tilequeue, I will create pull request for it.

I will also modify installation instrucions for vector-datasource for how to add contour data into postgresql database.


spreadsheets/collision_rank.yaml
```diff
@@ -108,6 +108,9 @@
 - {$layer: 'water', kind: 'sea'}
 - {$layer: 'pois', kind: 'nature_reserve'}
 - {$layer: 'water', kind: 'lake'}
+# for contours
+- _reserved: {count: 10}
+- {$layer: 'contours', kind: 'std'}
```
nvkelso: kind should be `contour`

yaml/contours.yaml
```diff
@@ -0,0 +1,15 @@
+global:
+  - &output_properties
+    source: {col: source}
+    min_zoom: {col: min_zoom}
+    ele: {col: ele}
```
nvkelso: We generally spell out property names, so `elevation` here instead of `ele`.

queries.yaml
```diff
@@ -9,6 +9,7 @@ all:
   - boundaries
   - transit
   - admin_areas
+  - contours
```
nvkelso: This would be an optional layer for the mainline build... and a fork could put it into the `all` grouping.

spreadsheets/collision_rank.yaml
nvkelso: I would put this at the very bottom of the file... just above the default value. 

These contour in this file are just for the label collisions, which should be least priority. If there are "index" contours (eg with a kind_detail) then those should have priority over generic contours.

spreadsheets/sort_rank/contours.csv
```diff
@@ -0,0 +1,2 @@
+kind,sort_rank
+*,1500
```
nvkelso: This should go between dam and breakwater, so reserve a slot there, and value more like `224` (with breakwater and other values in there getting +1).

https://github.com/tilezen/vector-datasource/blob/a53f1ef0b54c66da391584e25fa750f8aed1a9c9/spreadsheets/sort_rank/landuse.csv

yaml/contours.yaml
nvkelso: 1. Generally we want to assign an index contour that's every say 4th or 5th value... so **0**, 20, 40, 60, 80, **100** where 0 and 100 would be index contours. This allows the index contours to be "bold" (or thicker stroke) to establish better visual hierarchy on the map display, and the text can also be styled bold. 

So we'll probably need a new transform to assign `kind_detail` when it's an index contour.

queries.yaml
rwrx: Updated according to your comment.

spreadsheets/sort_rank/contours.csv
rwrx: Updated according to your comment.

spreadsheets/collision_rank.yaml
rwrx: Updated according to your first comment.

yaml/contours.yaml
rwrx: Updated according to your first comment.

rwrx: I am not sure how to add this new transfor to assign `kind_detail`. Is this also needed? Because in current styles I am doing this in function in scene styling files.

rwrx:
@nvkelso I have noticed that you have added this pull request and also other pull requests from me to 1.9.0. Which is great, thank you :). You do want to merge this before releasing 1.9.0? Also when is planned 1.9.0 to be released?

PR #1927, opened 2019-09-30
Add historic and castle_type tags

rwrx:
Add historic and castle_type tags

### DIFF
--- a/yaml/buildings.yaml
+++ b/yaml/buildings.yaml
@@ -19,7 +19,9 @@ globals:
       building_levels: {col: "building:levels"}
       building_min_levels: {col: "building:min_levels"}
       building_material: {col: "building:material"}
+      castle_type: {col: castle_type}
       height: {col: height}
+      historic: {col: historic}
       min_height: {col: min_height}
       layer: {col: layer}
       location: {col: location}


rwrx:
We have created a style similar to OpenTopoMap which also shows palaces, for example this one: https://www.openstreetmap.org/way/38975518 and we didn't have any possibility to filter these buildings because we cannot find these data into vector-datasource. We don't know how to add this correctly into vector-datasource so we tried this. How can I add this properly into vector-datasource?

yaml/buildings.yaml
```diff
@@ -19,7 +19,9 @@ globals:
       building_levels: {col: "building:levels"}
       building_min_levels: {col: "building:min_levels"}
       building_material: {col: "building:material"}
+      castle_type: {col: castle_type}
```
nvkelso, line 22: This would be better to put into a kind_detail value, with an allowlist.

yaml/buildings.yaml
```diff
@@ -19,7 +19,9 @@ globals:
...
       height: {col: height}
+      historic: {col: historic}
```
nvkelso, line 24: What values are present here in historic? Should they be turned into specific `kind` values instead?

yaml/buildings.yaml
rwrx: Yes, sure, I can do it. But I am not sure how. It is a very long time when I was changing it, so I cannot remember how this could be done. Thank you.

yaml/buildings.yaml
rwrx: Yes, sure, I can do it. But I am not sure how. It is a very long time when I was changing it, so I cannot remember how this could be done. Thank you.

PR #1928, opened 2019-09-30
Add man_made tag to roads layer

rwrx:
Add man_made tag to roads layer

### DIFF
--- a/yaml/roads.yaml
+++ b/yaml/roads.yaml
@@ -27,6 +27,7 @@ global:
       motor_vehicle: {col: motor_vehicle}
       access: {col: access}
       bicycle: {col: bicycle}
+      man_made: {col: man_made}
       hgv:
         case:
           - when: {hgv: [agricultural, delivery, designated, destination, local, 'no']}


nvkelso:
What is your aim here?

This PR is too broad as written, is there a specific property or feature you want to add? Having an example OSM feature to focus the discussion would be useful, please.

rwrx:
We have added this to filter man_made=dyke in roads. I was thinking about adding is as is added pier and quay: https://github.com/tilezen/vector-datasource/blob/master/yaml/roads.yaml#L838 but it can have multiple highway tag values not only path. So I am not sure how to add it properly.

rwrx:
Also here is one example: https://www.openstreetmap.org/way/516136046#map=16/48.2193/16.9755.

nvkelso:
If it's a dyke you're after then put this into the landuse layer instead, and just for the value(s) you're interested in.

(We've moved away from adding in any random values from an OSM key and allowlisting them, too.)

rwrx:
@nvkelso we have added this because of our style similar to OpenTopoMap which also have this for roads. What should I do or change in order for this to be allowed to merge?

PR #1929, opened 2019-09-30
Add dyke

rwrx:
Add dyke


yaml/landuse.yaml
```diff
@@ -891,14 +891,13 @@ filters:
       <<: *output_properties
       kind: groyne
       tier: 4
-  # dike
+  # dyke
   - filter:
-      man_made: dike
-      geom_type: polygon
-    min_zoom: { max: [ 13, *tier4_min_zoom ] }
```
nvkelso: Why is the `min_zoom` changing here?

nvkelso: This probably needs same treatment as `dam` where polygons and lines are treated differently.

nvkelso:
The CircleCI test failure is because we also need to change `dike` > `dyke` in the test:

- https://github.com/tilezen/vector-datasource/blob/master/integration-test/1794-demote-early-landcover.py#L289-L290

rwrx:
Now it seems that tests are ok.

PR #1930, opened 2019-09-30
Add chimney to pois

rwrx:
Add chimney to pois

### DIFF
--- a/yaml/pois.yaml
+++ b/yaml/pois.yaml
@@ -443,7 +443,7 @@ global:
         - landuse: quarry
         - leisure: [ dog_park, firepit, fishing, pitch, playground, slipway, swimming_area ]
         - lock: yes
-        - man_made: [ adit, communications_tower, crane, mast, mineshaft, obelisk, observatory,
+        - man_made: [ adit, communications_tower, chimney, crane, mast, mineshaft, obelisk, observatory,
             offshore_platform, petroleum_well, power_wind, telescope, water_tower,
             water_well, watermill, windmill ]
         - military: bunker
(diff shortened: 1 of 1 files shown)


nvkelso:
- [ ] Please also update POIs `layer` documentation in this PR (and the other ones).
    - https://github.com/tilezen/vector-datasource/blob/418ad3fb58ef18e5c39d22d1cdc3b88c64429914/docs/layers.md#poi-kind-values
- [ ] Since chimeny's don't usually have names, we need to add chimney to the PONI list.
    - https://github.com/tilezen/vector-datasource/blob/418ad3fb58ef18e5c39d22d1cdc3b88c64429914/yaml/pois.yaml#L446-L448

See also:

- https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dchimney

nvkelso:
These and the other PRs need to have tests added.

rwrx:
I have added chimney to docs. I would also create some tests but from looking into actual tests I am not sure how to do this. For example this is one chimney: https://www.openstreetmap.org/way/264214092.

yaml/pois.yaml
```diff
@@ -2640,6 +2640,11 @@ filters:
                 - shiploader
                 - tower_crane
             then: {col: "crane:type"}
+  - filter: {man_made: chimney}
+    min_zoom: 14
```
nvkelso: Please vary the min_zoom, like with like for obelisk:

```
  - filter: {man_made: obelisk}
    min_zoom:
      lookup:
        key: { call: { func: mz_to_float_meters, args: [ { col: height } ] } }
        op: '>='
        table:
          - [ 14, 20 ]  # z14 if height >= 20m
          - [ 15, 10 ]  # z15 if height >= 10m
        default: 16
```

rwrx: Ok, I have changed it as is in Obelisk.

rwrx:
@nvkelso I am not sure about the tests you mentioned. Should I copy these tests you mentioned? Thank you.

PR #1931, opened 2019-09-30
Add weir to water layer

rwrx:
- [ ] Update tests
- [ ] Update docs

### DIFF
--- a/spreadsheets/sort_rank/water.csv
+++ b/spreadsheets/sort_rank/water.csv
@@ -6,6 +6,7 @@ dam,-,201
 ditch,-,201
 drain,-,201
 playa,-,203
+weir,-,205
 lake,-,204
 water,-,204
 riverbank,-,204
--- a/tilejson/tilejson.json.erb
+++ b/tilejson/tilejson.json.erb
@@ -288,7 +288,7 @@
             "id" : "water",
             "fields" : {
                 "name" : "String. Localized name variants are in properties like 'name:en'. Common value.",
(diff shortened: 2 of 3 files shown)


nvkelso:
We put `dam` in **landuse** layer, so weir should probably also go into **landuse** layer? But if a dam is a point we put it into **pois** layer, and we should do that as well. The min_zoom should vary if point or line... and we'd probably want to vary the min_zoom for lines depending on the length.

https://wiki.openstreetmap.org/wiki/Tag:waterway=weir

rwrx:
I have moved weir into landuse layer. I am not sure also how to add different zooms depending on the line length.

spreadsheets/sort_rank/landuse.csv
```diff
@@ -21,6 +21,7 @@ dam,*,223
 land,*,222
 wetland,*,220
 mud,*,219
+weir,*,218
```
nvkelso: I would put this between `breakwater` and `dam`.

rwrx: I have put `weir` between `breakwater` and `dam`.

yaml/landuse.yaml
```diff
@@ -1252,6 +1252,14 @@ filters:
       <<: *output_properties
       kind: dam
     extra_columns: [way]
+  - filter:
+      waterway: weir
+      geom_type: line
+    min_zoom: 12
```
nvkelso, line 1258: Why 12 and not 13 or 14? In your style when do you show it?

rwrx: Yes, it is used in our style which is similar to OpenTopoMap as OpenTopoMap shows this in lower zooms.

PR #1932, opened 2019-09-30
Add fell kind to landuse

rwrx:
Add fell kind to landuse


queries.yaml
```diff
@@ -373,6 +373,7 @@ post_process:
         scree: barren
         shingle: barren
         stone: barren
+        fell: barren
```
nvkelso: Makes sense to me, thanks for taking care of the remapping!

nvkelso:
Are there also names for these, if so we need to add `collision_rank`.

rwrx:
I am not sure which number for `fell` should I set. I have looked at `bare_rock`, `desert`, `grassland`, `heath`, `sand`, `shingle` and they does not have set `sort_rank`. I haven't found a fell with a name. Is there some online tool to do osm query?

nvkelso:
Looks like a few `fell` features _do_ have names:

- https://taginfo.openstreetmap.org/tags/natural=fell#combinations

sort_order:

- https://github.com/tilezen/vector-datasource/blob/master/spreadsheets/sort_rank/landuse.csv
- Insert where `grass` is now, so `fell` takes 35 and all the rest above starting with grass increment +1 so `power_line` becomes 275. 
- There will be other test failures associated with this change that you'd need to also update, see CircleCI for which ones after you push the changes.

collision_order:

- https://github.com/tilezen/vector-datasource/blob/master/spreadsheets/collision_rank.yaml#L602
- Much simplier to just inset above grass, and build system takes care of auto incrementing everything.
- You may still have associated test failures in Circle that you'll update for after pushing, too.

rwrx:
I have updated sort and collision ranks, but now Circle is complaining that there is no 265 sort rank, because I have incremented it to 266: 

dam,LineString;MultiLineString,266

I am not sure how to change this Circle test.

rwrx:
I have added test, but I am not sure into which file I should have put it in and I am not sure whether it is correctly written.

spreadsheets/sort_rank/landuse.csv
```diff
@@ -1,115 +1,116 @@
...
+bridge,*,230
+pier,*,229
+dike,*,228
+cutline,*,227
+groyne,*,226
+breakwater,*,225
+dam,*,224
+land,*,223
+wetland,*,221
+mud,*,220
```
nvkelso, line 23: mud, wetland, land and above don't need to increment their numbers, as there is a discontinuity buffer between mud and footway.

rwrx: So, should I remove `dam`, `land`, `wetland` and `mud`? Will they sort properly when they will be removed from this list?

Issue #1939, opened 2020-08-17
So many kinds, so little clarity


nvkelso:
* **What did you see?** Lots of kinds, hardly any definitions.
* **What did you expect to see?** We pass thru OSM values for many different keys, but along the way the context is sometimes lost. Is `art` a place you can buy art, or an art installation? Let's add short, 1-2 sentence definitions for each kind. 
    - NOTE: some kinds are the same between landuse and pois layers, they should generally have the same definitions.
* **What map location are you having problems with?** https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#poi-kind-values and other layer sections.

bcamper:
Love this. I have been wanting this in the TileJSON too, for instance to provide context in UI.

stepps00:
Spent some time crafting definitions of Tilezen kinds, wikidata concordances, and splitting layers.. here are some files with definitions (existing kinds, as well as "proposed" kinds):

- Existing Tilezen kind definitions: https://gist.github.com/stepps00/2cde441ad59584ba34b70b8f1a5e3f88
- Proposed Tilezen kind definitions: https://gist.github.com/stepps00/afec2769f9227d6dc1cfb418974aa20b

Happy to put these in a markdown file/another format. I can also spend some time filling in any missing concordances/reworking definitions if that would be helpful.

In a follow-up issue, you might be able to get kind translations using the wikidata concordances.

nvkelso:
Thank you, @stepps00! I can take a crack at editing the docs to include table with icons and wikidata IDs for these.

Looks like a few of the definitions for proposed kinds are present in other layer already, and a few are totally new to Tilezen (but probably have an issue proposing their addition).

Noting here the descriptions were seeded with Wikidata text but heavily modified based on referencing OSM wiki and our own knowledge and are now an original work that is open.

nvkelso:
PR is up in #1940.

nvkelso:
This is [live](https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#poi-kind-values)!

Leaving it open to also update the tilejson per @bcamper's comment.

![image](https://user-images.githubusercontent.com/853051/91401830-c2d58b80-e7f5-11ea-82eb-77542bf2becb.png)

nvkelso:
@stepps00 there are a few missing wikidata IDs if you want to take another pass at that.

nvkelso:
[...]

PR #1940, opened 2020-08-27
[docs] add POI layer kind definitions and icons

nvkelso:
Docs only, connects to https://github.com/tilezen/vector-datasource/issues/1939 /cc @stepps00 and @bcamper.


nvkelso:
Looks good in preview, except for the images which need a merge to sort out and then quick follow to fix.

PR #1958, opened 2021-07-20
Fix wrong encoding for CA:PE road network

musculman:
Wrong encoding used for Prince Edward Island, Canada.
It was listed as `CA:PEI` but it should be `CA:PE`.

- [ ] Update tests
- [x] Update docs

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -1659,7 +1659,7 @@ Network value include:
 * `CA:NT`
 * `CA:ON:primary`
 * `CA:ON:secondary`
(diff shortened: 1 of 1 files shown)


nvkelso (approved):
LGTM

I verified this by looking at example route relations in OpenStreetMap and by reviewing in Bubble Wrap. Bubble Wrap doesn't show the custom shield in Prince Edward Island because the style config is set to CA:PEI instead of CA:PE like the data. We would match the upstream data.

PR #1956, opened 2021-06-14
Chinese parser for OSM

peitili:
https://github.com/tilezen/vector-datasource/issues/1955

- [x] Update tests
- [x] Update docs

PR #1960, opened 2021-07-26
Fix Chinese tags from OSM

peitili:
- [x] Update tests
- [x] Update docs

This is a follow-up PR for #1956 and it fixes an issue of Chinese labels in the vector tiles. There was a bug that if there is some name tags starts with "zh" such as "zh_pinyin", it would override the output "name:zh" feature, and our vector tiles would contain pinyin as the Chinese label for those sourced from OSM. However it is not ideal since pinyin is just a Romanisation for Chinese not actual Chinese characters.


vectordatasource/transform.py
```diff
@@ -592,14 +592,18 @@ def _normalize_country_code(x):
 
 osm_l10n_lookup = set([
     'zh-min-nan',
-    'zh-yue',
 ])
 
 # key is the name in OSM source; value is the Tilezen internal name
```
iandees, line 597: Update the comment here. Value is a tuple where the second field is a priority?

iandees (approved):
Looks OK to me. Let's make sure the logic makes sense to @nvkelso though.

vectordatasource/transform.py
peitili: > Update the comment here. Value is a tuple where the second field is a priority?

good point, will update in the next commit

vectordatasource/transform.py
```diff
@@ -592,14 +592,19 @@ def _normalize_country_code(x):
 
 osm_l10n_lookup = set([
     'zh-min-nan',
-    'zh-yue',
 ])
 
-# key is the name in OSM source; value is the Tilezen internal name
+
+# key is the name in OSM source; value is a tuple of Tilezen internal name and its priority value
```
nvkelso, line 598: Thank you for adding more detail here, I was scratching my head earlier

peitili:
@nvkelso also we only assume name:zh would contain "/" separated fields, but do we need to handle the case for all the fields to only select one of them from the "/" separated list in the final result?

nvkelso:
> @nvkelso also we only assume name:zh would contain "/" separated fields, but do we need to handle the case for all the fields to only select one of them from the "/" separated list in the final result?

Since we see this is the case for Papua New Guiana on Friday in Slack then yes I think we need to do this for each. https://www.openstreetmap.org/relation/307866

@peitili The unit tests you added inside transform generally work... but since the error was seen on a specific feature in tiles, we should test that tile in the associated unit test to make sure it doesn't regress, too. So in https://github.com/tilezen/vector-datasource/blob/master/integration-test/1955-chinese-parser.py add a test for USA using the full property data for https://www.openstreetmap.org/node/424317935 (related to https://www.openstreetmap.org/relation/148838).

peitili:
> > @nvkelso also we only assume name:zh would contain "/" separated fields, but do we need to handle the case for all the fields to only select one of them from the "/" separated list in the final result?
> 
> Since we see this is the case for Papua New Guiana on Friday in Slack then yes I think we need to do this for each. https://www.openstreetmap.org/relation/307866
> 
> @peitili The unit tests you added inside transform generally work... but since the error was seen on a specific feature in tiles, we should test that tile in the associated unit test to make sure it doesn't regress, too. So in https://github.com/tilezen/vector-datasource/blob/master/integration-test/1955-chinese-parser.py add a test for USA using the full property data for https://www.openstreetmap.org/node/424317935 (related to https://www.openstreetmap.org/relation/148838).

Sure, I tried to add a test to mimic the relation, however, it is non-trivial to figure out how the fixture work. For example. The test below would fail because it says  `layer 'places' was empty` and I debugged that all the layers are empty for this fixture set up. 

So instead I just create a OSM way in the integration test to represent United States. 

  ```
  def test_united_states_osm(self):
        # United States of America (osm city)
        z, x, y = (16, 0, 0)

        self.generate_fixtures(
            dsl.point(424317935, _tile_centre(z, x, y), {
                u'source': u'openstreetmap.org',
                u'name:zh': u'美國',
                u'name:yue': u'即美利堅合眾國',
                u'name:zh_pinyin': u'Měiguó',
            }),
            dsl.relation(148838, {
                'type': 'label'
            }, nodes=[424317935])
        )

        self.assert_has_feature(
            16, 0, 0, 'places',
            {'id': 424317935,
             'source': "openstreetmap.org",
             'name:zh': u'美國',
             'name:zht': u'美國'})

        self.assert_no_matching_feature(
            16, 0, 0, 'places',
            {'name:zh-default': u'美國'})
```

File: integration-test/1955-chinese-parser.py [hunk 3]
```diff
@@ -62,6 +68,130 @@ def test_san_francisco_osm(self):
...
+                u"short_name:es": u"EE.UU.",
+                u"short_name:vi": u"Mỹ",
+                u"source": u"openstreetmap.org",
+                u"sqkm": u"9826675",
+                u"wikidata": u"Q30",
+                u"wikipedia": u"en:United States"
+            }))
+
+        self.assert_has_feature(
+            16, 10482, 25330, 'places',
```
nvkelso, line 184: Since the rendering logic changes per zoom and by kind... please add to test
- as a low zoom tile (`4/3/6`) (change test to test `4, 3, 6` not the zoom 16 coord)
- for `kind: country` (add kind to the output test):

image

integration-test/1955-chinese-parser.py
```diff
@@ -62,6 +68,130 @@ def test_san_francisco_osm(self):
...
+            {'id': 26819236,
+             'source': "openstreetmap.org",
+             'name': 'United States of America',
+             'name:zh': u'美國',
+             'name:zht': u'美國'})
+
+        self.assert_no_matching_feature(
```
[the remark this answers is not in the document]
peitili: > Please add another `assert_no_matching_feature` for the Pinyin creaping into **zh** of `Měiguó`.

added in the next commit.

integration-test/1955-chinese-parser.py
peitili: > Since the rendering logic changes per zoom and by kind... please add to test
> 
> * as a low zoom tile (`4/3/6`) (change test to test `4, 3, 6` not the zoom 16 coord)
> * for `kind: country` (add kind to the output test):
> 
> image

I was able to find an example of how to do a low zoom integration test, so just copied some code, please see in the newest commit `test_united_states`

integration-test/1955-chinese-parser.py
```diff
@@ -1,10 +1,154 @@
...
+                u"sqkm": u"9826675",
+                u"wikidata": u"Q30",
+                u"wikipedia": u"en:United States",
+                '__ne_min_zoom': 1.7,
+                '__ne_max_zoom': 6.7,
+            }),
+        )
+        # should show up in zooms within the range 2-6
+        for zoom in xrange(2, 6):
+            x, y = deg2num(lat, lon, zoom)
```
nvkelso, line 134: neat!

integration-test/1955-chinese-parser.py
```diff
@@ -62,6 +206,132 @@ def test_san_francisco_osm(self):
             16, 10482, 25330, 'places',
             {'name:zh-default': u'旧金山/三藩市/舊金山'})
 
+        self.generate_fixtures(dsl.way(26819236, wkt_loads(
```
nvkelso, line 209: My last 2¢: Please remove this fixture and later 2 tests as I think they duplicate the `test_united_states` which starts on line 16, and that test has the corrected coordinates while this one still tests for zoom 16 in error.

peitili: that makes sense, I removed in the next commit.

nvkelso (approved):
Thanks for catching these edge cases!

I'll update my separate Natural Earth simplified and traditional Chinese logic in https://github.com/nvkelso/natural-earth-vector/issues/533.

PR #1963, opened 2021-08-09
Fix Chinese language parser

peitili:
- [x] Update tests
- [x] Update docs

Changes

1. We encountered an error (see below) during import and we are fixing that
1. We are also adding a change to reject any recognized Chinese properties if they are not in the predefined Chinese property list. This is to prevent unforeseen data from populating the Chinese result.

`Traceback (most recent call last): |   File \"/usr/src/tilequeue/tilequeue/command.py\", line 2352, in tilequeue_meta_tile_low_zoom    formatted_tiles, _ = processor.process_tiles() |   File \"/usr/src/tilequeue/tilequeue/process.py\", line 787, in process_tiles    log_fn=log_fn, |   File \"/usr/src/tilequeue/tilequeue/process.py\", line 512, in process_coord    output_calc_spec, log_fn=log_fn) |   File \"/usr/src/tilequeue/tilequeue/process.py\", line 388, in process_coord_no_format    shape, props, feature_id, nominal_zoom) |   File \"/usr/src/tilequeue/tilequeue/process.py\", line 27, in transform_fn    shape, properties, fid = fn(shape, properties, fid, zoom) |   File \"/usr/src/vector-datasource/vectordatasource/transform.py\", line 814, in tags_name_i18n    clean_backfill_zh(properties) |   File \"/usr/src/vector-datasource/vectordatasource/transform.py\", line 684, in clean_backfill_zh    properties['name:zh-Hans'] = properties['name:zh-Hans'].split('/')[0].strip().strip('\\\\') | AttributeError: 'NoneType' object has no attribute 'split'`

PR #1961, opened 2021-07-30
Add Chinese parser for NE and fix a few edge cases

peitili:
- [x] Update tests
- [x] Update docs

Add Chinese parser for NE and change the output property name from "zh" to "zh-hans" and "zht" to "zh-hant"
Also fixes a few edge cases for OSM and WOF Chinese fields.

https://github.com/tilezen/vector-datasource/issues/1955

PR #1964, opened 2021-08-13
Add separate NE config for Chinese simplified and traditional

nvkelso:
Extends work done in #1961 and #1963

During a build we discovered the live data isn't parsed correctly because the Natural Earth config is ignorant of the new `zht` name property. Because we reused `zh` for `zh-Hans` content that new data did flow thru correctly.

The fix is to extend the list of NE language codes with what's in the new beta data.

### DIFF
--- a/queries/ne.jinja2
+++ b/queries/ne.jinja2
@@ -1,5 +1,5 @@
 {% set ne_languages = ['ar', 'bn', 'de', 'el', 'en', 'es', 'fa', 'fr', 'he', 'hi', 'hu', 'id', 'it', 'ja', 'ko', 'nl', 'pl', 'pt', 'ru', 'sv', 'tr', 'uk', 'ur', 'vi', 'zh', 'zht'] %}
-{% set ne_viewpoints = ['iso', 'ar', 'bd', 'br', 'cn', 'de', 'eg', 'es', 'fr', 'gb', 'gr', 'id', 'il', 'in', 'it', 'jp', 'ko', 'ma', 'nl', 'np', 'pk', 'pl', 'ps', 'pt', 'ru', 'sa', 'se', 'tr', 'tw', 'us', 'vn'] %}
+{% set ne_viewpoints = ['iso', 'ar', 'bd', 'br', 'cn', 'de', 'eg', 'es', 'fr', 'gb', 'gr', 'id', 'il', 'in', 'it', 'jp', 'ko', 'ma', 'nl', 'np', 'pk', 'pl', 'ps', 'pt', 'ru', 'sa', 'se', 'tr', 'tw', 'ua', 'us', 'vn'] %}
 
 {% macro ne_common_properties() %}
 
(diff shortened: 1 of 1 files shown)


queries/ne.jinja2
```diff
@@ -1,4 +1,4 @@
-{% set ne_languages = ['ar', 'bn', 'de', 'en', 'es', 'fr', 'el', 'hi', 'hu', 'id', 'it', 'ja', 'ko', 'nl', 'pl', 'pt', 'ru', 'sv', 'tr', 'vi', 'zh'] %}
+{% set ne_languages = ['ar', 'bn', 'de', 'el', 'en', 'es', 'fa', 'fr', 'he', 'hi', 'hu', 'id', 'it', 'ja', 'ko', 'nl', 'pl', 'pt', 'ru', 'sv', 'tr', 'uk', 'ur', 'vi', 'zh', 'zht'] %}
```
nvkelso, line 1: `zht` is the critical one, but the new NE data also includes explicate English (instead of mostly English `name` singular), Farsi, Hebrew, Ukrainian, and Urdu.

peitili: this means our integration test misses this path?

nvkelso: The integration tests are based on "data fixtures" where you mock the **output** from a database query. 

So yes, they are missing this path. This is why we should **also** have (eg add) "build release tests" to make sure tiles have the right content for a full end-to-end prod promotion.

queries/ne.jinja2
```diff
@@ -1,5 +1,5 @@
-{% set ne_languages = ['ar', 'bn', 'de', 'en', 'es', 'fr', 'el', 'hi', 'hu', 'id', 'it', 'ja', 'ko', 'nl', 'pl', 'pt', 'ru', 'sv', 'tr', 'vi', 'zh'] %}
-{% set ne_viewpoints = ['iso', 'ar', 'bd', 'br', 'cn', 'de', 'eg', 'es', 'fr', 'gb', 'gr', 'id', 'il', 'in', 'it', 'jp', 'ko', 'ma', 'nl', 'np', 'pk', 'pl', 'ps', 'pt', 'ru', 'sa', 'se', 'tr', 'tw', 'us', 'vn'] %}
+{% set ne_languages = ['ar', 'bn', 'de', 'el', 'en', 'es', 'fa', 'fr', 'he', 'hi', 'hu', 'id', 'it', 'ja', 'ko', 'nl', 'pl', 'pt', 'ru', 'sv', 'tr', 'uk', 'ur', 'vi', 'zh', 'zht'] %}
+{% set ne_viewpoints = ['iso', 'ar', 'bd', 'br', 'cn', 'de', 'eg', 'es', 'fr', 'gb', 'gr', 'id', 'il', 'in', 'it', 'jp', 'ko', 'ma', 'nl', 'np', 'pk', 'pl', 'ps', 'pt', 'ru', 'sa', 'se', 'tr', 'tw', 'ua', 'us', 'vn'] %}
```
nvkelso, line 2: Technically Ukrainian POV was added to the data, so adding that here as well so we don't forget it later.

peitili (approved):

nvkelso:
The ZIP file of Natural Earth beta data we're loading does include POV, except for the populated places SHP which lost it along the way. I've restored the values and the next data delivery will include that. But not a blocker for this name work.

PR #1965, opened 2021-08-20
Fixes to build and consume new shapefiles tar

travisgrigsby:
* Point at newest shapesfiles tar in S3
* change import order to accommodate changes to Fiona, fix deprecation warning
* don't process WOF entities that have a '-' in the name
* update WOF manifest and download URLs (longer term we should update this to not use the 'legacy' section anymore)
* remove concept of a count from parsing of inventory.json in WOF

### DIFF
--- a/data/wof_snapshot.py
+++ b/data/wof_snapshot.py
@@ -12,7 +12,7 @@ import requests
 from tqdm import tqdm
 
 """
-expects s to look like "123456.geojson"
+expects input to look like "123456.geojson"
 """
 def _parse_wof_id(s):
     wof_id, ext = splitext(basename(s))
@@ -124,6 +124,7 @@ if __name__ == '__main__':
         print "Downloading %r" % (placetype)
         with tmpdownload(WOF_BUNDLE_PREFIX + fname, download_size) as fname:
             print "Parsing WOF data"
+            # 20210820: geocode.earth inventory files don't offer count, so count hardcoded to 1
             reader.add_archive(fname, version, 1)
 
     print "Writing output SQL"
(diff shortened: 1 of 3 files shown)


iandees (approved):
Looks good. I had some extra stuff related to getting it running in Python 3 and I completely removed the progress bar (`tqdm`) stuff where you just hard-code the progress max to 1.

data/wof_snapshot.py
```diff
@@ -11,7 +11,9 @@
 import requests
 from tqdm import tqdm
 
-
+"""
+expects s to look like "123456.geojson"
```
nvkelso, line 15: nit: extra `s` here

data/wof_snapshot.py
```diff
@@ -117,13 +119,12 @@ def __exit__(self, type, value, traceback):
...
         version = item['last_updated']
-        count = item['count']
         download_size = item['size_compressed']
 
-        print "Downloading %r with %d entries" % (placetype, count)
+        print "Downloading %r" % (placetype)
         with tmpdownload(WOF_BUNDLE_PREFIX + fname, download_size) as fname:
             print "Parsing WOF data"
-            reader.add_archive(fname, version, count)
+            reader.add_archive(fname, version, 1)
```
nvkelso, line 127: Add a comment here saying why the count is always 1, please. Something like `20210820: geocode.earth inventory files don't offer count, so set this to 1 all the time`

nvkelso (approved):
Two nits, otherwise LGTM

data/wof_snapshot.py
travisgrigsby: that's the name of the var.  I suppose it is unnecessary

PR #1971, opened 2021-09-01
Remove sharding

travisgrigsby:
It appears that one or more shards aren't finishing or failing to add min zooms to the planet_osm_point table.  Removing sharding to see if the situation improves.


iandees (approved):

iandees:
For posterity: we're doing this because my investigation of missing country label points showed that one or more of these shards didn't finish or failed to add min zooms to the planet_osm_point table.

PR #1972, opened 2021-09-02
Protect against abnormally large parking lots

iandees:
One of my global builds failed recently because of:

```
ERROR:  value "6472217472217" is out of range for type integer
CONTEXT:  PL/pgSQL function tz_estimate_parking_capacity(text,text,text,real) line 8 at RETURN
PL/pgSQL function mz_calculate_min_zoom_pois_(bigint,geometry,hstore,real,text) line 5 at RETURN
PL/pgSQL function mz_calculate_min_zoom_pois(planet_osm_point) line 5 at RETURN
```

This modifies the `tz_estimate_parking_capacity()` function to be more careful with the input data from OSM.

- [x] Update tests
- [ ] ~~Update docs~~

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1072,19 +1072,19 @@ DECLARE
   spaces_per_level INTEGER;
 BEGIN
   -- if the capacity is set to something between 0 and 99999, then use that.
-  IF capacity ~ '^[0-9]{1,5}+$' THEN
+  IF capacity ~ '^[0-9]{1,5}$' THEN
     RETURN capacity::integer;
   END IF;
   -- don't try to do this if way_area is abnormally large.
   -- (Epcot's parking lot is 647k m^2)
-  IF way_area > 700000 THEN
+  IF way_area > 2000000 THEN
     RETURN NULL;
   END IF;
   -- otherwise, try to use the information we have to guess the capacity
   spaces_per_level := (way_area / 46.0)::integer;
   levels_int := CASE
     -- limit levels to an integer between 0 and 99
-    WHEN levels ~ '^[0-9]{1,3}$' THEN levels::integer
+    WHEN levels ~ '^[0-9]{1,2}$' THEN levels::integer
     WHEN parking = 'multi-storey' THEN 2
     ELSE 1
   END;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1071,14 +1071,20 @@ DECLARE
   levels_int INTEGER;
   spaces_per_level INTEGER;
 BEGIN
-  -- if the capacity is set, then use that.
-  IF capacity ~ '^[0-9]+$' THEN
+  -- if the capacity is set to something between 0 and 99999, then use that.
+  IF capacity ~ '^[0-9]{1,5}+$' THEN
     RETURN capacity::integer;
+  END IF;
+  -- don't try to do this if way_area is abnormally large.
+  -- (Epcot's parking lot is 647k m^2)
+  IF way_area > 700000 THEN
+    RETURN NULL;
   END IF;
   -- otherwise, try to use the information we have to guess the capacity
   spaces_per_level := (way_area / 46.0)::integer;
   levels_int := CASE
-    WHEN levels ~ '^[0-9]+$' THEN levels::integer
+    -- limit levels to an integer between 0 and 99
+    WHEN levels ~ '^[0-9]{1,3}$' THEN levels::integer
     WHEN parking = 'multi-storey' THEN 2
     ELSE 1
   END;
--- a/integration-test/1972-big-parking-lots.py
+++ b/integration-test/1972-big-parking-lots.py
@@ -25,8 +25,10 @@ class BigParkingLotsTest(FixtureTest):
         )
 
         x, y = deg2num(lat, lon, 15)
-        self.assert_no_matching_feature(
-            15, x, y, 'places', {
+        self.assert_has_feature(
+            15, x, y, 'pois', {
                 'id': 8712714905,
+                'kin': 'parking',
+                'min_zoom': 18,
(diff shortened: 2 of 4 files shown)


nvkelso:
Looking good so far, I'll look closer once a test is added.

data/functions.sql
```diff
@@ -1071,14 +1071,20 @@ DECLARE
...
 BEGIN
-  -- if the capacity is set, then use that.
-  IF capacity ~ '^[0-9]+$' THEN
+  -- if the capacity is set to something between 0 and 99999, then use that.
+  IF capacity ~ '^[0-9]{1,5}+$' THEN
     RETURN capacity::integer;
   END IF;
+  -- don't try to do this if way_area is abnormally large.
+  -- (Epcot's parking lot is 647k m^2)
+  IF way_area > 700000 THEN
```
travisgrigsby, line 1080: just making sure this is the right number.  I understand Epcot's size, but is that the largest parking lot we want to support?

data/functions.sql
```diff
@@ -1071,14 +1071,20 @@ DECLARE
...
+    RETURN NULL;
+  END IF;
   -- otherwise, try to use the information we have to guess the capacity
   spaces_per_level := (way_area / 46.0)::integer;
   levels_int := CASE
-    WHEN levels ~ '^[0-9]+$' THEN levels::integer
+    -- limit levels to an integer between 0 and 99
+    WHEN levels ~ '^[0-9]{1,3}$' THEN levels::integer
```
travisgrigsby, line 1087: Isn't this going to allow between 0 and 999?  I thought this second number in the {x, y} was usually inclusive, no?  Same for line 1075

data/functions.sql
nvkelso: Thats about 173 acres which are kinda city block shaped. Generally ETL guards should allow for larger values, just not ridiculously large values. So 2x or 3x the number and call it good?

travisgrigsby (approved):

data/functions.sql
iandees: Yep. My local code has 2. I must have changed it and not saved/committed.

iandees: Fixed in 8c8823b.

vectordatasource/meta/function.py
```diff
@@ -407,13 +407,24 @@ def tz_estimate_parking_capacity(capacity, parking, levels, way_area):
...
         return capacity
 
     except (ValueError, TypeError):
         # sometimes people don't put integers in the capacity, which is kind of
         # annoying. it means we just have to fall back to estimating.
         pass
 
+    # don't try to do this if way_area is abnormally large
+    # (Epcot's parking lot is 647k m^2)
+    if way_area > 2000000:
```
peitili, line 425: Is it beneficial to print some log for these conditions? I don't see any log in this file, do we log in the caller?

iandees:
I added an integration test and a Python test, but it doesn't fail when I run it with my changes removed from the SQL. I'm going to spend some time digging into when/if the SQL functions are updated.

travisgrigsby:
> I added an integration test and a Python test, but it doesn't fail when I run it with my changes removed from the SQL. I'm going to spend some time digging into when/if the SQL functions are updated.

Bummer.  It must be calling the sql though, right?  Can you try breaking the sql and see if that does something?

integration-test/1972-big-parking-lots.py
```diff
@@ -0,0 +1,32 @@
...
+                u"amenity": u"parking",
+                u"capacity": u"6472217472217",
+                u"parking": u"surface",
+                u"source": u"openstreetmap.org",
+            })
+        )
+
+        x, y = deg2num(lat, lon, 15)
+        self.assert_no_matching_feature(
+            15, x, y, 'places', {
```
nvkelso, line 29: A zoom 15 tile should contain all the zoom 15+ content, so should we instead assert (like in `test_meta.py` change) that it be included but with a **min_zoom** of `18`?

nvkelso: Looks like most our tests are zoom 16 max zoom, so I'd update the assertion here to test 16 not 15 (and line 27 above).

iandees: I modified the test to check for the presence of a feature with a min_zoom of 18, but it doesn't seem to pass locally because the `pois` layer is completely empty:

```
AssertionError: Did not find feature including properties {'source': u'openstreetmap.org', 'min_zoom': 18, 'id': 8712714905} (because layer 'pois' was empty)
```

Am I missing something in the updated test?

integration-test/1972-big-parking-lots.py
```diff
@@ -0,0 +1,34 @@
...
+        self.generate_fixtures(
+            dsl.point(8712714905, (lon, lat), {
+                u"amenity": u"parking",
+                u"capacity": u"6472217472217",
+                u"parking": u"surface",
+                u"source": u"openstreetmap.org",
+            })
+        )
+
+        x, y = deg2num(lat, lon, 15)
```
nvkelso, line 27: Change 15 to 16

integration-test/1972-big-parking-lots.py
```diff
@@ -0,0 +1,34 @@
...
+        self.assert_has_feature(
+            15, x, y, 'pois', {
```
nvkelso, line 29: Change 15 to 16

iandees: That did it, thanks! Fixed in 94e0755a.

nvkelso: 16 is the max zoom for zoom 256 tiles (while 15 is max for 512 tiles)

nvkelso (approved):

PR #1977, opened 2021-10-06
Update the way we grab WOF assets

tgrigsby-sc:
They've changed the way they package assets over at WOF, so they wanted us to change what we're getting from them.  Now we grab a single global file, decompress it, open it as a sqlite db and read all the neighborhoods from it.  We no longer need to do things the old way, so I yanked out that code.


nvkelso:
@travisgrigsby can this be merged?

travisgrigsby:
> @travisgrigsby can this be merged?

Yep!  Was waiting to see if Peter had anything else.  Merged

Issue #1970, opened 2021-08-31
Show more disputed lines for POV in low zooms


nvkelso:
To solve "gaps" in the country boundary lines at low zooms sourced from Natural Earth, we need to be more liberal in NOT filtering by min_zoom all the time, just for that layer and features (`country`, `disputed`, `region` kinds in **boundary** layer).

![tilezen_v1d8_disputed_boundaries](https://raw.githubusercontent.com/tilezen/vector-datasource/master/docs/images/tilezen-v1.8.0-disputed-boundaries.gif)

Per https://github.com/nvkelso/natural-earth-vector/issues/301#issuecomment-894928535:

>  Adjust min_zoom of disputed lines to be smaller when one or more of their claimants needs them earlier to complete their own country boundary line at lower zooms. See China along India, Pakistan, Bhutan salients.

The simplest thing to do is in QGIS (and for vector tiles) is to _always_ include features from the scalerank theme (so remove the `min_zoom` filters) as for default the current `min_zoom` is ideal... but otherwise we'd need to include optional `min_zoom_{pov}` for all the POV and there are an insane number of columns as it is.

If we were to add various `min_zoom_{pov}` then downstream QGIS and vector tiles would need to take the minimum `min_zoom_{pov}`for inclusion at that zoom... but that's not something QGIS's basic layer filters are setup for, and the layer styling is already pretty complicated as it is.

nvkelso:
Relevant code:

- min_zoom globals: https://github.com/tilezen/vector-datasource/blob/v1.8.0/yaml/boundaries.yaml#L6
- min_zoom remapping: https://github.com/tilezen/vector-datasource/blob/v1.8.0/yaml/boundaries.yaml#L9-L31
- NE config: https://github.com/tilezen/vector-datasource/blob/v1.8.0/yaml/boundaries.yaml#L225-L247

One way to do this is to split apart the disputed lines from the international boundary lines (this may need to be a little more sophisticated – for instance you could test if there is a fclass_xx that's not null, then set to 0 instead of setting all to 0):

```
  # ne
  - filter:
      featurecla:
        - 'International boundary (verify)'
    min_zoom: { col: min_zoom }
    output:
      <<: *output_properties
      <<: *ne_localized_kind_properties
      kind: *ne_country_boundaries_kind
      kind_detail: '2'
    table: ne

# ne for disputed kind:xx POV we want to always include
# the disputed lines early so when that POV enables them
# the data is available in tiles
  - filter:
[...]

PR #1984, opened 2021-11-02
Split NE boundary filters

peitili:
In order for disputed boundaries to show at low zooms, we need to adjust the filters settings for boundaries layer.

- [x] Update tests
- [ ] Update docs

https://github.com/tilezen/vector-datasource/issues/1970

### Test
rebuild the metatile 1/1/0 which contains the tile 3/5/3, and it shows that 3/5/3 contains disputed boundary cn and min_zoom as 1

image


nvkelso (approved):
The 2 tests that were mod do pass, and the filter logic is good.

Issue #1982, opened 2021-10-28
Distinguish water reservoirs that are covered


nvkelso:
Add a data property to allow not rendering covered water features as blue.

Example feature:

- https://www.openstreetmap.org/way/28575415#map=17/37.75193/-122.45226
- 
`water=reservoir` and `covered=yes`

This one should definitely not appear as blue water as it is covered in real life.

![image](https://user-images.githubusercontent.com/853051/139192800-84f64e09-4e17-414e-a952-271342cd9c64.png)

nvkelso:
You'll want to modify the water.yaml to extend:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/water.yaml#L16-L22

Like:

```
  - &water_standard_properties_osm
      <<: *output_properties
      intermittent:
        case:
          - when:
              intermittent: 'yes'
            then: true
      covered:
        case:
          - when:
              covered: 'yes'
            then: true
```

Alternately it could go in the general section if values are not just "yes" or True:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/water.yaml#L2-L15

>     covered : {col: covered}

But then it could have any random value in it... so we'd want to at least sanitize those values.

Examples of sanitizing values:
- https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L266
- https://github.com/tilezen/vector-datasource/blob/31674b009f70c225e9b0e609882d63bbee7b167e/vectordatasource/transform.py#L4573-L4595

TagInfo can help understand value, but this is across all features not just water features. So you'd want a more refined search:

- https://taginfo.openstreetmap.org/keys/covered#values

And for bedtime reading, the OSM Wiki has hopes and dreams (rather than hard rules) about the state of editing:

- https://wiki.openstreetmap.org/wiki/Key%3Acovered#covered.3Dyes

peitili:
after the change and rebuild the reservoir looks like this using the tangram
![Snip20211114_8](https://user-images.githubusercontent.com/62725319/141734013-1639c6b9-b0b1-4b26-a725-c55ecf21c0e8.png)

PR #1989, opened 2021-11-09
Don't output feature to water layer if covered='yes'

peitili:
Don't output features to water layer when there is covered=yes tag to not display blue for covered reservoirs.

- [x] Update tests
- [x] Update docs

#1982


nvkelso (approved):
LGTM

Nit, please also update the docs to say `Covered water is not included.`

PR #1991, opened 2021-11-10
Add shield text length fields

travisgrigsby:
Implement *shield_text_length (as string) for all the *shield_text fields, including hgv_restrictions.  Added tests for all and updated docs.
- [X] Update tests
- [X] Update docs


integration-test/1979-shield-text-length.py
```diff
@@ -0,0 +1,200 @@
...
+            }, ways=[417097119])
+        )
+
+        self.assert_has_feature(
+            z, x, y, 'roads', {
+                'id': 417097119,
+                'shield_text': '123456',
+                'shield_text_length': '6',
+                'bicycle_shield_text': '1234567',
+                'bus_shield_text': '',
```
nvkelso, line 133: Why is `'bus_shield_text'` an empty string here?

nvkelso (approved):

docs/layers.md
```diff
@@ -1498,6 +1502,7 @@ To improve performance, some road segments are merged at low and mid-zooms. To f
 * `hgv_restriction`: optional property indicating limitations to heavy goods vehicle truck access. See below for list of values. Available on both point and line geometries. See also `hgv_restriction_shield_text`.
 * `hgv_time_restrictions` - optional property specifying when heavy goods vehicle truck access is restricted. See the `hgv_time_restrictions` values list below.
 * `hgv_restriction_shield_text`: optional and paired with `hgv_restriction` points with values like `5.1m`. Because the units are different per restriction an abbreviation should be provided. Values in meters can be specified with one decimal precision but value of 5.0m should be given as 5m.
+* `hgv_restriction_shield_text_length`: optional.  Returns length of `hgv_restriction_shield_text` as a string.  Missing if `hgv_restriction_field_text` is !(0 < length < 7) or field doesn't exist
```
nvkelso, line 1505: Nit, the `Missing if...` text should be replicated to all the other documentation changes, too.

travisgrigsby: can do

integration-test/1979-shield-text-length.py
travisgrigsby: because we're testing the case where we don't show a length when the shield text is empty.  That was one of the requirements we discussed yesterday - if len < 0 or len > 6 we don't include the length field

Issue #1995, opened 2021-11-16
Show more medium sized parks in cities


nvkelso:
In NYC we don't see the medium sized parks until zoom 15. Instead those should come in earlier at zoom 14. We can accomplish this by being more liberal in the size cutoff which is now 50,000 meters. 

Generic parks are configured here: 

 -https://github.com/tilezen/vector-datasource/blob/master/yaml/landuse.yaml#L433-L442

```
  - filter:
      any:
        leisure: park
        landuse: park
        boundary: national_park
    min_zoom: *tier2_min_zoom
    output:
      <<: *output_properties
      kind: park
      tier: 2
```

Which (along with 25 other kinds) use the `tier2_min_zoom` area filters. Most of these are similar "park" like features, though there are some "landcover" type kinds as well.

- https://github.com/tilezen/vector-datasource/blob/master/yaml/landuse.yaml#L55-L72

```
 - &tier2_min_zoom
    lookup:
      key: { col: way_area }
      op: '>='
      table:
        - [ 4, 1000000000 ]
        - [ 5, 1000000000 ]
        - [ 6,  150000000 ]
        - [ 7,  100000000 ]
        - [ 8,   10000000 ]
        - [ 9,    5000000 ]
        - [ 10,   1000000 ]
        - [ 11,    500000 ]
        - [ 12,    250000 ]
        - [ 13,    100000 ]
        - [ 14,     50000 ]
        - [ 15,      2000 ]
      default: 16
```

Looking at [Bryant Park](https://www.openstreetmap.org/way/22727025) and Madison Square park those are both around 40,000 meters. So a value between 25k and 35k should work at zoom 14. The next zoom break at 15 is 2k and the previous break at zoom 13 is 100k. 

In San Francisco we're also missing [Alamo Square](https://www.openstreetmap.org/way/745183964) at zoom 12 which is 173k area, so let's also try adjusting that down to 150k instead of 250k. 

Then zoom 14 needs to be adjusted down to accommodate zoom 13's change.

Something more like:

```
        - [ 11,    500000 ]
        - [ 12,    150000 ]
        - [ 13,     50000 ]
        - [ 14,     20000 ]
        - [ 15,      2000 ]
```

**But what about the labels for these features?**  (no-op)

[...]

PR #1996, opened 2021-11-16
including more medium sized parks

travisgrigsby:
Connects with https://github.com/tilezen/vector-datasource/issues/1995.

Updated landuse.yaml to show bryant park and alamo square as a proxy for other similarly sized parks in other geos.  

- [X] Update tests


nvkelso:
My guess is a few existing park zoom tests will fail, so look at Circle to find those.

PR #2002, opened 2021-11-17
Rename the parks integration test to match the issue number

iandees:
This renames the `1999-medium-sized-parks` integration test to `1995` to match #1995, the issue it's related to. This way I can use 1999 for #1999. 😄


nvkelso (approved):
LGTM

Issue #2004, opened 2021-11-19
introduce pre-commit linters to this repo


peitili:
We found that sometimes people mistakenly check-in bad code to the repo such as an invalid yaml file, we should enforce the repo by introducing pre-commit checks to lint the codes and enable the linter checks on the CI too.

peitili:
There are many good linters that can be applied to this repo, and I will add them by multiple PRs phase by phase to have a good cadence of changes and at the same time easier for reviewers.

peitili:
Most linters have been added and enabled.

PR #2005, opened 2021-11-19
Introduce pre-commit and add yaml linters

peitili:
Introduce the pre-commit to the repo and add yaml style check git hook to it. Also enable the check on the CI to guard.

The yaml files changes are all detected by the hook, and I manually fixed them. Most of the violations are the `<<` merge keys, those are legit violations since they don't conform to the [yaml spec](https://yaml.org/type/merge.html) . 

Another violation is a dup key of `source:` in the pois.yaml.

#2004 

- [ ] Update tests
- [x] Update docs

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -132,7 +132,9 @@ We'll cover the following topics in the next sections:
 - [Submit a Pull Request (PR)](CONTRIBUTING.md#10-submit-a-pull-request-pr)
 
 ### 1. Install prerequisite tool [pre-commit package manager](https://pre-commit.com) on your machine
+
 Since the repo right now uses python2.7, you need to install pre-commit using python2.7. 
+
 Then enable it by doing `pre-commit install`.
 
 ### 2. Choose an issue to work on
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -131,7 +131,11 @@ We'll cover the following topics in the next sections:
 - [Push your local branch to the server](CONTRIBUTING.md#9-push-your-local-branch-to-the-server)
 - [Submit a Pull Request (PR)](CONTRIBUTING.md#10-submit-a-pull-request-pr)
 
-### 1. Choose an issue to work on
+### 1. Install prerequisite tool [pre-commit package manager](https://pre-commit.com) on your machine
+Since the repo right now uses python2.7, you need to install pre-commit using python2.7. 
+Then enable it by doing `pre-commit install`.
+
+### 2. Choose an issue to work on
 
 We have a backlog of issues, but they are also grouped into milestones and tracked with [Waffle board](https://waffle.io/tilezen/vector-datasource).
 
@@ -139,7 +143,7 @@ When picking an issue from the Ready column for the active milestone, self assig
 
 If you propose to work on an issue in the Backlog but what to confirm some details add a comment to the issue or ask about it in Slack.
 
-### 2. Create a new branch
+### 3. Create a new branch
 
 Ensure you're on the master branch to establish a clean history:
 
@@ -173,7 +177,7 @@ NOTE: Your first push for a branch might require additional details:
 git push --set-upstream origin olga/875-camp-ground-zoom
 ```
 
-### 3. Create a new test
+### 4. Create a new test
 
 Create a new test for the issue in `integration-test` dir. Sometimes it's helpful to look thru the existing tests to find one that is a close match to the pattern and start there.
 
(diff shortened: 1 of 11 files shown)


nvkelso:
The errors in more detail:

```
- hook id: check-yaml
- exit code: 1

while constructing a mapping
  in "yaml/boundaries.yaml", line 238, column 7
found duplicate key "<<"
  in "yaml/boundaries.yaml", line 239, column 7

To suppress this check see:
   http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys

while constructing a mapping
  in "yaml/buildings.yaml", line 79, column 7
found duplicate key "<<"
  in "yaml/buildings.yaml", line 81, column 7

To suppress this check see:
   http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys

while constructing a mapping
  in "yaml/places.yaml", line 147, column 7
found duplicate key "<<"
  in "yaml/places.yaml", line 148, column 7

To suppress this check see:
   http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys

while constructing a mapping
  in "yaml/pois.yaml", line 6, column 5
found duplicate key "source" with value "{}" (original value: "{}")
  in "yaml/pois.yaml", line 14, column 5

To suppress this check see:
    http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys

while constructing a mapping
  in "yaml/landuse.yaml", line 450, column 7
found duplicate key "<<"
  in "yaml/landuse.yaml", line 452, column 7

To suppress this check see:
   http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys

while constructing a mapping
  in "yaml/roads.yaml", line 296, column 7
found duplicate key "<<"
  in "yaml/roads.yaml", line 297, column 7

To suppress this check see:
   http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys
```

nvkelso:
The `<<:` directive is used extensively to insert globals. Is there a more standard syntax we should switch to? Or set the linter to ignore those?

Example:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/places.yaml#L96

The `duplicate key "source" ` in https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L8-L14 I think is an error that should just be corrected with this PR.

peitili:
> The `<<:` directive is used extensively to insert globals. Is there a more standard syntax we should switch to? Or set the linter to ignore those?
> 
> Example:
> 
> * https://github.com/tilezen/vector-datasource/blob/master/yaml/places.yaml#L96
> 
> The `duplicate key "source" ` in https://github.com/tilezen/vector-datasource/blob/master/yaml/pois.yaml#L8-L14 I think is an error that should just be corrected with this PR.

I wish I could ignore the merge key errors, however the tool doesn't have the option. See more details in https://github.com/pre-commit/pre-commit-hooks/issues/468
https://github.com/pre-commit/pre-commit-hooks/issues/681

nvkelso:
The `<<:` is definitely valid YAML syntax:

- https://stackoverflow.com/questions/41063361/what-is-the-double-left-arrow-syntax-in-yaml-called-and-wheres-it-specced
- https://yaml.org/type/merge.html

Reading your links it seems like the linter suggestion is to change to wrap in brackets (which allows a list which we never use):

>       <<: [*output_properties]

Which to me is "allowed" but a perhaps very strict reading on the spec. Want to give that a try?

peitili:
> The `<<:` is definitely valid YAML syntax:
> 
> * https://stackoverflow.com/questions/41063361/what-is-the-double-left-arrow-syntax-in-yaml-called-and-wheres-it-specced
> * https://yaml.org/type/merge.html
> 
> Reading your links it seems like the linter suggestion is to change to wrap in brackets (which allows a list which we never use):
> 
> > ```
> >   <<: [*output_properties]
> > ```
> 
> Which to me is "allowed" but a perhaps very strict reading on the spec. Want to give that a try?

Actually, the linter only suggest to wrap in brackets if we have multiple merge keys under a parent key. It is not necessary when there is only a single merge key.

Yup you can try. I have tried these following two versions generate same result. Yup, you can try more.


```
 <<: [*output_properties1, *output_properties2]
```

```
 <<: *output_properties1
 <<: *output_properties2
```

nvkelso:
So the linter would rather we say:

      <<: [*output_properties, *alternate_fclass]

Versus:

```
      <<: *output_properties
      <<: *alternate_fclass
```

And so on when we are using multiple merge keys. Seems straight forward. You want to tackle that in this PR, too?

peitili:
> So the linter would rather we say:
> 
> ```
>   <<: [*output_properties, *alternate_fclass]
> ```
> 
> Versus:
> 
> ```
>       <<: *output_properties
>       <<: *alternate_fclass
> ```
> 
> And so on when we are using multiple merge keys. Seems straight forward. You want to tackle that in this PR, too?

yup; I think I tackled that already in the PR :)

docs/CONTRIBUTING.md
```diff
@@ -131,15 +131,19 @@ We'll cover the following topics in the next sections:
 - [Push your local branch to the server](CONTRIBUTING.md#9-push-your-local-branch-to-the-server)
 - [Submit a Pull Request (PR)](CONTRIBUTING.md#10-submit-a-pull-request-pr)
 
-### 1. Choose an issue to work on
+### 1. Install prerequisite tool [pre-commit package manager](https://pre-commit.com) on your machine
+Since the repo right now uses python2.7, you need to install pre-commit using python2.7.
```
nvkelso, line 135: Nit, GH markdown likes a newline between headers and next paragraph of content.

nvkelso (approved):
LGTM

Thanks!

PR #2011, opened 2021-11-20
Add more general file linters to pre-commit hook

peitili:
#2004

This is the second phase of linters to improve quality and styles of all type of files. Most of the linters in this phase auto-fixes the violations. The PR change includes all those fixed places by the linters.


iandees (approved):
Looks good, but adding newlines to the .prj and .cpg files might break them. Maybe.

nvkelso (commented):
I think the white space changes in the prj and cpg are fine. But easy enough to try them out in QGIS. I can do that on Monday. Otherwise LGTM

nvkelso:
> I think the white space changes in the prj and cpg are fine. But easy enough to try them out in QGIS. I can do that on Monday. Otherwise LGTM

The CPG loads fine into QGIS with or without the space.

From scratch a PRJ with or without a space loads fine into QGIS. (Some of these PRJ in the project seem to have unrelated problems even though they are in web mercator space correctly.)

nvkelso (approved):

Issue #2014, opened 2021-11-23
Simplify geometry of low zoom Natural Earth features


nvkelso:
At low zooms we don't currently simplify geometries, but we should. This will produce smaller file size.

Historically we've relied on switching back and forth between different Natural Earth scale sets (110m, 50m, then 10m), and that mostly works. But on the first zoom each of those are introduced that produces too much geometry, then on the next zoom it's a little bit much, and next next zoom it's usually fine (or we switched to the next NE scale set).

It may be a simple as setting these `simplify_start` zooms to 0, but there is a lot of OSM mixed up in those configs, so we might want to split the config for NE from OSM, too.

- water layer: https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L152-L167
- earth layer: https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L168-L181
- roads layer: https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L214-L238
- boundaries layer: https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L286-L302

nvkelso:
Fixed via https://github.com/tilezen/vector-datasource/pull/2015.

PR #2015, opened 2021-11-23
Always be simplifying (Natural Earth low zooms)

nvkelso:
Connects with https://github.com/tilezen/vector-datasource/issues/2014.

- **Shifts simplification mostly to zoom 0 from 8** – so Natural Earth features get newly generalized.
- **Start simplifying transit layer** – this was an error from way back when, no wonder that layer just gets bigger and bigger!
- **Truncate min_zoom to 1 decimal place always.** Will promote a little bit more feature merging. 

Todo:

- [ ] Update tests - visual and MVT based QA instead
- [x] Update docs - _we don't document level of generalization for geometries, and already say `min_zoom` is a float (but don't say the precision._

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -152,7 +152,7 @@ layers:
   water:
     geometry_types: [Point, MultiPoint, Polygon, MultiPolygon, LineString, MultiLineString]
     simplify_before_intersect: true
-    simplify_start: 8
+    simplify_start: 0
     transform:
       - vectordatasource.transform.tags_create_dict
       - vectordatasource.transform.tags_name_i18n
@@ -168,7 +168,7 @@ layers:
   earth:
     geometry_types: [Point, MultiPoint, Polygon, MultiPolygon, LineString, MultiLineString]
     simplify_before_intersect: true
-    simplify_start: 8
+    simplify_start: 0
     transform:
       - vectordatasource.transform.tags_create_dict
       - vectordatasource.transform.tags_name_i18n
@@ -192,7 +192,7 @@ layers:
       - vectordatasource.transform.add_id_to_properties
       - vectordatasource.transform.detect_osm_relation
       - vectordatasource.transform.remove_feature_id
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
     sort: vectordatasource.sort.places
     area-inclusion-threshold: 1
   landuse:
@@ -208,12 +208,12 @@ layers:
       - vectordatasource.transform.remove_feature_id
       - vectordatasource.transform.normalize_operator_values
       - vectordatasource.transform.major_airport_detector
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
     sort: vectordatasource.sort.landuse
     area-inclusion-threshold: 1
   roads:
     geometry_types: [LineString, MultiLineString]
-    simplify_start: 8
+    simplify_start: 4
     tolerance: 1.0
     transform:
       - vectordatasource.transform.tags_create_dict
@@ -233,7 +233,7 @@ layers:
       - vectordatasource.transform.road_trim_properties
       - vectordatasource.transform.remove_feature_id
       - vectordatasource.transform.tags_remove
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
     sort: vectordatasource.sort.roads
     area-inclusion-threshold: 1
   buildings:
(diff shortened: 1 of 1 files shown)


queries.yaml
```diff
@@ -208,12 +208,12 @@ layers:
...
       - vectordatasource.transform.normalize_operator_values
       - vectordatasource.transform.major_airport_detector
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
     sort: vectordatasource.sort.landuse
     area-inclusion-threshold: 1
   roads:
     geometry_types: [LineString, MultiLineString]
-    simplify_start: 8
+    simplify_start: 4
```
nvkelso, line 216: Features only come in starting at zoom 4, so simplifying before then is not necessary

queries.yaml
```diff
@@ -296,12 +296,14 @@ layers:
...
       - vectordatasource.transform.detect_osm_relation
       - vectordatasource.transform.remove_feature_id
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
       - vectordatasource.transform.remap_viewpoint_kinds
       - vectordatasource.transform.unpack_viewpoint_claims
     area-inclusion-threshold: 1
   transit:
     geometry_types: [LineString, MultiLineString, Polygon, MultiPolygon]
+    simplify_start: 5
```
nvkelso, line 305: Oops, the transit layer wasn't simplified at all!?

queries.yaml
```diff
@@ -152,7 +152,7 @@ layers:
   water:
     geometry_types: [Point, MultiPoint, Polygon, MultiPolygon, LineString, MultiLineString]
     simplify_before_intersect: true
-    simplify_start: 8
+    simplify_start: 0
```
nvkelso, line 155: All the transforms already run on Natural Earth and OSM features, the `simplify_start` is independent of those and evaluated on each feature on the layer.

queries.yaml
nvkelso: Zoom 5 because: 

- https://github.com/tilezen/vector-datasource/blob/2eb4486761062f67d473424e6526d1ebf7ba7643/yaml/transit.yaml#L26

peitili (approved):

nvkelso:
@peitili Can you generate some after tiles for this PR, please? These 4 coords:

- 0/0/0
- 2/2/1
- 4/4/10
- 5/16/10

queries.yaml
```diff
@@ -777,7 +779,7 @@ post_process:
   - fn: vectordatasource.transform.merge_line_features
     params:
       source_layer: boundaries
-      start_zoom: 8
+      start_zoom: 0
```
nvkelso, line 782: @peitili I added this config change since your review.

nvkelso:
This looks to be doing the right thing, yielding an average of 38% file size savings (ranges from 25% to 60%). Once the boundaries fix is also applied this is likely to be even better (see below).

The boundary merging is working, though the geometry precision wasn't matching the other simplified linework (more obvious in the overzoomed Bubble Wrap screenshot below). Fixed via https://github.com/tilezen/vector-datasource/pull/2015/commits/909d7a575ef6e811d324d3177cb6275e74b6ee63.

The geometry is degenerate over Natural Earth's raw features, but this isn't noticeable except in QGIS overzooming quite a bit. It does make the water and earth features inconsistent with each other, and with the NE coastline. But again not noticeable at 100% zoom or 200% zoom in web map. One solution is to disable the earth layer and draw that with map background only (applying water features over that). For OSM we generate water boundary lines, perhaps we should do that for NE now that it's simplified, too? Anon.

At zoom 5 when transit comes in things get a lot bigger because of unrelated complexity in feature properties (those should really be dropping out). Should address with separate PR.

Screenshots for the 2/1/2 tile at 512 px sizing:

image

Overzoomed in Bubble Wrap:

image

In QGIS:

image

Overzoomed:

image

Overzoomed, again again:

image

queries.yaml
```diff
@@ -280,13 +280,13 @@ layers:
...
       - vectordatasource.transform.normalize_operator_values
-      - vectordatasource.transform.truncate_min_zoom_to_2dp
+      - vectordatasource.transform.truncate_min_zoom_to_1dp
     sort: vectordatasource.sort.pois
     area-inclusion-threshold: 1
   boundaries:
     geometry_types: [Polygon, MultiPolygon, LineString, MultiLineString]
     simplify_before_intersect: true
-    simplify_start: 8
+    simplify_start: 0
```
nvkelso, line 289: @peitili this change is since PR review, based on build QA. The feature merging was working, but the linework detail was significantly more detailed that the water or earth layers on visual inspection. See PR comment for images.

nvkelso: Since boundaries appear in the 0/0/0 tile and zoom 1 tiles, this will have largest impact there on p50 size (not just p99).

peitili (approved):

Issue #2017, opened 2021-12-02
Remove early ferry and aeroway properties


nvkelso:
At zoom 10 even after https://github.com/tilezen/vector-datasource/pull/2008 I still see a lot of individual ferry lines because we're not dropping shield_text and shield_text_length, and network.

This new logic needs to add a `(kind == 'ferry' and zoom < 14) or` clause, like:

```
  - fn: vectordatasource.transform.drop_properties
    params:
      source_layer: roads
      start_zoom: 0
      end_zoom: 15
      properties:
        - network
        - shield_text
        - shield_text_length
      where: >-
        (kind_detail == 'motorway' and zoom <  7) or
        (kind_detail == 'trunk' and zoom <  9) or
        (kind_detail == 'primary' and zoom <  11) or
        (kind_detail == 'secondary' and zoom <  12) or
        (kind_detail == 'tertiary' and zoom <  13) or
        (kind_detail == 'motorway_link' and zoom <  13) or
        (kind_detail == 'trunk_link' and zoom <  13) or
        (kind_detail == 'primary_link' and zoom <  14) or
        (kind_detail == 'secondary_link' and zoom <  14) or
        (kind_detail == 'tertiary_link' and zoom <  14) or
        (kind == 'minor_road' and zoom < 15) or
        (kind == 'path' and zoom < 14) or
        (kind == 'ferry' and zoom < 14) or
        (kind == 'rail' and zoom < 15)
```

For `aeroway` I see too many individual ways at zooms 8-13. 

It's a few places like:

```
  - fn: vectordatasource.transform.drop_properties
    params:
      source_layer: roads
      start_zoom: 0
      end_zoom: 15
      # short-hand for "name" property in the list below means all name-like
      # properties.
      all_name_variants: true
      properties:
        - name
        - ref
        - all_networks
        - all_shield_texts
        - service
        - access
        - osm_relation
      where: >-
        (kind == 'highway' and zoom <  11) or    
        (kind_detail == 'trunk' and zoom <  12) or
        (kind_detail == 'primary' and zoom <  13) or
        (kind_detail == 'secondary' and zoom <  14) or
        (kind_detail == 'tertiary' and zoom <  14) or
        (kind_detail == 'motorway_link' and zoom <  14) or
        (kind_detail == 'trunk_link' and zoom <  14) or
        (kind_detail == 'primary_link' and zoom <  15) or
        (kind_detail == 'secondary_link' and zoom <  15) or
[...]

Issue #2018, opened 2021-12-02
Still more road merging


nvkelso:
After https://github.com/tilezen/vector-datasource/pull/2008 I'm still seeing some road segments with all the same properties as separate features instead of a single merged feature.

[340-new-v1.9-pre.mvt.zip](https://github.com/tilezen/vector-datasource/files/7639593/340-new-v1.9-pre.mvt.zip)

The highlighted geometries are one example from the above tile, where 3 of the 4 adjacent features all have the same properties and should be merged:

![image](https://user-images.githubusercontent.com/853051/144371394-db0acdce-0e4d-40b3-b087-bfb9b1ac1e56.png)

![image](https://user-images.githubusercontent.com/853051/144371385-2de6465e-8177-4ddf-9b8c-2b3fbaf6b6aa.png)

I suspect an order of operations where properties are being stripped after a merge instead of before a merge.

nvkelso:
When I dissolve across all properties in QGIS I see a 56% reduction in feature count:

![image](https://user-images.githubusercontent.com/853051/144372266-6d2f38b3-6be9-42b8-a5fa-e2430cda39b0.png)

nvkelso:
Let's also be more aggressive when names are dropped:

```
  - fn: vectordatasource.transform.drop_properties
    params:
      source_layer: roads
      start_zoom: 0
      end_zoom: 15
      # short-hand for "name" property in the list below means all name-like
      # properties.
      all_name_variants: true
      properties:
        - name
        - ref
        - all_networks
        - all_shield_texts
        - service
        - access
        - osm_relation
      where: >-
        (kind == 'highway' and zoom <  13) or    
        (kind_detail == 'trunk' and zoom <  13) or
        (kind_detail == 'primary' and zoom <  13) or
        (kind_detail == 'secondary' and zoom <  14) or
        (kind_detail == 'tertiary' and zoom <  14) or
        (kind_detail == 'motorway_link' and zoom <  15) or
        (kind_detail == 'trunk_link' and zoom <  15) or
        (kind_detail == 'primary_link' and zoom <  16) or
        (kind_detail == 'secondary_link' and zoom <  16) or
        (kind_detail == 'tertiary_link' and zoom <  16) or
        (kind == 'minor_road' and zoom < 15) or
        (kind == 'path' and zoom < 14) or
        (kind == 'rail' and zoom < 16) or
        (kind == 'ferry' and zoom < 16)
```

[...]

PR #2030, opened 2021-12-07
Remove early ferry and aeroway properties

peitili:
- [ ] Update tests
- [ ] Update docs

These two 512 px sized tiles for the ferry changes:

10/511/340 has 63 ferry lines now... but should just have 3 or 5 after (one per unique min_zoom value), though colour may come into play, there are 5 values there now.
12/2046/1362 has 58 ferry lines now... but should have 4 to 5 after, same deal
For the aeroway changes, this 512 px sized tile:

8/127/85 has 82 lines now, but should just have a couple (based if it's an international airport or not)

#2017

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -1240,6 +1240,7 @@ post_process:
         (kind == 'major_road' and zoom <  13) or
         (kind == 'minor_road' and zoom < 15) or
         (kind == 'path' and zoom < 14) or
+        (kind == 'ferry' and zoom < 15) or
         (kind == 'rail' and zoom < 15)
 
   # drop these "detail" tags to get better merging at zoom < 16
--- a/queries.yaml
+++ b/queries.yaml
@@ -1186,6 +1186,7 @@ post_process:
         (kind_detail == 'tertiary_link' and zoom <  14) or
         (kind == 'minor_road' and zoom < 15) or
         (kind == 'path' and zoom < 14) or
+        (kind == 'ferry' and zoom < 14) or
         (kind == 'rail' and zoom < 15)
 
   - fn: vectordatasource.transform.drop_properties
@@ -1216,6 +1217,7 @@ post_process:
         (kind_detail == 'secondary_link' and zoom <  15) or
         (kind_detail == 'tertiary_link' and zoom <  15) or
         (kind == 'minor_road' and zoom < 15) or
+        (kind == 'aeroway' and zoom < 14) or
         (kind == 'path' and zoom < 14) or
         (kind == 'rail' and zoom < 15) or
         (kind == 'ferry' and zoom < 15)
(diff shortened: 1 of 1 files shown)


peitili:
@nvkelso for this change, do we need to reimport(re-index min zoom stuffs like that) the DB to verify?

nvkelso:
No change to min_zoom or kind (that's done in different config files), this is just stripping out names with a later transform. So you can build using previous database state fine.

nvkelso:
This has a positive effect of dropping the properties, good job!

The tiles still have separate ferry and aeroway features, many more than expected.  For ferries this is more extreme, for aeroways it's much better.

There are also many short segments.

- https://github.com/tilezen/vector-datasource/blob/2017/queries.yaml#L1830-L1848

`drop_length_pixels: 0.1`

Try changing that to 1.0 and see what happens.

peitili:
@nvkelso please see my last commit, I changed this value for two sections.

nvkelso:
> The tiles still have separate ferry and aeroway features, many more than expected. For ferries this is more extreme, for aeroways it's much better.

This sounds like https://github.com/tilezen/vector-datasource/issues/2018

So let's look at the final 1px config change here in a build... but leave the "why still so many separate features" question for #2018.

peitili:
@nvkelso just finished the build with the latest change and here are the result
overall, the number of features dropped a lot, for example for tile 12/2046/1362 number of roads dropped from 1599 to 1442 from px 0.1 change. But if we filter only `ferry` then the change is not that much, only 1 difference.

Similarly tile 10/511/240, number of roads dropped from 1373 to 995. "kind"='ferry doesn't change, both are 46.


For tile 8/127/85 number of roads dropped from 690 to 459.

nvkelso:
Visually the change is not significant (which is good).

Looking at the data that changed (was dropped) in tile v1/512/12/2046/1362.mvt:

It's tiny line segments that are being dropped (shown in blue here, at around 150% enlargement):

image

Zooming in to around 300% enlargement:

image

These segments are sometimes includes in previously merged features, sometimes stand alone. But they are very small and it's OK to drop them – as long as any surrounding features that remain are stroked with a pixel then any geometric gap isn't obvious in the client.

nvkelso (approved):
LGTM

nvkelso:
Zoom 8 at 8/127/85 with 44% less features in roads layer over the property stripping reduction of 12%:

image

Zoom 10 at 10/511/340 with 28% reduction in feature count in roads layer over the property stripping reduction of 2% :

image

The earlier zoom 12 had 10% reduction in feature count over the property stripping reduction of 10%

peitili:
number of integration tests failures is 92 and matched on master, so OK to merge.

Issue #2023, opened 2021-12-05
Address points in buildings layer sometimes missing min_zoom


nvkelso:
Occasionally the `address` points exported in the **buildings** layer are missing `min_zoom` props. 

Besides the [vanilla address](https://github.com/tilezen/vector-datasource/blob/master/yaml/buildings.yaml#L138) nodes in OSM, Tilezen [harvests address points in queries.yaml](https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L937-L940) off ways as well via this transform.py logic:

- https://github.com/tilezen/vector-datasource/blob/2eb4486761062f67d473424e6526d1ebf7ba7643/vectordatasource/transform.py#L2505-L2578

```
        # we're only interested in a very few properties for
        # address points.
        label_properties = dict(
            addr_housenumber=addr_housenumber,
            kind='address')
```

We need to also set `min_zoom=17` in that section, like:

```
        # we're only interested in a very few properties for
        # address points.
        label_properties = dict(
            addr_housenumber=addr_housenumber,
            min_zoom=17,
            kind='address')
```

nvkelso:
PR is https://github.com/tilezen/vector-datasource/pull/2032.

nvkelso:
Relates to https://github.com/tilezen/tilequeue/pull/402.

PR #2032, opened 2021-12-09
Add min_zoom = 17 for generated address points

nvkelso:
Connects with https://github.com/tilezen/vector-datasource/issues/2023.

- [x] Update tests
- [x] ~Update docs~ just a bug fix

### DIFF
--- a/integration-test/2032-building-address-points-missing-min-zooms.py
+++ b/integration-test/2032-building-address-points-missing-min-zooms.py
@@ -1,0 +1,136 @@
+# -*- encoding: utf-8 -*-
+import dsl
+
+from . import FixtureTest
+
+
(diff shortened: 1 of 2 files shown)


nvkelso:
Relates to https://github.com/tilezen/tilequeue/pull/402 which removes overstuffed features from tiles when their `min_zoom` is larger than the tile's **nominal zoom**. But if the `min_zoom` is **null** then they aren't pruned! This sets the `min_zoom` to 17 so the generated address points will get pruned.

nvkelso:
LGTM

I'm the author so I can't approve, @travisgrigsby you want to do those honors?

travisgrigsby (approved):
Some of the best test code I've ever witnessed in all my years of witnessing test code

Issue #2021, opened 2021-12-02
Cleanup 8 remaining integration test failures


nvkelso:
Down to just 8 failures (down from 60)!

```
FAIL: test_cycleway_with_international_route (integration-test.1170-very-early-paths-and-bike-routes.VeryEarlyPathsAndBikeRoutes)
AssertionError: Did not find feature including properties {'kind': 'path', 'is_bicycle_related': True, 'bicycle_network': 'icn'} (because layer 'roads' was empty)

======================================================================
FAIL: test_path_with_international_route (integration-test.1170-very-early-paths-and-bike-routes.VeryEarlyPathsAndBikeRoutes)
AssertionError: Did not find feature including properties {'kind': 'path', 'walking_network': 'iwn'} (because layer 'roads' was empty)

======================================================================
FAIL: test_path_with_national_route (integration-test.1170-very-early-paths-and-bike-routes.VeryEarlyPathsAndBikeRoutes)
AssertionError: Did not find feature including properties {'kind': 'path', 'walking_network': 'nwn'} (because layer 'roads' was empty)

======================================================================
FAIL: test_full_lists_disappear_by_zoom_12 (integration-test.1194-bus-route-refs.BusRouteRefs)
AssertionError: Did not find feature including properties {'bus_network': <type 'NoneType'>, 'bus_shield_text': None} (because layer 'roads' was empty)

======================================================================
FAIL: test_tourism (integration-test.440-zoos-and-other-attractions-tourism.ZoosAndOtherAttractionsTourism)
AssertionError: Did not find feature including properties {'kind': 'theme_park', 'id': -1228099} (because layer 'pois' was empty)

======================================================================
FAIL: test_driveway_nwn (integration-test.596-add-hiking-routes.AddHikingRoutes)
AssertionError: Did not find feature including properties {'kind_detail': 'service', 'kind': 'minor_road', 'walking_network': 'nwn', 'service': 'driveway'} (because layer 'roads' was empty)

======================================================================
FAIL: test_minor_road_nwn (integration-test.596-add-hiking-routes.AddHikingRoutes)
AssertionError: Did not find feature including properties {'kind_detail': 'residential', 'kind': 'minor_road', 'walking_network': 'nwn'} (because layer 'roads' was empty)

======================================================================
FAIL: test_service_nwn (integration-test.596-add-hiking-routes.AddHikingRoutes)
[...]

PR #2036, opened 2021-12-13
fix linter on master

peitili:
There are some linter violations that might be brought by 

https://github.com/tilezen/vector-datasource/pull/2001
https://github.com/tilezen/vector-datasource/pull/2015

so fixing those


iandees:
FWIW, I had pre-commit enabled/installed when I was working on this code and it didn't catch any of these.

iandees (approved):

peitili:
> FWIW, I had pre-commit enabled/installed when I was working on this code and it didn't catch any of these.

that's because your PR branch didn't rebase some of the newer linters, so I suggested something like I mentioned in here

https://github.com/tilezen/vector-datasource/issues/2021#issuecomment-992996482

PR #2040, opened 2022-01-03
Make locality thinning grids smaller

iandees:
Extending #2001 with a few tweaks to the queries.yaml arguments for the locality thinning.

The `grid_width: 3` was way too small, resulting in way too few localities showing up:

image

Changing to `grid_width: 10` results in a more reasonable number of localities:

image


nvkelso (changes requested):
If this is a 512 px tile size versus 256 issue, then should we try doubling the grid_width instead? Your example from 3 > 10 looks far too dense to me.

queries.yaml
```diff
@@ -981,7 +981,7 @@ post_process:
       end_zoom: 9
       items_matching: { kind: locality }
       max_items: 1
-      grid_width: 3
+      grid_width: 10
```
nvkelso, line 984: To 6?

queries.yaml
```diff
@@ -994,7 +994,7 @@ post_process:
       end_zoom: 11
       items_matching: { kind: locality }
       max_items: 1
-      grid_width: 6
+      grid_width: 18
```
nvkelso, line 997: To 12?

iandees:
Sounds good. I was trying to roughly match the styled map from before thinning (where the client was doing all the thinning), and a plain doubling seemed to be missing some of the smaller localities. But here's a new round of screenshots with plain doubling:

 &nbsp;  | Before thinning | This PR as of this comment
---|---|---
z7 | image | image
z8 | image | image
z9 | image | image

queries.yaml
nvkelso: Resolved via https://github.com/tilezen/vector-datasource/pull/2040/commits/e0fe15091eec17cd2bd3cc23dfb853fd2d3d47d1.

queries.yaml
nvkelso: Resolved via https://github.com/tilezen/vector-datasource/pull/2040/commits/e0fe15091eec17cd2bd3cc23dfb853fd2d3d47d1.

nvkelso (commented):
Looking good, let's see it in a build :)

nvkelso:
Build says the numbers still aren't right.

If Tilequeue is operating on a larger window (like 4x4 grid of requested tiles), then we should we instead apply multiplier to vector-datasource for an allowlisted list of config props values?

queries.yaml
```diff
@@ -1052,7 +1052,7 @@ post_process:
       end_zoom: 9
       items_matching: { kind: locality }
       max_items: 1
-      grid_width: 3
+      grid_width: 24
```
nvkelso, line 1055: Try `12` (which is 4 times original value of 3)

queries.yaml
```diff
@@ -1065,7 +1065,7 @@ post_process:
       end_zoom: 11
       items_matching: { kind: locality }
       max_items: 1
-      grid_width: 6
+      grid_width: 48
```
nvkelso, line 1068: Try `24` (which is 4 times original value of 6)

queries.yaml
```diff
@@ -1078,7 +1078,7 @@ post_process:
       end_zoom: 13
       items_matching: { kind: locality }
       max_items: 1
-      grid_width: 12
+      grid_width: 96
```
nvkelso, line 1081: Try `48` (which is 4 times original value of 12)

queries.yaml
iandees: Fixed in 50bb8982.

queries.yaml
iandees: Resolved in 50bb8982.

queries.yaml
iandees: Resolved in 50bb8982.

queries.yaml
```diff
@@ -1052,7 +1052,7 @@ post_process:
       end_zoom: 9
       items_matching: { kind: locality }
       max_items: 1
-      grid_width: 3
```
nvkelso, line 1055: This one feels a little small in the build, so we're investigating increasing it to "4" or changing "max_items" to 2.

iandees: After we tried some variations locally, in d602ff1 I switched the first zoom range to grid_width "6" (x 4 for the metatile) with max_items 1.

Note that this makes the first and second zoom ranges in here do the same thing. If this ends up looking OK, we can consolidate the two ranges into one wider range.

iandees:
Sounds like we're happy with this most recent change. So last couple things here before we merge:

- [ ] Fix integration test 1999
- [ ] Merge the first two queries.yaml sections using gridded thinning function (for zooms z8-z9 and z9-z11) into a single range, since their values are the same

nvkelso:
We also need to fix the config and logic so it works for both Tilequeue and Tileserver, please.

iandees:
> We also need to fix the config and logic so it works for both Tilequeue and Tileserver, please.

I haven't found a path to do that yet. I'll keep thinking.

iandees:
After talking with Travis for a bit this afternoon we weren't able to find an approach that handles both tileserver and tilequeue. I'm going to spend the rest of the evening trying to get `tilequeue meta-tile` running locally so I can do some debugging and maybe find something we're missing.

The sticking point is that metatile-based builds run the functions in queries.yaml (like grid thinning) once for all the data in the metatile and the functions only get the `nominal_zoom` to work from. For metatile builds, that is the zoom level of the metatile that is being built, for tileserver-based requests that is the zoom level of the tile being requested (including adjustments for 512 or 256-sized tiles).

queries.yaml
```diff
@@ -1052,7 +1052,7 @@ post_process:
...
+      grid_width: 52181.0113
```
nvkelso, line 1055: It's not obvious this is in meters... should the param name be `grid_width_meters`?

Is the fractional meters necessary?

vectordatasource/transform.py
```diff
@@ -3189,8 +3183,8 @@ def keep_n_features_gridded(ctx):
         # Calculate the bucket to put this feature in.
         # Note that this purposefully allows for buckets outside the unpadded bounds
         # so we can bucketize the padding area, too.
-        bucket_x = int((shape.x - minx) / bucket_width)
-        bucket_y = int((shape.y - miny) / bucket_height)
+        bucket_x = int((shape.x - minx) / grid_width)
```
nvkelso, line 3186: Is this already flexible to whole number grid_width meters? It's rounding it to an int, so maybe? If it's already padded then seems like it's fine.

nvkelso (approved):
Two nits but otherwise LGTM and works in a QA build.

queries.yaml
iandees: I suppose they're not necessary, but using fractions means we get closer to exactly 3x3 subdivisions in a single 256px tile at this zoom level. If we go to integer values here, then the tiny subdivisions at the edge of the tile are slightly bigger and are more likely to get an extra feature in there.

vectordatasource/transform.py
iandees: It's rounding to an int after doing floating point math.

queries.yaml
iandees: Changed to `grid_width_meters` in 2eca280.

nvkelso (approved):

PR #2010, opened 2021-11-19
Tile size: water only changes

nvkelso:
Connects to #1998 and replaces PR #2000 (RIP).

- [ ] Update tests
- [ ] Update docs

Issue #2003, opened 2021-11-18
Remove early water labels


nvkelso:
Picking up from where https://github.com/tilezen/vector-datasource/issues/1477 leftoff... There are still a lot of water labels in the mid-zoom tiles.

This is controlled with this config: 

- https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L805-L837

Edit something like (adjusting low zooms and then moving others by a zoom to filter out along the way each zoom):

```
  # some water polygon features are too small to need a label - and if we label
  # them then the label is often so prominent that it hides the polygon it's
  # for! see the images here for examples of this:
  # https://github.com/tilezen/vector-datasource/issues/1477#issuecomment-447162484
  #
  # we want to reset min_zoom on these features and screen them out if they
  # don't meet minimum sizes per zoom.
  - fn: vectordatasource.transform.update_min_zoom
    params:
      source_layer: water
      start_zoom: 4
      end_zoom: 16
      # the inclusion of "None" in the list might seem odd, but None compares
      # numerically smaller than any integer, so acts as a default for this
      # table.
      min_zoom: >-
        [min_zoom for min_zoom, area_threshold in [
            (4,  120000000000),
            (5,   80000000000),
            (6,   40000000000),
            (7,   10000000000),
            (8,     500000000),
            (9,     200000000),
            (10,     50000000),
            (11,     10000000),
            (12,      1000000),
            (13,       500000),
            (14,        50000),
            (15,        10000),
            (16,         None),
        ] if area >= area_threshold][0]
      where: >-
        kind == 'lake' and label_placement
```

Then there are also removing line labels, which is controlled around, it should be done before label placements are generated:

- https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L508-L515

For the following kinds:

- kind: sea, zoom < 4
- kind: lake, zoom < 5
- kind: river, zoom < 12
- kind: [canal], zoom < 13
- kind: [dam,stream], zoom < 14
- kind: [ditch,drain], zoom < 15
- is_tunnel: true, zoom < 16

Something like:

```
  - fn: vectordatasource.transform.drop_properties
    params:
      source_layer: water
      start_zoom: 0
      end_zoom: 16
      # short-hand for "name" property in the list below means all name-like
      # properties.
      all_name_variants: true
      properties:
        - name
      where: >-
        (kind == 'sea' and zoom < 4) or
        (kind == 'lake' and zoom < 5) or
[...]

Issue #984, opened 2016-08-23
lakes lose `kind: lake` tag > z9


meetar:
* **What did you see?**

Past zoom 9, features of `kind: lake` lose that tag.

* **What did you expect to see?**

Tag would remain.

* **What map location are you having problems with?** _City and country are helpful, as well as tile coordinates or latitude / longitude._  

http://tangrams.github.io/explorer/#9.0/35.2348/136.0881/kind/lake

* **Screenshot?** _Props for animated gifs._

![biwako](https://cloud.githubusercontent.com/assets/459970/17895921/a5963a26-691c-11e6-9719-1b537569bf34.gif)

zerebubuth:
Looks like we're not picking up [this lake from OSM](http://www.openstreetmap.org/relation/63499), so it drops out when we transition from NE. The relation multi-polygon looks like it should come through as a lake (it's tagged `natural=water`), so further investigation is required.

zerebubuth:
Oh, wait... I see what's happening here. The lake is tagged `natural=water, water=lake`, and we don't have a special case for that, we're just assigning `kind: water`. There are probably many features out there which NE considers lakes but which are only tagged as more general "water" in OSM.

@nvkelso given that there's a `water=lake` additional tag on this particular `natural=water`, should we use that to "upgrade" it to a `kind: lake`?

nvkelso:
Yeah, I think so. Tagging for v1.1.

PR #2047, opened 2022-01-20
Restore early lake labels, add kind_detail for lakes and etc

nvkelso:
Connects with https://github.com/tilezen/vector-datasource/issues/2003 and followup of https://github.com/tilezen/vector-datasource/pull/2010 to restore "lake" water labels at low zooms.

Also non-breaking change implementation of #984 for the v1.x series (which should stay open as a separate breaking change for v2.x series).

- [x] Update tests
- [x] Update docs

### DIFF
--- a/integration-test/2047-restore-early-lake-labels.py
+++ b/integration-test/2047-restore-early-lake-labels.py
@@ -1,0 +1,117 @@
+# -*- encoding: utf-8 -*-
+import dsl
+from shapely.wkt import loads as wkt_loads
+
+from . import FixtureTest
+
+
+class BoundsTest(FixtureTest):
+    def generate_fixtures(self, *objs):
+        from dsl import Feature
+
+        boundaries = []
+        for feature in objs:
+            if feature.shape.geom_type in ('Polygon', 'MultiPolygon'):
+                props = feature.properties.copy()
+                props['boundary'] = True
+                boundary = Feature(
+                    feature.fid,
+                    feature.shape.boundary,
+                    props)
+                boundaries.append(boundary)
+
+        new_objs = list(objs)
+        new_objs.extend(boundaries)
+        FixtureTest.generate_fixtures(self, *new_objs)
+
+
+class WaterKinds(FixtureTest):
+    def test_water_reservoir(self):
+        self.generate_fixtures(dsl.way(1, wkt_loads(
+            'POLYGON ((-122.421611012467 37.80395991667648, '
+            '-122.420227786592 37.80413210674329, -122.420159873957 '
+            '37.80408277780808, -122.420117832801 37.80383748120349, '
+            '-122.42154480663 37.80365748294528, -122.421611012467 '
+            '37.80395991667648))'),
+            {
+            u'way_area': u'6936.45',
+            u'natural': u'water',
+            u'source': u'openstreetmap.org',
+            u'water': u'reservoir',
+            u'name': u'Francisco Reservoir',
+            u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {
+                'kind': 'water',
+                'reservoir': True,
+                'kind_detail': 'lake'
+            })
+
+    def test_water_lagoon(self):
+        self.generate_fixtures(dsl.way(1, wkt_loads(
+            'POLYGON ((-122.421611012467 37.80395991667648, '
+            '-122.420227786592 37.80413210674329, -122.420159873957 '
+            '37.80408277780808, -122.420117832801 37.80383748120349, '
+            '-122.42154480663 37.80365748294528, -122.421611012467 '
+            '37.80395991667648))'),
+            {
+            u'natural': u'water',
+            u'way_area': u'6936.45',
+            u'source': u'openstreetmap.org',
+            u'water': u'lagoon',
+            u'name': u'Francisco Reservoir',
+            u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {
+                'kind': 'water',
+                'alkaline': True,
+                'kind_detail': 'lake'
+            })
+
+    def test_water_ditch(self):
+        self.generate_fixtures(dsl.way(1, wkt_loads(
+            'POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'),
+            {
+            u'natural': u'water',
+            u'way_area': u'6936.45',
+            u'source': u'openstreetmap.org',
+            u'water': u'ditch',
+            u'name': u'Francisco Reservoir',
+            u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {
+                'kind': 'water',
+                'kind_detail': 'ditch'
+            })
+
+    def test_name_drop(self):
+        import dsl
+
+        for z in range(6, 13):
+            x, y = (3, 4)
+            area = 265548000
+
+            self.generate_fixtures(
+                dsl.way(1, dsl.box_area(z, x, y, area, include_boundary=True), {
+                    'name': u'Clear Lake',
+                    'natural': u'water',
+                    'type': u'multipolygon',
+                    'water': u'lake',
+                    'wikidata': u'Q1099503',
+                    'source': u'openstreetmap.org',
+                }),
+            )
+
+            self.assert_has_feature(
+                z, x, y, 'water', {
+                    'kind': 'water',
+                    'kind_detail': 'lake',
+                    'name': 'Clear Lake',
+                    'min_zoom': 6.0,
+                })
--- a/integration-test/2047-restore-early-lake-labels.py
+++ b/integration-test/2047-restore-early-lake-labels.py
@@ -119,12 +119,10 @@ class WaterKinds(FixtureTest):
                                                 })
 
             if z >= 9:
-                with self.features_in_tile_layer(z, x, y, 'water') as features:
-                    for f in features:
-                        if f['geometry']['type'] == 'Point':
-                            assert f['id'] is None
-                            assert f['properties']['name'] == 'Clear Lake'
-                            assert f['properties']['kind'] == 'water'
-                            assert f['properties']['kind_detail'] == 'lake'
-                            assert f['properties']['min_zoom'] == 9.0
-                            assert f['properties']['label_placement'] is True
+                self.assert_has_feature(z, x, y, 'water', {
+                    'kind': 'water',
+                    'name': 'Clear Lake',
+                    'kind_detail': 'lake',
+                    'label_placement': True,
+                    'min_zoom': 9.0,
+                })
--- a/integration-test/2047-restore-early-lake-labels.py
+++ b/integration-test/2047-restore-early-lake-labels.py
@@ -90,7 +90,7 @@ class WaterKinds(FixtureTest):
                 'kind_detail': 'ditch'
             })
 
-    def test_name_drop(self):
+    def test_name_not_drop(self):
         import dsl
 
         for z in range(6, 13):
--- a/integration-test/2047-restore-early-lake-labels.py
+++ b/integration-test/2047-restore-early-lake-labels.py
@@ -71,7 +71,7 @@ class WaterKinds(FixtureTest):
                 'kind_detail': 'ditch'
             })
 
-    def test_name_not_drop(self):
+    def test_prop_drop(self):
         import dsl
 
         lon, lat = (-122.417169, 37.769196)
@@ -93,31 +93,38 @@ class WaterKinds(FixtureTest):
                 self.assert_no_matching_feature(z, x, y, 'water',
                                                 {
                                                     'kind': 'water',
-                                                    'kind_detail': 'lake',
                                                 })
-            if z >= 6:
-                self.assert_has_feature(
-                    z, x, y, 'water', {
-                        'kind': 'water',
-                        'kind_detail': 'lake',
-                        'name': 'Clear Lake',
-                        'min_zoom': 6.0,
-                    })
+            if 6 <= z <= 14:
+                with self.features_in_tile_layer(z, x, y, 'water') as features:
+                    for f in features:
+                        if f['geometry']['type'] == 'Polygon':
+                            assert f['id'] is None
+                            assert f['properties']['min_zoom'] == 6
+                            assert 'name' not in f['properties']
+                            assert 'old_name' not in f['properties']
+            if z == 15:
+                with self.features_in_tile_layer(z, x, y, 'water') as features:
+                    for f in features:
+                        if f['geometry']['type'] == 'Polygon':
+                            assert f['id'] is None
+                            assert f['properties']['name'] == 'Clear Lake'
+                            assert f['properties']['kind'] == 'water'
+                            assert f['properties']['kind_detail'] == 'lake'
 
-            # the following assert will start to fail on z == 6
-            if z < 6:
+            if z < 9:
                 self.assert_no_matching_feature(z, x, y, 'water',
                                                 {
                                                     'kind': 'water',
                                                     'label_placement': True,
                                                 })
 
-            # if z >= 9:
-            #     self.assert_has_feature(
-            #         z, x, y, 'water', {
-            #             'kind': 'water',
-            #             'kind_detail': 'lake',
-            #             'name': 'Clear Lake',
-            #             'label_placement': True,
-            #             'min_zoom': 6.0,
-            #         })
+            if z >= 9:
+                with self.features_in_tile_layer(z, x, y, 'water') as features:
+                    for f in features:
+                        if f['geometry']['type'] == 'Point':
+                            assert f['id'] is None
+                            assert f['properties']['name'] == 'Clear Lake'
+                            assert f['properties']['kind'] == 'water'
+                            assert f['properties']['kind_detail'] == 'lake'
+                            assert f['properties']['min_zoom'] == 9.0
+                            assert f['properties']['label_placement'] is True
--- a/integration-test/2047-restore-early-lake-labels.py
+++ b/integration-test/2047-restore-early-lake-labels.py
@@ -104,7 +104,8 @@ class WaterKinds(FixtureTest):
                         'min_zoom': 6.0,
                     })
 
-            if z < 9:
+            # the following assert will start to fail on z == 6
+            if z < 6:
                 self.assert_no_matching_feature(z, x, y, 'water',
                                                 {
                                                     'kind': 'water',
--- a/integration-test/2047-restore-early-lake-labels.py
+++ b/integration-test/2047-restore-early-lake-labels.py
@@ -1,28 +1,9 @@
 # -*- encoding: utf-8 -*-
 import dsl
 from shapely.wkt import loads as wkt_loads
+from tilequeue.tile import deg2num
 
 from . import FixtureTest
-
-
-class BoundsTest(FixtureTest):
-    def generate_fixtures(self, *objs):
-        from dsl import Feature
-
-        boundaries = []
-        for feature in objs:
-            if feature.shape.geom_type in ('Polygon', 'MultiPolygon'):
-                props = feature.properties.copy()
-                props['boundary'] = True
-                boundary = Feature(
-                    feature.fid,
-                    feature.shape.boundary,
-                    props)
-                boundaries.append(boundary)
-
-        new_objs = list(objs)
-        new_objs.extend(boundaries)
-        FixtureTest.generate_fixtures(self, *new_objs)
 
 
 class WaterKinds(FixtureTest):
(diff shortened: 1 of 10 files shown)


peitili:
there is a linter violations, you might want to fix them using the pre-commit

peitili:
I am not sure why but the integration tests always failed with context deadline exceed.

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
         self.assert_has_feature(
             16, 10481, 25324, 'water',
             {'kind': 'lake', 'reservoir': True})
+
+    # this test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_reservoir(self):
```
peitili, line 28: @nvkelso please advice if we need some non-fake data and appreciate it if you can let me know which ones.

Also please advice whether we should add these tests here in this file or other issue number.

nvkelso (commented):
Please also add a test for the low zoom removals.

Clear Lake in California: [-4046481](https://www.openstreetmap.org/relation/4046481) with an area of `265548000`.

That features name was being stripped until zoom 12 in error, but with this change it should be visible in a zoom from zoom 5+ depending on it's area, which looks like results in a min_zoom of 6. Since it's OSM it'll not actually be in zoom 6, so test a zoom 8 tile and see if that feature is there with that min_zoom. 

This is the 256px tile I'd expect to see such a label in:

- https://tile.openstreetmap.org/10/162/391.png

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
         self.assert_has_feature(
             16, 10481, 25324, 'water',
             {'kind': 'lake', 'reservoir': True})
+
+    # This test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_reservoir(self):
+        # Francisco Reservoir, SF
+        self.generate_fixtures(dsl.way(386662458, wkt_loads('POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'), {
```
nvkelso, line 30: Is this the right OSM ID?

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+                               u'way_area': u'6936.45', u'source': u'openstreetmap.org', u'water': u'reservoir', u'name': u'Francisco Reservoir', u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {'kind': 'lake', 'reservoir': True, 'kind_detail': 'lake'})
+
+    # This test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_lagoon(self):
+        # Francisco Reservoir, SF
+        self.generate_fixtures(dsl.way(386662458, wkt_loads('POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'), {
```
nvkelso, line 40: Seems like wrong OSM ID.

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+                               u'way_area': u'6936.45', u'source': u'openstreetmap.org', u'water': u'lagoon', u'name': u'Francisco Reservoir', u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {'kind': 'lake', 'alkaline': True, 'kind_detail': 'lake'})
+
+    # This test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_ditch(self):
+        # Francisco Reservoir, SF
+        self.generate_fixtures(dsl.way(386662458, wkt_loads('POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'), {
```
nvkelso, line 50: Wrong OSM ID. If it's not relevant just make it be `1` instead.

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+                               u'way_area': u'6936.45', u'source': u'openstreetmap.org', u'water': u'ditch', u'name': u'Francisco Reservoir', u'covered': u'no'}))
```
nvkelso, line 51: I would expect another tag here: 

`'natural': u'water'`

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {'kind': 'lake', 'kind_detail': 'ditch'})
```
nvkelso, line 55: With the new `natural=water` tag addition above, I'd expect to see output more like:

>            {'kind': 'water', 'kind_detail': 'ditch'})

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+
+    # This test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_lagoon(self):
+        # Francisco Reservoir, SF
+        self.generate_fixtures(dsl.way(386662458, wkt_loads('POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'), {
+                               u'way_area': u'6936.45', u'source': u'openstreetmap.org', u'water': u'lagoon', u'name': u'Francisco Reservoir', u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {'kind': 'lake', 'alkaline': True, 'kind_detail': 'lake'})
```
nvkelso, line 45: With the new `natural=water` tag addition above, I'd expect to see output more like:

>             {'kind': 'water', 'alkaline': True, 'kind_detail': 'lake'})

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {'kind': 'lake', 'reservoir': True, 'kind_detail': 'lake'})
+
+    # This test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_lagoon(self):
+        # Francisco Reservoir, SF
+        self.generate_fixtures(dsl.way(386662458, wkt_loads('POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'), {
+                               u'way_area': u'6936.45', u'source': u'openstreetmap.org', u'water': u'lagoon', u'name': u'Francisco Reservoir', u'covered': u'no'}))
```
nvkelso, line 41: I would expect another tag here: 

`'natural': u'water'`

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
         self.assert_has_feature(
             16, 10481, 25324, 'water',
             {'kind': 'lake', 'reservoir': True})
+
+    # This test is for https://github.com/tilezen/vector-datasource/pull/2047
+    def test_water_reservoir(self):
+        # Francisco Reservoir, SF
+        self.generate_fixtures(dsl.way(386662458, wkt_loads('POLYGON ((-122.421611012467 37.80395991667648, -122.420227786592 37.80413210674329, -122.420159873957 37.80408277780808, -122.420117832801 37.80383748120349, -122.42154480663 37.80365748294528, -122.421611012467 37.80395991667648))'), {
+                               u'way_area': u'6936.45', u'source': u'openstreetmap.org', u'water': u'reservoir', u'name': u'Francisco Reservoir', u'covered': u'no'}))
```
nvkelso, line 31: I would expect another tag here: 

`'natural': u'water'`

integration-test/628-standardize-water-kinds.py
```diff
@@ -23,3 +23,33 @@ def test_reservoir(self):
...
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {'kind': 'lake', 'reservoir': True, 'kind_detail': 'lake'})
```
nvkelso, line 35: With the new natural=water tag addition above, I'd expect to see output more like:

>             {'kind': 'water', 'reservoir': True, 'kind_detail': 'lake'})

integration-test/628-standardize-water-kinds.py
nvkelso: It's ok, but a separate new test would be better since you're not otherwise modifying any of the other tests in this old file.

nvkelso: I mentioned in the general PR, but responding directly here too. 

Please also add a test for the low zoom removals.

Clear Lake in California: [-4046481](https://www.openstreetmap.org/relation/4046481) with an area of 265548000. You can copy the props but just use DSL to make a box that's as large as that area instead of copying all the detailed polygon in.

That features name was being stripped until zoom 12 in error, but with this change it should be visible in a zoom from zoom 5+ depending on it's area, which looks like results in a min_zoom of 6. Since it's OSM it'll not actually be in zoom 6, so test a zoom 10 tile and see if that feature is there with that min_zoom.

This is the 256px tile I'd expect to see such a label in:

https://tile.openstreetmap.org/10/162/391.png

integration-test/628-standardize-water-kinds.py
nvkelso: If it's made up, then use `1` instead.

integration-test/628-standardize-water-kinds.py
nvkelso: If it's made up, then use `1` instead.

integration-test/628-standardize-water-kinds.py
nvkelso: Please also add a test for the low zoom removals.

Clear Lake in California: [-4046481](https://www.openstreetmap.org/relation/4046481) with an area of 265548000. You can copy the props but just use DSL to make a box that's as large as that area instead of copying all the detailed polygon in.

That features name was being stripped until zoom 12 in error, but with this change it should be visible in a zoom from zoom 5+ depending on it's area, which looks like results in a min_zoom of 6. Since it's OSM it'll not actually be in zoom 6, so test a zoom 10 tile and see if that feature is there with that min_zoom.

This is the 256px tile I'd expect to see such a label in:

https://tile.openstreetmap.org/10/162/391.png

nvkelso: (It's OK but not great to edit the old test – but since you're not fixing anything in the old test, it's better to make a new file with the PR number in the name.)

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,117 @@
...
+            u'covered': u'no'}))
+
+        self.assert_has_feature(
+            16, 10481, 25324, 'water',
+            {
+                'kind': 'water',
+                'kind_detail': 'ditch'
+            })
+
+    def test_name_not_drop(self):
```
peitili, line 93: Traceback (most recent call last):
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/2047-restore-early-lake-labels.py", line 116, in test_name_drop
    'min_zoom': 6.0,
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/__init__.py", line 1492, in assert_has_feature
    self.test_instance.assert_has_feature(z, x, y, layer, props)
  File "integration-test/__init__.py", line 1325, in assert_has_feature
    self.assertions.assert_has_feature(z, x, y, layer, props)
  File "integration-test/__init__.py", line 1154, in assert_has_feature
    'layer %r was empty)' % (properties, layer))
AssertionError: Did not find feature including properties {'kind_detail': 'lake', 'kind': 'water', 'min_zoom': 6.0, 'name': 'Clear Lake'} (because layer 'water' was empty)

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,117 @@
...
+        import dsl
+
+        for z in range(6, 13):
```
peitili, line 96: @nvkelso I found that this test is strange it can only pass if the zoom is in this range

if I specify zoom <=5 it will complain the following, hmm why? I thought the water layer should still exist and only the name label shouldn't be there.

```
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/2047-restore-early-lake-labels.py", line 116, in test_name_drop
    'min_zoom': 6.0,
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/__init__.py", line 1492, in assert_has_feature
    self.test_instance.assert_has_feature(z, x, y, layer, props)
  File "integration-test/__init__.py", line 1325, in assert_has_feature
    self.assertions.assert_has_feature(z, x, y, layer, props)
  File "integration-test/__init__.py", line 1154, in assert_has_feature
    'layer %r was empty)' % (properties, layer))
AssertionError: Did not find feature including properties {'kind_detail': 'lake', 'kind': 'water', 'min_zoom': 6.0, 'name': 'Clear Lake'} (because layer 'water' was empty)
```

for any zoom >= 13 it will throw this error

```
FAIL: test_name_drop (integration-test.2047-restore-early-lake-labels.WaterKinds)
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/2047-restore-early-lake-labels.py", line 100, in test_name_drop
    dsl.way(1, dsl.box_area(z, x, y, area, include_boundary=True), {
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/dsl.py", line 184, in box_area
    _check(centre_x - 0.5 * size_x),
  File "/Users/pli2/Snapchat/Dev/vector-datasource/integration-test/dsl.py", line 180, in _check
    assert -half_earth_circum <= coord <= half_earth_circum
AssertionError
```

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,117 @@
+# -*- encoding: utf-8 -*-
+import dsl
+from shapely.wkt import loads as wkt_loads
+
+from . import FixtureTest
+
+
+class BoundsTest(FixtureTest):
```
peitili, line 8: this is unnecessary, should remove.

integration-test/628-standardize-water-kinds.py
peitili: fixed in the latest commit

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for z in range(6, 13):
+            x, y = (3, 4)
+            area = 265548000
+
+            self.generate_fixtures(
+                dsl.way(1, dsl.box_area(z, x, y, area, include_boundary=True), {
+                    'name': u'Clear Lake',
+                    'natural': u'water',
+                    'type': u'multipolygon',
```
nvkelso, line 104: Please remove this, just confuses things since we're faking a box for the shape.

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,117 @@
...
+                    'source': u'openstreetmap.org',
+                }),
+            )
+
+            self.assert_has_feature(
+                z, x, y, 'water', {
+                    'kind': 'water',
+                    'kind_detail': 'lake',
+                    'name': 'Clear Lake',
+                    'min_zoom': 6.0,
```
nvkelso, line 116: For the generated label point geometry (the cruz of this PR), you want to also test for:

>                     'label_placement': True,

Otherwise you're testing the polygon which has it's own min_zoom range.

To clarify... the polygon should be visible from zoom 6 conceptually based on it's area grading of `265548000` sq meters in https://github.com/tilezen/vector-datasource/blob/master/yaml/water.yaml#L23-L46.

```
        - [ 5,   500000000 ]
        - [ 6,   200000000 ]
        - [ 7,    50000000 ]
```

For the label we do a remapping of the min_zoom based on area grading of `265548000` sq metershere: https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L843-L878, so I expect min_zoom 9 for the label.

```
           (8,     500000000),
            (9,     200000000),
            (10,     40000000),
```

Since we don't transition from NE to OSM varies based on layer, but zoom 4 presumably for water from the osm polygons table:

https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L23-L26

So to recap: 
- the lake polygon shouldn't be visible until zoom 6, and it won't have a name property ever
- the lake label shouldn't be visible until zoom 9 (but wasn't visible until zoom 12 in error before), so we are testing for it's presence at zooms 9, 10, and 11 (and that it's not present at 8, and still present at zoom 12).

peitili: @nvkelso please see the latest commit about that integration test, the behavior is not what you described here. 

what the integ test told us are:

1. We are seeing `name` property appear since zoom 6 for the lake polygon
2. We are seeing label_placement since zoom 6

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,123 @@
...
+                self.assert_has_feature(
+                    z, x, y, 'water', {
+                        'kind': 'water',
+                        'kind_detail': 'lake',
+                        'name': 'Clear Lake',
+                        'min_zoom': 6.0,
+                    })
+
+            # the following assert will start to fail on z == 6
+            if z < 6:
```
peitili, line 108: @nvkelso  starting z==6 there are feature in the water layer


`{'label_placement': True, 'kind': 'water', 'collision_rank': 487, 'name': 'Clear Lake', 'area': 265548000, 'sort_rank': 204, 'kind_detail': 'lake', 'source': 'openstreetmap.org', 'min_zoom': 6.0, 'wikidata_id': 'Q1099503', 'id': 1}`

integration-test/2047-restore-early-lake-labels.py
nvkelso: Go for it!

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,123 @@
...
+                                                {
+                                                    'kind': 'water',
+                                                    'kind_detail': 'lake',
+                                                })
+            if z >= 6:
+                self.assert_has_feature(
+                    z, x, y, 'water', {
+                        'kind': 'water',
+                        'kind_detail': 'lake',
+                        'name': 'Clear Lake',
```
nvkelso, line 103: The name shouldn't be present, per https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L898-L911.

Why isn't this test failing?

integration-test/2047-restore-early-lake-labels.py
nvkelso: Hmm. For the label we do a remapping of the min_zoom based on area grading of 265548000 sq metershere: https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L843-L878, so I expect min_zoom 9 for the label.

```
           (8,     500000000),
            (9,     200000000),
            (10,     40000000),
```

integration-test/2047-restore-early-lake-labels.py
```diff
@@ -0,0 +1,123 @@
...
+                                                })
+
+            # if z >= 9:
+            #     self.assert_has_feature(
+            #         z, x, y, 'water', {
+            #             'kind': 'water',
+            #             'kind_detail': 'lake',
+            #             'name': 'Clear Lake',
+            #             'label_placement': True,
+            #             'min_zoom': 6.0,
```
nvkelso, line 122: The label's min_zoom should be `9.0` not `6.0`.

integration-test/2047-restore-early-lake-labels.py
nvkelso: This looks like an existing bug in that logic, as the last build shows the label for Clear Lake is in fact 6.0 not 9.0 like intended. Let's fix that.

nvkelso: The area threshold numbers look fine to me. Is this where clause is malformed?

> (properties is not None and properties.get('label_placement') is True)

integration-test/2047-restore-early-lake-labels.py
peitili: there is a bug in the yaml file processing in the update_min_zoom function, just fixed that and then fixed the integration tests

vectordatasource/transform.py
```diff
@@ -9471,6 +9470,9 @@ def update_min_zoom(ctx):
         local = defaultdict(lambda: None)
         local.update(props)
         local['zoom'] = zoom
+        # this is to make the name `properties` visible in the queries.yaml's
+        # where clause
+        local['properties'] = props
```
nvkelso, line 9475: This may impact build time, but generally makes things more consistent and obvious.

nvkelso (commented):
Final set of code, tests, and docs LGTM, including Q&A of a mini build.

@peitili this is ready to merge

peitili (approved):

PR #1993, opened 2021-11-13
Estimate population for null pop OSM places

nvkelso:
For #1992
Break out filters in landuse.yaml for different localities by whether they have a population listed or not.  If they don't, make the min_zoom higher and estimate the population.  
- [X] Update tests

PR #2048, opened 2022-01-27
OSM localities joins NE and prefers NE population to calculate population_rank

peitili:
### Background

#2020

When zooming across the NE <> OSM boundary around zoom 8, townspots shift position, population_rank values change, and min_zoom values change.

For country labels in the same places theme we do a join on WikidataID and apply the NE min_zoom to the OSM features.

We should do the same type of harvesting for localities. Namely:

OSM localities should get their population_rank and min_zoom values from NE.
The population value should not be adjusted, as that's for the incorporated area, while the NE population_rank is for the metro area and is more useful for label grading in the stylesheet.

### Changes
- Add export functions to export `__ne_max_zoom`, `__ne_min_zoom`, `__ne_pop_min` and `__ne_pop_max` to the json object
- Move the estimate population value that introduced in #1993 to a post_process function as the final fall-back values for population
- Use the population values from NE to calculate `population_rank` field
- Fix a bug that `{"population": None}` in the json object doesn't end up getting the default estimate population value.

#### Existing Test Changes
**1 Why there is a change necessary for
`integration-test.988-add-population-to-collision-rank.PopulationRankTest.test_0` ?**

In fact, this test breaks on master too, and this PR fixes it. 

**2 Why there is a change necessary for
`integration-test.988-add-population-to-collision-rank.PopulationRankTest.test_collision_rank_decreasing` ?**

When the data has tag `population: None` it falls into following filter line in places.yaml (surprisingly?!) https://github.com/tilezen/vector-datasource/blob/80799e74e0283a96b520c6fea8fa00455095e09b/yaml/places.yaml#L144 
and thus it won't have the hardcoded estimate population value `10000` for `kind_detail: city`, it will have population value be `0`. 

Because we moved the estimation population backfill logic from the yaml file to the transform function, and the transform function is executed at a later stage than the yaml file logic, now when tag `population: None` happens, it gets a chance to take the hardcoded estimate population value `10000`. And thus the `collision_rank` value changes and as a result, original test breaks. So one way to make this test pass again and more reasonable is to not start from tag `population: None` but from `population: 0`.

- [x] Update tests
- [x] Update docs

### DIFF
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1050,6 +1050,36 @@ BEGIN
 END
 $$ LANGUAGE plpgsql STABLE;
 
+
+-- returns a JSONB object containing __ne_pop_min and __ne_pop_max
+CREATE OR REPLACE FUNCTION tz_get_ne_min_max_pop(wikidata_id TEXT)
+RETURNS JSONB AS $$
+DECLARE
+  pop_min REAL;
+  pop_max REAL;
+BEGIN
+  IF wikidata_id IS NULL THEN
+    RETURN '{}'::jsonb;
+  END IF;
+
+  SELECT
+    pop_min, pop_max INTO pop_min, pop_max
+    FROM ne_10m_populated_places pp
+    WHERE pp.wikidataid = wikidata_id;
+
+  -- return an empty JSONB rather than null, so that it can be safely
+  -- concatenated with whatever other JSONB rather than needing a check for
+  -- null.
+  IF NOT FOUND THEN
+    RETURN '{}'::jsonb;
+  END IF;
+  RETURN jsonb_build_object(
+    '__ne_pop_min', pop_min,
+    '__ne_pop_max', pop_max
+  );
+END
+$$ LANGUAGE plpgsql STABLE;
+
 -- return the min zoom for a node that looks like a service area.
 CREATE OR REPLACE FUNCTION tz_looks_like_service_area(name TEXT)
 RETURNS INTEGER AS $$
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1035,7 +1035,7 @@ BEGIN
     SELECT
       pp.min_zoom, NULL INTO min_zoom, max_zoom
       FROM ne_10m_populated_places pp
-      WHERE pp.wikidataid = wikidata_id
+      WHERE pp.wikidataid = wikidata_id;
   END IF;
 
   -- return an empty JSONB rather than null, so that it can be safely
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1021,12 +1021,20 @@ BEGIN
       WHERE mu.wikidataid = wikidata_id;
   END IF;
 
-  -- finally, try states and provinces
+  -- try states and provinces
   IF NOT FOUND THEN
     SELECT
       min_label, max_label INTO min_zoom, max_zoom
       FROM ne_10m_admin_1_states_provinces sp
       WHERE sp.wikidataid = wikidata_id;
+  END IF;
+
+  -- finally, try localities
+  IF NOT FOUND THEN
+    SELECT
+      rank_min, rank_max INTO min_zoom, max_zoom
+      FROM ne_10m_populated_places pp
+      WHERE pp.wikidataid = wikidata_id
   END IF;
 
   -- return an empty JSONB rather than null, so that it can be safely
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1064,7 +1064,7 @@ BEGIN
   END IF;
 
   SELECT
-    pop_min, pop_max INTO pop_min, pop_max
+    pp.pop_min, pp.pop_max INTO pop_min, pop_max
     FROM ne_10m_populated_places pp
     WHERE pp.wikidataid = wikidata_id;
 
@@ -1078,7 +1078,7 @@ BEGIN
     '__ne_pop_min', pop_min,
     '__ne_pop_max', pop_max
   );
-END
+END;
 $$ LANGUAGE plpgsql STABLE;
 
 -- return the min zoom for a node that looks like a service area.
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1032,7 +1032,8 @@ BEGIN
   -- finally, try localities
   IF NOT FOUND THEN
     SELECT
-      rank_min, rank_max INTO min_zoom, max_zoom
+      -- There is no concept of max_zoom for ne_10m_populated_places
+      min_zoom, NULL INTO min_zoom, max_zoom
       FROM ne_10m_populated_places pp
       WHERE pp.wikidataid = wikidata_id
   END IF;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1033,7 +1033,7 @@ BEGIN
   -- There is no concept of max_zoom for ne_10m_populated_places
   IF NOT FOUND THEN
     SELECT
-      min_zoom, NULL INTO min_zoom, max_zoom
+      pp.min_zoom, NULL INTO min_zoom, max_zoom
       FROM ne_10m_populated_places pp
       WHERE pp.wikidataid = wikidata_id
   END IF;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1052,7 +1052,7 @@ $$ LANGUAGE plpgsql STABLE;
 
 
 -- returns a JSONB object containing __ne_pop_min and __ne_pop_max
-CREATE OR REPLACE FUNCTION tz_get_ne_min_max_pop(wikidata_id TEXT)
+CREATE OR REPLACE FUNCTION tz_get_ne_pop_min_max(wikidata_id TEXT)
 RETURNS JSONB AS $$
 DECLARE
   pop_min REAL;
--- a/data/functions.sql
+++ b/data/functions.sql
@@ -1030,9 +1030,9 @@ BEGIN
   END IF;
 
   -- finally, try localities
+  -- There is no concept of max_zoom for ne_10m_populated_places
   IF NOT FOUND THEN
     SELECT
-      -- There is no concept of max_zoom for ne_10m_populated_places
       min_zoom, NULL INTO min_zoom, max_zoom
       FROM ne_10m_populated_places pp
       WHERE pp.wikidataid = wikidata_id
@@ -1048,7 +1048,7 @@ BEGIN
     '__ne_min_zoom', min_zoom,
     '__ne_max_zoom', max_zoom
   );
-END
+END;
 $$ LANGUAGE plpgsql STABLE;
 
 
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -804,7 +804,7 @@ Places with `kind` values of `continent`, `country`, with others added starting 
 * `name`
 * `id`: The `osm_id` from OpenStreetMap or Natural Earth id
 * `kind`: normalized values between OpenStreetMap and Natural Earth
-* `population`: population integer values from OpenStreetMap or Natural Earth's minimum population value(`pop_min`) if the place has a join from OpenStreetMap and NatualEarth or an estimate value hardcoded in the transform function `tags_set_ne_pop_min_max_default` in this repo.
+* `population`: population integer values from OpenStreetMap or Natural Earth's minimum population value(`pop_min`) if the place has a join from OpenStreetMap and NatualEarth or an estimate based on the type of place.
 * `population_rank`: A value from 18 down to 0, indicating how large the population is on a particular place. A larger value indicates a bigger population. See "Population Rank" below for more details.
 * `source`: `openstreetmap`, `naturalearthdata.com`, or `whosonfirst.org`
 * `min_zoom`: a suggested minimum zoom at which the place should become visible based on scalerank and population values from Natural Earth, and invented for OpenStreetMap. Note that this is not an integer, and may contain fractional parts.
@@ -869,7 +869,7 @@ The values of population rank are derived from the `population` value as follows
 * 1: Less than 200
 * 0: No `population` value available or `population` value zero.
 
-OSM localities gets their population_rank from NaturalEarth's `pop_max` tag because NaturalEarth `pop_max` is for the metro area and is more useful for label grading in the stylesheet.
+When available, for the largest cities, OSM localities gets their population_rank from NaturalEarth's `pop_max` tag because NaturalEarth `pop_max` is for the metro area and is more useful for label grading in the stylesheet.
 
 ## Points of Interest
 
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -804,7 +804,7 @@ Places with `kind` values of `continent`, `country`, with others added starting 
 * `name`
 * `id`: The `osm_id` from OpenStreetMap or Natural Earth id
 * `kind`: normalized values between OpenStreetMap and Natural Earth
-* `population`: population integer values from OpenStreetMap or Natural Earth's maximum population value.
+* `population`: population integer values from OpenStreetMap or Natural Earth's minimum population value(`pop_min`) if the place has a join from OpenStreetMap and NatualEarth or an estimate value hardcoded in the transform function `tags_set_ne_pop_min_max_default` in this repo.
 * `population_rank`: A value from 18 down to 0, indicating how large the population is on a particular place. A larger value indicates a bigger population. See "Population Rank" below for more details.
 * `source`: `openstreetmap`, `naturalearthdata.com`, or `whosonfirst.org`
 * `min_zoom`: a suggested minimum zoom at which the place should become visible based on scalerank and population values from Natural Earth, and invented for OpenStreetMap. Note that this is not an integer, and may contain fractional parts.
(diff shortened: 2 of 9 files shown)


peitili:
I am going to do a mini build on California to generate some rendered tiles to verify the change.

data/functions.sql
```diff
@@ -1021,14 +1021,22 @@ BEGIN
...
     SELECT
       min_label, max_label INTO min_zoom, max_zoom
       FROM ne_10m_admin_1_states_provinces sp
       WHERE sp.wikidataid = wikidata_id;
   END IF;
 
+  -- finally, try localities
+  IF NOT FOUND THEN
+    SELECT
+      rank_min, rank_max INTO min_zoom, max_zoom
```
nvkelso, line 1035: The rank_min here is a red herring... but we can use this slot to solve a different problem where the min_zoom between Natural Earth and OSM is inconsistent.

TO:
>  **min_zoom**, rank_max INTO min_zoom, max_zoom

docs/layers.md
```diff
@@ -804,7 +804,7 @@ Places with `kind` values of `continent`, `country`, with others added starting
 * `name`
 * `id`: The `osm_id` from OpenStreetMap or Natural Earth id
 * `kind`: normalized values between OpenStreetMap and Natural Earth
-* `population`: population integer values from OpenStreetMap or Natural Earth's maximum population value.
+* `population`: population integer values from OpenStreetMap or Natural Earth's minimum population value(`pop_min`) if the place has a join from OpenStreetMap and NatualEarth or an estimate value hardcoded in the transform function `tags_set_ne_pop_min_max_default` in this repo.
```
nvkelso, line 807: `or an estimate based on the type of place.`

Don't need to get that technical for general audience documentation.

docs/layers.md
```diff
@@ -867,7 +867,9 @@ The values of population rank are derived from the `population` value as follows
 * 3: 1k to 2k
 * 2: 200 to 1k
 * 1: Less than 200
-* 0: No `population` value available.
+* 0: No `population` value available or `population` value zero.
+
+OSM localities gets their population_rank from NaturalEarth's `pop_max` tag because NaturalEarth `pop_max` is for the metro area and is more useful for label grading in the stylesheet.
```
nvkelso, line 872: When available, for the largest cities OSM localities...`

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+                'name': u'San Francisco',
+                'place': u'city',
+                'population': u'864816',
+                'rank': u'10',
+                'short_name': u'SF',
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q62',
+                'wikipedia': u'en:San Francisco',
+                '__ne_pop_max': u'20000000'
+            }),
```
nvkelso, line 22: Let's add for SF from Natural Earth:

>                '__ne_min_zoom': u'2.7'

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            })
+        # __ne_pop_max shouldn't appear in the final result
+        self.assert_no_matching_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                '__ne_pop_max': 20000000,
+            })
```
nvkelso, line 44: Let's add for SF from Natural Earth:

>                '__ne_min_zoom': u'2.7'

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+        # however the population_rank is calculated from '__ne_pop_max'
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                'population': 864816,
+                'population_rank': 15
+            })
```
nvkelso, line 35: Let's add for SF from Natural Earth:

>                 'min_zoom': 2.7

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            # https://www.openstreetmap.org/node/26819236
+            dsl.point(26819236, (-122.4199061, 37.7790262), {
+                'name': u'San Francisco',
+                'place': u'city',
+                'rank': u'10',
+                'short_name': u'SF',
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q62',
+                'wikipedia': u'en:San Francisco',
+                '__ne_min_zoom': 10,
```
nvkelso, line 137: Let's use real values, as the problem we're trying to solve is _earlier_ min_zoom values not later ones.

nvkelso: >                '__ne_min_zoom': 2.7,

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+        # without __ne_min_zoom, the min_zoom should be 8 which is from the
+        # https://github.com/tilezen/vector-datasource/blob/80799e74e0283a96b520c6fea8fa00455095e09b/yaml/places.yaml#L145
+        # but with __ne_min_zoom override, it is 10
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                'min_zoom': 10,
```
nvkelso, line 150: >                 'min_zoom': 2.7,

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            })
+
+        # __ne_min_zoom shouldn't appear in the final result
+        self.assert_no_matching_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                '__ne_min_zoom': 10,
```
nvkelso, line 160: >           '__ne_min_zoom': 2.7,

data/functions.sql
nvkelso, line 1035: - `min_zoom` instead of `rank_min`
- There is no concept of `max_zoom` for NE populated places... so you can use whatever you want here. But probably it's better to use 0 (min min_zoom) or None or NULL or 21 (max possible zoom) rather than random values from the rank_max of the population.

docs/layers.md
nvkelso: See also:

- https://github.com/whosonfirst/whosonfirst-properties/tree/main/properties/wof#population_rank

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            dsl.point(26819236, (-122.4199061, 37.7790262), {
+                'name': u'San Francisco',
+                'place': u'city',
+                'population': u'864816',
+                'rank': u'10',
+                'short_name': u'SF',
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q62',
+                'wikipedia': u'en:San Francisco',
+                '__ne_pop_max': u'20000000'
```
nvkelso, line 21: Let's use real values:

> 3450000

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+        # population is available directly from 'population' property
+        # however the population_rank is calculated from '__ne_pop_max'
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                'population': 864816,
+                'population_rank': 15
```
nvkelso, line 34: Let's use real values, based on update above:

>                 'population_rank': 12

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            })
+        # __ne_pop_max shouldn't appear in the final result
+        self.assert_no_matching_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                '__ne_pop_max': 20000000,
```
nvkelso, line 43: To

> 3450000

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            # https://www.openstreetmap.org/node/26819236
+            dsl.point(26819236, (-122.4199061, 37.7790262), {
+                'name': u'San Francisco',
+                'place': u'city',
+                'rank': u'10',
+                'short_name': u'SF',
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q62',
+                'wikipedia': u'en:San Francisco',
+                '__ne_pop_min': u'50000000'
```
nvkelso, line 58: Let's use real values:

>                 '__ne_pop_min': u'732072'

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+        # __ne_pop_min also __ne_pop_max is not available so
+        # population_rank is still calculated by the value of the backfilled
+        # population which is __ne_pop_min
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                'population': 50000000,
```
nvkelso, line 72: >                'population': 732072,

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+                'population_rank': 16
```
nvkelso, line 73: >                'population_rank': 11

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            })
+
+        # __ne_pop_min shouldn't appear in the final result
+        self.assert_no_matching_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                '__ne_pop_min': 50000000,
```
nvkelso, line 83: >            '__ne_pop_min': 732072

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+            # https://www.openstreetmap.org/node/26819236
+            dsl.point(26819236, (-122.4199061, 37.7790262), {
+                'name': u'San Francisco',
+                'place': u'city',
+                'rank': u'10',
+                'short_name': u'SF',
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q62',
+                'wikipedia': u'en:San Francisco',
+                '__ne_pop_max': u'20000000'
```
nvkelso, line 98: Use real values:

> 3450000

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+        # population is not available from source but backfilled by
+        # the estimate defined in tags_set_ne_pop_min_max_default
+        # population_rank is still calculated by __ne_pop_max
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                'population': 10000,
```
nvkelso, line 111: I think you want update the logic and then to split this into 2 tests? One that's more like what you have, and another one where it's null pop in OSM but with both `__ne_pop_min` and `__ne_pop_max`?

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+                'population_rank': 15
+            })
+        # __ne_pop_max shouldn't appear in the final result
+        self.assert_no_matching_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                '__ne_pop_max': 20000000,
```
nvkelso, line 121: To:

> 3450000

queries/planet_osm_point.jinja2
```diff
@@ -8,6 +8,7 @@ SELECT
     'source', 'openstreetmap.org'
   ) AS __properties__,
 
+  {# The `CASE WHEN mz_xxx_min_zoom IS NOT NULL` here is used to indicate whether the current row is from a particular layer #}
```
nvkelso, line 11: Please rewrite to add clarity.

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+        # the estimate defined in tags_set_ne_pop_min_max_default
+        # population_rank is still calculated by __ne_pop_max
+        self.assert_has_feature(
+            z, x, y, 'places', {
+                'id': 26819236,
+                'kind': 'locality',
+                'kind_detail': 'city',
+                'wikidata_id': 'Q62',
+                'population': 10000,
+                'population_rank': 15
```
nvkelso, line 112: To:

> 12

data/functions.sql
nvkelso: Or `ne_places_min_zoom` which is calculated from `scalerank` (which is odd).

peitili: I will use min_zoom since scalerank hasn't been maintained for a while according to you.

nvkelso: If you use `min_zoom` then you'd also need to change:

https://github.com/tilezen/vector-datasource/blob/master/queries.yaml#L35-L38

To have:

>      start_zoom: 2

Else the zoom 0/0/0 tile at 512px tile size will get a lot bigger because there are 11 features with min_zoom of `1.7`.

nvkelso: My preference is to move to `min_zoom` instead of `ne_places_min_zoom` as `scalerank` isn't maintained for 5 years in Natural Earth and `min_zoom` is.

integration-test/2020-locality-osm-ne-join.py
nvkelso: To figure out what values to use... you need to download linked file from the web, and view it in QGIS, from:

https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-populated-places/

And query this city.

queries.yaml
```diff
@@ -35,7 +35,9 @@ sources:
   ne:
     ## places
     # ne_10m_populated_places
+    # Note: we add start_zoom 2 here otherwise the zoom 0/0/0 tile at 512px(or zoom1 tile at 256px) tile size will get a lot bigger because there are 11 features with min_zoom of 1.7 as of Jan 2022.
```
nvkelso, line 38: Nit, add space: `512px(or` to `512px (or`

integration-test/2020-locality-osm-ne-join.py
nvkelso: Interesting:

```
            # don't overstuff features into tiles when they are in the
            # long tail of won't display, but make their min_zoom
            # consistent with when they actually show in tiles
            if min_zoom % 1 > 0.5:
                min_zoom = ceil(min_zoom)
            props['min_zoom'] = min_zoom
```

That's fine then.

nvkelso: So if it is 2.0, 2.1, 2.2, 2.3, 2.4. 2.5 then it'll go thru, else it'll round up to 3.0

integration-test/2020-locality-osm-ne-join.py
```diff
@@ -0,0 +1,161 @@
...
+                'source': u'openstreetmap.org',
+                'wikidata': u'Q62',
+                'wikipedia': u'en:San Francisco',
+                '__ne_min_zoom': 2.7,
+            }),
+        )
+
+        # without __ne_min_zoom, the min_zoom should be 8 which is from the
+        # https://github.com/tilezen/vector-datasource/blob/80799e74e0283a96b520c6fea8fa00455095e09b/yaml/places.yaml#L145
+        # but with __ne_min_zoom override, it is 2.7
```
nvkelso, line 143: Nit, add another line here about values rounding up?

nvkelso: > # because of the logic in tags_set_ne_min_max_zoom, the expected value is converted from 2.7 to 3.0

nvkelso:
LGTM but let's see it in a mini build

integration-test/2020-locality-osm-ne-join.py
peitili: added

nvkelso (approved):
LGTM included Q&A in a mini build.

PR #2050, opened 2022-02-01
change building drop to stop at 16

peitili:
- [ ] Update tests
- [ ] Update docs


integration-test/1732-restore-building-scale-rank.py
```diff
@@ -85,7 +85,7 @@ def test_merged_z15(self):
             z, x, y, 'buildings', {
                 'kind': 'building',
                 'scale_rank': 5,
-            }, 1)
+            }, 10)
```
peitili, line 88: there should be more changes to fix this, right now the integ test only passes if asserting 10 buildings.

PR #2053, opened 2022-02-04
admin level 3 + state/country = disputed_reference_line

travisgrigsby:
Need to handle hong kong / cyprus and a few other disputed admin_level 3 boundaries.  This turns any admin_level 3 place=state/country into a place=disputed_reference_line using a new filter in boundaries.yaml

### DIFF
--- a/yaml/boundaries.yaml
+++ b/yaml/boundaries.yaml
@@ -186,6 +186,7 @@ filters:
     output:
       <<: *output_properties
       kind: disputed_country
+      disputed: True
     table: osm
   - filter:
       all:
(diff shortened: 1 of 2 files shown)


nvkelso:
Please add some tests to ensure the point-of-view tags also get evaluated – so these new lines can be turned off and on in the stylesheet per different kind:xx values since a lot of these are disputed boundaries, too.

yaml/boundaries.yaml
```diff
@@ -175,6 +175,19 @@ filters:
       <<: *output_properties
       kind: county
     table: osm
+  - filter:
```
nvkelso, line 178: Please add a comment explaining why this is needed

nvkelso (approved):
One nit about comment, otherwise LGTM

PR #2056, opened 2022-02-08
roads layer thining z14 to z15

peitili:
- [x] Update tests
- [ ] Update docs

### DIFF
--- a/integration-test/1062-road-shield-cleanup.py
+++ b/integration-test/1062-road-shield-cleanup.py
@@ -15,14 +15,28 @@ class RoadShieldCleanup(FixtureTest):
             {'id': way_id, 'shield_text': expected_shield_text})
 
     def test_A151(self):
-        self._check_network_relation(
-            way_id=208288552, rel_id=1159812, tile=(16, 32949, 22362),
-            expected_shield_text='A151')
+        tile = (16, 32949, 22362)
+        self.load_fixtures([
+            'https://www.openstreetmap.org/way/%d' % (208288552,),
+            'https://www.openstreetmap.org/relation/%d' % (1159812,),
+        ], clip=self.tile_bbox(*tile))
+
+        z, x, y = tile
+        self.assert_has_feature(
+            z, x, y, 'roads',
+            {'id': type(None), 'shield_text': 'A151'})
 
     def test_E402(self):
-        self._check_network_relation(
-            way_id=121496753, rel_id=88503, tile=(16, 32975, 22371),
-            expected_shield_text='E402')
+        tile = (16, 32975, 22371)
+        self.load_fixtures([
+            'https://www.openstreetmap.org/way/%d' % (121496753,),
+            'https://www.openstreetmap.org/relation/%d' % (88503,),
+        ], clip=self.tile_bbox(*tile))
+
+        z, x, y = tile
+        self.assert_has_feature(
+            z, x, y, 'roads',
+            {'id': type(None), 'shield_text': 'E402'})
 
     def test_A52(self):
         self._check_network_relation(
--- a/integration-test/1194-bus-route-refs.py
+++ b/integration-test/1194-bus-route-refs.py
@@ -71,7 +71,7 @@ class BusRouteRefs(FixtureTest):
              'bicycle_network': 'lcn',
              'bicycle_shield_text': '50',
              'bus_network': type(None),
-             'bus_shield_text': type(None)})
+             'is_bus_related': True})
 
     def test_inbound_and_outbound_routes(self):
         # Jackson St. SF, part of trolley-bus route 3
@@ -114,9 +114,9 @@ class BusRouteRefs(FixtureTest):
             16, 10477, 25327, 'roads',
             {'id': 225516711,
              'bus_network': type(None),
-             'bus_shield_text': type(None),
+             'is_bus_related': True,
              'all_bus_networks': [type(None)],
-             'all_bus_shield_texts': ['3']})
+             'all_bus_shield_texts': [type(None)]})
 
     def test_full_lists_disappear_by_zoom_12(self):
         # make sure the all_* lists are gone by zoom 12 on major roads, but
--- a/integration-test/1194-bus-route-refs.py
+++ b/integration-test/1194-bus-route-refs.py
@@ -71,7 +71,7 @@ class BusRouteRefs(FixtureTest):
              'bicycle_network': 'lcn',
              'bicycle_shield_text': '50',
              'bus_network': type(None),
-             'bus_shield_text': '23'})
+             'bus_shield_text': type(None)})
 
     def test_inbound_and_outbound_routes(self):
         # Jackson St. SF, part of trolley-bus route 3
@@ -114,7 +114,7 @@ class BusRouteRefs(FixtureTest):
             16, 10477, 25327, 'roads',
             {'id': 225516711,
              'bus_network': type(None),
-             'bus_shield_text': '3',
+             'bus_shield_text': type(None),
              'all_bus_networks': [type(None)],
              'all_bus_shield_texts': ['3']})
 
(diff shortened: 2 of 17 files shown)


queries.yaml
```diff
@@ -1859,18 +1886,18 @@ post_process:
...
     params:
       source_layer: roads
       start_zoom: 8
-      end_zoom: 14
+      end_zoom: 16
       properties:
         - landuse_kind
       where: >-
-        landuse_kind in ('residential', 'industrial')
+        landuse_kind not in ['aerodrome', 'airfield', 'cemetery', 'college', 'conservation', 'forest', 'garden', 'golf_course', 'hospital', 'military', 'national_park', 'nature_reserve', 'park', 'parking', 'protected_area', 'runway', 'school', 'taxiway', 'university']
```
nvkelso, line 1900: This can be verified in 512 px tile **14-9612-6755** looking at شارع المطار street. The yellow segment on left has the same properties, except the `landuse_kind` value is null instead of residential.

We could have extended the end_zoom to accomplish the same thing for this tile... but the logic above also accomplishes the same thing while also removing `landuse_kind` = grass and other irrelevant details.

image

queries.yaml
```diff
@@ -1303,6 +1317,22 @@ post_process:
...
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 17
+      properties:
+        - ref
+        - osm_relation
+        - alt_name
+        - alt_name:ar
+        - alt_name:en
```
nvkelso, line 1332: This can be verified in 512 px tile 14-9612-6755, mostly in minor_roads but also in major_roads.

For example:  https://www.openstreetmap.org/way/24986420

image

queries.yaml
```diff
@@ -1278,6 +1276,22 @@ post_process:
...
+  # Because we have is_bus_route boolean, we don't need the
+  # details about bus network, operator, and shield text until
+  # later zooms when there is enough space to label them
+  - fn: vectordatasource.transform.drop_properties
+    params:
+      source_layer: roads
+      start_zoom: 0
+      end_zoom: 17
+      properties:
+        - bus_network
```
nvkelso, line 1288: Test with https://www.openstreetmap.org/way/190333972 which includes bus relations, visible in 16-17756-27464 and it's parent 15-8878-13732 tiles at 512px tile coordinates.

queries.yaml
```diff
@@ -2009,16 +2036,26 @@ post_process:
...
+  # drop labels on roads which are too short to render
+  # but with more flexibility for range of style text sizes
+  - fn: vectordatasource.transform.drop_names_on_short_boundaries
+    params:
+      source_layer: roads
+      start_zoom: 15
+      end_zoom: 17
+      # pixels (256px nominal) to require per letter in the name. if lines are shorter than this,
+      # the name gets dropped. larger values mean fewer lines are eligible for labelling.
+      pixels_per_letter: 8
```
nvkelso, line 2051: For 15-8878-13732, look at the street formerly known as Lafayette Street, which combined with property dropping above, and the pixels_per_letter going to 8 here, should yield labels restored versus the 1st draft of this PR:

image

queries.yaml
nvkelso: Change looks good with latest build, dropping these alt_names

queries.yaml
nvkelso: Fixed in latest build:

image

queries.yaml
nvkelso: 8 is still too big, still not seeing E. Kennedy labeled

nvkelso: 82 / 15 = 5 pixels per character

integration-test/1062-road-shield-cleanup.py
```diff
@@ -15,14 +15,28 @@ def _check_network_relation(
...
+        tile = (16, 32949, 22362)
+        self.load_fixtures([
+            'https://www.openstreetmap.org/way/%d' % (208288552,),
+            'https://www.openstreetmap.org/relation/%d' % (1159812,),
+        ], clip=self.tile_bbox(*tile))
+
+        z, x, y = tile
+        self.assert_has_feature(
+            z, x, y, 'roads',
+            {'id': type(None), 'shield_text': 'A151'})
```
peitili, line 27: @nvkelso basically I copied the implementation of `_check_network_relation` and the only diffrence I made here is to assert `'id': type(None)` instead.

nvkelso: LGTM

integration-test/1194-bus-route-refs.py
```diff
@@ -71,7 +71,7 @@ def test_one_bus_route(self):
              'bicycle_network': 'lcn',
              'bicycle_shield_text': '50',
              'bus_network': type(None),
-             'bus_shield_text': '23'})
+             'bus_shield_text': type(None)})
```
nvkelso, line 74: This should instead switch to `is_bus_related`: `True`

integration-test/1194-bus-route-refs.py
```diff
@@ -114,7 +114,7 @@ def test_inbound_and_outbound_routes(self):
             16, 10477, 25327, 'roads',
             {'id': 225516711,
              'bus_network': type(None),
-             'bus_shield_text': '3',
+             'bus_shield_text': type(None),
```
nvkelso, line 117: This should instead switch to `is_bus_related`: `True`

integration-test/1979-shield-text-length.py
```diff
@@ -70,8 +70,8 @@ def test_create_shield_text_length(self):
                 'shield_text_length': '2',
                 'bicycle_shield_text': '50',
                 'bicycle_shield_text_length': '2',
-                'bus_shield_text': '23',
-                'bus_shield_text_length': '2'
+                'bus_shield_text': type(None),
```
nvkelso, line 73: This should instead switch to `is_bus_related`: `True`

integration-test/1979-shield-text-length.py
```diff
@@ -120,7 +120,7 @@ def test_lengths_over_6_or_empty_are_not_reported(self):
                 'shield_text': '123456',
                 'shield_text_length': '6',
                 'bicycle_shield_text': '1234567',
-                'bus_shield_text': '',
+                'bus_shield_text': type(None),
```
nvkelso, line 123: This should instead switch to `is_bus_related`: `True`

integration-test/546-road-sort-keys-bridges.py
```diff
@@ -13,7 +13,7 @@ def test_motorway_bridge(self):
         self.assert_has_feature(
             16, 10472, 25323, 'roads',
             {'kind': 'highway', 'kind_detail': 'motorway', 'id': 28412298,
-             'name': 'Presidio Pkwy.', 'is_bridge': True, 'sort_rank': 443})
+             'name': type(None), 'is_bridge': True, 'sort_rank': 443})
```
nvkelso, line 16: This is because the feature is too short to fit the label, while at zoom 17 it would fit / we don't drop.

integration-test/546-road-sort-keys-tunnel.py
```diff
@@ -50,7 +50,7 @@ def test_secondary_level_0(self):
         self.assert_has_feature(
             16, 16812, 24391, 'roads',
```
nvkelso, line 51: If you updated this to 17, 33624, 48782 then the name would still be there, but it's fine this way.

queries.yaml
nvkelso: Success! 

image

nvkelso:
Estimated impact:

- z13: reduction of 5 - 10%
- z14: reduction of 20 - 75%
- z15: reduction of 15 - 45%

nvkelso (approved):
LGTM

There are a few test failures that this PR and the last road layer thinning PR introduced. Let's tackle them in a followup PR.

integration-test/1194-bus-route-refs.py
```diff
@@ -114,9 +115,10 @@ def test_inbound_and_outbound_routes(self):
...
+             'is_bus_related': True,
```
peitili, line 119: this is wrong, the integ test says it should be None

PR #2059, opened 2022-02-16
Add languages to the name: suffix include list

iandees:
This adds to the language include list for the label points in the `places` layer.


peitili (approved):

nvkelso (changes requested):
This `ne_languages` is only used for Natural Earth, and must exactly match the columns in the Natural Earth shapefiles as loaded into the Postgres table. It does not interact with OpenStreetMap data, which passes thru any valid name:xx tag.

Natural Earth does NOT support these other languages, so this would fail the build.

Suggest closing this PR in favor of modifying Tapalcatl to address any localization allowlist/blocklist funk.

iandees:
Closing as suggested!

PR #2060, opened 2022-02-17
Change OSM data source link

peitili:
openstreetmapdata.com is broken?, replace it with openstreetmap.org

- [ ] Update tests
- [x] Update docs


iandees:
I think the doc is right – we used to pull coastline data from openstreetmapdata.com, but then that went away (what you're seeing) and switched to https://osmdata.openstreetmap.de.

nvkelso (changes requested):
See @iandees comment.

Issue #2062, opened 2022-03-10
French regions and province boundary lines show too early


nvkelso:
* **What did you see?** In France the region lines show at zoom 3, and the province lines at 5.5.
* **What did you expect to see?** I expect the region lines to show at zoom 6, and the province lines at zoom 8. 

My expectations more closely match how the data is setup in Natural Earth in the `min_zoom` property. Instead the ETL here is using older `scalerank` values which aren't maintained in Natural Earth and have problems.

Generally we should be showing less features at early zooms, and overall.

See also:

- https://github.com/tilezen/vector-datasource/blob/master/yaml/boundaries.yaml#L17-L31

Instead we can probably do

- Straight lookup: 
    - `min_zoom: { col: min_zoom }`
- Clamped lookup to avoid putting 
    - `min_zoom: { clamp: { max: 11, min: 3, value: { col: min_zoom } } }`

New way (per zoom, not additive):

        - zoom 2 - 199 features
        - 3 @ 61
        - 4 @ 463 but they are all 4.6 and 4.7 so might be excluded to zoom 5
        - 5 @ 41
        - 6 @ 387 basic 
        - 6.6 @ 680
        - 6.7 @ 860
        - 7 @ 900
        -  7.7 @ 1024
        - 8 @ 1522
        - 8.7 @ 515
        - 9 and 10 and 11: 3,525 excluded since they are after OSM transition
        - zoom 2 to 7.7 @ 3,525 features

Old way:

        - [ 2,   2 ] @ 705 features
        - [ 3,   3 ] @ 636
        - [ 5,   4 ] @ 742
        - [ 5.5, 5 ] @ 1195
        - [ 6,   6 ] @ 1403
        - [ 6.7, 7 ] @ 1388
        - [ 6.8, 8 ] @ 1789
        - [ 7,   9 ] @ 1132
        -  [ 7,   0 ] @ 438
        - excluded: 750
        - zoom 2 to 7 @ 9,428 features

nvkelso:
For tests:

- min bounds check of zoom 2 (nothing there)
- max bounds check of zoom 9 (nothing there)
- min_zoom not scalerank

PR #2064, opened 2022-03-18
use NE min_zoom for min_zoom on region boundaries

travisgrigsby:
Fixes https://github.com/tilezen/vector-datasource/issues/2062.

This change moves to using the min_zoom field for NE region boundaries, clamped between 3 and 11.

- [X] Update tests
- [ ] Docs (N/A)


nvkelso (approved):
The new tests passes locally, though Overpass is struggling today for the rest.

Nit: It looks like the following 4 tests might also need min_zoom updates:

- 797-add-missing-boundaries.py
- 992-boundraies-min-zoom-and-name.py
- 1970-split-ne-boundaries-filter.py
- 1983-china-south-sea.py

travisgrigsby:
- [X] 797 (fixed)
- [ ] 992 (needs updates to test shp files)
- [X] 1970 (not broken)
- [X] 1983 (not broken)

nvkelso (approved):
Thanks for updating the other impacted tests!

PR #2067, opened 2022-03-25
Sql script to allow suppressing disputed names

jeffdefacto:
This adds a sql script for stripping names off of disputed island features in order to suppress potential vandalism.


iandees (commented):
We might want to move this to a different spot other than the root dir. Maybe in or near "[migrations](https://github.com/tilezen/vector-datasource/tree/master/data/migrations)"?

Also, it might be a good idea to truncate the number of decimal places so the SQL is a little bit easier to read.

travisgrigsby (approved):

nvkelso:
I like the idea of moving this to out of the root directory... though most the content in the `data/migrations/` dir is quite old and not relevant anymore. It was more relevant when we kept the same DB but needed to update it for a new schema – versus the global from scratch builds we do now.

If **tile-ops** runs the script on DB setup (which isn't clear to me reading the script alone, please add some inline comments to it at the top to establish better stand alone context), then another option is to put this script in the `data/` dir (where there are a few other SQL scripts now).

Consider a different file name that fits in more with existing SQL files like [apply-updates-non-planet-tables.sql](https://github.com/tilezen/vector-datasource/blob/master/data/apply-updates-non-planet-tables.sql). So something more like `apply-planet_disputed_area_features_name_suppression.sql`?

Why is it **tile-ops** responsibility to run this script versus always doing it with the rest of the SQL in the data dir on DB setup? Is it because we want to make it optional to scrub these disputed territories of names?

jeffdefacto:
@nvkelso I moved the script to the data folder and changed the trigger from tileops to the perform-sql-updates.sh script.

data/perform-sql-updates.sh
```diff
@@ -36,6 +36,10 @@ popd
 psql $PSQLOPTS  -f $sqlfile $@
 echo "done."
 
+# delete name tags from planet tables in certain disputed areas
```
nvkelso, line 39: Add more context here, please (keep your existing, add this too):

> # this removes a vector for name vandalism in contentious parts of the map.
> # While not addressed here, geometry vandalism can be mitigated by using a
> # OSM Planet distribution like Daylight.

data/apply-planet_disputed_area_features_name_suppression.sql
```diff
@@ -0,0 +1,84 @@
...
+           bbox
+      where st_within(l.way, bbox.geom)),
+
+ update_polygon as
+     (update planet_osm_polygon p
+         set tags = delete(p.tags, (select array_agg(key) from (select distinct skeys(tags) as key from suppress_polygon) a where key like '%name%'))
+         from suppress_polygon s
+         where s.osm_id = p.osm_id),
+
+ suppress_roads as
```
nvkelso, line 75: Do we dip into the roads table? Is this one necessary?

data/apply-planet_disputed_area_features_name_suppression.sql
```diff
@@ -0,0 +1,84 @@
+-- Script to delete names from certain disputed locations after osm2pgsql import.
+-- The "bbox" query creates polygon geometries to define which areas to select for name removal
+-- The "suppress" query selects features from the planet table that are fully within a "bbox" polygon
+-- The "update" query then deletes any key value pair within the selection with "name" somewhere in the key's string.
+
+with bbox as
+    (select *
+    from (values
+        (st_transform(st_geomfromtext('POLYGON ((-78.89 15.59,-78.89 16.12,-78.30 16.12,-78.30 15.59,-78.89 15.59))', 4326), 3857)),
```
nvkelso, line 9: Is there a way to comment which disputed areas each polygon is for? Inline in the SQL or at least in a comment above the SQL?

nvkelso: I assume these are almost all island disputes, given their bounding boxes. But would you support this method for larger arbitrary shapes for a land dispute, too?

data/apply-planet_disputed_area_features_name_suppression.sql
jeffdefacto: I wasn't sure if we use it, I just ran it on all tables for completeness. I can remove it, if its unneeded.

data/apply-planet_disputed_area_features_name_suppression.sql
jeffdefacto: Yes, these are currently all island or atoll disputes, but I believe it would also work for a land area just as well. 

The selection only takes features fully within the bounding box so I believe larger features like relations should not be unexpected casualties.

jeffdefacto: I updated the query with labels for each polygon

data/apply-planet_disputed_area_features_name_suppression.sql
nvkelso: Please remove this (as it would otherwise add processing time needlessly).

I don't see us using the roads table, in places like https://github.com/tilezen/vector-datasource/blob/master/data/apply-schema-update.sql or elsewhere.

We _do_ dip into the rel (relations) table for road shield information (but not for name values). No action needed there.

jeffdefacto: Sounds good. I removed that bit

nvkelso (approved):

PR #2069, opened 2022-04-20
more rigorous countries labels per ISO and point-of-view

jeffdefacto:
This adds sql to alter the tlc table to edit the featurecla column to mark certain geounits as country or unrecognized. OSM labels that match with unrecognized countries or are a non match will return as kind=unrecognized at min zoom 8. 

Countries updated as unrecognized
- Akrotiri Sovereign Base Area
- Ashmore and Cartier Islands
- Clipperton Island
- Coral Sea Islands
- Dhekelia Cantonment
- Somaliland
- Turkish Republic of Northern Cyprus
- Palestine (see Gaza Strip and West Bank instead)

Countries updated to Admin-0 dependency
- Gaza Strip
- Svalbard
- West Bank

Newly rendering countries from data, lua, and sql improvements

- Aland
- American Samoa
- Bermuda
- Bouvet Island
- Christmas Island
- Cocos (Keeling) Islands
- French Polynesia
- French Southern and Antarctic Lands
- Guam
- Heard Island and McDonald Islands
- Hong Kong
- Macau
- New Caledonia
- Norfolk Island
- Northern Mariana Islands
- Puerto Rico
- Réunion
- Saint Helena, Ascension and Tristan da Cunha
- Saint Barthélemy
- Saint Martin (France)
- Saint Pierre and Miquelon
- Sāmoa
- São Tomé and Príncipe
- Svalbard
- United States Minor Outlying Islands


vectordatasource/transform.py
```diff
@@ -8912,7 +8912,7 @@ def tags_set_ne_min_max_zoom(ctx):
             # way, we probably don't want to see them at zoom, which is lower
             # than most of the curated NE min zooms. see issue #1826 for more
             # information.
-            props['min_zoom'] = max(6, props['min_zoom'])
```
nvkelso, line 8915: Please make this 8 instead of 6.

And set the **kind** to `disputed_country`.

nvkelso: This will require more matching against `map_units` from Natural Earth, not `countries` table, so please also include that join change (keeping **wikidataid** as the join column) in this PR.

That table also has _more_ "countries" than we possibly want... so consider a related PR to strip out some of those (there are up to 70).

nvkelso: EG **Somaliland** and **Northern Cyprus**.

nvkelso: Looks like this will need to use the source DBF that makes the map_units (which includes countries) as some countries aren't included in the map units file alone?

nvkelso:
Relates to: 

#2075, #2076, #2077.

data/functions.sql
```diff
@@ -1006,28 +1006,31 @@ BEGIN
...
-    SELECT
-      min_label, max_label INTO min_zoom, max_zoom
-      FROM ne_10m_admin_0_map_units mu
-      WHERE mu.wikidataid = wikidata_id;
+      IF NOT FOUND THEN
+        SELECT
+            min_label, max_label INTO min_zoom, max_zoom
+        FROM ne_10m_admin_0_countries_tlc t
+        WHERE t.wikidataid = wikidata_id
+        AND fclass_tlc IN ('Admin-0 country' or 'Admin-0 dependency');
```
nvkelso, line 1021: This might not be the right SQL?

You will be missing the disputed areas recognized by Russia which are "Admin-0 breakaway and disputed", and you'll be missing the constituent countries of the United Kingdom (eg England) which are "Admin-1 region".

data/functions.sql
```diff
@@ -1006,28 +1006,31 @@ BEGIN
...
-  SELECT
-    min_label, max_label INTO min_zoom, max_zoom
-    FROM ne_10m_admin_0_countries c
-    WHERE c.wikidataid = wikidata_id;
+  -- if it's a country, only look it up in the iso and tlc tables
+  IF place_tag='country' OR place_tag='unrecognized' THEN
+      SELECT
+        min_label, max_label INTO min_zoom, max_zoom
+        FROM ne_10m_admin_0_countries_iso i
+        WHERE i.wikidataid = wikidata_id;
```
nvkelso, line 1014: This will get you Taiwan, even though it's not `IN ('Admin-0 country' or 'Admin-0 dependency');`, nice

nvkelso: It'll also yield the French overseas regions (mostly fine).

data/functions.sql
nvkelso: It does yield Somaliland, Northern Cyprus, and Kosovo

nvkelso: It also yields: Dhekelia Cantonment, Akrotiri Sovereign Base Area, Coral Sea Islands, Clipperton Island, Ashmore and Cartier Islands.

jeffdefacto: This is still WIP as I haven't been able to test it so this could be wrong. 

However, the breakaways should have a `kind=unrecognized` so I'm intentionally excluding them from the where statement. They should still get a `kind:xx = country` for the viewpoints that recognize them though. 

The [data/apply-ne_country_label_recasting.sql](https://github.com/tilezen/vector-datasource/pull/2069/files/4bd88709309eb97128e023c4ac32da4216c65506#diff-ebf5e928128c82976d4f33eb55aa8c331b9e5ce90b28e13b3487ff4da7ddab35) file makes some adjustments for specific countries or features that don't conform to our policies, such as Somaliland or northern Cyprus. 

I will take a look a the England situation.

nvkelso (commented):
I wonder if Tilezen should have its own view on this? What if you change the FEATURECLA column to equal the FCLASS_ISO values to start, and then modify that with your SQL override the FEATURECLA colum (and move join there) instead of mixing signals in the TLC column? Then the TLC is the most expansive.

Though if you wanted the option to show all these labels (not just your join filter limit then we should think about turning the join the other way, and using the lat/lng in the NE table, but harvesting name localizations from OSM where NE doesn't have them.

Separately... 

You'll need to setup remapping of the FCLASS_ISO, FCLASS_TLC, and other FCLASS_* values to valid Tilezen kinds.

See also:

- https://github.com/nvkelso/natural-earth-vector/blob/master/CHANGELOG#L22-L46

Values:

- `Admin-0 country` (eg United States)
    - **kind** = `country`
    - **kind_detail** = `country`
- `Admin-0 dependency` (eg Puerto Rico)
    - **kind** = `country`
    - **kind_detail** = `dependency`
- `Admin-1 region` (eg England in United Kingdom)
    - **kind** = `unrecognized`
    - **kind_detail** = `macroregion`
- `Admin-1 states provinces` (eg Martinique in France or Alaska in USA) 
    - **kind** = `unrecognized`
    - **kind_detail** = `region`
- `Admin-0 lease` (eg Baykonur leased to Russia)
    - **kind** = `unrecognized`
    - **kind_detail** = `lease`
- `Admin-0 breakaway and disputed` (eg Golan Heights)
    - **kind** = `unrecognized`
    - **kind_detail** = `breakaway_disputed`
- `Admin-0 claim area` (eg Crimea)
    - **kind** = `unrecognized`
    - **kind_detail** = `claim_area`
- `Admin-0 indeterminant` (eg Spratly Islands)
    - **kind** = `unrecognized`
    - **kind_detail** = `indeterminant`
- `Island group` (eg Midway Islands part of US Minor Outlying Islands dependency)
    - **kind** = `unrecognized`
    - **kind_detail** = `island_group`
- `Island` (eg Antigua a component of Antigua and Barbuda country)
    - **kind** = `unrecognized`
    - **kind_detail** = `island`

data/functions.sql
travisgrigsby: Looks like you need to replace your 'or' with a comma on this line or split it into 2 separate fclass_tlc=<blah> clauses joined with an OR

travisgrigsby:
@nvkelso note this change will make more sense once we merge https://github.com/tilezen/vector-datasource/pull/2078

vectordatasource/transform.py
nvkelso: Nagorno-Karabakh also shows up at min_zoom 1 now, because it starts as "kind" = `unrecognized`, so the if test wrapping this doesn't trigger, so the min_zoom isn't updated as part of the join.

nvkelso: As added protection, in https://github.com/tilezen/vector-datasource/blob/master/yaml/places.yaml#L141-L152 we should also change the min_zoom to a case:

```
 - filter: {name: true, place: [country,unrecognized]}
    # note: min_zoom needs to be smaller than any min_label in
    # ne_10m_admin_0_countries to make sure we're not dropping any by not being
    # included in the PostgreSQL index.
    #
    # note: also, we knock this down to 6 during post-processing for any country
    # which doesn't match an NE curated one.
    min_zoom: 
      case:
        - when: { place: country }
          then: 1
        - else: 8
    output:
      <<: [*output_properties, *ne_zooms]
      kind: {col: place}
    table: osm
```

Where the else can be 8 or some other value.

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,41 @@
+-- Recasts a few rows in the ne_10m_admin_0_countries_iso and ne_10m_admin_0_countries_tlc
+-- tables to either show or hide certain labels
+
+
```
nvkelso, line 4: Before here, set the value of **featurecla** to **fclass_tlc**

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,41 @@
...
+-- Akrotiri Sovereign Base Area
+update ne_10m_admin_0_countries_tlc set fclass_tlc = 'unrecognized' where ne_id = 1159320741;
```
nvkelso, line 6: set **featurecla** instead of **fclass_tlc**

repeat for other rows

data/functions.sql
```diff
@@ -1006,34 +1006,96 @@ BEGIN
...
 
-  -- if that fails, try map_units (which contains some sub-country but super-
-  -- state level stuff such as England, Scotland and Wales).
-  IF NOT FOUND THEN
+      IF NOT FOUND THEN
+        SELECT
+            min_label, max_label INTO min_zoom, max_zoom
+        FROM ne_10m_admin_0_countries_tlc t
+        WHERE t.wikidataid = wikidata_id
+        AND fclass_tlc IN ('Admin-0 country', 'Admin-0 dependency');
```
nvkelso, line 1021: AND **featurecla**

data/functions.sql
```diff
@@ -1006,34 +1006,96 @@ BEGIN
...
+        i.fclass_iso, i.fclass_tlc, i.label_x, i.label_y INTO fclass_iso_var, fclass_tlc_var, label_x_var, label_y_var
+    FROM ne_10m_admin_0_countries_iso i
+    WHERE i.wikidataid = wikidata_id;
+
+    IF NOT FOUND THEN
+        SELECT
+            t.fclass_iso, t.fclass_tlc, t.label_x, t.label_y INTO fclass_iso_var, fclass_tlc_var, label_x_var, label_y_var
+        FROM ne_10m_admin_0_countries_tlc t
+        WHERE t.wikidataid = wikidata_id
+        AND fclass_tlc IN ('Admin-0 country', 'Admin-0 dependency');
```
nvkelso, line 1076: AND `featurecla`

data/apply-ne_country_label_recasting.sql
nvkelso: And add note why you're doing it this way

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,38 @@
...
+-- Somaliland
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159321259;
+
+-- Turkish Republic of Northern Cyprus
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159320531;
+
+-- West Bank
+update ne_10m_admin_0_countries_tlc set featurecla = 'Admin-0 country' where ne_id = 1159320903;
+
+
```
nvkelso, line 36: NIt: It'd be helpful to know that this also yields **Kosovo** and **Taiwan** as `country` deliberately, and

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,38 @@
+-- Recasts a few rows in the ne_10m_admin_0_countries_iso and ne_10m_admin_0_countries_tlc
+-- tables to either show or hide certain labels
+
+-- sets featurcla to equal fclass_tlc values. We can then further modify the values while keeping the fclass intact.
```
nvkelso, line 4: NIt: It'd be helpful to know that this also yields **Kosovo** and **Taiwan** as `country` deliberately. Please add that as a comment here.

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,38 @@
...
+-- featurecla has a shorter varchar length and needs altering first
+alter table ne_10m_admin_0_countries_tlc alter column featurecla type varchar;
+update ne_10m_admin_0_countries_tlc set featurecla = fclass_tlc;
+
+-- Akrotiri Sovereign Base Area
```
nvkelso, line 9: Nit: Before here say something like:

> -- The TLC table yields around 10 country and dependency features over ISO which Tilezen will have opinions about, as follows:

data/apply-ne_country_label_recasting.sql
nvkelso: -- The TLC table yields around 10 country and dependency features over ISO (including Kosovo and Taiwan), which Tilezen sometimes has opinions about.

data/apply-ne_country_label_recasting.sql
nvkelso, line 36: I think you're missing `Indian Ocean Territories` for **ne_id** = `1159320363` being set to `unrecognized`.

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,38 @@
...
+update ne_10m_admin_0_countries_tlc set featurecla = 'Admin-0 country' where ne_id = 1159320901;
+
+-- Somaliland
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159321259;
+
+-- Turkish Republic of Northern Cyprus
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159320531;
+
+-- West Bank
+update ne_10m_admin_0_countries_tlc set featurecla = 'Admin-0 country' where ne_id = 1159320903;
```
nvkelso, line 34: Nit: I'd set this to `Admin-0 dependency` instead. It results in the same thing, but is more true to reality for later.

data/apply-ne_country_label_recasting.sql
```diff
@@ -0,0 +1,38 @@
...
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159320635;
+
+-- Coral Sea Islands
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159320359;
+
+-- Dhekelia Cantonment
+update ne_10m_admin_0_countries_tlc set featurecla = 'unrecognized' where ne_id = 1159320709;
+
+-- Gaza Strip
+update ne_10m_admin_0_countries_tlc set featurecla = 'Admin-0 country' where ne_id = 1159320901;
```
nvkelso, line 25: Nit: I'd set this to `Admin-0 dependency` instead. It results in the same thing, but is more true to reality for later.

nvkelso:
Please update the PR description, too

vectordatasource/transform.py
nvkelso: Only this comment remains unaddressed:

> Nagorno-Karabakh also shows up at min_zoom 1 now, because it starts as "kind" = unrecognized, so the if test wrapping this doesn't trigger, so the min_zoom isn't updated as part of the join.

Fix would be something like:

> elif props.get('kind') in ('country', 'unrecognized'):

data/apply-ne_country_label_recasting.sql
jeffdefacto: This isn't actually present in the table as Cocos and Christmas islands have their own rows.

nvkelso (commented):
Please list out which labels this suppresses (as unrecognized) and which labels it adds or otherwise changes.

osm2pgsql.lua
```diff
@@ -556,8 +560,18 @@ function osm2pgsql.process_node(object)
     if object.tags.place == 'territory' and object.tags['ISO3166-1'] == 'HM' then
         output_hstore['place'] = 'country'
     end
+-- Hong Kong
+    if object.tags.place == 'state' and object.tags['ISO3166-1'] == 'HK' then
+        output_hstore['place'] = 'country'
+        output_hstore['place:CN'] = 'state'
+    end
+-- Hong Kong
```
nvkelso, line 568: Nit: Macao

nvkelso:
Please also list out in the PR description the country labels this adds.

vectordatasource/transform.py
```diff
@@ -8906,13 +8906,14 @@ def tags_set_ne_min_max_zoom(ctx):
...
-        elif props.get('kind') == 'country':
+        elif props.get('kind') in ('country', 'unrecognized'):
             # countries and regions which don't have a min zoom joined from NE
             # are probably either vandalism or unrecognised countries. either
             # way, we probably don't want to see them at zoom, which is lower
             # than most of the curated NE min zooms. see issue #1826 for more
             # information.
-            props['min_zoom'] = max(6, props['min_zoom'])
+            props['kind'] = 'unrecognized'
+            props['min_zoom'] = max(8, props['min_zoom'])
```
nvkelso, line 8916: Consider also setting a max_zoom here to 10.

nvkelso:
Please ensure that all the countries are listed here, otherwise they'll come out "unrecognized" in high zooms for RAWR:

- https://github.com/tilezen/vector-datasource/blob/024909ed8245a4ad4a25c908413ba3602de6c335/spreadsheets/min_zoom/country.yaml

Specifically:

place | cc | min_zoom | max_zoom
----|--|----|----
Martinique | MQ | 5 | 10
Tuvalu | TV | 5 | 10
Réunion | RE | 5 | 10
Cook Islands | CK | 4 | 9
Marshall Islands | MH | 5 | 10
Palau | PW | 5 | 10
Tokelau | TK | 5 | 10
Cayman Islands | KY | 5 | 9.5
Guadeloupe | GP | 5 | 10
British Indian Ocean Territory | IO | 5 | 9.5
Mayotte | YT | 5 | 10
French Southern and Antarctic Lands  | TF | 4.5 | 9.5
French Guiana | GF | 5 | 10
France | FR | 1.7 | 6.7
Norway | NO | 3 | 7
Federated States of Micronesia | FM | 5 | 10
Solomon Islands | SB | 3 | 8

nvkelso (approved):

PR #2070, opened 2022-04-22
Recast Australia suburbs to towns

jeffdefacto:
This changes place=suburb features within an Australian polygon into place=town to allow rendering.


data/apply-planet_australia_suburb_recast.sql
```diff
@@ -0,0 +1,31 @@
...
+     suburb as
+         (select *
+          from planet_osm_point l where tags->'place'='suburb'),
+
+     selection as
+         (select * from suburb,
+                        bbox
+          where st_within(way, bbox.geom)),
+
+    trim_major_cities as
```
nvkelso, line 18: Explain that these are already there are cities, so we're trimming so we don't end up with duplicate labels on the map.

data/apply-planet_australia_suburb_recast.sql
```diff
@@ -0,0 +1,31 @@
+-- Suburbs in Australia are treated more like cities, so this recasts them to  place=town
+
+with bbox as
```
nvkelso, line 3: Nit: This looks not like a box to me... maybe say `bounds` instead of `bbox`?

Here and elsewhere...

nvkelso (approved):
Few nits but LGTM

PR #2072, opened 2022-04-26
SQL to hide unwanted NE borders

jeffdefacto:
This deletes any NE border feature that matches with the list of ne_ids in the `remove_lines` table.


nvkelso (changes requested):
These all need to be visible for different audiences (mostly Russia, Turkey, Taiwan), but should be `unrecognized` for the others. 

Can you switch this around so it's modifying the `fclass_*` attributes instead, please?

data/apply-ne_disputed_border_suppression.sql
```diff
@@ -0,0 +1,512 @@
...
+    union
+    select ne_id from south_ossetia
+    union
+    select ne_id from transnistria
+    )),
+
+-- Add Russian disputed view to Nagorno-Karabakh
+ab_do_so_tr_ne_10m_admin_0_boundary_lines_land as (
+update ne_10m_admin_0_boundary_lines_land l set
+            fclass_ru = ''
```
nvkelso, line 396: Is the intent this is "default" which happens to be `unrecognized`?

data/apply-ne_disputed_border_suppression.sql
```diff
@@ -0,0 +1,512 @@
...
+
+    )),
+
+
+-- Add Turkish view to Northern Cyprus
+tr_ne_10m_admin_0_boundary_lines_land as (
+update ne_10m_admin_0_boundary_lines_land l set
+            fclass_tr = 'country'
+where l.ne_id in (
+    select ne_id from nagorno_karabakh
```
nvkelso, line 439: Is this for Northern Cyprus or nagorno_karabakh?

data/perform-sql-updates.sh
```diff
@@ -43,6 +43,10 @@ echo "done."
 echo -e "\nDeleting disputed names"
 psql $PSQLOPTS  $@ -f apply-planet_disputed_area_features_name_suppression.sql
 
+# Deletes some of the Natural Earth dispute borders that we do not want to render
```
nvkelso, line 46: This is no longer deleting, please update comment and echo

data/apply-ne_disputed_border_suppression.sql
jeffdefacto: Do we want this to be unrecognized? I thought we wanted to show a disputed border there for russia to show that they acknowledge something existing, but not quite a country.

data/apply-ne_disputed_border_suppression.sql
jeffdefacto: Yeah good call. This isn't really ready for review yet. I'm just testing out the syntax

travisgrigsby (approved):

nvkelso (commented):
I really like the direction of this updated PR.

This does point that we need to do a similar `place=unrecognized`  for OSM boundaries so we have consistency on zoom in. Can be followup PR, just link them together please.

nvkelso (approved):

Issue #2075, opened 2022-04-28
Add top-level-countries DBF from Natural Earth


nvkelso:
The following archive:

https://naciscdn.org/naturalearth/5.1.0/housekeeping/ne_admin_0_details.zip

Includes a new table:

ne_admin_0_details_top_level_countries.dbf

That table needs to be added to Postgres as source of truth for "country-ness". Currently it's a join against the admin-0 countries table, which doesn't include all the data we want. This new file does.

travisgrigsby:
@nvkelso I'm looking at this - may need to meet with you to figure out how to wire this in

nvkelso:
You could load it using a tool like: https://pgloader.readthedocs.io/en/latest/ref/dbf.html

Though see my comment below about loading 2 separate shapefiles instead which pattern matches better with what we already do. 

Add it/them to the list of assets around here:

- https://github.com/tilezen/vector-datasource/blob/master/data/assets.yaml#L142-L146

You'll want to configure the extra TLC point-of-view around here:

- https://github.com/tilezen/vector-datasource/blob/734e48d74fd1b4d84f785af4cbd9074fd07574b6/queries/ne.jinja2#L2

But that assumes `fclass_tlc` is present on all related SHPs, which might not be true yet? I'll see if I need to unblock you there.

Then for the lookup for the "country" min_zoom, that's in:

- https://github.com/tilezen/vector-datasource/blob/215fd7a072b84368ce442d0480ca05a8cbffcea1/data/functions.sql#L996-L1052

Which spans across multiple tables today (across input Natural Earth themes across feature classes and kinds), which is problematic because the same Wikidata can be present in multiple tables, sometimes for the same conceptual feature and sometimes not (eg fuzzy versus strict matching) – take France for example. 

So the reason we want to load this new table is we want to get only the curated set of strict matches. 

There is an alternate way of doing this, which is to load 2 new shapefiles instead of 1 DBF, and then splitting the current function so admin-0 is treated separately from admin-1, localities & etc. For the new admin-0 it'd could be a coalesce across the ISO and TLC shapefiles instead (I'd first look in ISO, then in TLC). That should include 100% of what's in the source DBF (which is used as input for both these SHP).

- http://naciscdn.org/naturalearth/5.1.0/10m/cultural/ne_10m_admin_0_countries_iso.zip (smaller, most conservative list)
[...]

Issue #2074, opened 2022-04-28
Upgrade to NE v5.1.0


nvkelso:
And switch to versioned URLs from the unversioned ones, eg:

https://naciscdn.org/naturalearth/5.1.0/housekeeping/ne_admin_0_details.zip

nvkelso:
Fixed via #2078.

PR #2078, opened 2022-04-28
Move to NE 5.1.2 - Add Config for new POVs

travisgrigsby:
* #2074 - update to use NE 5.1.0
* #2075 - Add new NE TLC (top level countries) and ISO tables and remove support for admin-0-countries.
* #2081 - Add TLC POV for NE
* #2082 - Export `fclass_iso` and `fclass_tlc` and `label_x` and `label_y` for OSM countries that match NE countries

For #2082, I added the functionality to drop it into the output, but we aren't doing anything with it yet, so I haven't written any tests for it.  Once we decide to adopt a label_x and label_y if present, for example, there will be compelling function to test.


nvkelso (changes requested):
Please update the versioned asset URLs to match the new pattern. #2074

And also add the new NE table to the data assets list and wire it into the NE tables. #2075

travisgrigsby:
Updated for #2074

data/assets.yaml
```diff
@@ -1,5 +1,5 @@
 bucket: nextzen-tile-assets
-datestamp: 20220426
+datestamp: 20220429
```
nvkelso, line 2: NIt: but 5.1.0 was only officially released on May 4th, so bring this forward?

data/assets.yaml
```diff
@@ -55,98 +55,104 @@ shapefiles:
     url: http://s3.amazonaws.com/tilezen-assets/curated/admin_areas_20180409.zip
 
   - name: ne_110m_lakes
-    url: http://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_lakes.zip
+    url: https://naciscdn.org/naturalearth/5.1.0/110m/physical/ne_110m_lakes.zip
```
nvkelso, line 58: Thanks for switching these to versioned CDN links... much more obvious which version we're on and makes updating NE a deliberate process.

data/functions.sql
```diff
@@ -1006,28 +1006,30 @@ BEGIN
...
+        WHERE t.wikidataid = wikidata_id;
+      END IF;
+
+      IF NOT FOUND THEN
+        RETURN '{}'::jsonb;
+      END IF;
   END IF;
 
-  -- try states and provinces
-  IF NOT FOUND THEN
```
nvkelso, line 1025: Don't you want to keep this in here to bail early if found?

data/functions.sql
```diff
@@ -1051,6 +1053,65 @@ BEGIN
...
+    END IF;
+
+    IF NOT FOUND THEN
+        RETURN '{}'::jsonb;
+    END IF;
+  END IF;
+
+  -- There is no label_x and label_y for the non-countries
+  SELECT
+    fclass_iso, fclass_tlc, NULL, NULL INTO fclass_iso, fclass_tlc, label_x, label_y
```
nvkelso, line 1089: This is confusingly called: "latitude" (label_y) and "longitude" (label_x) in the NE admin-1 table. Please update to pull them out.

data/functions.sql
```diff
@@ -1051,6 +1053,65 @@ BEGIN
...
+  FROM ne_10m_admin_1_states_provinces sp
+  WHERE sp.wikidataid = wikidata_id;
+
+  -- finally, try localities
+  IF NOT FOUND THEN
+  SELECT
+    fclass_iso, fclass_tlc, NULL, NULL INTO fclass_iso, fclass_tlc, label_x, label_y
```
nvkelso, line 1096: Same deal: This is confusingly called: "latitude" (label_y) and "longitude" (label_x) in the NE populated places table. Please update to pull them out.

nvkelso (commented):
Can you add NE and OSM unit tests for TLC, please?

data/functions.sql
nvkelso: (so keep the output as label_{x,y} but the input would be `longitude, latitude`

data/functions.sql
travisgrigsby: You're right.  I thought I was returning on green line 1024, but that's only if we didn't find it.  Updated to check if it's a country else do something else, then return once at the end.

data/assets.yaml
travisgrigsby: I agree.  Much clearer

data/functions.sql
travisgrigsby: fixed in 31a5b4a27b76b89b24214157b4758b3ddebad030

data/functions.sql
travisgrigsby: fixed in 31a5b4a27b76b89b24214157b4758b3ddebad030

data/functions.sql
travisgrigsby: Fixed in 31a5b4a27b76b89b24214157b4758b3ddebad030

data/assets.yaml
travisgrigsby: updated to most recent (which I think @jeffdefacto found an issue in for the iso countries.zip)

data/assets.yaml
```diff
@@ -141,12 +141,15 @@ shapefiles:
 
   - name: ne_10m_admin_0_countries
     url: https://naciscdn.org/naturalearth/5.1.0/10m/cultural/ne_10m_admin_0_countries.zip
+    prj: 3857
```
nvkelso, line 144: These files aren't web mercator (3857). Instead they need to be re-projected into web mercator from WGS84 lat/lng

data/assets.yaml
nvkelso: That zero length char definition is really odd. Opens fine in QGIS but PostGIS barfs on it (saying it doesn't support 0 length varchar fields).

Filed https://github.com/mbloch/mapshaper/issues/541 upstream.

nvkelso: With https://github.com/mbloch/mapshaper/issues/541 fixed upstream, I'll update Natural Earth to v5.1.1 with the fix.

data/assets.yaml
nvkelso: These will all need to switch to 5.1.1 to pick up the other change.

data/assets.yaml
nvkelso: Natural Earth 5.1.1 is now out, please update here and see if the problem persists.

data/assets.yaml
travisgrigsby: updated here: 2635448740ded3a712fbb9cc5c61d108ac734ec9

nvkelso (approved):

PR #2079, opened 2022-04-28
Add Dispute IDs everywhere

travisgrigsby:
This adds dispute ID everywhere it was missing and takes advantage of @jeffdefacto's lua magic to no longer need to do any of the disputed border cutting we were doing before.  

NE dispute_ids
* update NE.jinja to export components for dispute_id
* update boundaries.yaml to export components for dispute_ids for NE-zoom features

OSM dispute_ids
* Change the filter that exported `mz_internal_dispute_mask` to just export a `disputed_reference_line` instead.  The other code path will now handle applying all the right dispute-related properties.
* remove all the code for handling `mz_internal_dispute_mask` since it is no longer needed.

- [X] Update tests
- [ ] Update docs


queries/ne.jinja2
```diff
@@ -32,7 +32,9 @@ SELECT
 {% endif %}
     'min_zoom', mz_boundary_min_zoom,
     'featurecla', featurecla,
-    'scalerank', scalerank
+    'scalerank', scalerank,
+    'ne_id', ne_id,
```
nvkelso, line 36: Should ne_id and BRK_A3 be prefixed with `tz_` since they should only be used for internal processing?

queries/ne.jinja2
```diff
@@ -32,7 +32,9 @@ SELECT
...
+    'BRK_A3', BRK_A3
```
nvkelso, line 37: Tilezen's convention is lowercase only... so:

>     'BRK_A3', brk_a3

queries/ne.jinja2
nvkelso: >     'ne_id', tz_ne_id,

yaml/boundaries.yaml
```diff
@@ -173,7 +177,7 @@ filters:
     min_zoom: 8
     output:
       <<: [ *output_properties, *dispute_properties ]
-      kind: mz_internal_dispute_mask
+      kind: disputed_reference_line
```
nvkelso, line 180: Is there a reason to keep this as a separate filter now, versus consolidating it with the filter immediately above?

queries/ne.jinja2
nvkelso: Nevermind... this is the DB ETL part, the `tz_` part happens in the query step.

yaml/boundaries.yaml
```diff
@@ -77,16 +77,20 @@ global:
           - geom_type: line
           - mz_boundary_from_polygon: true
 
-  - &dispute_id_components
+  - &osm_dispute_id_components
     tz_breakaway_code: { col: 'ne:brk_a3' }
     tz_ne_id: { col: 'ne_id' }
 
+  - &ne_dispute_id_components
+    tz_breakaway_code: { col: 'BRK_A3' }
```
nvkelso, line 85: this should be lowercase, see my other comment above.

>    tz_breakaway_code: { col: 'brk_a3' }

integration-test/2022-disputed-borders.py
```diff
@@ -321,3 +321,104 @@ def test_places_with_viewpoints(self):
...
+                'kind:xx': 'unrecognized'
+            })
+
+    def test_ne_breakaway_has_dispute_id(self):
+        z, x, y = 8, 0, 0
+
+        self.generate_fixtures(
+            dsl.way(1, dsl.tile_diagonal(z, x, y), {
+                'featurecla': 'Breakaway',
+                'BRK_A3': '456',
```
nvkelso, line 357: Nit: There may be a setting in the shapefile importing step to lowercase all field names? In DBF the field names are case insensitive. But then we might be tripping up ourselves later it looks like.

yaml/boundaries.yaml
travisgrigsby: This filter catches mainly border=administrative from what I can tell.  I can combine with the above, but it will basically look like two filters in one: 
`if (it's admin level [2,4,6] AND boundary disputed) OR (it has disputed_by AND (dispute=yes or disputed=yes))`.  
I'm not sure it's any cleaner to combine into one filter ... if you're jazzed about it I can do it though

integration-test/2022-disputed-borders.py
travisgrigsby: fixed in [4624bff](https://github.com/tilezen/vector-datasource/pull/2079/commits/4624bff282298e346af4fb4f60dda5f1d1b86a7d)

yaml/boundaries.yaml
travisgrigsby: fixed in [4624bff](https://github.com/tilezen/vector-datasource/pull/2079/commits/4624bff282298e346af4fb4f60dda5f1d1b86a7d)

queries/ne.jinja2
travisgrigsby: fixed in [4624bff](https://github.com/tilezen/vector-datasource/pull/2079/commits/4624bff282298e346af4fb4f60dda5f1d1b86a7d)

travisgrigsby:
@nvkelso Looks like the disputed borders cutting is still needed - we're seeing partial overlaps of border pieces with disputed pieces.  Alas! Someday we can remove this.

travisgrigsby:
@nvkelso added tests in 761ae3ca007175e0fd3d9ccb2abe59ae0eaf3027

nvkelso (approved):

PR #2080, opened 2022-05-03
Mark unwanted disputed lines as unrecognized

jeffdefacto:
This adds some lua to mark currently unwanted disputed lines to receive a `unrecognized_dispute=yes` tag that `boundaries.yaml` will turn into `kind: unrecognized`


nvkelso (commented):
This LGTM, though the diff is noisy because it includes changes from other active PRs. Please make sure those other PRs are merged and then the diff here is just the meaningful changes.

osm2pgsql.lua
```diff
@@ -786,6 +789,15 @@ function osm2pgsql.process_relation(object)
         return
     end
 
+-- Mark some disputes as unrecognized to hide them by default
```
nvkelso, line 792: Nit: add incline comments to say plainly what disputed each of these `B##` are for

yaml/boundaries.yaml
```diff
@@ -123,6 +123,19 @@ filters:
...
   #
+
+  # Additionally, we filter out disputes that are not currently desired for widespread display.
+  # These get a kind of `unrecognized` in order to suppress their display
+  # These have been marked in the lua with unrecognized_dispute=yes to identify them.
+  - filter:
+      all:
+        - geom_type: line
+        - unrecognized_dispute: yes
+    min_zoom: 8
```
nvkelso, line 134: This is OK to be 8 here, because it for OSM features which generally come in at zoom 8. The Natural Earth features separately come in earlier with per feature min_zooms.

nvkelso (commented):
I see some spare `tz_` properties leaking thru reviewing a build. Please remove those and then I think this PR is ready to merge.

PR #2086, opened 2022-05-25
Lua change to turn off somaliland admin1 features to match NE behavior

jeffdefacto:
Somaliland and Somalia have differing views on admin1 (OSM admin_level = 4) borders within Somaliland's borders. NE displays no borders within the area so setting the OSM data to match will align the views. 

This adds some lua to select the admin_level 4 features and set their admin_level to nil


nvkelso (approved):
Verified in a build.

PR #2087, opened 2022-06-10
disable reimports on the ci

peitili:
We are using a very old version of reimport that still supports python2, but his version on the CI sometimes generate inconsistent results on reimports. So we just disable the reimports on the CI.


kevinkreiser:
this looks good to me but im assuming the integration tests sometimes have network issues?

peitili:
> this looks good to me but im assuming the integration tests sometimes have network issues?

Yup, those failed on master branch always.

kevinkreiser (approved):

PR #2088, opened 2022-06-23
Use NE names when available

jeffdefacto:
Names from Natural Earth will overwrite OSM places features names where available


data/functions.sql
```diff
@@ -1264,3 +1264,157 @@ BEGIN
...
+    ne_name_tags HSTORE DEFAULT NULL;
+BEGIN
+    IF wikidata_id IS NULL THEN
+        RETURN '{}'::jsonb;
+    END IF;
+
+  -- if it's a country, only look it up in the iso and tlc tables
+    IF place_tag='country' OR place_tag='unrecognized' THEN
+        SELECT hstore(ARRAY[
+            'ne_name_ar', i.name_ar,
```
travisgrigsby, line 1281: I'm curious about the choice to do array - wouldn't just a straight hstore be better?

data/functions.sql
```diff
@@ -1264,3 +1264,157 @@ BEGIN
...
+                'ne_name_zht', t.name_zht
+                ]) INTO ne_name_tags
+            FROM ne_10m_admin_0_countries_tlc t
+            WHERE t.wikidataid = wikidata_id
+                AND featurecla IN ('Admin-0 country', 'Admin-0 dependency');
+        END IF;
+
+    ELSE
+            SELECT hstore(ARRAY[
+                'ne_name_ar', sp.name_ar,
```
travisgrigsby, line 1347: I'm not sure if this is possible, but it seems preferable to have this arg list in a var somewhere and reuse it across all these queries - you'd just have to change the alias for each of the tables to the same thing.  It's just visually difficult to verify that these are all the same

data/functions.sql
```diff
@@ -1264,3 +1264,157 @@ BEGIN
...
+                'ne_name_zh', t.name_zh,
+                'ne_name_zht', t.name_zht
+                ]) INTO ne_name_tags
+            FROM ne_10m_admin_0_countries_tlc t
+            WHERE t.wikidataid = wikidata_id
+                AND featurecla IN ('Admin-0 country', 'Admin-0 dependency');
+        END IF;
+
+    ELSE
+            SELECT hstore(ARRAY[
```
travisgrigsby, line 1346: there's enough visual distance from the last comment that it's probably worth saying 'not a country' etc in a comment here

data/functions.sql
```diff
@@ -1264,3 +1264,157 @@ BEGIN
...
+        END IF;
+    END IF;
+
+      -- return an empty JSONB rather than null, so that it can be safely
+      -- concatenated with whatever other JSONB rather than needing a check for
+      -- null.
+    IF NOT FOUND THEN
+        RETURN '{}'::jsonb;
+    END IF;
+    RETURN to_jsonb(coalesce(ne_name_tags, ''::hstore));
```
travisgrigsby, line 1418: oh does this coalesce turn an array into a map?   if so, then ignore my previous comment about arrays

vectordatasource/transform.py
```diff
@@ -9739,3 +9739,21 @@ def create_dispute_ids(shape, props, fid, zoom):
...
+    Override the name:xx properties with __ne_* variants from
+    Natural Earth, if there are any.
+    """
+
+    name_prefix = '__ne_name_'
+
+    for k in props.keys():
+        if k.startswith(name_prefix):
+            language = k[len(name_prefix):]
+            name = props.pop(k)
```
travisgrigsby, line 9755: nit: maybe `ne_name` ? Makes it a little clearer what's going on.

data/functions.sql
travisgrigsby: It would also mean fewer places to update when we decide to add a country

travisgrigsby (approved):
This all makes sense to me!   I have some minor cosmetic suggestions but, assuming there are no issues when one of the rows is bad (e.g. a null ne lang value should be ok), ship it!

Issue #2092, opened 2022-07-05
For buildings, add root_id = id


travisgrigsby:
For performance on the rendering side for buildings, it'd be nice to set the root_id equal to its id.  This saves us from having to check the kind before determining whether to look at root_id or id.

nvkelso:
Fixed via #2093

PR #2093, opened 2022-07-05
Travisg/20220630 add root id to buildings

travisgrigsby:
Addresses #2092 .  Adds a `root_id` equal to the `id` to buildings to help rendering performance.

- [x] Update tests
- [ ] Update docs


bcamper (approved):
lgtm

PR #2095, opened 2022-07-13
Support for On-The-Fly GeoJson Layers and a new Mutator Transform

kevinkreiser:
The basic gist here is I want to add 2 features.

1. i want to be able to drop a new layer into a tile build and be able to reference that layer from queries. if the layer is not there on the filesystem i want the transforms that reference it to do nothing (same thing that happens when a layer isnt found). the bulk of this was accomplished over in: https://github.com/tilezen/tilequeue/pull/414
2. i want a new transform that allows me, with injected python, to be able to mutate what the properties or geometry of a given feature becomes. it seemed to me we already had some of these things especially for properties but geometry was lacking.

Once I had those two features implemented I added a landmarks layer to the queries which simply allows you to drop a geojson file into your tile build and mark buildings with it and become a point feature layer.


vectordatasource/transform.py
```diff
@@ -1291,12 +1295,57 @@ def _f(shape, cutting_shape):
     return _f
 
 
+# we can load external geojson layers from file, all we need is a directory and geojson file named after the layer
+# we expect it to already be in mercator projection and we only fill out the minimal structure of the feature tuple
+# there's no ID, most of the datum stuff is missing, we are mostly worried about the shape and properties
+# we return None if we couldnt load the layer for any reason
+def _load_external_layer(params, name):
```
kevinkreiser, line 1302: this is moving over to `tilequeue` where the rest of the layers are loaded and ill add a new section to queries.yaml instead of having it be part of the transform

queries.yaml
```diff
@@ -338,6 +339,15 @@ layers:
...
       - vectordatasource.transform.remove_feature_id
     area-inclusion-threshold: 1
+  # optional inline layer, if the file is found on disk it will be used
+  landmarks:
+    geometry_types: [Point, Polygon, MultiPolygon]
+    clip: false
+    area-inclusion-threshold: 1
+    simplify_before_intersect: false
+    simplify_start: 0,
+    layer_path: ./metatile-layers/landmarks.geojson
```
kevinkreiser, line 349: this property signifies an "inline" layer. when the features are loaded from file we expect them to be already processed. that is we will not try to load a `landmarks.yaml` and do processing on the data that is in the geojson file. we expect the tags and values to be there and the coordinates to already be in mercator projection

queries.yaml
```diff
@@ -542,6 +552,23 @@ post_process:
       source_layer: roads
       properties: [layer]
 
+  # supersede existing osm buildings with data from landmark geojson layer
```
kevinkreiser, line 555: landmarks mark buildings with a superseded property so that when the two overlap the downstream consumer of the tiles will be able to tell

queries.yaml
```diff
@@ -542,6 +552,23 @@ post_process:
...
+  - fn: vectordatasource.transform.overlap
+    params:
+      base_layer: buildings
+      cutting_layer: landmarks
+      attribute: superseded
+      target_attribute: superseded
+      linear: false
+
+  # turn the landmark geojson polygons into point features
```
kevinkreiser, line 564: the final output of the landmarks layer is a point layer, basically landmark pois with a bunch of metadata about the landmark. the mutate transform lets us both modify the geom and the properties with python code that gets `eval`'d directly. pretty powerful stuff. you can reference existing `properties` of a feature using the `{properties}` tag and you can access the geometry using the `{geometry}` tag. in the future we might want to allow more stuff to be referenced.

queries.yaml
```diff
@@ -1865,6 +1892,7 @@ post_process:
         - addr_housenumber
         - addr_street
         - osm_relation
+      exclude: ['superseded']
```
kevinkreiser, line 1895: if a building gets marked by a landmark, we exclude it from merging with other buildings, keeping it a single entity

requirements.txt
```diff
@@ -7,13 +7,15 @@ future==0.16.0
...
 hiredis==0.2.0
+kdtree
 lxml==4.6.2
 mapbox-vector-tile==1.2.0
 ModestMaps==1.4.7
 protobuf==3.4.0
 psycopg2==2.7.3.2
 pyclipper==1.0.6
 pycountry==17.9.23
+pyshp==2.3.0
```
kevinkreiser, line 18: apparently my system thought i needed these, i didnt check to see if that was the case. i can revert

vectordatasource/transform.py
```diff
@@ -9758,3 +9758,43 @@ def override_with_ne_names(shape, props, fid, zoom):
...
     return shape, props, fid
+
+
+def mutate(ctx):
+    """
+    We take a layer and we modify the geometry using the python expression in the query. The expressions have access to
+    both the existing shape and existing properties via python string format replacements {shape} and {properties}
+    respectively. Each expression is then eval'd to replace the existing feature in the layer with the result of the
+    expressions. By default the expressions are a no-op
+    """
```
kevinkreiser, line 9769: pretty good explanation here. basically we loop over the features and eval the python in the queries.yaml to compute new properties or geometry. if you omitted one or the other (they are both optional) it just leaves the existing values unchanged.

queries.yaml
peitili: Could we make the name more explicit about its behavior, like `pre_processed_layer_path` might be a better name?

integration-test/2095_inline_layers.py
```diff
@@ -0,0 +1,78 @@
+# -*- encoding: utf-8 -*-
```
peitili, line 1: did you check whether your PR introduces new test failures than the master branch?

vectordatasource/transform.py
```diff
@@ -9758,3 +9758,43 @@ def override_with_ne_names(shape, props, fid, zoom):
...
+
+    layer = ctx.params.get('layer')
+    assert layer, 'regenerate_geometry: missing layer'
+    geometry_expression = ctx.params.get('geometry_expression', '{shape}')
+    properties_expression = ctx.params.get('properties_expression', '{properties}')
+    assert geometry_expression or properties_expression, \
+        'regenerate_geometry: missing one of geometry or properties expression'
```
peitili, line 9776: The comment is confusing here, if the assert is false, then both are missing, correct?

integration-test/2095_inline_layers.py
kevinkreiser: do we have a list of ones that are failing? i can check on my machine which ones fail and pass before and after my change and report back

vectordatasource/transform.py
kevinkreiser: yes both are missing but you need at least one hence: "missing one of" as in you need at least one of them. i can reword it though

integration-test/2095_inline_layers.py
peitili: you can also check it on the CI report

kevinkreiser: the report is identical to masters last failed build, the tests fail at: `test.test_transform.BuildingsUnifyTest`

peitili: oh no, this is not good, it means none of integration tests got a chance to even run yet. I found the unit test failure was introduced recently https://github.com/tilezen/vector-datasource/pull/2093

@travisgrigsby would you help fix that unit test so that other integration tests can be run and checked?

travisgrigsby: done #2099

kevinkreiser: cool ill wait for that to merge, then merge master into this branch then compare the two outputs from CI

integration-test/2095_inline_layers.py
```diff
@@ -0,0 +1,80 @@
...
+    def test_landmarks_buildings_unmerged(self):
+        # we can temporarily add the external layer, here we build a square structure over the center of the map
+        # note that the origin of the feature is in the upper right quadrant but should appear in all quadrants
+        null_island = {
+            'type': 'FeatureCollection',
+            'name': 'foo',
+            'crs': {'type': 'name', 'properties': {'name': 'urn:ogc:def:crs:EPSG::3857'}},
+            'features': [
+                {'type': 'Feature',
+                 'properties': {'name': 'null island hut', 'superseded': True, 'height': math.pi, 'origin': [1, 1], 'id': 42},
```
nvkelso, line 43: 🙇

queries.yaml
```diff
@@ -9,6 +9,7 @@ all:
   - boundaries
   - transit
   - admin_areas
+  - landmarks
```
nvkelso, line 12: Add docs

nvkelso: See the traffic layers for inspiration for optional layer descriptions.

queries.yaml
```diff
@@ -542,6 +552,23 @@ post_process:
       source_layer: roads
       properties: [layer]
 
+  # supersede existing osm buildings with data from landmark geojson layer
+  - fn: vectordatasource.transform.overlap
+    params:
+      base_layer: buildings
+      cutting_layer: landmarks
+      attribute: superseded
+      target_attribute: superseded
```
nvkelso, line 561: In all the places...

`superseded_by` instead of `superseded`.

Please use the same family of property names as WOF in Mapzen tradition: 

https://github.com/whosonfirst/whosonfirst-properties/blob/main/properties/wof/superseded_by.json

queries.yaml
kevinkreiser: ha! i was looking around for docs, it seemed like even the changelog isnt up to date. ill grep around to see if i can find them!

queries.yaml
kevinkreiser: ah wow i had no idea this existed, i'll swap it out

queries.yaml
nvkelso: https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#layer-reference
and
https://github.com/tilezen/vector-datasource/blob/master/docs/layers.md#traffic-flow

nvkelso: The CHANGELOG is out of date. Historically that's been updated as releases are tagged – and it's past time to tag v1.9. Thanks for the nudge.

vectordatasource/transform.py
```diff
@@ -9758,3 +9759,44 @@ def override_with_ne_names(shape, props, fid, zoom):
...
+
+    zoom = ctx.nominal_zoom
+    start_zoom = ctx.params.get('start_zoom', 0)
+    end_zoom = ctx.params.get('end_zoom')
+    if zoom < start_zoom:
+        return None
+    if end_zoom is not None and zoom >= end_zoom:
+        return None
+
+    # for the max zoom a transform needs to be re-entrant so we take a copy here
```
peitili, line 9789: what is `max zoom` here?

kevinkreiser: ah sorry the last zoom in the tile pyramids that we create (z16). there is code over in tilequeue here: https://github.com/tilezen/tilequeue/blob/master/tilequeue/process.py#L613 where it basically calls into the transforms a second time (hence re-entrancy)

integration-test/2095_inline_layers.py
kevinkreiser: i was finally able to get the build to run al the way and it has the same number of failures as master for integration tests:

master:
```
Ran 2198 tests in 1000.156s

FAILED (failures=109)
```

this branch:
```
Ran 2200 tests in 734.782s

FAILED (failures=109)
```

travisgrigsby (approved):

travisgrigsby:
I'd remove the strikethrough in the description before merging

PR #2098, opened 2022-07-26
Change min_zoom of canals

jeffdefacto:
lower min zoom of canals to 11 when boat=yes is not present


travisgrigsby (approved):

nvkelso:
@jeffdefacto Please file a followup PR with documentation changes and an integration test.

PR #2101, opened 2022-08-03
Fix NE country names

jeffdefacto:
Some country names in Natural Earth are undesirable so this changes them back to their OSM names which are generally shorter. Most of these changes can be deprecated once NE is updated.

There was also an issue with the overwriting of chinese names with NE as the column name was incorrect that has been fixed and added to the test.

This also addresses some broken country related tests that needed fixing


travisgrigsby (approved):
awesome thanks for doing this

PR #2103, opened 2022-08-11
Fix sea labels

jeffdefacto:
This fixes a bug that caused sea labels to not actually be rendered if they were just label points rather than converted polygons.

The one zoom they worked properly at was z16 which was the only zoom we tested at. I changed to the test to check different zoom levels.


nvkelso:
Generally LGTM, but screenshots or build details to verify please.

queries.yaml
```diff
@@ -1036,9 +1037,9 @@ post_process:
   - fn: vectordatasource.transform.drop_features_where
     params:
       source_layer: water
-      start_zoom: 3
+      start_zoom: 2
       where: >-
-        'label_placement' not in properties and kind in ('sea', 'bay', 'strait', 'fjord')
+        'label_placement' not in properties and kind in ('sea', 'bay', 'strait', 'fjord', 'marina', 'harbour')
```
nvkelso, line 1042: This is to fix an error that Gretchen found in Chicago

integration-test/148-sea-ocean-labels-water-layer.py
```diff
@@ -10,33 +10,163 @@
 # layer.
 class SeaOceanLabelsWaterLayer(FixtureTest):
 
-    def test_gulf_of_california(self):
-        z = 16
-        x, y = deg2num(27.5578100671435, -111.704650024786, z)
+    # NOTE: No ocean points in the North America extract :-(
+    def test_ocean_point(self):
```
nvkelso, line 14: Added ocean unit test

integration-test/148-sea-ocean-labels-water-layer.py
```diff
@@ -10,33 +10,163 @@
...
+        self.assert_no_matching_feature(
+            0, x, y, 'water',
+            {'kind': 'ocean', 'name': 'Atlantic Ocean'})
+        x, y = deg2num(1.0, 1.0, 1)
+        self.assert_has_feature(
+            1, x, y, 'water',
+            {'kind': 'ocean', 'name': 'Atlantic Ocean', 'min_zoom': 1.0,
+             'label_placement': True})
+
+    def test_gulf_of_california_point(self):
```
nvkelso, line 42: Sea unit test for points

integration-test/148-sea-ocean-labels-water-layer.py
```diff
@@ -10,33 +10,163 @@
...
+        self.assert_no_matching_feature(
+            4, x, y, 'water',
+            {'kind': 'sea', 'name': 'Greenland Sea'})
+        x, y = deg2num(1.0, 1.0, 5)
+        self.assert_has_feature(
+            5, x, y, 'water',
+            {'kind': 'sea', 'name': 'Greenland Sea', 'min_zoom': 5.0,
+             'label_placement': True})
+
+    def test_hudson_bay_large_polygon(self):
```
nvkelso, line 101: Sea unit test for polygons

integration-test/148-sea-ocean-labels-water-layer.py
```diff
@@ -10,33 +10,163 @@
...
+            {'kind': 'sea', 'name': 'Hudson Bay', 'min_zoom': 2.0,
+             'label_placement': True})
+
+        # ensure the zoom range is correct vis-a-vis the size
+        x, y = deg2num(1.0, 1.0, 1)
+        self.assert_no_matching_feature(
+            1, x, y, 'water',
+            {'kind': 'sea', 'name': 'Hudson Bay',
+             'label_placement': True})
+        # Even though the feature has min_zoom 2, it isn't exported in a tile
```
nvkelso, line 130: This was super confusing so added a note – basically even though the feature _should_ be in a zoom 2 tile because it's min_zoom is 2, for build performance and tile size reasons we actually exclude it elsewhere until zoom 4 tiles.

queries.yaml
```diff
@@ -903,7 +905,7 @@ post_process:
   - fn: vectordatasource.transform.update_min_zoom
     params:
       source_layer: water
-      start_zoom: 4
+      start_zoom: 2
```
nvkelso, line 908: This could be zoom 4 like before, but future proofing as it's low cost and if the other config is changed to 2 from 4, this would already allow it to work

queries.yaml
```diff
@@ -929,8 +931,9 @@ post_process:
         ] if area >= area_threshold][0]
       # note the following line cannot be written as
       # properties is not None and properties.get('label_placement') is True
+      # and we exclude sea and ocean because those are pre-graded
       where: >-
-        properties is not None and label_placement
+        properties is not None and label_placement and kind not in ('sea', 'ocean')
```
nvkelso, line 936: This exclusion is required else the OSM points for oceans and seas will get re-graded to min_zoom 16!

queries/planet_osm_polygon.jinja2
```diff
@@ -160,12 +160,22 @@ FROM (
...
+-- don't conflict with the Natural Earth data at zoom < 8.
 {% elif zoom >= 7 %}
     ({{ bounds['polygon']|bbox_filter('way', 3857) }} AND
       (mz_earth_min_zoom < {{ zoom + 1 }} OR
        mz_landuse_min_zoom < {{ zoom + 1 }} OR
        mz_poi_min_zoom < {{ zoom + 1 }} OR
-       mz_transit_level < {{ zoom + 1 }}))
+       mz_transit_level < {{ zoom + 1 }} OR
+       mz_water_min_zoom < {{ zoom + 1 }}))
+{% elif zoom >= 2 %}
```
nvkelso, line 173: So technically our min_zoom 1 ocean points will also pop in at zoom 2 tiles...

vectordatasource/transform.py
```diff
@@ -1608,10 +1608,10 @@ def intracut(ctx):
     'province': 4,
     'state': 4,
 
-    'sea': 3,
+    'sea': 4,
```
nvkelso, line 1611: These are junk DNA at this point? But now at least they are consistent with everything else, and this section does occasionally get triggered for other places, so didn't feel comfortable removing totally (even just for ocean and sea).

yaml/water.yaml
```diff
@@ -47,13 +47,56 @@ globals:
         - [ 15,       1000 ]
         - [ 16,        400 ]
       default: 17
+  - &water_sea_min_zoom
```
nvkelso, line 50: This is the meat of the PR, all the rest is annoying config and bug fixes

nvkelso:
@jeffdefacto @peitili @travisgrigsby This should be ready to roll now :)

jeffdefacto:
I see you expanded upon my PR a little bit haha @nvkelso

peitili (approved):

nvkelso:
@jeffdefacto Thanks for the initial work here! It turned out to be a very meaty problem to dig into, and good way to keep busy now that I have more time on my hands ;)

PR #2106, opened 2022-09-21
fix a broken integ test

peitili:
- [ ] Update tests
- [ ] Update docs

### DIFF
--- a/integration-test/1425-osm-features.py
+++ b/integration-test/1425-osm-features.py
@@ -63,7 +63,7 @@ class FeatureTests(FixtureTest):
         self._run_poi_test(
             'http://www.openstreetmap.org/node/4991752549',
             '16/33891/21798',
-            {'kind': 'charging_station', 'scooter': True, 'motorcar': False})
+            {'kind': 'charging_station', 'scooter': True, 'truck': False})
 
         self._run_poi_test(
             'http://www.openstreetmap.org/node/4882913388',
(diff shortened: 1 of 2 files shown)


integration-test/1425-osm-features.py
```diff
@@ -59,10 +59,12 @@ def test_amenities(self):
         self._run_poi_test(
             'http://www.openstreetmap.org/node/3656027075',
             '16/19337/24632', {'kind': 'charging_station'})
+
         self._run_poi_test(
             'http://www.openstreetmap.org/node/4991752549',
             '16/33891/21798',
-            {'kind': 'charging_station', 'scooter': True, 'car': False})
+            {'kind': 'charging_station', 'scooter': True, 'truck': False})
```
kevinkreiser, line 66: based on the tagging maybe it should have been `motorcar`, but `truck` should work as well

kevinkreiser (approved):

integration-test/1425-osm-features.py
kevinkreiser: these kind of tests are like whack-a-mole if you dont freeze a version of osm to test on, my 2 cents

peitili: > these kind of tests are like whack-a-mole if you dont freeze a version of osm to test on, my 2 cents

how to get a freeze osm? is there a "permlink" or something?

peitili: yup changed to motorcar

kevinkreiser: yeah the only way to freeze it would be to take a snapshot ourselves and run against that. that could be done but then we'd have to make sure it was lots of tiny little snippets so that it would be something we could actually check in size wise.. better would be to mock the data in the tests that use the real data and stop using the real data probably. either way a lot of work

PR #2121, opened 2024-05-31
Update dependencies

rwrx:
Add missing dependencies into requirements.txt be able to install vector-datasource properly

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ argparse==1.4.0
 ASTFormatter==0.6.4
 boto==2.49.0
 edtf==2.6.0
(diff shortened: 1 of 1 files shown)


iandees:
I think the Fiona dependency is only needed to generate shapefiles, which shouldn't normally need to run. Is that where you ran into it?

rwrx:
Yes, it is possible, that when generating shapefiles

iandees:
Looks like the CI checks are failing because the repo is using an extremely out of date Python version. I'll merge as-is, but we should probably look at doing the Python 3 upgrade at some point.

PR #2122, opened 2024-07-01
Fix missing boundary property for water layer

rwrx:
Fix missing boundary property for water layer which caused that water boundary lines was not shown in Mapzen map styles


rmarianski (commented):
thanks. This looks good to me, but I defer to @nvkelso in case I'm missing something.

